PR #8, opened 2016-12-03
Unit test framework and some tests for editText. Also, fix Backspace on a Mac.

paulcwatts:
Testing editText meant moving a lot of its dependencies into their own files. 

A couple of additions:

1. I fixed backspace for us Mac users.
2. I added tox testing, and fixed an issue with packaging that tox found.

There are still some TODOs, but I can finish those while this is being reviewed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.?sv
 __pycache__
 .idea/
+visidata.egg-info/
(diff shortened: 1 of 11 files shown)


paulcwatts:
Another thing I can add is also making sure Travis is running the tox tests.

saulpw:
There were too many changes in develop since the base of this branch, so I applied the key changes here manually.  Thanks, Paul!

PR #17, opened 2017-04-27
Extend Path API

zormit:
Small changes to the Path API to support the "journal" code example in the developer quickstart doc.

### DIFF
--- a/visidata/vd.py
+++ b/visidata/vd.py
@@ -1471,6 +1471,9 @@ class Path:
     def is_dir(self):
         return os.path.isdir(self.resolve())
 
+    def exists(self):
+        return os.path.exists(self.resolve())
+
     def iterdir(self):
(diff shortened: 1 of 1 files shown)


visidata/vd.py
```diff
@@ -1471,6 +1471,9 @@ def read_bytes(self):
     def is_dir(self):
         return os.path.isdir(self.resolve())
 
+    def exists(self):
+        return os.path.exists(self.resolve())
```
zormit: this might not be the desired function in some cases, see [here](http://stackoverflow.com/questions/82831/how-do-i-check-whether-a-file-exists-using-python). But for the example, where we know that it *will* be a file, it should be good enough?

saulpw: Yes, this is good enough (focus in VisiData and this support class is simplicity/straightforwardness).  Thanks for getting the example to work, @zormit!

PR #39, opened 2017-06-02
add CLI option "-f" to specify filetype

jhrmnn:
Hi. Excellent software! Just a tiny change. My SQL databases have the extension `.db` and I thought this is the simplest change to make that work with visidata. adding `open_db = open_sqlite` to the SQL addon seemed bad, since `.db` is quite generic.

### DIFF
--- a/bin/vd
+++ b/bin/vd
@@ -13,6 +13,7 @@ def main():
     parser.add_argument('inputs', nargs='*', help='initial sources')
     parser.add_argument('-d', dest='debug', action='store_true', default=False, help='abort on error')
     parser.add_argument('-r', dest='readonly', action='store_true', default=False, help='disable editing')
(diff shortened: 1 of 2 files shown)


saulpw (changes requested):
This is a good idea.  Please resubmit PR against develop branch and I will accept.  Thanks for contributing!

saulpw:
As a second option, what if it auto-loaded a `~/.vdrc`, and then you could put `open_db = open_sqlite` in there?  Do you have other `.db` files yourself, or is your concern for other people?

jhrmnn:
Ok, I rebased against `develop`.

My concern was for other people, my `.db` files are always SQLite. So I'd welcome `~/.vdrc`. Or yet better `~/.config/visidata/config` to go with [XDG](https://wiki.archlinux.org/index.php/XDG_Base_Directory_support).

PR #40, opened 2017-06-02
Dpb docstrings

brannerchinese:
Add docstrings for principal classes and some methods in `addons/`.

@saulpw: I am especially curious as to whether I've summarized the functionality of `metasheets.SheetJoin` correctly.

### DIFF
--- a/visidata/addons/editlog.py
+++ b/visidata/addons/editlog.py
@@ -9,7 +9,7 @@ def open_vd(p):
     return vs
 
 class EditLog(Sheet):
-    """Populate log of commands for current session."""
+    """Maintain log of commands for current session."""
     current_replay_row = None  # must be global, to allow replay
 
     def __init__(self, name, *args):
@@ -29,7 +29,7 @@ class EditLog(Sheet):
         self.current_exec_row = None
 
     def undo(self):
-        """Delete last command and replay entire log."""
+        """Delete last command, reload sources, and replay entire log."""
         if len(self.rows) < 2:
             error('no more to undo')
 
@@ -47,7 +47,7 @@ class EditLog(Sheet):
     def before_exec_hook(self, sheet, keystrokes, args=''):
         """Declare initial sheet before any undos can occur.
 
-        This is done when source is initiallylopened."""
+        This is done when source is initially opened."""
         assert sheet is vd().sheets[0], (sheet.name, vd().sheets[0].name)
         if EditLog.current_replay_row is None:
             self.current_active_row = [ sheet.name, keystrokes, args, None ]
--- a/visidata/addons/editlog.py
+++ b/visidata/addons/editlog.py
@@ -9,7 +9,7 @@ def open_vd(p):
     return vs
 
 class EditLog(Sheet):
-    """Populate log of commands for current session"""
+    """Populate log of commands for current session."""
     current_replay_row = None  # must be global, to allow replay
 
     def __init__(self, name, *args):
@@ -29,7 +29,7 @@ class EditLog(Sheet):
         self.current_exec_row = None
 
     def undo(self):
-        """Delete last command and replay entire log"""
+        """Delete last command and replay entire log."""
         if len(self.rows) < 2:
             error('no more to undo')
 
@@ -47,14 +47,14 @@ class EditLog(Sheet):
     def before_exec_hook(self, sheet, keystrokes, args=''):
         """Declare initial sheet before any undos can occur.
 
-        This is done when source is initiallylopened"""
+        This is done when source is initiallylopened."""
         assert sheet is vd().sheets[0], (sheet.name, vd().sheets[0].name)
         if EditLog.current_replay_row is None:
             self.current_active_row = [ sheet.name, keystrokes, args, None ]
             self.rows.append(self.current_active_row)
 
     def after_exec_sheet(self, vs, escaped):
-        """Declare ending sheet for the most recent command"""
+        """Declare ending sheet for the most recent command."""
         if self.current_active_row:
             if escaped:
                 del self.rows[-1]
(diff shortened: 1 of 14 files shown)


visidata/addons/editlog.py
```diff
@@ -9,6 +9,7 @@ def open_vd(p):
     return vs
 
 class EditLog(Sheet):
+    """Populate log of commands for current session."""
```
saulpw: s/Populate/Maintain

visidata/addons/editlog.py
```diff
@@ -28,6 +29,7 @@ def __init__(self, name, *args):
         self.current_exec_row = None
 
     def undo(self):
+        """Delete last command and replay entire log."""
```
saulpw: Include 'reload sources' between delete and replay.

visidata/addons/editlog.py
```diff
@@ -43,13 +45,16 @@ def undo(self):
         status('undid "%s"' % deleted_row[1])
 
     def before_exec_hook(self, sheet, keystrokes, args=''):
+        """Declare initial sheet before any undos can occur.
+
+        This is done when source is initiallylopened."""
```
saulpw: typo

visidata/addons/metasheets.py
```diff
@@ -12,12 +12,14 @@ def _getattrname(o, k):
     return v.__name__ if v else None
 
 def ColumnGlobal(name):
+    """Return Column object with given name."""
```
saulpw: This is semi-vestigial but still used in one place.  I'd like to remove or rename.

visidata/addons/metasheets.py
```diff
@@ -60,6 +62,7 @@ def __getattr__(self, colname):
 
 
 def ColumnExpr(sheet, expr):
+    """Assign expression to column object."""
```
saulpw: Create new Column from Python expression

visidata/addons/metasheets.py
```diff
@@ -132,6 +137,13 @@ def reload(self):
 
 #### slicing and dicing
 class SheetJoin(Sheet):
+    """Implement four kinds of JOIN.
+
+     * "full": full outer JOIN
```
saulpw: Since the jointype is a special character, include the jointype character here in the comments (e.g. '&' is inner)

visidata/addons/pivot.py
```diff
@@ -3,18 +3,21 @@
...
         self.variableCols = variableCols
         for colnum, col in enumerate(srcsheet.keyCols):
             if col not in variableCols:
-                newcol = Column(col.name, getter=lambda r,colnum=colnum: r[0][colnum])
+                newcol = Column(col.name, getter=lambda r, colnum=colnum: r[0][colnum])
                 newcol.srccol = col
                 self.nonpivotKeyCols.append(newcol)
 
-        self.command(ENTER, 'vd.push(source.copy(cursorCol.aggvalue)).rows = cursorRow[1].get(cursorCol.aggvalue, [])', 'push sheet of source rows aggregated in this cell')
+        self.command(ENTER, '''vd.push(source.copy(cursorCol.aggvalue)).rows='''
```
saulpw, line 18: I think this is probably a very good idea, to split up some more complicated commands into multiple lines like this.  However, I would put the `.rows=` on the second line, and use single quotes instead of triple quotes (since the strings don't span multiple lines).

visidata/addons/pyobj.py
```diff
@@ -87,11 +91,13 @@ def reload(self):
 
 
 def open_json(p):
+    """Handle JSON file as a whole, via `json.load`."""
```
saulpw: "JSON file as a single object"

visidata/addons/pyobj.py
```diff
@@ -87,11 +91,13 @@ def reload(self):
...
     import json
     return load_pyobj(p.name, json.load(p.open_text()))
 
 # one json object per line
 def open_jsonl(p):
+    """Handle JSON file incrementally via `json.loads`."""
```
saulpw: JSONL file as a list of objects (one JSON object per line)

saulpw (commented):
Overall good.  Can single-line """docstrings""" become 'docstrings' ?

visidata/addons/pivot.py
DataBranner: I avoided proofreading of this sort, which is usual in Python to keep every line below 80 characters.

Custom is to use triple quotes for long strings (if the strings occur naturally within `()`, `[]`, or `{}` they will be treated as a single string).

Custom is to break _after_ operators, rather than before. So 

```python
self.command(ENTER, '''vd.push(source.copy(cursorCol.aggvalue)).'''
                    '''rows=...''')
```

or 

```python
self.command(ENTER, '''vd.push(source.copy(cursorCol.aggvalue)).rows='''
                    '''...''')
```

but not

```python
self.command(ENTER, '''vd.push(source.copy(cursorCol.aggvalue)).rows'''
                    '''=...''')
```

Also, the custom is not to fiddle excessively with perfect indentation. I usually use eight spaces for a continuing line, regardless of content.

DataBranner:
> Can single-line """docstrings""" become 'docstrings' ?

Custom is trebled double quotes. https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring:

> For consistency, always use `"""triple double quotes"""` around docstrings.

DataBranner:
Thanks for the other comments. Will make changes as soon as I'm able.

anjakefala:
```quote
Custom is trebled double quotes. https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring:
```

@DataBranner Trebled single quotes also work. :) I think they would more aesthetically mesh with [saulpw's existing doc style](https://github.com/saulpw/visidata/blob/stable/visidata/vd.py#L3).

DataBranner:
Closed because revised version of this pull request was merged, instead.

PR #44, opened 2017-06-06
Fixed path command. It is missing PATH$:

mtelesha:
If you follow the directions path will be change to ONLY include visidata PATH.


saulpw:
I believe this is already fixed in the `develop` branch, which will be merged into `stable` soon.  Thanks for your attention to detail!

PR #49, opened 2017-06-07
Dpb small fixes

DataBranner:
Make three small fixes previously discussed with @saulpw.

### DIFF
--- a/visidata/vd.py
+++ b/visidata/vd.py
@@ -1747,9 +1747,9 @@ class OptionsObject:
 options = OptionsObject(base_options)
 
 # Generator => A .. Z AA AB ...
-chars = string.ascii_uppercase
 defaultColNames = (''.join(j) for i in range(options.maxlen_col_hdr)
-                              for j in itertools.product(chars, repeat=i+1)
+                             for j in itertools.product(string.ascii_uppercase,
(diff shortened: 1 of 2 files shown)


visidata/vd.py
```diff
@@ -1752,6 +1746,12 @@ def __setitem__(self, k, v):
 
 options = OptionsObject(base_options)
 
+# Generator => A .. Z AA AB ...
+chars = string.ascii_uppercase
```
saulpw: Don't create this at the global scope.  It's only used once in the following generator, so just use string.ascii_uppercase there.

visidata/vd.py
```diff
@@ -350,12 +351,15 @@ def chooseOne(choices):
     else:
         return input('/'.join(str(x) for x in choices) + ': ')
 
+
 def regex_flags():
+    """Return flags to pass to regex functions from options"""
     return sum(getattr(re, f.upper()) for f in options.regex_flags)
 
 # A .. Z AA AB .. ZZ
 defaultColNames = list(itertools.chain(string.ascii_uppercase, [''.join(i) for i in itertools.product(string.ascii_uppercase, repeat=2)]))
```
saulpw: This version should be removed.

PR #53, opened 2017-06-12
Add basic Sphinx support and document the process

DataBranner:
This Sphinx installation is built and visible at http://forked-visidata.readthedocs.io. 

The only actual documentation written so far is a description of installing Sphinx and setting up Read the Docs.

~~Original `.md` files have been left in place, side by side with their `.rst` equivalents.~~ It's not too late to decide to stay with `.md`, although `.rst` has more power and is more standard in use with Sphinx.


saulpw:
Thank you!  Please remove _build/ and other files which are not source material.

DataBranner:
Unfortunately, `docs/_build/` is where the actual active HTML and other material is, under Sphinx. I see clearly that it is not a very Unixful tool, alas. (There are even binary files!) I left the older `.md` and `.html` files where they were in docs/ until I had a chance to talk to you — properly speaking, those should be removed as part of this pull request.

DataBranner:
I guess it's possible that RTD will generate all of that stuff each time it builds the documentation. But I haven't tested it with `_build/` removed. Will try it tomorrow morning.

DataBranner:
Couldn't wait. You were right.

saulpw:
Okay, this looks a lot cleaner.  The last 4 files (your notes, requirements.txt, reverse-sort-commands.vd, and vd.py) should not need to be added/changed in this PR.

DataBranner:
Why not requirements.txt?

saulpw:
The only two requirements are already listed.  The other requirements aren't necessary for VisiData to be installed.  If those are all necessary for the documentation (are they?) then they should be listed somewhere else.

DataBranner:
This is an interesting question. `requirements.txt` is used by `pip`. Without it, Sphinx and sphinx-rtd-theme (not part of the standard library) are not available locally, and Sphinx can't be used. Since users sometimes want to build Sphinx locally in order to check the documentation formatting visually, before pushing to repo, Sphinx should be available locally.

Read the Docs, however, seems to collect all the libraries it needs from the code itself. I created a [new repository without `requirements.txt`](https://github.com/databranner-forks/vd-no-requirements), and [RTD was able to generate the docs correctly](http://vd-no-requirements.readthedocs.io/en/latest/). There's a line in the build-record:

```bash
python /home/docs/checkouts/readthedocs.org/user_builds/vd-no-requirements/envs/latest/bin/pip install --use-wheel -U --cache-dir /home/docs/checkouts/readthedocs.org/user_builds/vd-no-requirements/.cache/pip sphinx==1.5.3 Pygments==2.2.0 setuptools==28.8.0 docutils==0.13.1 mkdocs==0.15.0 mock==1.0.1 pillow==2.6.1 readthedocs-sphinx-ext<0.6 sphinx-rtd-theme<0.3 alabaster>=0.7,<0.8,!=0.7.5 commonmark==0.5.4 recommonmark==0.4.0
```

That this is something generated by RTD is clear because the libraries listed are not the same as those in `requirements.txt`.

DataBranner:
There are two other approaches I can think of. 

1. The simpler one is to pare `requirements.txt` down to only the most necessary libraries, probably just Sphinx and `sphinx-rtd-theme`. `setup.py` as you have originally configured it already looks to `requirements.txt`:

   ```python
   setup(...
         install_requires=open('requirements.txt').read().split(),
         ...
         )
   ```

2. The other is to use the `install_requires` keyword in `setup.py` to specify just those libraries explicitly. That would be sufficient for installing VD with pip, I think, but it will cost some users extra work, as installation of libraries via `pip` from `requirements.txt` is normal practice.

docs/documentation.rst
```diff
@@ -0,0 +1,80 @@
...
+
+      Read the Docs project: Admin => Advanced Settings => Default branch
+
+#. **Set Python interpreter version**: Also on the "Advanced Settings" page, it may be useful to set the Python interpreter to the correct version of Python for the project (CPython 3.x).
+
+#. **Submit and build**: On submit, assuming a webhook is found, Read the Docs will build and deploy a new version, using the branch you have set.
+
+   The build may fail if you make too many attempts to build within a very short time.
+
+   If a build attempt fails, Python or other tracebacks are visible if you click on the attempt in question, on the Builds pages. Currently (20170611) the presence of ``pkg-resource==0.0.0`` in ``requirements.txt`` causes such a failure, so edit it out if you have updated that file from ``pip freeze``.
```
saulpw: remove this paragraph.

PR #59, opened 2017-06-26
Add a tour of VisiData

anjakefala:


anjakefala:
https://github.com/anjakefala/visidata/blob/tours/docs/tours.rst

PR #62, opened 2017-06-26
Adds missing __init__.py to addons and updates README

### LINKED ISSUE #61: Problems installing visidata
I ran into two issues when following the install instructions for visidata:

1) The requirements.txt is not where the readme says it will be
2) Running `vd` after install throws an ImportError "No module named 'visidata.addons".

I've attached my install and run logs.  The run logs show the full stack trace for the import error.

[install.txt](https://github.com/saulpw/visidata/files/1102131/install.txt)
[run.txt](https://github.com/saulpw/visidata/files/1102132/run.txt)

gusennan:
Fixes #61

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ $ pip3 install visidata
 ```
 $ git clone http://github.com/saulpw/visidata.git
 $ cd visidata
-$ pip install -r requirements.txt
+$ pip install -r docs/requirements.txt
 $ python setup.py install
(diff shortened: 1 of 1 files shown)


saulpw:
Thanks for submitting this, @gusennan.  I think requirements.txt belongs in the toplevel in general so I made a separate commit to fix these issues.  Please keep reporting issues and sending patches!

PR #73, opened 2017-06-30
Added support for reading gzip-compressed files.

khughitt:
This PR allows files which have been gzip-compressed and end in a `.gz` extension to be handled by `vd`.


saulpw:
Thanks, @khughitt!  If you submit the PR against `develop` branch, I will merge it straightaway and make sure it works with my larger datasets.

khughitt:
Sounds good! Will do.

PR #75, opened 2017-07-02
Bugfix: User-guide was not being translated

anjakefala:


DataBranner:
I don't understand the meaning of "translated" in the commit message.

anjakefala:
https://github.com/saulpw/visidata/blob/develop/docs/user-guide.rst

It was not being properly rendered.

DataBranner:
Oh, that was the beginning of a change in design, not fully implemented. But it seems you're working on the user guide, so I'll leave it alone and stick to other parts of the documentation.

PR #76, opened 2017-07-02
Add documentation section for devs on handling the screen

brannerchinese:
Proposed new section is visible on line at http://forked-visidata.readthedocs.io/en/latest/screen.html#. Note:

1. Now that we have a tour for users, there is no need for separate "screen layout" sections for users and devs. The existing content has been revised for devs alone, combining the old `screen-layout.rst` and `dev-cursors.rst` files.

2. There are a few small fixes to other parts of the docs — one small change to the user docs, @anjakefala, in 6ce693e.

3. @saulpw: One thing I have not been able to clarify, because I don't understand it, is

    > Key columns are ‘___’.

    (in http://forked-visidata.readthedocs.io/en/latest/screen.html#key-columns). I'm happy to clarify it before PR-merge if you can explain what you meant.


docs/about/contributing.rst
```diff
@@ -9,22 +9,24 @@ there is an easy way to improve the tool for both of us.
...
+* `develop <https://github.com/saulpw/visidata/tree/develop>`__ has the
+  most up-to-date version of visidata (which will eventually be merged to
+  stable).
+
+If you wish to contribute, please fork from `develop
+<https://github.com/saulpw/visidata/tree/develop>`__ and submit a `pull request
+<https://github.com/saulpw/visidata/pulls>`__ against it.
+
+A developer's guide can be found `here
+<http://visidata.readthedocs.io/en/stable/dev-guide>`__.
```
anjakefala: Oh, that link has been broken.
It should be: https://visidata.readthedocs.io/en/stable/dev-guide.html

brannerchinese:
Thanks. I've cut it down to https://visidata.readthedocs.io, which will find the "stable" version by default.

PR #77, opened 2017-07-03
Removes a bit of code duplication

gusennan:
Reading through the code and saw this duplicated block.  Think this is a reasonable way to handle the 'search wrapped' message without the duplicate block.


saulpw:
Thank you, @gusennan, this is definitely cleaner code.   However, how does this perform with 5-10 million rows?  It seems like this could substantially increase the time for each search.  If it is only a few percent, I would be willing to take it for the cleaner code.  Would you be willing to time it?  I can provide a large dataset if you need.

gusennan:
I thought about that a bit too, but I think because it's the branch of the if-statement that exits the function (returns), this check is only going to be executed once.  Is that not the case?

saulpw:
Yes, but it is executed once 'per incremental search'.  That is, you first type `/foo` and then it moves the cursor to the first row which has 'foo' in the current column, then every subsequent `n` or `p` would execute this code again.  The code is structured so that if you hit a block of rows with `foo` in them, that each subsequent `n` takes an unnoticeable amount of time.  My concern is that this added `r in rng` check would noticeably detract from this snappiness for large datasets.  If it were a simple arithmetic comparison, that would re-complicate the code a bit, but alleviate my concerns.  Does that seem worthwhile to remove this duplication?

gusennan:
Ah, I see your concern.  Well that shouldn't be an issue because `int` membership in `range` is [an O(1) arithmetic operation](https://docs.python.org/3/library/stdtypes.html#typesseq-range) in Python 3.2+ (the documentation about this is near the bottom of the section on range membership).  If you're curious, [here's](https://github.com/python/cpython/blob/6cca5c8459cc439cb050010ffa762a03859d3051/Objects/rangeobject.c#L338) the implementation of the arithmetic check in cpython.

I'm happy to profile this though if you'd like to send a large dataset.  Would be good to have for future tests for future contributions anyway.

saulpw:
Okay, if that's constant time then that's fine.  I would like a smoke test with a large dataset though; try with this dataset (5.6m rows, 1.4GB): https://data.cityofnewyork.us/Public-Safety/NYPD-Complaint-Data-Historic/qgea-i56i  Thank you!

gusennan:
moveRegex executed 10000 times:

[before.txt](https://github.com/saulpw/visidata/files/1121918/before.txt) took 33.92336903799878 sec
[after.txt](https://github.com/saulpw/visidata/files/1121919/after.txt) took 33.82868683500419 sec

```python
import sys
import time

from visidata.vd import vd, openSource

def perf_test(argv):
    vs = openSource(argv[1])
    singleton = vd()
    singleton.push(vs)
    vs.vd = singleton
    time.sleep(10)
    start = time.perf_counter()
    print("found on rows:", end=' ')
    for i in range(10000):
        vs.moveRegex(regex="Murder", columns=vs.columns[7])
        print(vs.cursorRowIndex, end=' ')
    stop = time.perf_counter()
    print("\n\ntook {} sec".format(stop - start))

if __name__ == "__main__":
    perf_test(sys.argv)
```

PR #79, opened 2017-07-23
Integrate SchemaType into VisiData

ingydotnet:
I have a project called SchemaType that is a Schema and Type definition
language for all tabular data. It started out as a schema language for YAML
(and JSON (since all JSON *is* YAML)), but then realized that it fits at least
15 data models (including csv, tsv, rdbms, spreadsheets, etc) fairly well.

To give you an idea of things, I added a `sample_data/birdsdiet.stp` file that
quite concisely and exactly defines the data in `sample_data/birdsdiet.tsv`.

I can imagine that if you ran: `vd sample_data/birdsdiet.tsv` it could see the
associated `.stp` file, and then do lots of cool things for you. The most
obvious things are:

* Assign column types
* Validate data
  * Possibly showing invalid data in red
  * Possibly give hints on what's wrong

Note: there is no need to merge this PR. I just thought it was a good way to
start a potentially fascinating conversation.

### DIFF
--- a/sample_data/birdsdiet.stp
+++ b/sample_data/birdsdiet.stp
@@ -1,0 +1,19 @@
+-spec: schematype.org/v0.0.1
+-from: github:schematype
+
++birdsdiet:
+  Family: +bird-name
+  MaxAbund: +float 0..500
+  AvgAbund: +float 0..50
+  Mass: +float 0..5500
+  Diet: +bird-food
+  Passerine: +bool/int
+  Aquatic: +bool/int
+
++bird-name: +str // {Word} ( {sep} {word} ){0,3} //
++bird-food: .[Insect PlantInsect Vertebrate]
+
+-defs:
+  Word: / {uc} {word}? /
+  word: / {wc} ( {dash} (= {lc}) | {lc} )* /
+  sep: /( {sp}{amp}{sp} | {sp} | {amp} )/


saulpw:
Wow, this is awesome.  Can the source data file be given in the .stp too?  Or is it just known from the .stp filename stem?  Either way, let's make an open_stp() addon which sets up the data for consumption in vd.  Should be relatively easy if you've already got parsing code for .stp.

saulpw:
And/or visidata could save a .stp description file on command, from the information it already has.

ingydotnet:
I don't think if makes sense for a schema file (.stp) to know anything about a
specific instance file (.tsv). Is that what you meant?

I think you could certainly add a --schema=... CLI option.

.stp format is a YAML DSL format. ie it's valid YAML but has some DSLs in the
string content. It compiles to an stpc file that is JSON. Still, I should
probably write a Python3 compiler for this. I'll port my Pegex
language-agnostic parser framework to Python to do it.

Having vd spit out a .stp file is a great idea. It's something that I always
wanted to do. Use existing (valid) data to make an approximate schema, then
possibly refine the schema by hand, then use it to generate various software
for working with that kind of data. VisiData would be the perfect tool for
doing that.

VisiData can be a great user-friendly entrypoint for SchemaType.

PR #90, opened 2017-12-02
[curses] Add option to use default term colors

anjakefala:
- Set to False (default): the "normal" and "background" curses colors will be
  set by VisiData
- When set to True, curses will instead use the default values for colors
  currently specified by the terminal
- To set to True on default, add `options.use_default_colors=True` to
  ~/.visidatarc


saulpw (approved):
Thank you, @anjakefala!  Looks good to me.

PR #120, opened 2018-03-06
Support bz2 and xz (LZMA) compression

jpgrayson:
The Path module is modified to recognize and transparently decompress bzip2
(.bz2) and LZMA (.xz) files in addition to the already-supportedd gzip
(.gz) format.

The Path.read_bytes() method is also repaired to correctly handle
compressed files (it previously only worked with uncompressed files).

### DIFF
--- a/visidata/Path.py
+++ b/visidata/Path.py
@@ -1,6 +1,5 @@
 import os
 import os.path
-import gzip
 
 from .vdtui import *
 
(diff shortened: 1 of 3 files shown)


saulpw:
Hi @jpgrayson, thanks very much for this PR.  We'd like to take it and it looks exactly right.  However, we've been trying out a new development workflow and just pushed the 1.1 release (which is not in the current develop branch) to stable.  Can you submit this same commit as a PR against the current stable?

Apologies for the inconvenience, we need to update the contributing.md with the new process and will be doing so soon.

PR #144, opened 2018-05-09
Portable clipboard handling

### LINKED ISSUE #143: Portable clipboard handling
Currently, copying to the system clipboard must be configured manually for every platform apart from macOS. Although it's documented (e.g. [here](http://visidata.org/man/)), I assumed the option was only needed for advanced customization (e.g. on non-standard platforms), and didn't realize that it must be set on Linux until it errored when I tried to use it (and I saw it mentioned in a [comment on a closed issue](https://github.com/saulpw/visidata/issues/98#issuecomment-379616663)).

Why not set it up automatically on all systems, rather than just Macs? There's a library which supports writing to and reading from the clipboard <s>portably and painlessly</s>: [pyperclip](https://github.com/asweigart/pyperclip). It could either be added as an optional dependency, or — better still — copy 'n' pasted into the project (it's a [single file](https://github.com/asweigart/pyperclip/commit/df268c11d1b9bdfcb322e7fdcb61191e572e316c)). The current `options.clipboard_copy_cmd` option could be kept for advanced usage/non-standard platforms.

chocolateboy:
This PR closes #143 by providing sensible defaults for the `clipboard_copy_cmd` option on:

- macOS (using pbcopy)
- Windows (using clip)
- Linux etc. (using xclip or xsel)

There are issues (chiefly https://github.com/asweigart/pyperclip/issues/67, which causes a segfault on Python 3) which make pyperclip unusable on Linux in its current form, so a small helper class is used instead.

### DIFF
--- a/visidata/clipboard.py
+++ b/visidata/clipboard.py
@@ -1,3 +1,6 @@
+import shutil
+import subprocess
+import tempfile
 
 from visidata import *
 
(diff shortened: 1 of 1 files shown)


saulpw:
Thanks for doing this, @chocolateboy!  Sorry for the delay in merging, we're at PyCon this week and haven't had time to test the change ourselves.  We'll merge it as soon as we can and then it will be in the next release.  Thanks again!

visidata/clipboard.py
```diff
@@ -20,35 +24,76 @@
...
-# or 'xclip -selection primary'
-# or 'pbcopy w' for mac
+option('clipboard_copy_cmd', '', 'command to copy stdin to system clipboard')
 
-import tempfile
-import subprocess
+commands = {
+    ('clip', 'win32'):    [],                                      # Windows Vista+
+    ('pbcopy', 'darwin'): ['w'],                                   # macOS
+    ('xclip', None):      ['-selection', 'clipboard', '-filter'],  # Linux etc.
```
anjakefala, line 32: Hey @chocolateboy! I was wondering about some of the changes you made here.

From what I can tell, within`xclip` the `-filter` flag seems to make some output from xclip overlay the VisiData interface. So i have a preference for removing it.

Additionally, I was wondering about the choice for `primary` -> `clipboard ` for both `xclip` and `xsel`? They seem to work well but what is the underlying difference between them?

Otherwise, works beautifully and I am really glad you did the work for this.. =)

anjakefala (changes requested):

visidata/clipboard.py
chocolateboy: If `-filter` is omitted, xclip runs as a daemon, which [blocks when run in a pipeline](https://github.com/asweigart/pyperclip/issues/116#issuecomment-387825778). This may not matter in VisiData (I haven't run it in a pipeline), but it shouldn't hurt. I think a better solution than launching xclip as a daemon every time would be to ensure the command's stdout is consumed and discarded (which is what I assumed was happening).

`clipboard` is the system clipboard targeted by <kbd>Ctrl-C</kbd>, <kbd>Ctrl-V</kbd>, and <kbd>Ctrl-X</kbd>. This is the standard clipboard used by browsers, text editors etc i.e. the apps most likely to be on the receiving end of pastes from VisiData. `primary` is the X clipboard accessible by clicking the left and middle mouse buttons. I believe that use case is less common (my guess is that most users are unaware of it), particularly for a keyboard-oriented tool, but anyone who has a preference for it can still, of course, configure it via the `clipboard_copy_cmd` option.

chocolateboy: Just pushed a fix which discards the clipboard command's stdout.

anjakefala: Thanks for elaborating @chocolateboy and for the fix which discards the clipboard's stdout! Looks good.

anjakefala (approved):

PR #163, opened 2018-06-27
Adds instructions for installing via pipsi

TheLonelyGhost:
Not sure if you want this pointed at the `dev` branch instead, since they're docs-only changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,12 @@ To install VisiData, with loaders for the most common data file formats (includi
 
     $ pip3 install visidata
 
+### Install via pipsi
+
+For the best results in a platform-agnostic way, use [pipsi](https://github.com/mitsuhiko/pipsi) to create a virtualenv entirely for visidata. It's all the benefits of sandboxing with virtualenv without having to do that pesky `. {dir}/bin/activate` step.
+
+    $ pipsi install --python=python3 visidata
+
 ### Install via conda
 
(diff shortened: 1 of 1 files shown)


anjakefala:
Hm! I actually did not even know about pipsi (I was reading up on pipenv) and this seems like a really useful thing to flesh out. I agree with you that we should try to make it as easy as possible for the pip3 download option to be as accessible to non-python users. (Homebrew and Debian packages both force us to pin down mandatory dependencies which is not feasible with the amount of loaders we currently support. Pip3 is the most flexible and versatile installation option and is painful as-is to some of our users.)

So this is wonderful! I cannot merge it in yet because we are actually working to move the installation instructions [off the README](https://github.com/saulpw/visidata/blob/develop/www/install.md). 

Also, you know what, I think this belongs more in a /topic piece on pip3 wrangling (it feels to me like it is not an installation instruction per se...more a tutorial which assists users who need the flexibility of pip3 but do not already have an existing python3 setup which they are comfortable with).

If you want to be part of reviewing that piece, or helping write the pipsi part of it, I can bring you on board for that! In any case, thank you so much for drawing my attention to pipsi and for providing the tutorial for how to use it. :D

TheLonelyGhost:
Yeah I can help with that. I didn’t see a repo for VisiData.org to make a PR for the instructions there.

If you need any help with cleaning up docs from a fresh set of eyes, I’m happy to assist.

Issue #164, opened 2018-07-02
Trouble with pyenv after python was upgraded to 3.7


deinspanjer:
I'm struggling with my Mac environment here, and I would greatly appreciate it if you have any pointers.

I tried to run visidata today, and got the following error:
```dyld: Library not loaded: @executable_path/../.Python
  Referenced from: /usr/local/Cellar/visidata/1.2/libexec/bin/python3.6
  Reason: image not found
```

I tried reinstalling visidata using `brew reinstall visidata` and the error changed to:
```Traceback (most recent call last):
  File "/usr/local/bin/vd", line 10, in <module>
    from visidata import *
  File "/usr/local/Cellar/visidata/1.2/libexec/lib/python3.7/site-packages/visidata/__init__.py", line 7
    from .async import *
              ^
SyntaxError: invalid syntax
```

So I started digging and I found that it seems something has gotten goofed up between having pyenv on my system (which had not yet had an explicit 3.7.0 installed) and having the Homebrew version of python 3.7.0 installed via brew upgrade yesterday.

```
which python3
~/.pyenv/shims/python3

pyenv
pyenv 1.2.5

echo $PATH
/usr/local/Cellar/pyenv-virtualenv/1.1.3/shims:~/.pyenv/shims:/usr/local/bin:~/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin

pyenv version
system (set by ~/.pyenv/version)

python3
Python 3.7.0 (default, Jun 29 2018, 20:13:13)
```

So I tried a few different fumblings and finally I decided to try removing pyenv from the path entirely.  Did that, reinstalled visidata, but still the same error about invalid syntax.

I understand this likely isn't a "visidata" problem, but if there is anything you can suggest, I'd greatly appreciate it.

anjakefala:
Hi Daniel!

Thank you so much for bringing this to our attention! :100: 

Our current hypothesis is that the original error occurred because brew removed Python 3.6 and replaced it with Python 3.7. The VisiData installation could no longer find the 'image' for Python 3.6.

This was 'fixed' when you reinstalled VisiData, however the next error that occurred was due with the fact that Python3.7 introduced `async` as a [reserved keyword](https://docs.python.org/3/whatsnew/3.7.html). We currently have a file `async.py` and a decorator `@async`. Both will likely need to be changed. 

The things we can do:

1) We find a workaround using brew.
2) We find a workaround using pyenv on a system with a brewed and non-brewed python.
[...]

PR #165, opened 2018-07-04
 [python3.7] Change async decorator to asyncthread and rename async.py 

anjakefala:
Python3.7 introduced async as a [reserved keyword](https://docs.python.org/3/whatsnew/3.7.html).

As a result async.py needed to be renamed to asyncthread.py. The decorator `async` also needed to be renamed to `asyncthread`.

Closes https://github.com/saulpw/visidata/issues/164


saulpw (approved):
Looks good!

PR #167, opened 2018-07-07
[docs] Update motd and release notes for 1.2.1

anjakefala:


saulpw (commented):
motd should be a copy instead of rename.

anjakefala:
Thanks! Updated!

PR #170, opened 2018-07-15
[commands] Commands, options and keybindings overhaul

anjakefala:
**Settings architecture**

The first component of this PR is that the API for setting a keybinding, option and creating a command have been refined.

The notes associated with this change have been written up [here](https://github.com/saulpw/visidata/blob/commands/dev/design/169-settings.md).

This permits the layering of settings: there are global, sheet-specific and instance level options/keybindings. 

**Longnames**

The longnames are a major VisiData API.  They will be used for remapping keybindings to a specific behaviour. The cmdlog will use longnames instead of keystrokes by default, which will make .vd scripts more shareable/publishable.  As more commands are added, some will be used infrequently enough that they are unbound and only executed with exec-command (^A). 

In this PR,  longnames have been settled on, with the goal being that at least 90% of them will be consistent from version to version, thus minimising user friction.

The following are the design notes for the longnames:

1) two-three words max if possible, should be short and fit on a keymap (verb - object - input)

2) command classes should be unique in their first 3 chars and ideally mostly in their first 2.

3) command longnames should be intuitively understandable, or at least not jargony

4) longnames should evoke interface, when possible

The following is the (current) longname glossary:
- syscopy: copy to system clipboard
- sysopen: open with $EDITOR or other external program.  'launch'?
- cells: this column, selected rows (or all rows if none selected)
- expr: python expression
- regex: python regex
- all: all sheets or all visible columns
- col: cursorCol (will be marked in sheet)
- cols: all visible columns
- subst: regex A/B
- capture: regex match A(B) yielding capture groups
- dup: push copy of sheet
- searchr: search reverse
- go: move the cursor
- scroll: change the visible screen area without changing the cursor
- select: all or selected
- fill: (from fill null cells)
- dive: conceptual -> go deeper
- slide: move row/column
- show: displays on status
- cancel-: kill thread/async process
- setcol: set selected cells in this column
- addcol: add new column to this sheet
- search: on default with regex
- aggregate: set an aggregator
- delete: cuts and moves to clipboard


anjakefala:
The foundation and architectural decisions are in a solid place. What is left is to compose tests, refine naggling issues and update the documentation. Moving that work to develop.

PR #180, opened 2018-08-21
pcapng support

layertwo:
This was a lot easier than I thought.

Moved out pcap file reader to `open_pcap`. I'm not thrilled dpkt uses a ValueError to indicate it can't open a file (it should be more specific), but it works for now. There are a handful of other types of captures, but those don't seem to be very popular.

Added some additional transport types (we could probably do better on this), but it was low risk and easy to implement.

Lastly, ntar is an older pcapng format, so created that mapping.

Wireshark has great sample captures here where I pulled most of the captures I used to test with. https://wiki.wireshark.org/SampleCaptures

### DIFF
--- a/visidata/loaders/pcap.py
+++ b/visidata/loaders/pcap.py
@@ -90,6 +90,14 @@ def init_pcap():
     load_iana(url_iana)
 
 
+def read_pcap(f):
+
+    try:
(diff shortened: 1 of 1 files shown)


saulpw:
Thanks @lmessenger!

PR #181, opened 2018-08-21
Limit gz=/setcol-range parameter to the number of rows to be filled

ssiegel:
This protects users who naïvely input e.g. itertools.count()
and expect it to work, from eating up all system memory.


saulpw (approved):
Looks good, thanks @ssiegel!  Sorry this bit you.

PR #185, opened 2018-09-11
Handle a difference file with real cell differences

mhosken:
This code implements true row differencing between a --diff sheet and the main sheet, only showing cells that have changed.

This fixes a bug whereby if a row is inserted or deleted, all rows following are considered different. It also removes the diff sheet from the main list of sheets.

### DIFF
--- a/bin/vd
+++ b/bin/vd
@@ -100,7 +100,7 @@ def main():
 
     if args.diff:
         vs = openSource(args.diff)
-        vd().push(vs)
+#        vd().push(vs)
         setDiffSheet(vs)
 
     if args.batch:
--- a/bin/vd
+++ b/bin/vd
@@ -100,7 +100,7 @@ def main():
 
     if args.diff:
         vs = openSource(args.diff)
-#        vd().push(vs)
+        vd().push(vs)
         setDiffSheet(vs)
 
     if args.batch:
--- a/visidata/diff.py
+++ b/visidata/diff.py
@@ -1,21 +1,49 @@
 from visidata import *
+from difflib import SequenceMatcher
 
 theme('color_diff', 'red', 'color of values different from --diff source')
 theme('color_diff_add', 'yellow', 'color of rows/columns added to --diff source')
 
 globalCommand(None, 'setdiff-sheet', 'setDiffSheet(sheet)')
 
+class DiffRow(list):
+    def __hash__(self):
+        return hash(u"\uFDD0".join(self))
+
+def loaddiff(base, dif):
+    base.rows = [DiffRow(x) for x in base.rows]
+    dif.rows = [DiffRow(x) for x in dif.rows]   # or do this before?
+    m = SequenceMatcher(a=dif.rows, b=base.rows)
+    for t, astart, aend, bstart, bend in m.get_opcodes():
+        alen = aend - astart
+        blen = bend - bstart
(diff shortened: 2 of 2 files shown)


saulpw (commented):
Hi @mhosken, thanks for the PR!  I like the concept and your approach seems workable at first glance.  I'll have to review it more thoroughly tomorrow. There are a few details that would need to be fixed regardless though:

- no leftover print() or comments
- merge into develop branch, not directly to stable
- test with large input file--loaddiff should probably be tagged `@asyncthread`

Thanks for thinking of this and coding it up.

mhosken:
I'm not sure how to move this to another branch.

Hmm. removing the diff sheet from vd().push() stops it from working. No idea why that is. Too deep for me!

saulpw:
@mhosken Sheets are loaded on first push, so if the sheet is never pushed, it would have to be loaded explicitly (with `vs.reload()`).

But I would prefer the diff sheet to be pushed anyway, so I can `Ctrl-^` back and forth between the two sheets.  If it's not pushed automatically, we would need some way to push the diff sheet, I think.

saulpw:
Okay, I've had a chance to give this a more thorough review.  My previous plan for the future of sheet diff was to allow rows to be grouped and diffed by key columns, so that row order wouldn't matter.  But I was holding off, because I couldn't account for the use case you have here (like an interactive 'git diff' for tabular data).  I'm glad you submitted this PR because you've shown me that we need both kinds of functionality.  Your proposed implementation has also convinced me that difflib is a reasonable way to go about getting this kind of diff (collapsing removed/inserted whole rows so that other diffs are still evident).

While this proposed implementation will work for small data sheets, it has a fundamental flaw which prevents it from integrating with VisiData's functionality as a whole: it replaces .rows with a new list and each row with a new DiffRow object.  This would cause the following major problems:

- This will not work for non-sequence rowtypes.  Most source sheets (csv, tsv, etc) do use a sequence rowtype, but derived sheets, meta (internal) sheets, and some more esoteric source sheets use a non-sequence rowtype.  It is fundamental to the design of VisiData that a Sheet's rowtype be opaque, and only its columns and reload() know how to reach into the row's internals.  DiffRow unfortunately breaks that separation.

- 'remote' sheets like PandasSheet which use a non-list .rows object (in the case of PandasSheet it's a forwarding wrapper) lose all advantage of having the rows already in memory.  Using the proposed differ with a pandas sheet as either the base or the diff will significantly and unnecessarily increase memory usage.

To fix this PR will require a substantial rewrite.  The relevant diff components would have to be stored in a separate data structure associated with the sheet, and the individual rows would have to be referenced instead of duplicated.  (VisiData uses `id(row)` in other places, for better or worse).  Let's create an issue on Github and endeavor to do this more diligently in the near future.  Are you interested in being part of that process?

In the meantime, you can copy this code into your `.visidatarc` to continue to use it yourself.  If `makeDiffColorizer` is overridden in .visidatarc, then `setDiffSheet` will use the overridden version.  Note that in the next version (released this week or next, likely 1.4), the return value of colorizeDiffs will have to be `"color_diff"` instead of `options.color_diff` (and same for `color_diff_add`).

mhosken:
I'm happy to close this PR and let you take it from here. Just glad I was able to encourage you in a good direction.

PR #258, opened 2019-03-03
Paste text from system clipboard functionality

kovasap:
I was having a hard time finding the best python way of interacting with the clipboard, so I added a dependency to do it.  Curious to talk about what the best way of doing this is!  

Also I just picked an arbitrary user action (ctrl-click), since DOUBLE_CLICK was being captured by the single click code.  Happy to discuss more about what the final action should be.

### DIFF
--- a/visidata/clipboard.py
+++ b/visidata/clipboard.py
@@ -4,7 +4,6 @@ import subprocess
 import sys
 import tempfile
 import functools
-import pyperclip
 
 from visidata import vd, asyncthread, status, fail, option, options
 from visidata import Sheet, saveSheets, Path, Column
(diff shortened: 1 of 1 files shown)


visidata/clipboard.py
```diff
@@ -106,6 +110,14 @@ def save(self, vs, filetype):
                 close_fds=True)
             p.communicate()
 
+def pasteFromClipboard(col_idx, row_idx, sheet):
+    text = pyperclip.paste()
```
saulpw: Hi @kovasap!  Thanks for submitting this PR.  We tried using pyperclip for the `syscopy` commands but there were issues (https://github.com/saulpw/visidata/pull/144).  Probably what we should do, is add a fourth element to `__clipboard_commands` for the "paste args" (so the third element becomes the "copy args".  Then a `_Clipboard.paste()` method would return the value on the system clipboard using the same kind of `subprocess.Popen` as in `copy()` (but in reverse).

visidata/clipboard.py
```diff
@@ -36,6 +37,9 @@ def setslice(L, a, b, M):
 Sheet.addCommand('zY', 'syscopy-cell', 'copyToClipboard(cursorDisplay)')
 Sheet.addCommand('gzY', 'syscopy-cells', 'copyToClipboard("\\n".join(vd.sheet.cursorCol.getDisplayValue(r) for r in selectedRows))')
 
+Sheet.addCommand('CTRL-BUTTON1_PRESSED', 'paste-mouse',
```
saulpw: `CTRL-BUTTON1` might be okay for a default binding.  I think the longname should be `syspaste-mouse` though, for symmetry with `syscopy` since the data is coming from the system clipboard.

visidata/clipboard.py
```diff
@@ -106,6 +110,14 @@ def save(self, vs, filetype):
                 close_fds=True)
             p.communicate()
 
+def pasteFromClipboard(col_idx, row_idx, sheet):
+    text = pyperclip.paste()
+    for t in text.split('\n'):
+        try:
+            sheet.columns[col_idx].setValue(sheet.rows[row_idx], t)
+        except IndexError:
+            break
```
saulpw, line 139: I can see how this `try/except` is useful for when pasting goes past the end of the sheet.  I think it would be better as an `if` though, so that any `IndexError`s caused by setValue aren't swallowed up.  Otherwise I think this should work!

saulpw (changes requested):

visidata/clipboard.py
kovasap: that makes sense - I'll change the name.  what about middle click (kinda like one of the possible pastes on linux systems)?

visidata/clipboard.py
kovasap: sounds good!

visidata/clipboard.py
kovasap: Ok cool.  I can throw that together.

kovasap:
Sorry for the delays on the changes, i will probably have time to finalize everything this weekend!

PR #262, opened 2019-03-12
Add an option to disable sorted keys when saving to JSON

chocolateboy:
This PR adds an option to disable sorting object keys when saving to JSON.

By default, sheets are saved to JSON with sorted keys. This is fine in many cases, but I often find myself having to postprocess the output (e.g. with [jq](https://github.com/stedolan/jq)) to restore the meaningful, human-friendly order I've arranged in the sheet(s), particularly when dealing with records that have a `name`, `title` or `type` field, which tend to get shunted to the bottom of each record while unimportant fields and (sometimes) long arrays of `addresses`, `categories`, `dates` etc. are given pride of place. (If you've ever tried to look at a [JSON AST](https://astexplorer.net/) with sorted keys, you'll know how tricky it can be to grasp the structure when the all-important `type` field is relegated to the bottom of each node.)

Providing an option to disable sorting means I can skip this postprocessing step. It also saves me the trouble of re-arranging the columns again whenever I re-import these files i.e. it makes it quicker/easier to roundtrip data and resume processing it.

### DIFF
--- a/visidata/loaders/json.py
+++ b/visidata/loaders/json.py
@@ -6,6 +6,7 @@ from visidata import wrapply, TypedExceptionWrapper, TypedWrapper
 
 
 option('json_indent', None, 'indent to use when saving json')
+option('json_sort_keys', True, 'sort object keys when saving to json')
 
 
 def open_json(p):
@@ -79,7 +80,7 @@ class Cell:
 
 class _vjsonEncoder(json.JSONEncoder):
     def __init__(self, **kwargs):
-        super().__init__(sort_keys=True, **kwargs)
+        super().__init__(sort_keys=options.json_sort_keys, **kwargs)
         self.safe_error = options.safe_error
 
     def default(self, cell):


saulpw (approved):
Looks good, @chocolateboy!  Happy to make it an option, thanks for submitting a clean PR!

saulpw:
WRT #429, I'd like to change it so that the order of the keys is always the order of the columns.  I think this means we can remove the `json_sort_keys` option.  Does that work for you?  (Does anyone ever want their keys to be actually sorted?  If so we can leave the option but default to False).

chocolateboy:
Preserving the column order by default works for me and would be my preference, thanks :+1: But I'd vote to keep the sorted option as well, since I use both. (As a general rule, if I know the data/schema, there's often a "natural", prioritized order I want to preserve, but If I'm exploring unfamiliar data, sorted keys/columns can make it easier to find things.)

saulpw:
Okay.  We'll keep the option but change the default to be `False`.  Thanks for quick feedback!

Issue #266, opened 2019-03-21
pandas: selecting rows does not work


azjps:
To reproduce, open any `PandasSheet` and try pressing `s`; no rows appear as selected.

The problem comes down to that `selection.py` heavily uses `id()` to track rows, and it will either use `__iter__()` or `__getitem__()` to access the rows. However, with `pandas`, these two methods yield different objects with different ids:
```python
$ ipython
Python 3.7.2 (default, Dec 29 2018, 06:19:36)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.3.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import pandas as pd

In [2]: df = pd.DataFrame({"a": [1, 2], "b": [3, 4]})

In [3]: [id(r) for r in df]
Out[3]: [140290703622920, 140290703910032]

In [4]: id(df.iloc[0]), id(df.iloc[1])
Out[4]: (140290637990936, 140290637990936)
```
In general, `id()` is not reliable for `pandas` rows as for heterogeneous-dtype dataframes a copy will be made when accessing rows (and probably also for some homogeneous-dtype):
```python
In [5]: df = pd.DataFrame({"a": [1, 2], "b": ["c", "d"]})

In [6]: df
Out[6]:
   a  b
0  1  c
1  2  d

In [7]: id(df.iloc[0])
Out[7]: 140290624720968

In [8]: id(df.iloc[0])
Out[8]: 140290663594864
```
I think the right way to work around this would be to allow customizing how to define an id for a row. For pandas, we can usually use the row label assuming a default `Int64Index`. If the DataFrame has duplicated entries in its index then its a bit hopeless without significant refactoring.

azjps:
Actually I take that back, it would probably make more sense to override all of the selection-related API and instead use a boolean mask instead to represent the selected rows. That would avoid any issues with working with individual rows or duplicate index entries. To do so though we would need to pass the integer index of the row as opposed to the row itself into `select()`; this could either be done by passing the information through `DataFrameAdapter.__getitem__` or modifying the keybinding, and I'm not sure which option makes more sense.

azjps:
Closed by #267

Issue #250, opened 2019-02-06
Use value_counts()/describe()/etc for derived sheets of PandasSheet


azjps:
I haven't done timings, but for a medium-sized `pandas.DataFrame` of ~1 million rows, computing a frequency or describe table for a column of a `PandasSheet` takes a significant amount of time (I'd estimate tens of minutes). It would be nice if `pandas/numpy` vectorized operations were used such as `DataFrame.value_counts()` and `DataFrame.describe()`, since this computation should be very fast to compute for a DataFrame of this size:

```python
In [11]: df.shape
Out[11]: (968981, 73)

In [12]: %timeit df.int_column.value_counts()
58.8 ms ± 1.73 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [13]: df.int_column.value_counts()
Out[13]: 
103296875000    57661
103187500000    55473
103218750000    53819
103234375000    53511
# ...
```

(Happy to work on a PR but would probably need a couple of pointers.)

saulpw:
Hi @azjps, thanks for the suggestion!  I think it would be great if pandas were better supported.  The current [PandasAdapter](https://github.com/saulpw/visidata/blob/develop/visidata/loaders/_pandas.py) is really just a glorious hack, so I'm glad it's working as well as it is. 

It wouldn't be that difficult to make an optimized PandasFreqTable, which would be a little less exciting than the regular FreqTable because it wouldn't be able to provide feedback when calculating, and you wouldn't be able to dive into or select or modify the source rows (unless there is a pandas method that returns a list of the rows or row indexes instead of just the counts?).  But if all you need is a quicker non-interactive summary, I can see this being a great addition.

If you're interested in submitting a PR, check out [how to create a loader](http://visidata.org/docs/loaders/).  Basically you'd start with:

```
PandasSheet.addCommand('F', 'freq-col', 'vd.push(PandasFreqTable(sheet, cursorCol))')

class PandasFreqTable(Sheet):
   def reload(self):
       # populate self.columns and self.rows according to aggregators and groups
```

and then write the reload() function.  See [freqtbl.py](https://github.com/saulpw/visidata/blob/develop/visidata/freqtbl.py) for how it does the columns, and you'd use pandas to create the rows.  I can't drive it myself but I'm happy to pair with you and/or provide and support you would need.

PR #267, opened 2019-03-21
[#250] Use value_counts() for PandasSheetFreqTable

azjps:
**do not merge yet**

Put in a bit of initial work into exploring using `value_counts()` for a Frequency Table for pandas. The surface area of the frequency table was considerably larger than I expected so I only implemented a small subset of features, but I'll note that a lot of the expected features weren't working prior either (for example, #266). Creating this PR for some initial feedback.

An initial hacky implementation for using value_counts()
to generate frequency table for pandas, which is vectorized
and hence significantly faster than the approach in
PivotTable. Added support for [Enter] > sourcerows.
Toggling/selection support is still broken here because
id(row) will not work with pandas.

Also tried to do some small cleanup with dtype
handling in PandasSheet.

This commit includes a lot of experimental and debugging
code which can be removed.

todo: support toggle in PandasSheet/PandasSheetFreqTable

### DIFF
--- a/visidata/loaders/_pandas.py
+++ b/visidata/loaders/_pandas.py
@@ -3,6 +3,8 @@ from visidata import *
 
 class DataFrameAdapter:
     def __init__(self, df):
+        import pandas as pd
+        assert isinstance(df, pd.DataFrame)
         self.df = df
 
     def __len__(self):
@@ -19,24 +21,48 @@ class DataFrameAdapter:
 
 # source=DataFrame
 class PandasSheet(Sheet):
-    def reload(self):
-        import numpy
-        import pandas
-        def dtypeToType(df, colname):
-            t = df[colname].dtype
-            if t == numpy.int64: return int
-            if t == numpy.float64: return float
-#            if t == pandas.Timestamp: return date
-            return anytype
+    def dtype_to_type(self, dtype):
+        import numpy as np
+        try:
+            if np.issubdtype(dtype, np.int):
+                return int 
+            if np.issubdtype(dtype, np.float):
+              return float
+            if np.issubdtype(dtype, np.datetime64):
+                return date
+        except TypeError:
+            # For categoricals and other pandas-defined dtypes
+            pass
+        return anytype
+   
+    def read_tsv(self, path, **kwargs):
+        """Partial function for reading TSV files using pd.read_csv"""
+        import pandas as pd
+        return pd.read_csv(path, sep='\t', **kwargs)
 
+    # @property
+    # def selectedRows(self):
+    #     if len(self._selectedRows) <= 1:
+    #         return list(self._selectedRows.values())
+    #     return [self.rows.iloc[i] for i in range(self.rows.shape[0]) if id(self.rows.iloc[i]) in self._selectedRows]
+
+    # def selectRow(self, row):
+    #     super().selectRow(row)
+    #     raise ValueError(self.selectedRows)
+
+    def reload(self):
+        import pandas
         if isinstance(self.source, pandas.DataFrame):
             self.df = self.source
         elif isinstance(self.source, Path):
             filetype = getattr(self, 'filetype', self.source.ext[1:])
-            readfunc = getattr(pandas, 'read_'+filetype) or error('no pandas.read_'+filetype)
+            if filetype == 'tsv':
+                readfunc = self.read_tsv 
+            else:
+                readfunc = getattr(pandas, 'read_'+filetype) or error('no pandas.read_'+filetype)
             self.df = readfunc(self.source.resolve(), **options('pandas_'+filetype+'_'))
 
-        self.columns = [ColumnItem(col, type=dtypeToType(self.df, col)) for col in self.df.columns]
+        self.columns = [ColumnItem(col, type=self.dtype_to_type(self.df[col])) for col in self.df.columns]
         self.rows = DataFrameAdapter(self.df)
 
     @asyncthread
--- a/visidata/loaders/_pandas.py
+++ b/visidata/loaders/_pandas.py
@@ -24,9 +24,9 @@ class PandasSheet(Sheet):
     def dtype_to_type(self, dtype):
         import numpy as np
         try:
-            if np.issubdtype(dtype, np.int):
+            if np.issubdtype(dtype, np.integer):
                 return int 
-            if np.issubdtype(dtype, np.float):
+            if np.issubdtype(dtype, np.floating):
               return float
             if np.issubdtype(dtype, np.datetime64):
                 return date
@@ -39,16 +39,6 @@ class PandasSheet(Sheet):
         """Partial function for reading TSV files using pd.read_csv"""
         import pandas as pd
         return pd.read_csv(path, sep='\t', **kwargs)
-
-    # @property
-    # def selectedRows(self):
-    #     if len(self._selectedRows) <= 1:
-    #         return list(self._selectedRows.values())
-    #     return [self.rows.iloc[i] for i in range(self.rows.shape[0]) if id(self.rows.iloc[i]) in self._selectedRows]
-
-    # def selectRow(self, row):
-    #     super().selectRow(row)
-    #     raise ValueError(self.selectedRows)
 
     def reload(self):
         import pandas as pd
@@ -82,23 +72,30 @@ class PandasSheet(Sheet):
     def _checkSelectedIndex(self):
         if self._selectedMask.index is not self.df.index:
             # self.df was modified inplace, so the selection
-            # is no longer valid. Likely to heavy but erring
+            # is no longer valid -- just delete it. Maybe too heavy but erring
             # on side of correctness here.
+            vd.status('pd.DataFrame.index updated, clearing selected {} rows'
+                      .format(self._selectedMask.sum()))
             self._selectedMask = pd.Series(False, index=self.df.index)
 
-    # TODO: add vectorized implementation for multiple rows
+    # Base selection API. Refer to GH #266: using id() will not identify
+    # pandas rows since iterating on rows / selecting rows will return
+    # different copies. Instead, re-implement the selection API by
+    # keeping a boolean pd.Series indicating the selected rows.
     def isSelected(self, row):
         if row is None:
             return False
         self._checkSelectedIndex()
         return self._selectedMask.loc[row.name]
     def selectRow(self, row):
-        'Select given row. O(log n)'
+        'Select given row'
         self._checkSelectedIndex()
         self._selectedMask.loc[row.name] = True
     def unselectRow(self, row):
         self._checkSelectedIndex()
+        is_selected = self._selectedMask.loc[row.name]
         self._selectedMask.loc[row.name] = False
+        return is_selected
     @property
     def nSelected(self):
         self._checkSelectedIndex()
@@ -106,7 +103,23 @@ class PandasSheet(Sheet):
     @property
     def selectedRows(self):
         self._checkSelectedIndex()
-        return self.df.loc[self._selectedMask]
+        return DataFrameAdapter(self.df.loc[self._selectedMask])
+    # Vectorized implementation of multi-row selections
+    def select(self, rows, status=True, progress=True):
+        self._checkSelectedIndex()
+        self._selectedMask.loc[[row.name for row in rows]] = True
+    def unselect(self, rows, status=True, progress=True):
+        self._checkSelectedIndex()
+        self._selectedMask.loc[[row.name for row in rows]] = False
+    def selectByIndex(self, start=None, end=None):
+        self._checkSelectedIndex()
+        self._selectedMask.loc[start:end] = True
+    def unselectByIndex(self, start=None, end=None):
+        self._checkSelectedIndex()
+        self._selectedMask.loc[start:end] = False
+    def toggleByIndex(self, start=None, end=None):
+        self._checkSelectedIndex()
+        self._selectedMask.loc[start:end] = ~select._selectedMask
 
 def view_pandas(df):
     run(PandasSheet('', source=df))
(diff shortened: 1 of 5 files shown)


anjakefala:
I really appreciate that you decided to open a WIP PR as the work started out, @azjps, because it helps us be more involved in the process along the way. 

Also, a general thank you for helping us improve the pandas loader. :blush: :blush: 

It is good timing because the Frequency Table / Pivot Table are also undergoing a design overview. It will be interesting to see how the pieces fit together. We will be able to take a more detailed look at the contents of your initial work, soon. =)

azjps:
Glad to assist, do let me know if there's anything I can clarify about this commit since it is pretty exploratory/messy. :)

visidata/loaders/pandas_freqtbl.py
```diff
@@ -0,0 +1,155 @@
...
+        # add default bonus columns
+        for c in [
+                    Column('count', type=int, getter=lambda col,row: len(row.sourcerows)),
+                    Column('percent', type=float, getter=lambda col,row: len(row.sourcerows)*100/df.shape[0]),
+                    Column('histogram', type=str, getter=lambda col,row: options.disp_histogram*(options.disp_histolen*len(row.sourcerows)//value_counts.max()), width=options.disp_histolen+2),
+                    ]:
+            self.addColumn(c)
+
+        # TODO: can make this part async w/ progress bar
+        for element in value_counts.index:
```
saulpw: if `.index` has a len(), then you can use `Progress(value_counts.index)` and that will take care of it.  Otherwise `with Progress(total=n) as prog: ... prog.addProgress(1)`

visidata/loaders/pandas_freqtbl.py
```diff
@@ -0,0 +1,155 @@
...
+        # (1) this only works if len(groupByCols) == 1 since it relies on value_counts()
+        # (2) this assumes some amount of non-degeneracy (non-empty dataframe with non-null values)
+        super().initCols(use_range=False)
+        # columns now the same as the original table
+        # raise ValueError(str([c.name for c in self.columns]))
+
+        df = self.source.rows
+
+        if len(self.groupByCols) > 1:
+            raise NotImplementedError("Frequency Table for > 1 columns currently not supported for PandasSheet")
```
saulpw: I use `error()` or `fail()` for these (both of which raise an ExpectedException).  error() is red; fail() is yellow and less severe.

saulpw:
Hey @azjps, I think you're on the right track here.  A faster and more functional PandasSheet would be appreciated by a bunch of people.

I did not even realize that selection doesn't work at all on pandas sheets. We'll probably have to overload the whole selection mechanism for PandasSheet.  What other selection mechanism would work for Pandas, do you think?

I also just cleaned up the selection API a bit, so at develop HEAD, the functions and properties from Sheet that would need to be overloaded are:

- `selectRow(sheet, row)`
- `unselectRow(sheet, row)`
- `unselectAll(sheet)`
- `isSelected(sheet, row)`: bool
- `selectedRows`: list of selected rows in sheet order
- `nSelected`: int

All other selection features should stem from those functions.

visidata/loaders/pandas_freqtbl.py
azjps: 👍 

I think it should be possible to do multiple `groupByCols` here by either using https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.crosstab.html and `stack()`, or by dropping down to `PivotTable`'s implementation. I'm not sure if there are gotchas to using `pd.crosstab().stack()` but I'll give it a shot.

azjps:
Thanks for the comments, I'll look into cleaning those up when I get a chance.

Re: selection, unfortunately in the general case, I don't believe it is possible to uniquely identify a row of a pandas.DataFrame only given the row itself, as is done in `selectRow(sheet, row)`, namely if the index of the DataFrame contains non-unique entries. Non-unique index elements can be annoying to deal with in other places too though so it may very well be reasonable to require that the index is unique (or even stricter that it must be a default `RangeIndex`, e.g. after calling `reset_index()`). The other alternative I had in mind was to manipulate the row object to pass a bit more information such as the exact position of the row in the DataFrame. Either way I think the idea would be to rely on the row's position as opposed to the row object itself. A potential downside I see from this is selecting might interact poorly with editing (adding / removing rows), although I don't believe editing is supported currently for `PandasSheet`.

saulpw:
> I don't believe it is possible to uniquely identify a row of a pandas.DataFrame only given the row itself

Yeah, this sounds problematic.  VisiData uses id(row) as a really effective and general-purpose optimization strategy that has worked well, except for pandas.  I don't think we can get around the need for something like this (dealing with millions of rows, with potentially millions of rows selected, would otherwise require a substantial rearchitecture), but what if we at least made it easier to experiment?  Like calling a Sheet.id(row) method which by default returns the builtin id(), but in the case of a PandasSheet could return the index or some other annotation on the row?  This would allow the API to stay the same and require only a minimum change for other Sheets which can't use the builtin id.  I think your potential solutions would work within this framework, should we give it a try?

azjps:
Yeah that might work, I have a similar idea in mind that I'll give a shot implementing.

On an unrelated note, is there some way to view output logging or some other form of debugging? I've been debugging by "logging" the information I want in an exception message and using Ctrl+E to see it, but its a pretty inefficient workflow especially since I'm still pretty new to the code.

saulpw:
Ah, that's a great question. Traditional printf-style debugging is more difficult in VisiData, it's true; I usually use `status()` (or `fail()` to do status-with-raise), which has a couple of nice features like msg dedup and a more efficient history with `Ctrl+P`.  I only use `Ctrl+E` for internal errors when I need to see the backtrace.

For development these days I get a lot more mileage out of `Ctrl+Y`, which lets you inspect the row object itself (or sheet object with `g Ctrl+Y`), and then you can drill down from there.  I'm thinking that might be what you're looking for.

Also, if I'm debugging something, I tend to create a .vd script and replay whatever commands I'm doing (whether the whole test or just the setup).  Then I have two windows, one for editor and one for visidata, and in the vd window I can rerun the test with `gq Up Enter` and it's a really quick turnaround time.

visidata/loaders/_pandas.py
```diff
@@ -75,6 +79,34 @@ def sort(self):
...
+        self._checkSelectedIndex()
+        self._selectedMask.loc[row.name] = False
+    @property
+    def nSelected(self):
+        self._checkSelectedIndex()
+        return self._selectedMask.sum()
+    @property
+    def selectedRows(self):
+        self._checkSelectedIndex()
+        return self.df.loc[self._selectedMask]
```
azjps: `selectedRows` is probably wrong here, the result doesn't have the right iterable output. Probably need to wrap in a DataFrameAdapter

visidata/loaders/pandas_freqtbl.py
```diff
@@ -92,60 +92,76 @@ def unselectRow(self, row):
...
+            this_column = df.loc[:, str(self.groupByCols[0].name)]
+            value_counts = this_column.value_counts()
+        elif len(self.groupByCols) >= 1:
+            this_column = df.loc[:, str(self.groupByCols[0].name)]
+            # Compute cross-tabulation to get counts, and sort/remove zero-entries.
+            # Note that this is not space-efficient: the initial cross-tabulation will
+            # have space on the order of product of number of unique elements for each
+            # column, even though its possible the combinations present are sparse
+            # and most combinations have zero count.
+            value_counts = pd.crosstab(this_column, [df.df[c.name] for c in self.groupByCols[1:]])
```
azjps: on second thought, it will likely make much more sense to use `pandas.pivot_table` for this. Possibly can even use to implement `PivotSheet`, albeit ambitious

visidata/loaders/_pandas.py
saulpw: ah yeah, or a different kind of adapter.  One thing to keep in mind, is that selectedRows should return a list of rows in the same order as the sheet itself.  This is for things like `dup-sheet` (`"`) to feel right. I think your approach already does this, but I wanted to make sure it's explicit.

saulpw:
Hey @azjps, how are you doing with this PR?  I was thinking, maybe it would be useful if we made a time to work on this together.  We could have a shared tmate session and either voice or text chat going at the same time.  With your pandas skills and my VisiData knowledge we might be able to get it working in fairly short order.  What do you think?

azjps:
Sounds like a good idea, I'll see when I can pull out some time to do such (will need to figure out if I can get tmate working on WSL first 😅 ). For this PR I still want to try changing crosstab() to pivot_table(), but other then that I'm pretty happy with it as far as my local testing goes. It would be good to get some unit tests added too which I'm not sure how to go about, as well as maybe trying a larger dataset for some quick benchmarks (which doesn't need to be committed, could just be some random data generated by a script).

visidata/loaders/_pandas.py
```diff
@@ -19,25 +21,43 @@ def __getattr__(self, k):
...
 # source=DataFrame
 class PandasSheet(Sheet):
+    def dtype_to_type(self, dtype):
+        import numpy as np
+        try:
+            if np.issubdtype(dtype, np.integer):
+                return int 
+            if np.issubdtype(dtype, np.floating):
+              return float
+            if np.issubdtype(dtype, np.datetime64):
```
saulpw, line 43: does this work with datetime columns?  I'm not sure if I commented it out because it was the wrong mechanism (thanks for showing me the right idiom, btw), or because visidata.date and np.datetime64 aren't compatible.

visidata/loaders/_pandas.py
```diff
@@ -19,25 +21,43 @@ def __getattr__(self, k):
...
             filetype = getattr(self, 'filetype', self.source.ext[1:])
-            readfunc = getattr(pandas, 'read_'+filetype) or error('no pandas.read_'+filetype)
+            if filetype == 'tsv':
+                readfunc = self.read_tsv 
+            else:
+                readfunc = getattr(pd, 'read_'+filetype) or error('no pandas.read_'+filetype)
             self.df = readfunc(self.source.resolve(), **options('pandas_'+filetype+'_'))
 
-        self.columns = [ColumnItem(col, type=dtypeToType(self.df, col)) for col in self.df.columns]
+        # TODO: should we reset the index here and add it as a key column?
```
saulpw, line 67: Ah, this is exactly what @danlat was suggesting in his PR.  I'm not so familiar with pandas DF and its index, but it sounds like it's something that pandas users want/need access to.  So let's do that (maybe merging them both in separately).

visidata/loaders/_pandas.py
```diff
@@ -49,6 +69,57 @@ def sort(self):
             ascending += [not reverse] * len(cols)
         self.rows.sort_values(by=by_cols, ascending=ascending, inplace=True)
 
+    def _checkSelectedIndex(self):
+        if self._selectedMask.index is not self.df.index:
+            # self.df was modified inplace, so the selection
+            # is no longer valid -- just delete it. Maybe too heavy but erring
+            # on side of correctness here.
+            vd.status('pd.DataFrame.index updated, clearing selected {} rows'
+                      .format(self._selectedMask.sum()))
```
saulpw, line 95: Let's use `.nSelected` as you've redefined below. and a minor text change, to `{} selected rows`.

visidata/loaders/_pandas.py
```diff
@@ -61,3 +132,19 @@ def open_dta(p):
     return PandasSheet(p.name, source=p, filetype='stata')
 
 open_stata = open_pandas
+
+undoSheetSelection = undoAttrCopy('[sheet]', '_selectedMask')
+
+# Override with vectorized implementations
+PandasSheet.addCommand('gt', 'stoggle-rows', 'toggleByIndex()', undo=undoSheetSelection),
```
saulpw: I *think* these command redefinitions should be `None` for the keybinding, so that if someone binds `gt` globally to another longname it won't get rebound here on the PandasSheet.  But we'd have to make sure that `gt` etc still properly bind to the longname on the PandasSheet, of course (and I'll have to fix the keybinding code if they don't).

saulpw (commented):
I only had a few comments; I think if this is working well for you, let's clean it up (mostly removing commented out code and TODO comments) and merge it in and see how it works for other people.

Thanks so much for doing this, I don't have the experience with pandas to do it well, and I'm glad that the pandas loader is getting some love!

visidata/loaders/_pandas.py
danlat: It's similar to a database index column in that it's meant to be sorted and unique for quick access and querying, but pandas does not include the index in DataFrame.columns (from the list comprehension in this code)

visidata/loaders/_pandas.py
azjps: It does work with datetime columns, I work quite a bit with time series and have used them in visidata before 😄

visidata/loaders/_pandas.py
azjps: It would lead to a cyclic reference here, since `nSelected` is calling `_checkSelectedIndex` itself, although the calling of `_checkSelectedIndex` is arguably overly defensive.

visidata/loaders/_pandas.py
azjps: Done, I haven't looked at the keybinding code but setting to `None` seems to work fine for what I've tried.

azjps:
I've left some code comments in the hope that they will be instructive, since I wasn't 100% sure about some implementation choices I made. Besides that, I'm pretty satisfied with the state of these commits. Here's the snippet I used for testing / performance timings; it creates a 1 million row dataframe with various column dtypes. I wasn't sure where in the repository to include it, if at all:
```python
#!/usr/bin/env python3
"""
Run visidata PandasSheet on a dataframe with columns
of different datatypes and a length controlled by
the command-line option --repeat. The default size
of the loaded dataframe is 1 million rows.
"""

import argparse

import pandas as pd
import numpy as np

import visidata

visidata.options.disp_date_fmt = "%Y%m%d %H:%M:%S.%f"

if __name__ == "__main__":
    group_count = 10
    parser = argparse.ArgumentParser()
    parser.add_argument(
        "--repeat", default=100000, type=int,
        help="Number of times to repeat rows, total # of rows = {} * repeat. "
             "Default=100000 for a million rows total."
             .format(group_count)
    )
    args = parser.parse_args()

    repeat_count = args.repeat
    df = pd.DataFrame({
        "date1": pd.date_range(start="2020/01/01", freq="D", periods=group_count),
        "date2": pd.date_range(start="2020/01/01 02:15:30.012345", freq="D", periods=group_count),
        "int1": np.arange(group_count),
        "float1": np.arange(group_count) + 0.5,
        "cat1": [str(i) for i in np.arange(group_count)],
        "repeat1": np.repeat(1, group_count),
    })
    df["cat1"] = df["cat1"].astype("category")
    df["timedelta1"] = df["date1"].diff()
    # Repeat so that we are a significant amount of rows
    big_df = df.iloc[np.repeat(np.arange(0, group_count), repeat_count)]
    big_df = big_df.copy().reset_index(drop=True)
    # To make the frequencies a bit more interesting, shift the int2 & date2 columns 20% "upwards"
    big_df["int2"] = big_df["int1"].shift(-repeat_count // 5)
    big_df["date2"] = big_df["date2"].shift(-repeat_count // 5)
    big_df["int3"] = big_df["int1"].shift(repeat_count // 5)

    visidata.run(visidata.PandasSheet('pandas', source=big_df))
```

anjakefala:
Hi @azjps! 

# Your PR
Thank you so much for all of the work you have done on improving the pandas loader. As we were testing it, we discovered some performance issues which we are going to look into fixing after we merge. Once we make them, we will leave a comment outlining them, so you can see what we have done.

# Testing
Thanks so much for the testing script. 

We mostly have functional tests with `.vd` scripts. We are going to incorporate your python script into this setup, because testing on a pandas Dataframe would be super useful.

PR #269, opened 2019-03-28
Remove extra copy of man page.

QuLogic:
It's already installed in `/usr/share/man/man1`. This should also handle the case of a user install and an editable install, but I only tested the first and the last.


QuLogic:
Hmm, actually, this is not perfect as manpages in packages like on Debian/Fedora get compressed.

anjakefala:
Hi @QuLogic! I do appreciate you taking the time to make these PRs. 

It is quite likely that I am going to not merge this one in. I want to give you an opportunity to explain it, and change my mind. =) If that sounds like a time sink that you do not want to invest in, for a PR that will possibly not be included, I respect that and you are free to just close it.

Part of my nervousness is that VisiData is installed through various [means](https://visidata.org/install) on Mac and Linux machines. Currently, man page installation works reasonably well, and I am hesitant to accept a PR which changes how that works without understanding the benefits behind that change.

I suspect that you are right that the packages_data section is redundant, since it is part of MANIFEST.in. (Is that where its redundancy lies?) I am unsure what the goal is behind your other changes. Can you please expand?

anjakefala:
Hi @QuLogic, I am going to go ahead and close this PR for now. 

If you would like to reopen it with an overview with the benefits and intent behind this change, along with an enumeration of possible installation risks if there are any, you are more than welcome to!

QuLogic:
Sorry, I never seemed to have commented on this. The man page in `package_data` is not redundant with `MANIFEST.in`, but with the entry in `data_files`.

The rest of the code changes find the man page in the location installed by `data_files`, running through possible global install, user install, and editable install locations.

PR #272, opened 2019-04-02
Allow saving pcaps to json

layertwo:
Ran into an issue where saving PCAPs to JSON or JSONL didn't want to work properly. CSV or TSV was typically fine.

The solution I found was to define the column types before saving. This is now done in the loader.

### DIFF
--- a/visidata/loaders/pcap.py
+++ b/visidata/loaders/pcap.py
@@ -234,9 +234,9 @@ class EtherSheet(Sheet):
     rowtype = 'packets'
     columns = [
         ColumnAttr('timestamp', type=date, fmtstr="%H:%M:%S.%f"),
-        Column('ether_manuf', getter=lambda col,row: mac_manuf(macaddr(row.src))),
-        Column('ether_src', getter=lambda col,row: macaddr(row.src), width=6),
-        Column('ether_dst', getter=lambda col,row: macaddr(row.dst), width=6),
+        Column('ether_manuf', type=str, getter=lambda col,row: mac_manuf(macaddr(row.src))),
+        Column('ether_src', type=str, getter=lambda col,row: macaddr(row.src), width=6),
+        Column('ether_dst', type=str, getter=lambda col,row: macaddr(row.dst), width=6),
         ColumnAttr('ether_data', 'data', type=vlen, width=0),
     ]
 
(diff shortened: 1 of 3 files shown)


visidata/loaders/pcap.py
```diff
@@ -245,18 +245,18 @@ class IPSheet(Sheet):
...
+        ColumnAttr('ip_hdrlen', 'ip.hl', type=int, width=0, helpstr="IPv4 Header Length"),
         ColumnAttr('ip_proto', 'ip.p', type=lambda v: protocols['ip'].get(v), width=8, helpstr="IPv4 Protocol"),
-        ColumnAttr('ip_id', 'ip.id', width=0, helpstr="IPv4 Identification"),
-        ColumnAttr('ip_rf', 'ip.rf', width=0, helpstr="IPv4 Reserved Flag (Evil Bit)"),
-        ColumnAttr('ip_df', 'ip.df', width=0, helpstr="IPv4 Don't Fragment flag"),
-        ColumnAttr('ip_mf', 'ip.mf', width=0, helpstr="IPv4 More Fragments flag"),
-        ColumnAttr('ip_tos', 'ip.tos', width=0, type=FlagGetter('ip_tos'), helpstr="IPv4 Type of Service"),
-        ColumnAttr('ip_ttl', 'ip.ttl', width=0, helpstr="IPv4 Time To Live"),
-        ColumnAttr('ip_ver', 'ip.v', width=0, helpstr="IPv4 Version"),
+        ColumnAttr('ip_id', 'ip.id', type=int, width=10, helpstr="IPv4 Identification"),
```
saulpw, line 253: The typing all makes sense.  (I'm a bit surprised that we need to declare them `str`, but if it works, it's fine).  Do we want to made these unhidden by default though?  That seems cluttered, and you can expose all hidden columns with `gv`.  Though I will defer to you on this, since you are likely the primary consumer of pcap files in VisiData =)

visidata/loaders/pcap.py
```diff
@@ -300,21 +300,21 @@ class PcapSheet(Sheet):
...
+        Column('dstmanuf', type=str, getter=lambda col,row: manuf(macaddr(row.dst))),
+        Column('dsthost', type=str, getter=lambda col,row: row.dsthost),
         Column('dstport', type=int, getter=lambda col,row: get_port(row, 'dport')),
         ColumnAttr('ether_proto', 'type', type=lambda v: protocols['ethernet'].get(v), width=0),
         ColumnAttr('tcp_flags', 'ip.tcp.flags', type=FlagGetter('tcp'), helpstr="TCP Flags"),
-        Column('service', getter=lambda col,row: getService(getTuple(row))),
+        Column('service', type=str, getter=lambda col,row: getService(getTuple(row))),
         ColumnAttr('data', type=vlen),
-        ColumnAttr('ip.len', type=int),
+        ColumnAttr('ip.len', 'ip_len', type=int),
```
saulpw: Seems like these should be reversed (such that `ip_len` is the name of the column, and `ip.len` is the deep attr to get).

saulpw:
Thanks for the PR, @layertwo!  We'll definitely merge this; let me know what you think about the couple comments I had and if anything needs to be changed.

visidata/loaders/pcap.py
layertwo: Correct, will fix that.

visidata/loaders/pcap.py
layertwo: I think by default, I'd rather show them. Didn't realize it was all there until I reviewed source and hidden columns. If someone doesn't want those columns, they're welcome to hide them.

visidata/man/vd.inc
```diff
@@ -770,7 +770,7 @@ These are the supported sources:
 .El
 .Pp
 .Bl -inset -compact -offset xxx
-.It Sy pcap No ( requires Sy xpkt Ns , Sy dnslib Ns )
+.It Sy pcap No ( requires Sy dpkt Ns , Sy dnslib Ns )
```
layertwo, line 773: Package is actually **dpkt** not **xpkt**

PR #277, opened 2019-04-27
[pandas] Include DataFrame index by moving to columns

danlat:
Visidata is awesome for doing some quick browsing of parquet files using pandas, but wasn't including the index. This adds the index as long as it's not the default RangeIndex (0-n)

### DIFF
--- a/visidata/loaders/_pandas.py
+++ b/visidata/loaders/_pandas.py
@@ -36,6 +36,9 @@ class PandasSheet(Sheet):
             readfunc = getattr(pandas, 'read_'+filetype) or error('no pandas.read_'+filetype)
             self.df = readfunc(self.source.resolve(), **options('pandas_'+filetype+'_'))
 
+        if type(self.df.index) is not pandas.RangeIndex:
+            self.df = self.df.reset_index()
+
         self.columns = [ColumnItem(col, type=dtypeToType(self.df, col)) for col in self.df.columns]
         self.rows = DataFrameAdapter(self.df)
 


saulpw (commented):
Thanks for this PR, @danlat.  If we make a new DF, will edits or other changes be propagated back to the origina DF?  (i.e. should this have inplace=True?)  Also @azjps, does this interact with your changes at all?

danlat:
The original DF will be gone and self.df will reference our new one with the index reset to columns.

anjakefala:
Hi @danlat! Would you mind opening this PR again on the `develop` branch? See [here](https://github.com/saulpw/visidata/blob/develop/CONTRIBUTING.md#submitting-code). Once you do, I think this looks good.

azjps:
> Thanks for this PR, @danlat. If we make a new DF, will edits or other changes be propagated back to the origina DF? (i.e. should this have inplace=True?) Also @azjps, does this interact with your changes at all?

@saulpw no, it won't be conflict with my PR. I do think maybe there could be an option somewhere to toggle what to do with the index -- as you mentioned it could cause some confusion with any editing features, and another possibility which is probably most aesthetically appropriate is to set the index columns as keys.

anjakefala:
@azjps
```
as you mentioned it could cause some confusion with any editing features
```
Thanks for drawing my attention to this.

```
another possibility which is probably most aesthetically appropriate is to set the index columns as keys.
```
:purple_heart:

PR #288, opened 2019-05-28
Windows support?

scholer:
Hi,

I usually use vd from WSL on Windows. However, I sometimes find myself in a regular Windows "Command Prompt", typing `vd` to open e.g. a tsv file. This obviously fails, because "visidata cannot be installed on Windows". But I figured I would give it a go to see how much I would need to change visidata to be able to open a tsv file inside a Windows Command Prompt. Turns out, not much change is actually needed. 

This commit enables rudimentary Windows support (using e.g. `windows-curses` package). The changes involves the following:

* Using `entry_points` instead of `scripts` in `setup.py` (changed 1 line of code)
* Converted `vd` script to a entrypoint-compatible module/function. (added 2 lines of code) 
* Catching/ignoring Windows-specific curses error in colors.py. (Added a try...except clause)
* Using 32-bit csv.field_size_limit on Windows platforms in loaders/csv.py. (changed 1 line of code)
* Catching import errors for modules that are missing on Windows in shell.py (Added a try...except clause).

I'm not sure if this is actually something worth merging - adding windows-specific code obviously adds a little bit of clutter (even though it is only a small hand-full of lines). Also, this is just to get to the point of being able to open and navigate a tsv file, and browsing the filesystem using `vd`. There may be further changes needed to make e.g. editing work.

Anyways, I did the change for myself, and figured maybe other Windows-people would be interested as well. Even if you don't want to merge, at least windows-people can look at this PR and add experimental windows support themselves.


aborruso:
Great idea, thank you

saulpw:
Thanks for putting this together, @scholer!  This will make some people very happy.  We'll review it and make sure installation and everything still works properly on Linux/Mac and then take the changes for 1.6.

One question though, how might we create a VisiData installer for Windows?  Ideally someone could download an .exe (or .msi, I don't know how Windows apps are packaged these days), and click on it to install Python and vd and everything they need so that it just works.  Do you know if there's a reasonably easy way to do this?

saulpw:
We dug into this PR today, and for a couple reasons we decided to port the changes manually (giving you credit of course).  However, it turns out that the wrapper script that is installed via `entry_points` adds about 250ms of time to startup, which would be double the current time.  The difference may seem minor but it is absolutely perceptible (I noticed it anyway) and adds unnecessary friction.  So we're going to install the slower wrapper script as `visidata` and Windows users can create an alias for `vd` if they want.   Also sometimes new users install visidata the package and then try to run it as `visidata` and this solves that problem too. Thanks for doing the legwork on this!  It is much appreciated.

anjakefala:
Commits are https://github.com/saulpw/visidata/commit/c8dba50d52ed46fe08b2e9cb4fd90dc77b5c9c13, https://github.com/saulpw/visidata/commit/2a11d5c0f83ecd7d122c7cd39c000b9688a4f15f, and https://github.com/saulpw/visidata/commit/de29366c2860136496819594ef3b80510210467a =)

jcmkk3:
I'm really looking forward to this Windows support. Thanks! As for how to distribute to Windows users, the trend that I've been seeing is using Scoop for these sorts of command line tools. I would love to be able to install it through scoop. https://scoop.sh/

anjakefala:
@jcmkk3 Can you open an investigation issue for scoop? This way your suggestion will not be lost. =) Do you have experience with creating Scoop packages?

PR #294, opened 2019-06-05
Accept ndjson and ldjson as aliases for jsonl

tsibley:


saulpw:
Thanks for this, @tsibley!  I had never heard of ndjson and ldjson before, but if they exist anywhere in the wild, these are good aliases for vd to have.  Also thank you for keeping the manpage and other dev tools up-to-date.

tsibley:
Thanks for merging! :-)

PR #299, opened 2019-06-20
Add JIRA compatible table save

layertwo:
I commonly use `vd` for a task list or status board of items I am working on. Since I'm also typically doing some project tracking in JIRA or Confluence, it would be nice to take the data I have and write it out to a JIRA table.

The JIRA loader is pretty much the same as the Markdown loader just with a different header column style.

Adding the file extension `.jira` will save the file out as a JIRA compatible table (Confluence differentiates Confluence Wiki from markdown).

Bonus fix: I've also changed the default markdown header row indicator from `+` to `|` which seems much more common and is the Github markdown table style.


saulpw:
Thanks for doing this, @layertwo!  Though, `save_jira` seems almost identical to `save_md`, just the header separator and the missing second header row.  Maybe we could merge the code?  I'm thinking either a `markdown_jira` (True/False) option, or a more general `markdown_style` (`jira` and `orgmode` would be the current values, with others possible in the future).  If you really want to have a `jira` filetype, let's try to find a way to merge the code, anyway.

Thanks also for noticing the more prolific `|` separator, that's a better default I agree.

layertwo:
Yeah, I agree. Using the same markdown code makes the most sense. I've updated to reflect that, and can be extended if there's other markdown styles to support (not sure if there are?).

Also, screwed up the commit history with `git commit --amend`. You'll probably want to squash when you merge in.

PR #302, opened 2019-06-21
Add support for opening tar files

layertwo:
Still learning how to add new loaders, so here's an archive loader (includes the older `zip`). Supports .tar and compressed variants (.tgz, .txz, .tbz2).

Enjoy!


saulpw:
Thanks @layertwo, this is great!  I think you're definitely getting the hang of it.  A few minor changes:

- Move the .columns assignment into reload(); assigning columns in init is deprecated.
   - then even the `__init__` can be elided; open_zip can become `return read_archive(p.name, source=p, archive_type='zip')`.  This follows the predominant open_ pattern of "return FooSheet(p.name, source=p)" which I'm trying to standardize on for 2.0.
- Include a rowdef comment (type information of row contents) close to the rowtype.
- Please rename read_archive to ArchiveSheet (I've been meaning to clean this up for awhile myself).

Otherwise, though, spot on.  And keep 'em coming!  You get 10 points for every loader, and a VisiData sticker if you want one :)

visidata/loaders/archive.py
```diff
@@ -0,0 +1,73 @@
...
+                ColumnAttr('name'),
+                ColumnAttr('size', type=int),
+                ColumnAttr('mtime', type=date),
+                ColumnAttr('type', type=int),
+                ColumnAttr('mode', type=int),
+                ColumnAttr('uname'),
+                ColumnAttr('gname')
+            ]
+
+    def reload(self):
```
saulpw: also, now that I'm looking at this, we should try this with a very large tarfile.  I'm thinking we want the reload to be `@asyncthread` at minimum, and then start rows with an empty list and iterate over getmembers, calling addRow on each element.  The zip loader should really be doing the same thing, but it's less impactful because the .zip has a directory at the end of the file, and .tar.gz I believe has to be read linearly.

layertwo:
Thanks, @saulpw. Spent some time working on this. The progress for async row adding is based on the `csv` loader. 

Pretty sure I understand all that you had requested. Let me know if there are any other nitpicks you have.

And yes, would abolustely love a `vd` sticker.

visidata/loaders/archive.py
layertwo: Sound be taken care of in the last few commits.

visidata/loaders/archive.py
```diff
@@ -0,0 +1,89 @@
...
+            ]
+
+        self.recalc()
+        if self.archive_type == 'zip':
+            contents = zipfile.ZipFile(self.source.resolve(), 'r').infolist()
+
+        if self.archive_type == 'tar':
+            contents = tarfile.open(name=self.source.resolve()).getmembers()
+
+        with Progress(total=len(contents)) as prog:
```
saulpw: Heh, this is actually a lot simpler:
```
for row in Progress(contents):
    self.addRow(row)
```

Everything else is handled for you.  The csv loader is more complicated because the number of rows isn't known beforehand, and the number of header rows can be variable.  Here, though, the `contents` has a proper `len` and there is no header row (we specify the columns manually).

saulpw (approved):

Issue #297, opened 2019-06-14
Add the ability to place sparklines in the cells of a column.


frosencrantz:
Visidata allows you to graph results in a separate canvas, but it would be useful if it could condense the values of several columns to single column with a sparkline.  The sparkline would combine the values of the other columns.

 I have some data with multiple columns containing 0 or 1 (N or Y).  With the headers and the line separators it can be hard to get a feel for where different rows match on particular column or are different.  It seems like this could provide a visual map into the other columns.

(There is a new plugin suggestion in my future...)

saulpw:
This is an awesome idea, @frosencrantz.  If you can write `def sparkline(values)` which takes a list of numeric values and returns a unicode braille sparkline, then we can make a nice VisiData plugin to bring it all together.  I'm thinking a command `addcol-sparkline` on the columns sheet that would add a sparkline column from the selected columns (maybe override `&`).  Or on the source sheet, from the visible numeric columns.  It would be a fun little feature.

frosencrantz:
I won't be able to provide any code.  I'd be willing to provide feedback.  I have data that this feature would be useful for.

anjakefala:
Hi @frosencrantz! We currently have quite a few things we are working on, and have to prioritise features where we get collaborative help over ones where we do not. 

This is a really cool feature though, and if anyone is willing to provide the sparkline function code, we can jump right on it. 

Some possibly helper material: https://shapr.github.io/posts/2019-06-16-tmux-sparklines.html

layertwo:
Hi @frosencrantz, I have a first implementation for a sparkline column in [8be3612](https://github.com/layertwo/visidata/commit/8be361231bbffa8dc3ed0a04e7bc551f194302ed) if you would like to review.

If you'd also like to test it out and don't want to run the custom branch, I can also provide as a `vd` plugin.

frosencrantz:
Hi @layertwo I tried the latest version as a plugin, but it was not working for me. There are three issues:

1. There was a missing addGlobals() call to export the SparklineColumn, 
2. There is a missing self. before sparkline on line 9.
3. An issue I haven't tracked down, type issues on line 14,  at least one value in values is a SparklineColumn, when it should be a number. 

[...]

PR #305, opened 2019-06-23
Sparklines

layertwo:
Sparklines in response to #297.


visidata/sparkline.py
```diff
@@ -0,0 +1,29 @@
+from visidata import *
```
anjakefala: We try to avoid wildcard imports where possible. =) (The exception being __init__)

Can you please change this to `from visidata import Sheet, Column`?

visidata/sparkline.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class SparklineColumn(Column):
+    def __init__(self, name, **kwargs):
+        super().__init__(name, **kwargs)
+
+    def calcValue(self, rows):
```
anjakefala: For @saulpw, I noticed that usually our `calcValue`s takes a single `row` (not `rows`). Is that important to keep consistent?

visidata/sparkline.py
```diff
@@ -0,0 +1,29 @@
...
+        return self.sparkline(*rows)
+
+    def sparkline(self, *values):
+        lines = "▁▂▃▄▅▆▇█"
+        values = [int(v) for v in values]
```
anjakefala: From testing, this works really well for sheets where every column is an `int`. =)

This may not be worth polishing for the merge (it does not seem to be a requirement in the original issue), but this is a direction for further development of this feature.

Atm, if any column, regardless if it is hidden or not, cannot be coerced to integer the sparklines column will error out.

Ways we may consider handling this: 
* have the sparklines only be drawn for `visibleCols` (Columns that have not been hidden). 
* have the sparklines only be drawn only for the integer columns in the current sheet. (See `plot` for examples of this.)

anjakefala (approved):

anjakefala:
I left one comment! Once that is fixed, we are good to merge. =) Thank you so much for taking this on!

layertwo:
Ok, that's been updated accordingly. Had a few other fixes I put in now that this is a vd plugin.

anjakefala:
Thanks @layertwo! Looks good for merging. :blush:

Issue #286, opened 2019-05-27
Add immediate filtering like a fuzzy finder


frosencrantz:
Tools like [fzf](https://github.com/junegunn/fzf) filter their input immediately.  It would be useful if visidata had a feature where you could filter the rows of a table faster than you can currently with the select ('g|' or '|')/filter ('"') workflow available now.

The immediate filtering could have the same options (e.g. current column, visible rows) that are used now for select/unselect commands. You wouldn't need to hit enter to see the results of your query or have to hit the filter button to reduce the table to the matching/selected rows.  Speeding up how fast can search your data.

saulpw:
Thanks for the request, @frosencrantz.  I wrote a little plugin to do this, and I'm thinking about how best to incorporate it.

frosencrantz:
Great.  What's the best way to try it? How do you load plugins? I tried to import from visidatarc file.

anjakefala:
@frosencrantz Our documentation on manually installing plugins is here =) https://github.com/saulpw/visidata/blob/develop/docs/plugins.md#manually

(The automated installation method has not been shipped yet.)

Let me know if you still have trouble!

In summary:

```
Move the plugin module into ~/.visidata.
Ensure its dependencies are installed.
Import the module within ~/.visidatarc.
```

anjakefala:
On second thought @frosencrantz  these instructions are too vague. I only just caught up on what this issue entails.

Here is the specific how to:
1. Install VisiData from the `develop` branch `pip3 install git+https://github.com/saulpw/visidata.git@develop` (welcome to the cutting edge of VisiData =)
2. `mkdir -p ~/.visidata/plugins`
3. Copy [plugins/livesearch.py](https://raw.githubusercontent.com/saulpw/visidata/develop/plugins/livesearch.py) into `~/.visidata/plugins`
4. Add `import plugins.livesearch` to `~/.visidatarc`

Currently, the instasearch commands are mapped to `Alt-s` and `gAlt-s`. You would then press `Enter` to open a duplicate sheet with those rows. If you would like to change those keybindings, open up `~/.visidata/livesearch.py` and locate the `Sheet.addCommand`s at the bottom. Those are the modules that create the actual commands. The keystrokes are the first argument. You can change them to `/` and `g/`, or anything else that you like. =)

There may be a couple of bugs in the display. If you have any improvements for the plugin, we would be happy to take them!

aborruso:
[...]

PR #308, opened 2019-07-07
Update notes about installing a plugin

frosencrantz:
Based on comments from @anjakefala in https://github.com/saulpw/visidata/issues/286#issuecomment-500272731


anjakefala:
Awesome, thanks @frosencrantz! :blush: This also reminded me that we need to re-think how to handle dependencies and incorporate that into the manual written strategy.

Issue #333, opened 2019-08-16
[loader] Support for S3 paths


ajkerrigan:
Hi there, I've been using visidata for local files recently and loving it! Would you be open to having an S3 loader so objects could be opened using S3 paths directly?

I've done some local fiddling inspired by the http loader, and taking an optional extra dependency on [s3fs](https://github.com/dask/s3fs/). If this seems like something you'd actually want in the product I'd like to talk it through, get some feedback and try to help make it happen.

Thanks for a great tool!

saulpw:
Hey @ajkerrigan-mdsol, an s3 transport/loader would be pretty great.  I would use it myself.  Have you been working with 1.5.2 or the develop branch?  You should be able to package it as a plugin, and we can add it to the plugins sheet that will be in 2.x.

ajkerrigan:
Hey @saulpw, thanks for the reply! I was working off 1.5.2 (by accident really, I was just poking around before opening an issue and got carried away). Getting a plugin together sounds like the right call and I see some good examples there, thanks for the nudge :).

saulpw:
For sure!  I'd love to help any way I can. We're on Freenode #visidata if you do the IRC thing, or you can email me at `vd@saul.pw` if you want to chat a different way.

ajkerrigan:
Great! I've got a minimally viable plugin started at https://github.com/ajkerrigan/visidata-plugins, trying to follow cues from https://github.com/jsvine/visidata-plugins and deviate from the core VisiData bits as little as possible. I'll catch up with you through freenode/mail to chat :).

saulpw:
This is great, @ajkerrigan!  We'll add it to our plugins.tsv whenever it's ready for wider distribution.

PR #351, opened 2019-09-07
[plugins] add vds3

ajkerrigan:
Initial support for browsing S3 paths and read-only access to objects.

This should be a good first step toward addressing #333


ajkerrigan:
I realized that I've been working off `develop`, and `v2.-1` doesn't work with this plugin because of some path changes. I've put the required version as `v2.-2`... which isn't released yet. I'm not sure if this PR should hang open until `v2.-2` gets cut or if there's a better way to handle it.

saulpw:
Hm, we're planning to release v2.-2 next weekend (~9/14), not sure what the best strategy is.  Always tricky to be developing with an unstable API :)  but it should be getting more stable with every pre-2.0 release.

ajkerrigan:
Hi @saulpw - I've been using this plugin locally with recent pre-release visidata versions, and it's been behaving pretty well. I cut a new release and updated this PR to match the updated plugins.tsv format with the explicit version link and file hash. Can you take a peek when you get a chance, and see if it's ripe enough for inclusion in the plugin sheet?

saulpw (approved):

PR #355, opened 2019-09-10
Minor plugin management tweaks

ajkerrigan:
I ran into a couple issues when testing out plugin installs/uninstalls:

- 53d66fc fixes a couple minor bugs during uninstalls
- 90b2d5e handles pip installs more smoothly in cases where the `pip3` binary doesn't target VisiData's pip environment (in my case, due to [pipx](https://github.com/pipxproject/pipx))

### DIFF
--- a/visidata/plugins.py
+++ b/visidata/plugins.py
@@ -62,7 +62,7 @@ class PluginsSheet(TsvSheet):
                 outfp.write(pyfp.read())
 
         if plugin.requirements:
-            p = subprocess.Popen(['pip3', 'install']+plugin.requirements.split())
+            p = subprocess.Popen([sys.executable, '-m', 'pip', 'install']+plugin.requirements.split())
             status(tuple(p.communicate()))
 
         with Path(options.config).open_text(mode='a') as fprc:
--- a/visidata/plugins.py
+++ b/visidata/plugins.py
@@ -75,10 +75,10 @@ class PluginsSheet(TsvSheet):
 
     def removePlugin(self, plugin):
         vdrc = Path(options.config)
-        oldvdrc = vdrc+'.bak'
+        oldvdrc = vdrc.with_suffix('.bak')
         try:
             shutil.copyfile(vdrc, oldvdrc)
-            vdrc_contents = Path(oldvdrc).read_text().replace('\n'+_plugin_import(plugin), '')
+            vdrc_contents = oldvdrc.read_text().replace(_plugin_import(plugin)+'\n', '')
 
             with Path(options.config).open_text(mode='w') as fprc:  # replace without import line
                 fprc.write(vdrc_contents)


visidata/plugins.py
```diff
@@ -75,10 +75,10 @@ def removePluginIfExists(self, plugin):
 
     def removePlugin(self, plugin):
         vdrc = Path(options.config)
-        oldvdrc = vdrc+'.bak'
+        oldvdrc = vdrc.with_suffix('.bak')
         try:
             shutil.copyfile(vdrc, oldvdrc)
-            vdrc_contents = Path(oldvdrc).read_text().replace('\n'+_plugin_import(plugin), '')
+            vdrc_contents = oldvdrc.read_text().replace(_plugin_import(plugin)+'\n', '')
```
saulpw, line 81: The newline was put before the plugin import, to make sure we were removing an actual import (and not e.g. commented out), and also to allow comments following the import.

visidata/plugins.py
```diff
@@ -75,10 +75,10 @@ def removePluginIfExists(self, plugin):
 
     def removePlugin(self, plugin):
         vdrc = Path(options.config)
-        oldvdrc = vdrc+'.bak'
+        oldvdrc = vdrc.with_suffix('.bak')
```
saulpw, line 78: Does this result in `.visidatarc.bak` or `.bak`?

saulpw (commented):
Thanks for these tweaks, @ajkerrigan!  I had a couple of comments/questions before we can merge.

visidata/plugins.py
ajkerrigan: Ahhh good call. Would it be more reliable then maybe to read all lines as an iterable, then write back only lines that start with the import line? Seems like that would address both concerns, but might cause another problem I'm not considering.

visidata/plugins.py
ajkerrigan: That leaves you with `.visidatarc.bak`.

ajkerrigan: Although... if it were possible to have a config file like `~/my.visidatarc` then it would turn it into `~/my.bak`. We could dodge that with `vdrc.with_suffix(vdrc.suffix + '.bak')` but I'm not sure if that's a good or bad idea. Thoughts?

visidata/plugins.py
saulpw: Since reading https://medium.com/@sdboyer/so-you-want-to-write-a-package-manager-4ae9c17d9527  I'm thinking hard about whether we should be automatically modifying the user-editable .visidatarc at all.  I wonder if we should make an autogenerated `.visidata/plugins/__init__.py` (or similar) in which the plugin imports can be maintained, and have a simple `from plugins import *` in `.visidatarc`.

anjakefala: >  I wonder if we should make an autogenerated .visidata/plugins/__init__.py (or similar) in which the plugin imports can be maintained, and have a simple from plugins import * in .visidatarc.

I agree with this. I think this is a great idea.
Users can still maintain their plugins manually in .visidatarc.

anjakefala: **Update**: I decided to take on reviewing this PR and I am now working on implementing the change @saulpw proposed as a prototype.

ajkerrigan: Oh that's great @anjakefala! This PR was intended to be "minor tweaks", but the discussion pointed to deeper changes. Think it's worth closing this PR entirely and sucking anything worthwhile into a new one? I'm happy to work with you on it if you want a partner, or just cheer afterward 😁.

anjakefala: Sure! Let us see where things are once I push the updated plugins.py, and then integrate for e.g. the pipx work. =)

I would love to include you! I already started working on it, so I am going to clean up what I have and push it into a new PR today. Then maybe we can take a look and figure out what should be done next?

Issue #346, opened 2019-09-01
Visidata 2 feedback:  Changelog example for stdout/pipe redirect


frosencrantz:
The example for stdout/pipe redirect purpose could be better.  The entry does have some helpful notes about how to save output or not save output.  It does give a sense of what can be done.

My issue is with the example command line `ls|vd|lpr`.  The issues:
  * you won't be able to select the first line, since vd will use it as a header.
  * if any files contain spaces, they might end up in a separate column.
  * By default visidata does give you a listing of the files in the current directory with attributes like size.  Why does the example pipe `ls` into vd?
  * why not just use fzf, which is faster and optimized for selecting filenames?

Where visidata shines over fzf is the ability to select a line based on values other than just names like numeric values.  And the ability to only select a particular column of data.  The fzf wiki has an example where visidata would be a better choice:  https://github.com/junegunn/fzf/wiki/Examples#processes  They have some examples where they pipe ps to fzf, but then need to use awk to strip a column and send it to xargs.

So why not make the example: ps -ef | vd | xargs -r kill 
Also, add a helpful note about how to save the values from a single column, I'm not sure how to do that.

anjakefala:
This is really useful, @frosencrantz! Thanks. You are right that the example could be better and that that is where VisiData shines. =)

We have not created a page in our documentation for piping. I am really glad you did the work of putting this together and if you could create a short page for pipes.md in here https://github.com/saulpw/visidata/tree/develop/docs and create a PR on develop, I would love that. We can then figure out where the rightful place is for the content. 

(We used to have a command for saving a single column, but it may have fallen through the cracks. That is definitely a useful one to have for efficient piping, so we should re-add that in.)

anjakefala:
@frosencrantz added the promised documentation! ^^

frosencrantz:
@saulpw of @anjakefala not sure if you want to address this or not. To me it seems useful to have a way to save the values of a column (or multiple columns) to stdout, but without the column header. It's not clear there is a way to do that. This seems useful to me when visidata is in the middle of a pipeline.

like if you select some files to print, the header is not useful data to send to lpr.

PR #358, opened 2019-09-15
Add docs for using vd in a pipeline.

frosencrantz:
As promised for issue #346

There are a couple questions I can't answer, see the items with a leading checkbox.


frosencrantz:
@saulpw not sure if you want to address this or not.  To me it seems useful to have a way to save the values of a column (or multiple columns) to stdout, but without the column header.  It's not clear there is a way to do that.  This seems useful to me when visidata is in the middle of a pipeline.

PR #367, opened 2019-09-27
Plugin Refactor

anjakefala:
Join collaboration between @anjakefala and @ajkerrigan. =)

The following changes were made to the plugin system:
1. Plugins are now imported into `~/.visidata/plugins/__init__.py`. Users manually add `from plugins import *` to their `~/.visidatarc`.
2. If a module is already imported into `plugins/__init__.py` it will not be added again. (This is O(n) where n is the number of already imported plugins; contentious.)
3. Bug is fixed where module at the top of `plugins/__init__.py` would not be removed.
4. Ensure consistent Python exe for plugin installs.

### DIFF
--- a/visidata/plugins.py
+++ b/visidata/plugins.py
@@ -63,20 +63,29 @@ class PluginsSheet(TsvSheet):
 
     @asyncthread
     def _install(self, plugin):
-        outpath = _plugin_path(plugin)
+        exists = False
+        with Path(_plugin_init()).open_text(mode='r') as fprc:
+            r = re.compile(r'^{}\W'.format(_plugin_import(plugin)))
+            for line in fprc.readlines():
+                if r.match(line):
+                    exists = True
+                    warning("plugin already loaded")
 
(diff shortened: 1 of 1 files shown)


visidata/plugins.py
```diff
@@ -54,37 +63,49 @@ def installPlugin(self, plugin):
...
+                    warning("plugin already loaded")
 
-        with urlcache(plugin.url, 0).open_text() as pyfp:
-            with outpath.open_text(mode='w') as outfp:
-                outfp.write(pyfp.read())
+        if not exists:
+            outpath = _plugin_path(plugin)
 
-        if plugin.requirements:
-            p = subprocess.Popen(['pip3', 'install']+plugin.requirements.split())
```
saulpw: It looks like requirements won't be installed anymore; will the user need to install them manually now?

saulpw (approved):
Looks great!  This will make plugin management a lot cleaner.

PR #370, opened 2019-10-08
[http] Use options.encoding when no encoding is provided by response headers

tsibley:
The alternative is a characters vs. bytes error later on because no encoding means response.iter_lines() returns bytes instead of chars.

This provides a nice fallback encoding of UTF-8 in the default case.


saulpw:
Thanks for this, @tsibley!  Definitely an improvement.

PR #396, opened 2019-11-14
Make unit test sample path relative to the package root

ajkerrigan:
Hopefully using Python's [pkg_resources](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#resource-extraction) module will ensure that unit tests can run and find sample data based on the package root path.


anjakefala:
Wow, what a goof!

I tested this locally, and it works reliably. Thanks for the fix @ajkerrigan!

PR #410, opened 2019-11-26
Added option to control the number of digits to be displayed for floa…

khughitt:
Greetings!

I attempted to create a small PR to add an option (`disp_float_ndigits`) to allow the user to control how many digits are displayed for float columns.

I think the PR should be close -- the option now appears with the correct default and can be set in the usual manner, however, changing the value doesn't seem to update the current sheet. Any ideas what I'm missing?

I tried to follow the conventions used for `disp_date_fmt` since that seemed to be similar in spirit.

### DIFF
--- a/visidata/_types.py
+++ b/visidata/_types.py
@@ -4,16 +4,16 @@ import collections
 import functools
 import datetime
 import locale
-from visidata import options, theme, TypedWrapper, undoAttr
+from visidata import option, options, theme, TypedWrapper, undoAttr
 
 #__all__ = ['anytype', 'vdtype', 'typemap', 'getType', 'typemap']
+
+option('disp_float_fmt', '%.02f', 'default fmtstr to format for float values', replay=True)
 
 try:
     import dateutil.parser
 except ImportError:
     pass
-
-theme('disp_float_ndigits', 2, 'default number of digits to display for float column')
 
 # _vdtype .typetype are e.g. int, float, str, and used internally in these ways:
 #
@@ -54,7 +54,7 @@ vdtype(None, '∅')
 vdtype(anytype, '', formatter=lambda _,v: str(v))
 vdtype(str, '~', formatter=lambda _,v: v)
 vdtype(int, '#', '%.0f')
-vdtype(float, "%", "%.0{}f".format(options.disp_float_ndigits))
+vdtype(float, '%', formatter=lambda _,v: options.disp_float_fmt % v)
 vdtype(dict, '')
 vdtype(list, '')
 
--- a/visidata/_types.py
+++ b/visidata/_types.py
@@ -13,6 +13,7 @@ try:
 except ImportError:
     pass
 
+theme('disp_float_ndigits', 2, 'default number of digits to display for float column')
 
 # _vdtype .typetype are e.g. int, float, str, and used internally in these ways:
 #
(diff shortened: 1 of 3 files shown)


visidata/_types.py
```diff
@@ -53,7 +54,7 @@ def getType(typetype):
 vdtype(anytype, '', formatter=lambda _,v: str(v))
 vdtype(str, '~', formatter=lambda _,v: v)
 vdtype(int, '#', '%.0f')
-vdtype(float, '%', '%.02f')
+vdtype(float, "%", "%.0{}f".format(options.disp_float_ndigits))
```
saulpw: Using `options` at the toplevel like this only uses the default, because it's evaluated before options parsing (including .visidatarc) happens.  Also it looks like `vtask` is just resetting the default.  Do you have a use for this configurability yourself?

saulpw:
Thanks for putting this together, @khughitt.  I'm not sure it works as intended (see my code comment); how do you intend to use it?

I think if we were to offer configurability like this, it should be `disp_float_fmt` for symmetry with the existing `disp_date_fmt`.  The `disp_date_fmt` method evaluates the option at formatting time so it can be changed at any point.

As a separate issue, we need to rename `disp_date_fmt`, since the format string affects saving in certain text formats, but that's a separate issue.  Then it can also be set from the command-line (`disp_` and `color_` options can't be set that way).

khughitt:
@saulpw Sorry for the delayed response and thank you for the feedback!

I realized the initial PR wasn't fully-working, but thought it should be close so I was hoping to get your feedback -- I probably should have made that more clear though.

I spent a bit more talk walking through the initialization and handling of settings this morning though, and I think I have it working now.

I also updated the name of the option, per your suggestion.

In general, I think this is useful to be able to control since the some users may prefer more precision for the default float formatting. Some REPL's (e.g. R) provide similar support for specifying the default number of digits to be displayed when printed to the screen.

I tested it with different settings, and also with no user setting for `disp_float_fmt` and it all seemed to behave well.

Just let me know if I missed anything!

visidata/_types.py
```diff
@@ -53,7 +54,7 @@ def getType(typetype):
 vdtype(anytype, '', formatter=lambda _,v: str(v))
 vdtype(str, '~', formatter=lambda _,v: v)
 vdtype(int, '#', '%.0f')
-vdtype(float, '%', '%.02f')
+vdtype(float, '%', formatter=lambda _,v: options.disp_float_fmt % v)
```
saulpw: almost there...the formatter should be like the disp_date_fmt usage:

`vdtype(date, '@', '', formatter=lambda fmtstr,val: val.strftime(fmtstr or options.disp_date_fmt))`

The first parameter is the given fmtstr, as set by the user on the Column sheet.  So we have to take that into account similarly with disp_float_fmt or the user won't be able to override fmtstr on float columns.

khughitt:
Hey! Sorry for the slow response.. Finally got around to looking into this. Thanks for catching that! I updated the file once more to check for user customization, per your suggestion. I tested it locally and it seems to work for all scenarios (default, config, user-override).

Let me know if I missed anything else!

anjakefala:
Looks good @khughitt! I am going to merge and move this to the `develop` branch. =)

PR #419, opened 2019-12-13
Use HTTPS protocol for git sibmodules

tombh:
This allows installation of VisiData in automated environments such as
Dockerfiles where the git user is not logged into Github.


anjakefala (approved):
Thanks @tombh! I should also ensure this syntax is updated everywhere.

PR #428, opened 2019-12-27
Migrate unit tests to pytest

ajkerrigan:
A little while back @anjakefala and I kicked around the idea of migrating unit tests to pytest. It sounded like consolidating some redundant setup logic into a central location could make the move worthwhile.

This PR serves as a proof of concept for that move, setting up a couple shared fixtures in [conftest.py](https://docs.pytest.org/en/latest/fixture.html#conftest-py-sharing-fixture-functions).

Changes beyond that are minimal for now. Let's figure out if the switch makes sense, or if there are still some kinks to iron out first :)

### DIFF
--- a/dev/requirements-dev.txt
+++ b/dev/requirements-dev.txt
@@ -1,3 +1,4 @@
 markdown
+pytest
 pyyaml
 tox
--- a/visidata/tests/conftest.py
+++ b/visidata/tests/conftest.py
@@ -1,0 +1,25 @@
+import pytest
+from unittest.mock import Mock
+
+
+@pytest.fixture(scope="class")
+def curses_setup():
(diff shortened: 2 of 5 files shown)


anjakefala:
Tested it and it runs like a dream! Tagging @saulpw for final merge decision.

anjakefala (approved):

dev/requirements-dev.txt
```diff
@@ -1,3 +1,4 @@
 markdown
+pytest
```
anjakefala, line 2: This was added to the right place! This made me happy.

saulpw (approved):
Thanks for putting this together, @ajkerrigan!

Issue #432, opened 2020-01-04
[develop] -P longname is not working


anjakefala:
**Small description**
`vd -P longname` was introduced in v2.-2. It replays *longname*. 

Since 751de48351533b3f5fb0ebde35ec638129dc09fd it instead resuts in this stacktrace:
```
Traceback (most recent call last):
  File "visidata/basesheet.py", line 116, in exec_command
    vd.cmdlog.beforeExecHook(self, cmd, '', keystrokes)
  File "visidata/cmdlog.py", line 143, in beforeExecHook
    sheetname = sheet.name
  File "visidata/basesheet.py", line 142, in name
    return self._name or self.source.name + '_'+self.rowtype
AttributeError: 'NoneType' object has no attribute 'name'
```

**Steps to reproduce with sample data and a .vd**
`vd -P open-plugins`

**Additional context**
`develop` branch

PR #436, opened 2020-01-06
Add empty sheet name fallback

ajkerrigan:
In some cases a sheet has no name information available (`_name` or a `source` to pull from). The `name` property can error out in that case.

Since we don't have enough information to return a useful name property in these cases, one option is to fall back to an empty name. Callers may have additional local context that can serve as a sheet name substitute (such as a command longname in the case of #432).

This approach may clash with the "no empty sheet name" intent of 751de48, in which case we can handle this differently :).

### DIFF
--- a/visidata/basesheet.py
+++ b/visidata/basesheet.py
@@ -139,7 +139,7 @@ class BaseSheet(Extensible):
 
     @property
     def name(self):
-        return self._name or self.source.name + '_'+self.rowtype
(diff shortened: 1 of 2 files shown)


anjakefala:
This work was done in response to this bug. =) https://github.com/saulpw/visidata/issues/432

saulpw (approved):

saulpw:
These one-liners do get onerous though don't they :)

PR #440, opened 2020-01-07
Vfake extra providers

ajkerrigan:
The `vfake` plugin is incredibly useful! One way it can become even more handy is by loading additional user-defined providers. @saulpw and I discussed this briefly, initially thinking we could use a whitespace-separated list of provider modules and pass them into the `includes` kwarg of the `Faker()` constructor. After a bit of local fiddling, I found that approach to be less flexible and a bit trickier than adding individual providers via `add_provider()`.

By allowing an option to specify a list of `Provider` classes, we open the door for `.visidatarc` configurations like this:

```python
from faker.providers import BaseProvider
from faker_cloud import AmazonWebServicesProvider
from string import ascii_uppercase, digits

class VdCustomProvider(BaseProvider):
    def ws_computer_name(self):
        return self.lexify(f"EC2AMAZ-{'?' * 7}", ascii_uppercase + digits)

options.vfake_extra_providers = [AmazonWebServicesProvider, VdCustomProvider]

Sheet.bindkey("zf", "setcol-fake")
```

In this case `vfake` has access to:

* The default Faker base providers
* AWS providers from the `faker_cloud` module (installed via `pip`)
* Some exploratory extras defined locally right inside `.visidatarc`

Custom providers will be more sustainable and shareable if they are promoted to a proper vd plugin or even a pip-installable package. Being able to define and add them right from the vd config file makes the initial customization bar much lower though, without sacrificing long-term maintainability.

### DIFF
--- a/plugins/vfake.py
+++ b/plugins/vfake.py
@@ -5,21 +5,37 @@ from visidata import vd, Column, Sheet, option, options, asyncthread, Progress
 __version__ = '0.9'
 
 option('locale', 'en_US', 'default locale to use for Faker', replay=True)
+option('vfake_extra_providers', None, 'list of additional Provider classes to load via add_provider()', replay=True)
 
-# See also: https://faker.readthedocs.io/en/master/communityproviders.html
-option('vfake_extra_providers', [], 'list of additional Provider classes to load via add_provider()', replay=True)
+def addFakerProviders(fake, providers):
+    '''
+    Add custom providers to Faker. Provider classes typically derive from
+    faker.providers.BaseProvider, so check for that here. This helps to
+    highlight likely misconfigurations instead of hiding them.
+
+    See also: https://faker.readthedocs.io/en/master/communityproviders.html
+
+    fake: Faker object
+    providers: List of provider classes to add
+    '''
+    import faker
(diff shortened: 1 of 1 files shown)


plugins/vfake.py
```diff
@@ -6,11 +6,20 @@
 
 option('locale', 'en_US', 'default locale to use for Faker', replay=True)
 
+# See also: https://faker.readthedocs.io/en/master/communityproviders.html
+option('vfake_extra_providers', [], 'list of additional Provider classes to load via add_provider()', replay=True)
```
saulpw: It's not implemented yet, but I'm thinking that having a list as option default will mean "one of these enumerated values".  Maybe change the default to `None`? (which allows any type)

saulpw:
This looks good, @ajkerrigan!  I think we should change the default (and any fallout), but otherwise I think it's good to merge.

plugins/vfake.py
ajkerrigan: Ahh good tip, thanks! I changed the default, and then moved some stuff around because it was starting to make `setValuesFromFaker` look noisy.

PR #445, opened 2020-01-16
[loaders] Add loader for s3:// URLs

tsibley:
This loader daisy chains with the https:// loader by generating
pre-signed S3 URLs using the S3 API.  Daisy chaining like this keeps the
code simpler and reuses the content-encoding and content-type logic in
the http loader rather than re-implementing against S3 object
properties.

Authentication is handled in the normal way for AWS clients, either
through environment variables¹ or credentials files².

Tests for this will require first setting up an accessible S3 bucket.

¹ https://boto3.readthedocs.io/en/latest/guide/configuration.html#environment-variables
² https://boto3.readthedocs.io/en/latest/guide/configuration.html#shared-credentials-file

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,5 +20,6 @@ dpkt                # pcap
 dnslib              # pcap
 namestand           # graphviz
 datapackage         # frictionless .json
+boto3               # s3
 
 -e git+https://github.com/saulpw/pyxlsb.git@master#egg=pyxlsb # xlsb
--- a/visidata/__init__.py
+++ b/visidata/__init__.py
@@ -98,6 +98,7 @@ from .loaders.markdown import *
 from .loaders.pcap import *
 from .loaders.png import *
 from .loaders.ttf import *
+from .loaders.s3 import *
 from .loaders.sas import *
 from .loaders.spss import *
 from .loaders.xml import *
--- a/visidata/loaders/s3.py
+++ b/visidata/loaders/s3.py
@@ -1,0 +1,18 @@
+from visidata import openurl_https, Path
+from urllib.parse import urlparse
+
+def openurl_s3(path, filetype=None):
+    import boto3
+
+    s3_url = urlparse(path.given)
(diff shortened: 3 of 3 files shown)


tsibley:
I'm happy to add a test file for this loader, but there's a few things needed first:

- [ ] Decision on which AWS account / S3 bucket to use. Seems like ideally this would be "official" and controlled by the project maintainers.
- [ ] Provisioning of limited AWS credentials for the CI infrastructure.
- [ ] Consideration of how to skip the test when credentials aren't available.

saulpw:
Hi @tsibley, thanks for putting this together!  I appreciate how simple the code is, though boto is a large dependency.  Have you seen @ajkerrigan's [vds3 plugin](https://github.com/ajkerrigan/visidata-plugins), which uses s3fs to browse the files in an s3 bucket?

tsibley:
Ah, the vds3 plugin looks great! Thanks for the pointer, I wasn't aware of it. (I haven't tried out vds3 yet, as I ran into problems installing it via the plugins sheet, but I think it's an underlying install issue with plugins in general. I'll make a separate issue.)

I thought the boto3 dependency wouldn't be an issue since it's optional, like pandas or lxml. In any case, I'm happy to drop this PR in favor of recommending the vds3 plugin. :-)

Would you be receptive to helping folks discover plugins via status line suggestions? e.g. "Install the vds3 plugin to read s3:// URLs." I'm thinking of situations like the one I ran into, where someone runs `vd s3://foo/bar.csv` and finds it doesn't work. If VisiData had prompted me to check out the plugin (instead of throwing a `KeyError`), that would have been really slick. I'd be willing to implement this, if you're receptive.

tsibley:
(Followed up on the underlying plugin install issue I ran into in #448.)

saulpw:
Closing this in light of `vds3`; reopen if you'd like to explore the boto plugin concept further.

PR #446, opened 2020-01-19
[plugins] update vds3 to 0.3

ajkerrigan:
Add support for custom/local S3 endpoints and glob-matching.

For the record, I've been using a local utility script ([gist](https://gist.github.com/ajkerrigan/7bd8f0c14543b7a78a68b71e95b8d5a1)) to quickly generate or compare plugin hashes. Figured I'd link it here in case it's useful to anyone else.


saulpw (approved):
This is exactly how this should work, nice job!

Issue #444, opened 2020-01-16
Display columns are duplicated when loading JSON with a single object


ajkerrigan:
**Small description**

If I open a JSON file which contains a single JSON object rather than an array of objects, its columns are duplicated in the VisiData window. Sample contents (`singleitem.json`):

```json
{
  "key1": "val1",
  "key2": "val2"
}
```

**Expected result**

Opening a JSON file with a single object should appear the same as opening a JSON file containing an _array_ containing only that object.

In VisiData 1.5.2, this works as expected:

```
 key1   | key2   ‖
 val1   | val2   ‖
```

**Actual result with screenshot**

In VisiData 2.-3dev:

```
 key1   ╵ key2   ╵ key1   ╵ key2   ║
 val1   ╵ val2   ╵ val1   ╵ val2   ║
```

**Steps to reproduce with sample data and a .vd**

```
echo '{
  "key1": "val1",
  "key2": "val2"
}' > singleitem.json

vd singleitem.json
```

**Additional context**

Version: VisiData 2.-3dev, most recently 0164c10.

Note that this issue appears to be only a _display/UI_ quirk. Saving the file does _not_ cause duplicated columns in the output.

ajkerrigan:
For the record, this feels like a pretty low priority issue and I'm happy to take a peek at it myself. Just wanted to get it documented before I forgot!

saulpw:
Nice find, @ajkerrigan, and thanks for fixing it!

PR #447, opened 2020-01-20
Avoid adding columns twice when loading JSON dicts

ajkerrigan:
It looks like when loading a JSON file containing a single object rather than an array, `JsonSheet` ends up adding columns twice: once in `addRow()` and once in `iterload()`. If I'm looking at things properly, handling that in `addRow()` is sufficient.

Addresses #444


saulpw:
This is awesome, @ajkerrigan, thanks for fixing this!

PR #453, opened 2020-01-29
[melt-] use addColumn api instead of manually appending

anjakefala:
- this fixes a bug where the *ColumnsSheet* for melted sheets did not
contain column values
- this is due to the columns on the melted sheet not containing a sheet attribute
- `addColumn()` takes care of that, along with a host of other goodies

*Please do not merge if approved because I would like to edit the commit message before this goes into `develop`!*

I am unsure of when `self.columns.append()` is preferred over `self.addColumn()`, and if it is actually preferred here for a reason. For that reason, I would like this change reviewed. =) 

Reproduction for bug: run `vd -p tests/gMelt.vd` and then press `C`.

### DIFF
--- a/visidata/tidydata.py
+++ b/visidata/tidydata.py
@@ -22,10 +22,12 @@ class MeltedSheet(Sheet):
 
     @asyncthread
     def reload(self):
+        self.columns = []
         isNull = isNullFunc()
 
         sheet = self.source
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Yup, makes total sense to me.

anjakefala:
Thanks! =)

PR #469, opened 2020-02-16
Update Canvas delete- commands with current API

anjakefala:
Also, a bonus test fix commit. =)

This fixes delete on Canvas.


visidata/canvas.py
```diff
@@ -676,10 +676,10 @@ def render_sync(self):
 Canvas.addCommand('t', 'stoggle-cursor', 'source.toggle(list(rowsWithin(plotterCursorBox)))')
 Canvas.addCommand('u', 'unselect-cursor', 'source.unselect(list(rowsWithin(plotterCursorBox)))')
 Canvas.addCommand(ENTER, 'dive-cursor', 'vs=copy(source); vs.rows=list(rowsWithin(plotterCursorBox)); vd.push(vs)')
-Canvas.addCommand('d', 'delete-cursor', 'source.delete(list(rowsWithin(plotterCursorBox))); reload()')
+Canvas.addCommand('d', 'delete-cursor', 'toDelete=list(rowsWithin(plotterCursorBox)); source.copyRows(toDelete); source.deleteBy(lambda r,toDelete=toDelete: r in toDelete); reload()')
```
saulpw: This is workable for now, but this is O(N\*M) and we can do O(N\*log(M)) (where N is the number of rows and M is the number of rows to delete).  We ought to have a helper function that takes an iterable of rows and deletes them; it would create a dict of `rowid(r): row` (like `_selectedRows` is maintained as) from the iterable, and give that to the closure passed to deleteBy.  If you have a set of 2M data points and highlight 1M with the cursor and delete them from the canvas, the difference will be noticeable.

Also notice that `deleteSelected` is `@asyncthread`, and `deleteBy` is not.  So this would actually cause VisiData to hang for awhile in the above scenario.  So our new helper function should be async, and we could then call it with `rowsWithin(plotterCursorBox)`.  But then we need to figure out how to reload() the canvas after the delete thread finishes.  This is needed for Canvas.deleteCursor but not for Sheet, so for now we should skip the generic helper function and just refactor this execstr into its own canvas.deleteCursor() function and mark it asyncthread.

visidata/canvas.py
```diff
@@ -676,10 +681,10 @@ def render_sync(self):
 Canvas.addCommand('t', 'stoggle-cursor', 'source.toggle(list(rowsWithin(plotterCursorBox)))')
 Canvas.addCommand('u', 'unselect-cursor', 'source.unselect(list(rowsWithin(plotterCursorBox)))')
 Canvas.addCommand(ENTER, 'dive-cursor', 'vs=copy(source); vs.rows=list(rowsWithin(plotterCursorBox)); vd.push(vs)')
-Canvas.addCommand('d', 'delete-cursor', 'toDelete=list(rowsWithin(plotterCursorBox)); source.copyRows(toDelete); source.deleteBy(lambda r,toDelete=toDelete: r in toDelete); reload()')
+Canvas.addCommand('d', 'delete-cursor', 'toDelete=list(rowsWithin(plotterCursorBox)); deleteCursor(toDelete); reload()')
```
saulpw: - The reload() needs to be inside deleteCursor(), so it runs properly after the delete has completed (otherwise it may reload the Canvas from a Sheet with a delete-in-progress.
- toDelete doesn't have to be a separate variable anymore.
- the list() should be done inside deleteCursor, because it will need to become a dict in order to achieve O(N*logM).  better to pass the iterator from rowsWithin() directly.

anjakefala: > The reload() needs to be inside deleteCursor(), so it runs properly after the delete has completed (otherwise it may reload the Canvas from a Sheet with a delete-in-progress.

Can an async function be called from within an async function?

anjakefala: > toDelete doesn't have to be a separate variable anymore.

Should there then be a Canvas().deleteCursor() and Canvas().deleteVisible()?

saulpw: I guess it should be `Canvas.deleteSourceRows()` and then you pass the rowsWithin() to it.  Or maybe it should be `Canvas.deleteBox()` and do the `rowsWithin` itself?

saulpw: > Can an async function be called from within an async function?

Yes, of course :)

anjakefala:
I decided to go with `deleteSourceRows`! I moved reload() and the listifier to within it.

saulpw (approved):

Issue #471, opened 2020-02-18
Errors while calculating sheet name in view/view_pandas


ajkerrigan:
**Small description**

When opening VisiData from a REPL using `visidata.view()` or `visidata.view_pandas()`, errors evaluating the `sheet.name` property prevent VisiData from loading properly.

**Expected result**

VisiData can be used from a REPL to view arbitrary Python objects with `visidata.view()`, or Pandas DataFrames with `visidata.view_pandas()`.

**Actual result with screenshot**

Error invoking `visidata.view_pandas()`:

```
~/code/visidata/visidata/basesheet.py in name(self)
    142         if self._name:
    143             return self._name
--> 144         if not self.source:
    145             return ''
    146         return self.source.name + '_'+self.rowtype

~/.pyenv/versions/3.8.0/envs/vd-dev/lib/python3.8/site-packages/pandas/core/generic.py in __nonzero__(self)
   1550
   1551     def __nonzero__(self):
-> 1552         raise ValueError(
   1553             "The truth value of a {0} is ambiguous. "
   1554             "Use a.empty, a.bool(), a.item(), a.any() or a.all().".format(

ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
```

Error invoking `visidata.view()` (tested with similar results against lists, dicts, and lists of dicts):

```
~/code/visidata/visidata/basesheet.py in name(self)
    144         if not self.source:
    145             return ''
--> 146         return self.source.name + '_'+self.rowtype
    147
    148     @name.setter

AttributeError: 'list' object has no attribute 'name'
```

**Steps to reproduce with sample data and a .vd**

From a Python REPL in an environment with VisiData installed from `develop` (tested against commit 5b72a2cdc77ff21e3dc0dbd8e8a11bfed5de911e):

```python
import visidata
import pandas as pd

obj = [{'col1': 10, 'col2': 'fish', 'col3': True}]

visidata.view(obj)
visidata.view_pandas(pd.DataFrame(obj))
```

**Additional context**
Please include the version of VisiData.

2.-3dev (commit 5b72a2cdc77ff21e3dc0dbd8e8a11bfed5de911e)

Note that the view/view_pandas functionality is related to #470. And in case it's useful, here is some historical context around recent changes to `sheet.name` that seems relevant here:

9c00900a7ae5a267d6085b62c72f42cb724e758b
https://github.com/saulpw/visidata/pull/436
[...]

PR #472, opened 2020-02-19
Tweak logic for the sheet.name property

ajkerrigan:
This is one possible solution to the #471 which doesn't cause regressions with other related issues. It may not be the _optimal_ solution, but that's what PRs are for eh? :)

Switch to an 'easier to ask forgiveness than permission' approach, since a
sheet name can vary a bit depending on which attributes are present.
Falling back to `rowtype` (which always exists) is a more useful default
name than an empty string, and catching missing attributes once is
simpler to follow than a chain of conditionals.

### DIFF
--- a/visidata/basesheet.py
+++ b/visidata/basesheet.py
@@ -139,11 +139,10 @@ class BaseSheet(Extensible):
 
     @property
     def name(self):
-        if self._name:
-            return self._name
-        if not self.source:
-            return ''
-        return self.source.name + '_'+self.rowtype
+        try:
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Sure, let's try it, see if any weird corner cases come up.  Since every sheet should have a palatable name anyway, maybe we add a `vd.exceptionCaught()` and let it fire for view() until we fix that properly.

anjakefala:
Thanks for the PR @ajkerrigan! I verified that this fixed the pandas case, tested a couple of different VisiData cases, and ran the existing tests. Everything seems fine for merging. =)

PR #482, opened 2020-03-09
[loaders] handle password-protected zip files

ajkerrigan:
If we wrap zipfile's `open()` method, we can catch and handle "password required" errors with VisiData's typical input flow.

This stems from a question/discussion with geekscrapy in the IRC channel. Some important details:

* `display=False` prevents the password from echoing back or being recorded in the command log for later replay.
* Leaving out a `type` for `vd.input()` prevents the password from being stored in a history buffer.

Thanks for the tips along the way @saulpw !

### DIFF
--- a/visidata/loaders/archive.py
+++ b/visidata/loaders/archive.py
@@ -16,19 +16,19 @@ class ZipSheet(Sheet):
         ColumnAttr('compress_size', type=int)
     ]
 
-    def open_zipfile(self, fp, *args, **kwargs):
+    def openZipFile(self, fp, *args, **kwargs):
         '''Use VisiData input to handle password-protected zip files.'''
         try:
             return fp.open(*args, **kwargs)
         except RuntimeError as err:
             if 'password required' in err.args[0]:
                 pwd = vd.input(f'{args[0].filename} is encrypted, enter password: ', display=False)
-                return fp.open(*args, **kwargs, pwd=pwd.encode())
+                return fp.open(*args, **kwargs, pwd=pwd.encode('utf-8'))
             vd.error(err)
 
     def openRow(self, fi):
             zfp = zipfile.ZipFile(str(self.source), 'r')
-            decodedfp = codecs.iterdecode(self.open_zipfile(zfp, fi),
+            decodedfp = codecs.iterdecode(self.openZipFile(zfp, fi),
                                           encoding=options.encoding,
                                           errors=options.encoding_errors)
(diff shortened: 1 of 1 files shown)


saulpw (changes requested):
I wonder whether it's better to inline it in `openRow` instead.  With monkey-patching like this, any VisiData plugin which uses `ZipFile.open` in a command will automatically use this same workflow, which is nice. But a non-VisiData tool that imports visidata and tries to open a password-protected .zip file will then also try to use `vd.input`, which won't work outside of `vd.run()`.

ajkerrigan:
> I wonder whether it's better to inline it in `openRow` instead. With monkey-patching like this, any VisiData plugin which uses `ZipFile.open` in a command will automatically use this same workflow, which is nice. But a non-VisiData tool that imports visidata and tries to open a password-protected .zip file will then also try to use `vd.input`, which won't work outside of `vd.run()`.

That's a good call, I wasn't sure where to slot this functionality myself. I _had_ to use the monkeypatching wrapper for local testing in `.visidatarc`, but I see the point about avoiding surprises elsewhere. Here's another option that avoids throwing too much logic directly into `openRow`, but also keeps any additions local to `ZipSheet`.

visidata/loaders/archive.py
```diff
@@ -16,9 +16,19 @@ class ZipSheet(Sheet):
         ColumnAttr('compress_size', type=int)
     ]
 
+    def open_zipfile(self, fp, *args, **kwargs):
```
saulpw: We should use something other than `open_`, unless we want `-f zipfile` to work somehow (and then this would have to return a sheet of some kind).  Maybe `openZipFile`?

visidata/loaders/archive.py
```diff
@@ -16,9 +16,19 @@ class ZipSheet(Sheet):
...
+        '''Use VisiData input to handle password-protected zip files.'''
+        try:
+            return fp.open(*args, **kwargs)
+        except RuntimeError as err:
+            if 'password required' in err.args[0]:
+                pwd = vd.input(f'{args[0].filename} is encrypted, enter password: ', display=False)
+                return fp.open(*args, **kwargs, pwd=pwd.encode())
```
saulpw: Let's make `'utf-8'` an explict parameter to `encode()`.  (It's the default option but would be helpful I think, especially given that this is a password.)

saulpw (approved):
looks good, let's ship it! :ship:

PR #491, opened 2020-03-27
[path-] avoid infinite recursion when _path does not have attr

### LINKED ISSUE #489: [Bug] [v2.-4dev] dup-rows/-deep does not copy rows
I found that if I try and do a `dup-rows-deep` with a DirSheet I get the following error with no rows in the sheet (columns are correct and as expected).

This only appears to happen with a DirSheet, I just tried with a TextSheet which works as expected.

```
  File "/path/vd_plugins/radare2/r2/lib/python3.7/site-packages/visidata-2._4dev-py3.7.egg/visidata/threads.py", line 201, in _toplevelTryFunc
  File "/path/vd_plugins/radare2/r2/lib/python3.7/site-packages/visidata-2._4dev-py3.7.egg/visidata/threads.py", line 80, in _async_deepcopy
    newlist.append(deepcopy(r))
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/copy.py", line 180, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/copy.py", line 281, in _reconstruct
    if hasattr(y, '__setstate__'):
  File "/path/vd_plugins/radare2/r2/lib/python3.7/site-packages/visidata-2._4dev-py3.7.egg/visidata/path.py", line 77, in __getattr__
    r = getattr(self._path, k)
  File "/path/vd_plugins/radare2/r2/lib/python3.7/site-packages/visidata-2._4dev-py3.7.egg/visidata/path.py", line 77, in __getattr__
    r = getattr(self._path, k)
  File "/path/vd_plugins/radare2/r2/lib/python3.7/site-packages/visidata-2._4dev-py3.7.egg/visidata/path.py", line 77, in __getattr__
    r = getattr(self._path, k)
  [Previous line repeated 492 more times]
RecursionError: maximum recursion depth exceeded
```
### Replicate:

1. Use v2.-4dev branch
2. Open a DirSheet with `vd .`
3. Do a deep copy of the sheet with `dup-rows-deep`

### Result:
A sheet with no rows, columns as expected for a DirSheet, and the error message above.

### Expected:
A deep copy of the original DirSheet with all rows (including those selected)

### cmdlog:
```
	SqliteSheet	header	set-option	0
	UsvSheet	delimiter	set-option	␞
	UsvSheet	row_delimiter	set-option	␟
	override	disp_date_fmt	set-option	%Y-%m-%d %H:%M:%S
			open-file	.	o
_files			dup-rows-deep		gz"	open duplicate sheet with deepcopy of all rows
```

Also, am I correct in understanding that if I make a deep copy, modifications I make to that copy should propagate to the original sheet? And this should include selecting/deselecting rows?

anjakefala:
- previously assumed that every time vd`Path` does not have the
attribute, `_path` (`pathlib.Path`) would
- when `_path` lacked the attribute, `AttributeError` was not being raised
- instead VisiData would look for that attribute in an additional layer
of `_path` (since looking for it would invoke `__getattr__` *again*), and look for `_path` forever until Python put it out of its
misery
- this commit first checks for the existence of _path

Fixes #489, which was triggering this recursion since `deepcopy()` was
checking for the `__setstate__` attribute, which `pathlib.Path` objects do
not have implemented.

### DIFF
--- a/visidata/path.py
+++ b/visidata/path.py
@@ -74,7 +74,10 @@ class Path(os.PathLike):
         if hasattr(self.__dict__, k):
             r = getattr(self.__dict__, k)
         else:
-            r = getattr(self._path, k)
+            if self.__dict__.get('_path', None) is not None:
+                r = getattr(self._path, k)
+            else:
+                raise AttributeError(k)
         if isinstance(r, pathlib.Path):
             return Path(r)
         return r


visidata/path.py
```diff
@@ -74,7 +74,10 @@ def __getattr__(self, k):
         if hasattr(self.__dict__, k):
             r = getattr(self.__dict__, k)
         else:
-            r = getattr(self._path, k)
+            if self.__dict__.get('_path', None) is not None:
```
saulpw, line 77: Let's use the same `hasattr` form as above for this.

visidata/path.py
```diff
@@ -74,7 +74,10 @@ def __getattr__(self, k):
...
+                r = getattr(self._path, k)
+            else:
+                raise AttributeError
```
saulpw: I think this should be `raise AttributeError(k)`, no?

anjakefala: Great idea!

visidata/path.py
anjakefala: We cannot use `hasattr`, regrettably, because `hasattr` uses `getattr`, which is what we are trying to avoid using this convoluted method. :sweat_smile:

saulpw (approved):
Looks good, thanks for fixing this!

Issue #485, opened 2020-03-14
[develop] Can't connect to postgresql db


sfranky:
**Small description**

I try to connect to a postgresql database with
```
vd postgres://<user>:<passwd>@IP:PORT/dbname
```

**Actual result with screenshot**

```
Traceback (most recent call last):
  File "/usr/local/bin/vd", line 4, in <module>
    __import__('pkg_resources').run_script('visidata===2.-3dev', 'vd')
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 654, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 1441, in run_script
    exec(script_code, namespace, namespace)
  File "/usr/local/lib/python3.6/site-packages/visidata-2._3dev-py3.6.egg/EGG-INFO/scripts/vd", line 6, in <module>
  File "/usr/local/lib/python3.6/site-packages/visidata-2._3dev-py3.6.egg/visidata/main.py", line 215, in vd_cli
  File "/usr/local/lib/python3.6/site-packages/visidata-2._3dev-py3.6.egg/visidata/main.py", line 139, in main_vd
  File "/usr/local/lib/python3.6/site-packages/visidata-2._3dev-py3.6.egg/visidata/vdobj.py", line 37, in _vdfunc
  File "/usr/local/lib/python3.6/site-packages/visidata-2._3dev-py3.6.egg/visidata/data.py", line 193, in openSource
  File "/usr/local/lib/python3.6/site-packages/visidata-2._3dev-py3.6.egg/visidata/data.py", line 158, in openPath
  File "/usr/local/lib/python3.6/site-packages/visidata-2._3dev-py3.6.egg/visidata/loaders/postgres.py", line 21, in openurl_postgres
  File "/usr/local/lib/python3.6/site-packages/visidata-2._3dev-py3.6.egg/visidata/path.py", line 77, in __getattr__
AttributeError: 'PosixPath' object has no attribute 'path'
```

**Additional context**
Latest dev version, installed with pipx. and all dependencies.
Python 3.6.8
CentOS 7

anjakefala:
This should now be fixed on the `develop` branch! https://github.com/saulpw/visidata/commit/984b03a9049c9803b09f1e80735e4424832f7307

Can you please verify, @sfranky? =)

PR #493, opened 2020-03-31
[postgres-] parse given url

anjakefala:
This updates the postgres loader using the new vdPath. I am not sure where these properties got lost.

I know that this is kinda gross, and I am mostly opening this PR to start and track a conversation. This...does work if you adhere to the syntax.

Addresses #485

### DIFF
--- a/visidata/loaders/postgres.py
+++ b/visidata/loaders/postgres.py
@@ -18,13 +18,17 @@ def codeToType(type_code, colname):
 def openurl_postgres(url, filetype=None):
     import psycopg2
 
-    dbname = url.path[1:]
+    url = urlparse(url.given)
+
+    if url.scheme != 'postgres':
+        error('postgres loader requires a postgres url scheme')
+
     conn = psycopg2.connect(
-                user=url.username,
-                dbname=dbname,
-                host=url.hostname,
-                port=url.port,
-                password=url.password)
+                user=url.netloc.split(':')[0],
+                dbname=url.path[1:],
+                host=url.netloc.split(':')[1].split('@')[1],
+                port=url.netloc.split(':')[-1],
+                password=url.netloc.split(':')[1].split('@')[0])
 
     return PgTablesSheet(dbname+"_tables", sql=SQL(conn))
 


visidata/loaders/postgres.py
```diff
@@ -18,13 +18,12 @@ def codeToType(type_code, colname):
...
     import psycopg2
 
-    dbname = url.path[1:]
     conn = psycopg2.connect(
-                user=url.username,
-                dbname=dbname,
-                host=url.hostname,
-                port=url.port,
-                password=url.password)
+                user=url.given.split('/')[2].split(':')[0],
```
saulpw: Probably we can do something like `url = urlparse(url.given)`.  Maybe even this should be done automatically for the `openurl_` calls.

anjakefala: We do! This is in `vdPath().scheme`, but the result was very weird, lemme paste it. One moment.

anjakefala: >ParseResult(scheme='postgres', netloc='kefala:password@localhost:56714', path='/test', params='', query='', fragment='')

Ah, the issue is that `vdPath().scheme` only contains that `scheme` attribute.

anjakefala: Okay, this is slightly better to work with. I can work with this. I shall return tomorrow with a better update.

sfranky:
The error i get now when trying to connect to a postgres db is:
```
  File "/home/fragsot/.local/pipx/venvs/visidata/bin/vd", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/fragsot/visidata/bin/vd", line 6, in <module>
    visidata.main.vd_cli()
  File "/home/fragsot/visidata/visidata/main.py", line 213, in vd_cli
    rc = main_vd()
  File "/home/fragsot/visidata/visidata/main.py", line 133, in main_vd
    vs = openSource(src)
  File "/home/fragsot/visidata/visidata/vdobj.py", line 37, in _vdfunc
    return func(visidata.vd, *args, **kwargs)
  File "/home/fragsot/visidata/visidata/data.py", line 156, in openSource
    return vd.openPath(Path(p), filetype=filetype)  # convert to Path and recurse
  File "/home/fragsot/visidata/visidata/data.py", line 121, in openPath
    return getGlobals()[openfunc](p, filetype=filetype)
  File "/home/fragsot/visidata/visidata/loaders/postgres.py", line 28, in openurl_postgres
    return PgTablesSheet(dbname+"_tables", sql=SQL(conn))
NameError: name 'dbname' is not defined
```

anjakefala:
Closing PR because we are going to take a different approach.

Issue #495, opened 2020-04-01
Errors sorting pandas sheet in 2.x


ajkerrigan:
**Small description**

When attempting to sort a pandas sheet, I get a `RecursionError` followed by a `TypeError`.

**Expected result**

Pandas sheets are sortable.

**Actual result with screenshot**

Sorting a pandas sheet produces this stack trace:

```
Traceback (most recent call last):
   File "/Users/akerrigan/.local/pipx/venvs/visidata/lib/python3.7/site-packages/visidata/basesheet.py", line 123, in execCom…
     exec(code, vdglobals, LazyChainMap(vd, self))
   File "sort-asc", line 1, in <module>
     import math
   File "/Users/akerrigan/.local/pipx/venvs/visidata/lib/python3.7/site-packages/visidata/sort.py", line 15, in orderBy
     vd.addUndo(setattr, sheet, 'rows', copy(sheet.rows))
   File "/Users/akerrigan/.local/homebrew/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/copy.p…
     return _reconstruct(x, None, *rv)
   File "/Users/akerrigan/.local/homebrew/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/copy.p…
     if hasattr(y, '__setstate__'):
   File "/Users/akerrigan/.local/pipx/venvs/visidata/lib/python3.7/site-packages/visidata/loaders/_pandas.py", line 20, in __…
     return getattr(self.df, k)
   File "/Users/akerrigan/.local/pipx/venvs/visidata/lib/python3.7/site-packages/visidata/loaders/_pandas.py", line 20, in __…
     return getattr(self.df, k)
   File "/Users/akerrigan/.local/pipx/venvs/visidata/lib/python3.7/site-packages/visidata/loaders/_pandas.py", line 20, in __…
     return getattr(self.df, k)
   [Previous line repeated 489 more times]
 RecursionError: maximum recursion depth exceeded while calling a Python object
```

Working through that error with local modifications reveals a new error:

```
Traceback (most recent call last):
   File "/Users/akerrigan/code/visidata/visidata/threads.py", line 201, in _toplevelTryFunc
     t.status = func(*args, **kwargs)
   File "/Users/akerrigan/code/visidata/visidata/loaders/_pandas.py", line 99, in sort
     by_cols += [col.name for col in cols]
 TypeError: 'Column' object is not iterable
```

[...]

Issue #360, opened 2019-09-16
[pandas develop] PandasSheets cannot be modified


anjakefala:
**Small description**
When the work was done to enable selecting in pandas, [this line was added that effectively copies the dataframe](https://github.com/saulpw/visidata/blob/develop/visidata/loaders/_pandas.py#L71). This means that the source dataframe is not being modified.

**Expected result**
PandasSheets should be modifiable.

**Steps to reproduce with sample data and a .vd**
vd sample_data/co3.dta

And then try to edit any value.

**Additional context**
develop branch

anjakefala:
Fixed in #496!

PR #496, opened 2020-04-01
Pandas updates - sort and modify

ajkerrigan:
I hit an issue while trying to sort a pandas sheet (#495) and was working on it locally. While trying to avoid stomping on other related issues or previous fixes, I came across the modification issue in #360. This PR attempts to resolve both issues, though the changes are in separate commits so we can tweak or divide the PR as needed.

The modification issue appeared to stem from the fact that the pandas `read_stata` function returns a DataFrame with an Int64Index instead of the default RangeIndex. The warning during edits doesn't seem to show up when using the pandas loader with CSV/JSON/Parquet files. I dug into the difference between Int64Index and RangeIndex a bit ([reference](https://pandas.pydata.org/pandas-docs/stable/user_guide/advanced.html#int64index-and-rangeindex)), and it seems like converting Int64Index to RangeIndex could cover what `reload()` is already trying to do without forcing a `reset_index()`. There is probably a better way to handle it, but this seems like forward progress at least :)

### DIFF
--- a/visidata/loaders/_pandas.py
+++ b/visidata/loaders/_pandas.py
@@ -70,7 +70,9 @@ class PandasSheet(Sheet):
             self.df = readfunc(str(self.source), **options('pandas_'+filetype+'_'))
 
         # reset the index here
-        if type(self.df.index) is not pd.RangeIndex:
+        if type(self.df.index) is pd.Int64Index:
+            self.df.index = pd.RangeIndex(len(self.df.index))
+        elif type(self.df.index) is not pd.RangeIndex:
             self.df = self.df.reset_index()
 
         self.columns = [
--- a/visidata/loaders/_pandas.py
+++ b/visidata/loaders/_pandas.py
@@ -6,30 +6,20 @@ class DataFrameAdapter:
     def __init__(self, df):
         import pandas as pd
         assert isinstance(df, pd.DataFrame)
-        self._df = df
+        self.df = df
 
     def __len__(self):
-        return len(self._df)
+        return len(self.df)
 
     def __getitem__(self, k):
         if isinstance(k, slice):
-            return DataFrameAdapter(self._df.iloc[k])
-        return self._df.iloc[k]
+            return DataFrameAdapter(self.df.iloc[k])
+        return self.df.iloc[k]
 
     def __getattr__(self, k):
-        return getattr(self._df, k)
-
-    def __setattr__(self, k, val):
-        if k.startswith('_'):
-            super().__setattr__(k, val)
-        else:
-            setattr(self._df, k, val)
-
-    def __getstate__(self):
-        return self._df
-
-    def __setstate__(self, df):
-        self.__init__(df)
+        if 'df' not in self.__dict__:
+            raise AttributeError(f"'{self.__class__.__name__}' has no attribute '{k}'")
+        return getattr(self.df, k)
 
 # source=DataFrame
 class PandasSheet(Sheet):
--- a/visidata/loaders/_pandas.py
+++ b/visidata/loaders/_pandas.py
@@ -17,6 +17,9 @@ class DataFrameAdapter:
         return self.df.iloc[k]
 
     def __getattr__(self, k):
+        # This check helps avoid infinite recursion trouble
+        if k == 'df':
+            raise AttributeError
         return getattr(self.df, k)
 
 
@@ -90,9 +93,9 @@ class PandasSheet(Sheet):
         'Sort rows according to the current self._ordering.'
         by_cols = []
         ascending = []
-        for cols, reverse in self._ordering[::-1]:
-            by_cols += [col.name for col in cols]
-            ascending += [not reverse] * len(cols)
+        for col, reverse in self._ordering[::-1]:
+            by_cols.append(col.name)
+            ascending.append(not reverse)
         self.rows.sort_values(by=by_cols, ascending=ascending, inplace=True)
 
     def _checkSelectedIndex(self):
--- a/visidata/loaders/_pandas.py
+++ b/visidata/loaders/_pandas.py
@@ -6,22 +6,30 @@ class DataFrameAdapter:
     def __init__(self, df):
         import pandas as pd
         assert isinstance(df, pd.DataFrame)
-        self.df = df
+        self._df = df
 
     def __len__(self):
-        return len(self.df)
+        return len(self._df)
 
     def __getitem__(self, k):
         if isinstance(k, slice):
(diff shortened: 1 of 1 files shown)


visidata/loaders/_pandas.py
```diff
@@ -17,6 +17,9 @@ def __getitem__(self, k):
         return self.df.iloc[k]
 
     def __getattr__(self, k):
+        # This check helps avoid infinite recursion trouble
+        if k == 'df':
+            raise AttributeError
```
anjakefala:
```
if hasattr(y, '__setstate__'):
```

I bumped into, too!

This occurs because we do not handle missing attributes. I think I prefer we checking if self has the attribute `df`, since that captures the root of the problem, and raise AttributeError otherwise.

Like here:
https://github.com/saulpw/visidata/pull/491/files

anjakefala: @saulpw, we should audit the code for all implementations of `__getattr__`, and ensure that each handles what happens when attributes are missing.

visidata/loaders/_pandas.py
```diff
@@ -67,7 +70,9 @@ def reload(self):
             self.df = readfunc(str(self.source), **options('pandas_'+filetype+'_'))
 
         # reset the index here
-        if type(self.df.index) is not pd.RangeIndex:
+        if type(self.df.index) is pd.Int64Index:
```
anjakefala, line 71: Really nice find!

Should we keep an eye out for other Indexing types?

visidata/loaders/_pandas.py
```diff
@@ -90,9 +95,9 @@ def sort(self):
         'Sort rows according to the current self._ordering.'
         by_cols = []
         ascending = []
-        for cols, reverse in self._ordering[::-1]:
-            by_cols += [col.name for col in cols]
-            ascending += [not reverse] * len(cols)
+        for col, reverse in self._ordering[::-1]:
```
anjakefala, line 96: @saulpw, can `self._ordering[::]` contain multiple columns? Probably when you have multiple key columns, and do `g[`?

anjakefala:
This is AMAZING and WELCOME. It makes me really, really happy! :purple_heart:

visidata/loaders/_pandas.py
ajkerrigan: True @anjakefala - calling out `df` explicitly looks hacky and I'm sure we can make it cleaner. It would probably make sense to have `DataFrameAdapter` be a `pd.DataFrame` subclass if not for the initial loading time. The way you linked could work and be more flexible (in case we need to add other local attributes someday).

visidata/loaders/_pandas.py
saulpw: Yes; each item in `_ordering` should be a tuple of `(list(columns), reverse_bool)`.

anjakefala: Judging from @ajkerrigan's linked issue report: https://github.com/saulpw/visidata/issues/495

It has not been a `list(columns)`, so we should look into that!

visidata/loaders/_pandas.py
ajkerrigan: Not really sure to be honest. Converting Int64-->Range here seems clear enough here (assuming we want/need a RangeIndex), but I don't know enough to extrapolate beyond that.

visidata/loaders/_pandas.py
ajkerrigan: I did do some tests locally with `z[`/`z]` to make sure this behaved when sorting by multiple columns. Because in those cases, `self._ordering` looks like:

```[(<Column 1>, True), (<Column 2>, False)]```

Still definitely possible that I missed a piece of behavior, but as far as I can tell sorting works with single or multiple columns.

saulpw: My apologies, I changed it a few months ago in 9da019ca4, to be `(column, reverse_bool)`.  So that is the issue, that the "interface" for `_ordering` changed and the pandas loader did not change with it.  I wonder if `_ordering` should lose its private `_` and become a documented interface?  Other loaders/plugins that override provide custom sort functionality would need access to it as well, so we should probably formalize it.

saulpw (changes requested):
Thanks for figuring this out, @ajkerrigan!  Let's find the right pattern for the AttributeError raise (since we need it in a couple places like this), and apply it here.  Otherwise looks great, I think you saved us hours of debugging pandas indexes!

ajkerrigan:
> Thanks for figuring this out, @ajkerrigan! Let's find the right pattern for the AttributeError raise (since we need it in a couple places like this), and apply it here. Otherwise looks great, I think you saved us hours of debugging pandas indexes!

Thanks for the review! I'm with you, sorting out the attribute behavior in a repeatable way seems handy. During my local edits I experimented both with tweaks to `__getattr__` and overriding `__copy__`.

After a bit more reading, I think we can address the recursion issue in a more general way by overriding `__getstate__` and `__setstate__`. That last commit is a working sample, but it's probably worth talking through a bit. We'll get there! :)

visidata/loaders/_pandas.py
anjakefala: I can also confirm that it does work!

anjakefala: Agreed about formalizing `_ordering`, and properly documenting it..

visidata/loaders/_pandas.py
anjakefala: Okay, perhaps this is the best we can do for now, and we can mature it from there. I did test a few formats, and things seem to be fine. =) Perhaps folks who care about other formats can chip in if they notice something. Thanks!

anjakefala:
@ajkerrigan, I am fine with implementing `__getstate__` and `__setstate__`! I am a bit confused why that is more general than handling the potential of a missing attribute? Could not another attribute be invoked that is missing from `_df`?

ajkerrigan:
I don't think that an attribute missing from `_df` is the problem, it's more the death spiral of trying to direct `__getattr__` to `self.df` when that hasn't been set yet. So another way to handle that would be to raise that `AttributeError` from `__getattr__` when `self.df` doesn't exist yet. That's closer to my original change, the difference is that we don't really care which attribute we're looking for (we're not checking `k`). We just know that `__getattr__` is a fallback lookup, and we're making sure `self.df` exists before we look there. Maybe that's less weird than the custom state logic?

anjakefala:
Looks great! Merging in. =) Thanks *so* much, @ajkerrigan!

PR #497, opened 2020-04-06
Account for all visible rows when expanding a column

ajkerrigan:
This changes the behavior of the `expand-cols` family of commands to account for all visible rows by default, rather than just the active/cursor row.

This change is something I need, but forgot about until @geekscrapy mentioned it recently. That said, it's a bunch of changes you all didn't ask for. So I'm perfectly happy to migrate it locally to `~/.visidatarc` or a plugin if the default case should only look at the active/cursor row.

For the record, this is the sort of behavior that tripped me up and that I'm looking to change. Given a file like:

```json
[
  {
    "AvailabilityZones": [
      "us-east-1c",
      "us-east-1e"
    ]
  },
  {
    "AvailabilityZones": [
      "us-east-1e",
      "us-east-1d",
      "us-east-1c"
    ]
  }
]
```

I will get two new columns if I expand from the first row:

image

image

But three if I expand from the second row:

image

Ideally, I want the number of new columns to be independent of the active row. I also want it to create as many columns as needed to host the longest sequence (for list/tuple columns) or union of all keys (for dict columns).

### DIFF
--- a/visidata/pyobj.py
+++ b/visidata/pyobj.py
@@ -1,6 +1,13 @@
+from functools import singledispatch
+
 from visidata import *
 
 option('visibility', 0, 'visibility level')
+isNull = lambda v: any((
+    v is None,
+    v is options.null_value,
+    isinstance(v, TypedWrapper)
+))
 
 
 class PythonSheet(Sheet):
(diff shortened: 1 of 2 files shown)


visidata/pyobj.py
```diff
@@ -1,6 +1,13 @@
+from functools import singledispatch
+
 from visidata import *
 
 option('visibility', 0, 'visibility level')
+isNull = lambda v: any((
```
saulpw: isNull must be created at runtime or else changing `null_value` has no effect.  The idiom in VisiData is to call `isNullFunc()` to return `isNull` just before using it (but only once if using it in a loop, because getting an option value is expensive); that way it should be handled consistently across all cases considering nulls

saulpw (changes requested):
This makes sense to me, @ajkerrigan. I've not used `@singledispatch` before so it's good to see it in action.  I think we can take this if we clean up the options issue.

visidata/pyobj.py
ajkerrigan: Thanks @saulpw, good call! I had tried to use `isNullFunc` but ran into issues trying to null-check columns with non-hashable values like lists. I was trying to keep this change local, but now that you mention it maybe I'll try tweaking that `isNullFunc` to cover this case without changing its intent.

visidata/wrappers.py
```diff
@@ -10,7 +10,10 @@
 option('null_value', None, 'a value to be counted as null', replay=True)
 
 def isNullFunc():
-    return lambda v,nulls=set([None, options.null_value]): v in nulls or isinstance(v, TypedWrapper)
+    return lambda v,nulls=set([None, options.null_value]): any((
+        *(v == n for n in nulls),
```
ajkerrigan: Handling it this way, we don't need to see if a non-hashable column value (like a list) is in the `nulls` set.

saulpw: We need to be really careful when modifying isNullFunc; if the returned function is even a little slower, this will make several operations (save, freq, describe, etc) significantly slower for larger datasets. So we'd have to measure this change carefully to see how it performs.  By inspection though I think it's definitively slower, as the isinstance() passed to any() is calculated every time, whereas in the original there is an "early out" if the v value is already null.

Maybe this should be `lambda v, nullv=options.null_value: v is None or v == nullv or isinstance(v, TypedWrapper)` with a special case of `lambda v: v is None or isinstance(v, TypedWrapper)` for when nullv is just None.  We'd have to measure to see if there's a performance penalty for this.  I agree though that it would be best for null checks on lists/dicts/objects to not raise an exception.

ajkerrigan: Yikes you're right, I did some testing with moderately sized files and a describe was ~5 times slower. I updated it based on your feedback and it looks like the performance is back to what it was before, but I'll have to give it a bit more of a workout to be sure.

visidata/wrappers.py
```diff
@@ -10,10 +10,10 @@
...
 
 def isNullFunc():
-    return lambda v,nulls=set([None, options.null_value]): any((
-        *(v == n for n in nulls),
-        isinstance(v, TypedWrapper)
-    ))
+    nullv = options.null_value
+    if nullv is None:
+        return lambda v: v is None or isinstance(v, TypedWrapper)
+    return lambda v: v is None or v == nullv or isinstance(v, TypedWrapper)
```
saulpw: I think we need to bind `nullv=nullv` in the lambda, due to Python's lexical scoping?  `nullv` is unique to this purpose now but that would be a frustrating bug to come across later :)

ajkerrigan: True, surprises like that are no fun. Good catch :)

visidata/wrappers.py
ajkerrigan: With some local benchmarks on a bigger/slower dataset, I was seeing timings of just over 7 seconds to describe a sheet with my "slow" change. That same describe reliably took ~5 seconds before this PR, and the same with your suggestions. Happy to test anything else, but from what I can tell the performance looks good now.

saulpw (approved):

frosencrantz:
Looking forward to this, @ajkerrigan.  Thanks for fixing this!

PR #498, opened 2020-04-06
[undo] add undo for rename-col-x family

anjakefala:


saulpw (approved):
Looks great, nicely done!

PR #506, opened 2020-04-13
[postgres-] fix loader bugs

anjakefala:
Bonus: add fix for postgres loader.

Pull request is in draft mode:
* so that we can review the design for the change
* so we can test the postgres bugfix

### DIFF
--- a/visidata/loaders/postgres.py
+++ b/visidata/loaders/postgres.py
@@ -18,6 +18,7 @@ def codeToType(type_code, colname):
 def openurl_postgres(url, filetype=None):
     import psycopg2
 
+    url = urlparse(url.given)
     dbname = url.path[1:]
     conn = psycopg2.connect(
                 user=url.username,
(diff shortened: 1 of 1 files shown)


visidata/loaders/postgres.py
```diff
@@ -18,6 +18,10 @@ def codeToType(type_code, colname):
 def openurl_postgres(url, filetype=None):
     import psycopg2
 
+    if not url.is_url() and url.scheme != 'postgres':
```
saulpw: These should be guarantees for `openurl_` functions.  If they're not in some case then let's fix/error closer to the root--loaders shouldn't have to check this.

visidata/path.py
```diff
@@ -178,6 +178,16 @@ def exists(self):
     def scheme(self):
         if self.is_url():
             return urlparse(self.given).scheme
+        else:
+            error('trying to urlparse a non-url')
+
+    @property
```
saulpw: Let's not commit to a `Path.urlparse`property in the API right now. Maybe we instead adjust the `openurl_` API to take a `url` in the first place.  For this commit let's inline urlparse in the postgres loader and we can revisit later.

saulpw (changes requested):

saulpw (approved):

saulpw:
Thanks @anjakefala for fixing these!

PR #514, opened 2020-04-21
[cmdlog-all] do not log in gD the global cmdlog or internal sheets

### LINKED ISSUE #510: [develop] 2.x gD CommandLog should behave similar to the 1.5.2 CommandLog
**Small description**
The `gD` `commandlog-all` is meant to be a comprehensive CommandLog for the full session. It provides the functionality of what was the standard CommandLog in 1.5.2.

However, it has some logs that are unexpected.

Examples include:
*  `delete-row` when performed on a CommandLog sheet
*  `cmdlog-all`

**Expected result**
I think that the command filters that are configured to occur on regular `cmdlog-sheet`s should also occur on `cmdlog-all`.

I think it is important that 2.x `cmdlog-all` mimic as much of the old behaviour of 1.5.2 `cmdlog` as possible.

Screenshot from 2020-04-18 22-13-32

**Additional context**
v2.-4

This is a sister issue to #426

anjakefala:
- global cmdlog behaviour should be consistent with VisiData v1.5.2
cmdlog

Closes #510

### DIFF
--- a/visidata/cmdlog.py
+++ b/visidata/cmdlog.py
@@ -64,6 +64,9 @@ def isLoggableCommand(longname):
             return False
     return True
 
+def isLoggableSheet(sheet):
+    return sheet is not vd.cmdlog and not isinstance(sheet, (OptionsSheet, ErrorSheet))
+
 
 @Sheet.api
 def moveToRow(vs, rowstr):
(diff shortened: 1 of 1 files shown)


anjakefala:
The location was chosen so that the global cmdlog could have its own cmdlog.

Also, it seems like there was an accidental double-indent? That is why the other rows were affected.

saulpw:
> Also, it seems like there was an accidental double-indent? That is why the other rows were affected.

Double-indent usually means an if-statement and removed, and the de-indent was elided to minimize the diff.  (This is also how half-indents happen.)

saulpw (approved):

Issue #530, opened 2020-04-25
VisiData only recognizes keys present in the first object of JSON/YAML lists


jsvine:
**Small description**

As described in issue title; details below.

**Expected result**

VisiData should render all key-value information for top-level JSON/YAML dicts.

**Actual result with screenshot**

See below.

**Steps to reproduce with sample data and a .vd**

Command: `vd https://raw.githubusercontent.com/simonw/museums/master/museums.yaml -f yaml`

Compare the keys for the VisiData-loaded sheet vs. the keys in the second item in the YAML file. It seems that VisiData checks only the first item in a JSON/YAML list before determining the columns for the entire sheet. To me, that's unexpected and has, a few times, given me an incorrect impression of the data in a file. 

Ideally, these loaders would, as they iterated through a file, append columns to the sheet as they encountered new keys.

**Additional context**

`v2.-4dev`, revision `96e5ab66` (Sunday, April 19)

ajkerrigan:
I can take a look at this. I ran into similar behavior when expanding columns, so the changes from #497 could be useful as a reference here.

anjakefala:
@ajkerrigan @jsvine Can this be closed? =)

jsvine:
Yes, thanks! And my apologies: I appear to have been mistaken re. the JSON aspect. Thank you both!

PR #533, opened 2020-04-26
[loaders] Have the yaml loader derive from json

ajkerrigan:
This addresses #530.

It looks to me like the JSON loader already handles progressively adding columns, while the YAML loader only takes columns from the first row. Thinking about it for a minute, I'm wondering if it makes sense for the YAML loader to derive from the JSON one.

On the plus side, this would mean some more consistency between the two loaders. We also get row dives for free in YAML since the JSON loader derives from `PythonSheet`. There may be good reasons that this change is a bad idea though -  feedback welcome one way or the other :smile:

### DIFF
--- a/visidata/loaders/yaml.py
+++ b/visidata/loaders/yaml.py
@@ -1,21 +1,19 @@
 from visidata import *
 
 
-class YamlSheet(Sheet):
+class YamlSheet(JsonSheet):
     def iterload(self):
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Looks good to me, AJ!  This is perfect. Thanks for adding the Progress too.

Issue #522, opened 2020-04-23
[wishlist] Add support for specifying postgres connection paramaters not over URL


danielcynerio:
Currently all paramaters for postgres are over a URL, this is problematic for passwords for that can sometimes not be encoded over the URL.

When trying to access RDS using an AWS access token the password is a whole URL of itself and this clashes with url parsing for passwords.

I think this can be fixed by adding support for an rds:// url that will support the parameters that RDS needs and use the AWS python library to generate the password. Alternatively, it could be useful to get the password from a separate command line argument.

I wouldn't mind making the patch, just would like to know what is preferred. I am running a patched version locally to allow me to work with it for now.

saulpw:
Hi @danielcynerio,

I think this is a reasonable suggestion.  If you can get it working as a patch to the postgres loader (adding an `openurl_rds` function that recomposes the postgres url for the postgres loader), then that should work fine.

In the meantime, one possible workaround is to put the password in quotes:

`postgres://you:"the whole password"@...`

This worked for me with the recent imap loader, so it might work for you in this case too.  Let me know if it does!  And we'd be interested in an rds loader regardless.

danielcynerio:
Hey @saulpw,
I added a openurl_rds function. The current format of the URL is:
 rds://db_user@hostname:port/region/dbname
since RDS needs a region in addition to the db name.

Currently it only supports Postgres over RDS. Maybe the URL should account for different DBs (f.e. MySQL) in the future?

I haven't made any changes to the documentation or anything like that. Say if you need anything else.

The work around to putting the password in quotes wouldn't work due to all the weird things in the token. The password looks something like:
`rdspostgres.cdgmuqiadpid.us-west-2.rds.amazonaws.com:5432/?Action=connect&DBUser=jane_doe&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Expires=900...`

saulpw:
Wow, that is really gross :(

PR #536, opened 2020-04-26
Add support for connecting directly to RDS

danielcynerio:
The URL has the following format:
    rds://db_user@hostname:port/region/dbname
And assumes that the AWS IAM for the user is configured properly.

### DIFF
--- a/visidata/loaders/postgres.py
+++ b/visidata/loaders/postgres.py
@@ -15,6 +15,25 @@ def codeToType(type_code, colname):
     return anytype
 
 
+def openurl_rds(url, filetype=None):
+    import boto3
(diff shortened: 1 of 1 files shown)


danielcynerio:
Closes #522

saulpw:
Hi @danielcynerio, thanks for putting this together!  I've been mulling this over, and in general I don't want to take on boto3 as a dependency, but in this case it's such a tidy little change that I think we'll merge it.  At some point we may want to consider consolidating AWS plugins like this and @ajkerrigan's `vds3`.  Thanks again!

danielcynerio:
Sure! Thank you for the quick response!

PR #539, opened 2020-04-30
[undo] add undo for reload-sheet

### LINKED ISSUE #528: [wishlist] Allow undo-ing a reload?
Command log and data: [undo-reload-log-and-data.zip](https://github.com/saulpw/visidata/files/4532253/undo-reload-log-and-data.zip) 

(Note: To fully reproduce, you'll have to press `U` to run the final step, which  seems not to have been logged.)

In the command log above, I load a minimal CSV, `addcol-expr ="a"`, press reload, and then try to undo. My expectation had been that the reload would have been undone, returning me to the pre-reload state ... but instead the statusline tells me that addcol-expr is "undone":

> 1› test-100-nums| addcol-expr undone

... which is unexpected, since that column does not exist on the (not-undone) reloaded sheet.

Ideally (for me), VisiData would simply be able to undo a reload. That said, I realize there may be other constraints / architectural priorities. So, as a backup, perhaps VisiData could just tell me that "reload commands cannot be undone"? That still leaves the question of how to handle undoing/redoing actions that came before the reload, though...

anjakefala:
Closes #528

It will be important to review, if there are other relevant elements of the state, or if everything is derived from `rows` and `columns`.

So far tested:
* modification
* expression columns
* selection
* added rows


visidata/undo.py
```diff
@@ -95,6 +95,15 @@ def _undo():
             c.name = name
     vd.addUndo(_undo)
 
+@VisiData.api
+def addUndoReload(vd, rows, cols):
+    oldrows = rows
+    oldcolumns = [copy(c) for c in cols]
```
saulpw: I think just `oldcolumns = cols` is fine, like `rows`, since reload assigns a new list to both rows and columns.

saulpw (approved):
Simple enough!

Issue #529, opened 2020-04-25
Plain JSON lists cannot be loaded


jsvine:
**Small description**

VisiData seems unable to load/render JSON files that are "plain" lists, such as:

```json
[
    "tt0177489",
    "tt0098168",
    "tt0233146",
    "tt0100816",
    "tt0122427"
]

```

**Expected result**

VisiData should be able to handle any valid JSON file.

**Actual result with screenshot**

See below.

**Steps to reproduce with sample data and a .vd**

[test.json.zip](https://github.com/saulpw/visidata/files/4532256/test.json.zip)

```sh
vd test.json
```

When I try running this, the right-statusline correctly tells me that there are 5 rows, but the core view is entirely blank. The main statusline says this:

> 1› test| [19x] AttributeError: 'NoneType' object has no attribute 'keycol'

The traceback for which is:

```sh
Traceback (most recent call last):
  File "[...]/src/visidata/visidata/basesheet.py", line 183, in checkCursorNoExceptions
    return self.checkCursor()
  File "[...]/src/visidata/visidata/sheets.py", line 508, in checkCursor
    if self.cursorCol.keycol:
AttributeError: 'NoneType' object has no attribute 'keycol'
```

The issue seems to be that VisiData is expecting the list items to be dicts, when instead they are just strings. Running vd test-2.json (see below; identical to test.json except that it wraps those strings in a simple dict) does work just fine.

[test-2.json.zip](https://github.com/saulpw/visidata/files/4532258/test-2.json.zip)

**Additional context**

`v2.-4dev`, revision `96e5ab66` (Sunday, April 19)

ajkerrigan:
Let me take a look at this one - I just ran into the same error in a few cases where I'm piping JSON output to VisiData. Saving the same JSON and opening it as a file worked with no trouble. Could be related.

anjakefala:
![image](https://user-images.githubusercontent.com/7489659/80781878-de22d980-8b28-11ea-8b06-7cf1b39098fc.png)
                             :1st_place_medal:

ajkerrigan:
Hey @jsvine I think with #541 landed this should be kicked, but let me know if you notice any fishiness :)

jsvine:
A big thanks!

PR #541, opened 2020-05-01
[loaders] handle non-dict rows in json data

ajkerrigan:
Wrap non-dict JSON rows in an object with a consistent key name, so they can be
treated the same as dict rows.

I considered a couple ways this change could blow things up and have tried to account for them, which probably means there are other funky edge cases I _haven't_ thought of. That said, this change renders a weirdo JSON file like:

```json
[
  [1, 2, 3],
  "meh",
  2,
  true,
  {"meh": "moo"},
  {"meh": [1, 2, 3]}
]
```

as:

```
      | meh ║                                                                                                                                                                                   
 [3]  |    ⌀║
 meh  |    ⌀║
 2   #|    ⌀║
 True |    ⌀║
     ⌀| moo ║
     ⌀| [3] ║
```

And handles @jsvine's test case from #529 as we'd expect:

```
[
    "tt0177489",
    "tt0098168",
    "tt0233146",
    "tt0100816",
    "tt0122427"
]
```

```
           ║                                                                                                                                                                                    
 tt0177489 ║
 tt0098168 ║
 tt0233146 ║
 tt0100816 ║
 tt0122427 ║
```

The default column name for non-dict data will be an empty string, though I added an option for it since it seems like something you'd want to be able to customize.

Tangential change: don't check for `self.cursorCol.keyCol` if `self.cursorCol` is `None`.

Alternate approaches or suggested tweaks to this one are welcome, thanks!

### DIFF
--- a/visidata/loaders/json.py
+++ b/visidata/loaders/json.py
@@ -6,7 +6,7 @@ from visidata import *
 
 option('json_indent', None, 'indent to use when saving json')
 option('json_sort_keys', False, 'sort object keys when saving to json')
-option('json_default_colname', '', 'column name to use for non-dict rows')
+option('default_colname', '', 'column name to use for non-dict rows')
 
 
 def open_json(p):
@@ -22,7 +22,7 @@ class JsonSheet(PythonSheet):
     def iterload(self):
         self.colnames = {}  # [colname] -> Column
         self.columns = []
-        self.defaultColName = options.json_default_colname
+        self.defaultColName = options.default_colname
 
         try:
             with self.source.open_text() as fp:
--- a/visidata/loaders/json.py
+++ b/visidata/loaders/json.py
@@ -6,6 +6,7 @@ from visidata import *
 
 option('json_indent', None, 'indent to use when saving json')
 option('json_sort_keys', False, 'sort object keys when saving to json')
+option('json_default_colname', '', 'column name to use for non-dict rows')
 
 
 def open_json(p):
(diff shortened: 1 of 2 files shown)


saulpw (commented):
Thanks for fixing these issues, @ajkerrigan!  This looks good.  The only thing I'm not sure about is the option for the default colname.  I'm wondering if we shouldn't have a more generic option that would work across multiple loaders, or somehow ties into the `force_valid_colnames` option.  Also, note that options prefixed with the loader filetype are usually passed through to the Python library for that loader.  Sometimes this is a complete passthrough (like in the csv loader), even here the `indent` and `sort_keys` options are passed through to the respective objects that use those options.  So maybe just `default_colname`?  I'm not completely sure, but it seems somewhat more consistent.  The option can stay in the json loader while it's the only one using it.

ajkerrigan:
> Thanks for fixing these issues, @ajkerrigan! This looks good. The only thing I'm not sure about is the option for the default colname. I'm wondering if we shouldn't have a more generic option that would work across multiple loaders, or somehow ties into the `force_valid_colnames` option. Also, note that options prefixed with the loader filetype are usually passed through to the Python library for that loader. Sometimes this is a complete passthrough (like in the csv loader), even here the `indent` and `sort_keys` options are passed through to the respective objects that use those options. So maybe just `default_colname`? I'm not completely sure, but it seems somewhat more consistent. The option can stay in the json loader while it's the only one using it.

Ah, good thoughts @saulpw . I figured it was equally likely that I was over or underthinking the option piece :). A more generic name that lives in the json loader at least for now seems like a fine compromise.

saulpw (approved):
Looks good!

PR #542, opened 2020-05-01
Start fixing tests

tsibley:
Tests seem to have been failing for a while on `develop`. :worried:  This PR starts to fix them. Details are in the commit messages, but so far this:

1. Addresses a few missing `tests/golden/` files.
2. Adds a test runner for running tests individually so that the extent of failures can be seen more easily.

My motivation for starting this work is that I wanted to fix a tricky bug, but I couldn't use the test suite to see if the fix unexpectedly broke things. I'll be submitting my proposed fix in a separate PR.

PR #543, opened 2020-05-01
Fix column name crashes

### LINKED ISSUE #344: TSV file with column name "length" causes TypeError
**Small description**
Files in TSV format containing a column named `length` cannot be loaded.

**Expected result**
See content of TSV file.

An empty file is shown. In the footer line it says:
```
TypeError: 'property' object is not callable
```

Create a file named `test.tsv` with this content:
```
length
1
```
Then, try to open it:
```
vd test.tsv
```

**Additional context**
version 1.5.2

tsibley:
Fixes crashes that occur in `SequenceSheet` loaders (csv, tsv, xls(x), fixed width) when particular column names like `length` or `_fields` or `__name__` are present. Lots of details in the commit messages. Re-fixes #344.

~This PR is based off of #542 so that I could check that no new tests failed; it includes the commits in that PR. A listing of just the commits unique to this PR is here: https://github.com/tsibley/visidata/compare/start-fixing-tests...fix-column-name-crashes.~ _Now independent._

An alternate solution to these crashes could be to not use `namedlist` for user-provided data and remove its use in `SequenceSheet`. It could still be used for hardcoded, VisiData-provided sheets. This seemed liked a bigger change to the internals to me, so I didn't do it, but it is perhaps less fraught with peril in the long run.


anjakefala:
Hey @tsibley! 

Very nice catch. Thank you so much for looking into this so deeply. 

So far, I have validated the bugs by using your tests. Thank you for including a test for each case. Since each has a unique error, I agree that it was the right thing to do. 

`length` is definitely the most reasonable one to handle. 

In terms of including the test-individually.sh commits I *think* it might make the merge a bit messy. Let's remove them before we press the button. =) I am now going to review the actual change.

anjakefala:
Oh! One thing, would you mind moving that `sample_data` to `tests/`?

We like to keep the data that is specific for testing niche use cases in the `tests/` folder. The sample data folder is more for trying out VisiData with various file formats.

anjakefala:
@saulpw, for when we all decide whether to go with these changes, or remove namedlist from `SequenceSheet`, here is where `namedlist` is used:

```
class SequenceSheet(Sheet):                                                                                                                                                                  'For sheets which use ColumnItem on rows that are Python sequences (list, namedtuple, etc).'                                                                                        
     def setCols(self, headerrows):                                                                                                                                                               self.columns = []                                                                                                                                                               
         [...]                                                                                                                                                 
         self._rowtype = namedlist('tsvobj', [(c.name or '_') for c in self.columns])                                                                                                    
                                                                                                                                                                                         
     def newRow(self):                                                                                                                                                                   
         return self._rowtype()                                                                                                                                                          
                                                                                                                                                                                         
     def addRow(self, row, index=None):                                                                                                                                                  
             [..]                                                                                                                      
             self._rowtype = namedlist('tsvobj', [(c.name or '_') for c in self.columns])                                                                                                
         if type(row) is not self._rowtype:                                                                                                                                              
             row = self._rowtype(row)                                                                                                                                                      
```

https://github.com/saulpw/visidata/pull/543/commits/d566fa8408040df6675f1e7f729369fc5cf68a87 seems like a very reasonable change to me. I am less comfortable making a call for https://github.com/saulpw/visidata/pull/543/commits/6e250da0ccee44af1a0b2e673c605ca86784187d.

tests/golden/column-name-_fields.csv
```diff
@@ -0,0 +1,4 @@
+_fields,foo,bar,baz
```
saulpw, line 1: Is it reasonable to combine these into a single test?  Or do they have to be the first column?

visidata/utils.py
```diff
@@ -95,8 +91,4 @@ def __setattr__(self, k, v):
             except ValueError:
                 super().__setattr__(k, v)
 
-    for i, attrname in enumerate(fieldnames):
```
saulpw, line 98: Wow, this has been there since 2017.  Thanks for cleaning this up.

tests/golden/column-name-_fields.csv
anjakefala: They do not have to be! A single test would be less laborious to scale. The reason why i thought separate tests made sense was bc if it regresses, their individual error messages are not very clear. But it is a lot of files and I really am fine either way.

tsibley: Exactly. The different files cover the different ways the column names can conflict in the implementation. While they could all be put into one file, it would make debugging a regression harder. In fact, I started my local debugging that way, but it wasn't immediately clear when I fixed one way because the others still failed.

tsibley:
Repushed without the commits from #542 and with data moved from `sample_data/` into `tests/`.

saulpw (approved):
Thanks @tsibley for fixing this!  I'm sorry if you were burnt by that bug in the first place.

saulpw:
By the way, I think it's a great idea to fix TsvSheet/etc to not use namedlist, but I've been putting it off because of the meta-sheets.  Maybe once 2.0 is out..

Issue #500, opened 2020-04-07
[defermods] move back into vdcore


anjakefala:
We are going to bring back defermods and `vsh` into vdcore. `vsh` will possibly be a plugin.

Defermods is going to be integrated with:
* DirSheet
* SqliteSheet

# Design Proposals
`z^S` will save to original source. `^S` will save to provided input source.

* We set `defer = True` in Sheets whose deferred modifications we want to track. 
    * Pros: If we regret this later, it will be easier to pull out the logic into a plugin. 
* We created a `DeferredSheet` that deferred Sheets will co-inherit from.

Cons for `defer = True` design.

The plugin code works by overriding core functions and commands.

E.g.

```
 @Sheet.api                                                                                                                                                                              
 def newRow(self):                                                                                                                                                                       
     row = type(self)._rowtype()                                                                                                                                                         
     self._deferredAdds[self.rowid(row)] = row                                                                                                                                           
     return row  
```
```
Sheet.addCommand('d', 'delete-row', 'vd.cliprows = [(sheet, cursorRowIndex, deleteRows([cursorRow]))]; not sheet.defermods or cursorDown(1)') 
```

This makes sense for a plugin, since plugins was **all-or-nothing**, but it makes it tricky for the re-integration. If we use the `defer = True` strategy, we will end up needing to do something like this.

```
@Sheet.api
def newRow(self):
    if self.defermods:
           [do defermods stuff]
     else:
           [do regular sheet stuff]
```

I am not sure if that is actually better/easier to excise than keeping all of the deferred logic in its own sheet.

saulpw:
If possible, something like this would be ideal:

```
@Sheet.api
def newRow(self):
    if not self.defermods:
        return newRow.__wrapped__(self)
    [do defermods stuff]
```

This is how plugins should extend functions in general, calling the original for the base case.

anjakefala:
```
plugins/defermods.py:def newRow(self):
plugins/vtask.py:    def newRow(self, **kwargs):
vgit/diff.py:    def newRow(self, linenum, refnum, line):
visidata/cmdlog.py:    def newRow(self, **fields):
visidata/loaders/csv.py:    def newRow(self):
visidata/loaders/json.py:    def newRow(self):
visidata/loaders/png.py:    def newRow(self):
visidata/metasheets.py:    def newRow(self):
visidata/sheets.py:    def newRow(self):
visidata/sheets.py:    def newRow(self):
visidata/sheets.py:    def newRow(self):
```

[...]

PR #545, opened 2020-05-04
[defer] move defermods back into vdcore

anjakefala:
- [X] move deferadds back in
- [X] move deferdels back in
- [x] move defermods back in
- [x] undo for deferadds
- [x] undo for deferdels
- [x] undo for defermods
- [x] ensure that reload clears deferDicts
- [ ] write an sqlite saver test
- [x] decide command structure for `commit-sheet` and `save-sheet`
- [ ] document the exact minimum boilerplate for an extension that wants to use defermods

Current outstanding bugs:
* once committed, added rows fall to the bottom
* you cannot undo a committed defer

Closes https://github.com/saulpw/visidata/issues/500


visidata/column.py
```diff
@@ -179,6 +179,13 @@ def _calcIntoCache(self, row):
     def getValue(self, row):
         'Memoize calcValue with key sheet.rowid(row)'
 
+        if getattr(self.sheet, 'defer', False):
```
anjakefala, line 182: I decided to have the defer=True case first because it was so much shorter than the defer=False case.

visidata/data.py
```diff
@@ -183,31 +186,40 @@ def loadInternalSheet(klass, p, **kwargs):
     return vs
 
 @Sheet.api
-def deleteBy(self, func):
+def deleteBy(self, func, commit=False):
```
anjakefala, line 190: This `commit` variable got added because `addNewRows` (see above) uses `deleteBy` in its undo. It essentially forces a committed delete, even if defer=True.

We want that if a user undoes a deferredAdd, that that added row just disappears. We do not want it turning red. 

It could possibly be handled better than this. Or the variable could be better named.

visidata/defermods.py
```diff
@@ -0,0 +1,198 @@
...
+    if oldval != val:
+        rowid = col.sheet.rowid(row)
+        if rowid not in col.sheet._deferredMods:
+            rowmods = {}
+            col.sheet._deferredMods[rowid] = (row, rowmods)
+        else:
+            _, rowmods = col.sheet._deferredMods[rowid]
+        rowmods[col] = val
+
+        def _undoCellChanged(col, row, oldval):
```
anjakefala, line 52: Is the logic flow in this `_undo` clear?

visidata/defermods.py
```diff
@@ -0,0 +1,198 @@
...
+
+    if not cstr:
+        fail('no diffs')
+
+    if options.confirm_overwrite:
+        confirm('really %s? ' % cstr)
+
+    sheet.putChanges(path, adds, mods, deletes)
+
+Sheet.addCommand('z^S', 'commit-sheet', 'commit()')
```
anjakefala, line 198: Is this the best way to make `commit-sheet` available? And then we teach users that to save an sqlite file to source, you press `z^S`? That feels funky. 

We could also bind it to `^S` on SqliteSheet. But that would mean that every deferred sheet would need to remember to bind it.

visidata/loaders/sqlite.py
```diff
@@ -44,7 +44,7 @@ def iterload(self):
             r = self.execute(conn, 'SELECT COUNT(*) FROM "%s"' % tblname).fetchall()
             rowcount = r[0][0]
             for row in Progress(self.execute(conn, 'SELECT * FROM "%s"' % tblname), total=rowcount-1):
-                yield row
+                yield list(row)
```
anjakefala, line 45: This needs to be mutable.

visidata/sheets.py
```diff
@@ -954,7 +954,7 @@ def quit(vd, *sheets):
 globalCommand('S', 'sheets-stack', 'vd.push(vd.sheetsSheet)', 'open Sheets Stack: join or jump between the active sheets on the current stack')
 globalCommand('gS', 'sheets-all', 'vd.push(vd.allSheetsSheet)', 'open Sheets Sheet: join or jump between all sheets from current session')
 
-BaseSheet.addCommand('^R', 'reload-sheet', 'reload(); recalc(); status("reloaded")', 'reload current sheet'),
+BaseSheet.addCommand('^R', 'reload-sheet', 'reload(); _dm_reset(); recalc(); status("reloaded")', 'reload current sheet'),
```
anjakefala: Every sheet now does the `_dm_reset()` upon reload, regardless if `defer=True` or not. I also feel uncomfortable with this.

anjakefala:
Current outstanding bugs:
*  once committed, added rows fall to the bottom
* you cannot undo a committed defer

visidata/clipboard.py
```diff
@@ -52,8 +52,14 @@ def syscopyCells(sheet, col, rows):
 
 @Sheet.api
 def delete_row(sheet, rowidx):
-    oldrow = sheet.rows.pop(rowidx)
-    vd.addUndo(sheet.rows.insert, rowidx, oldrow)
+    oldrow = sheet.rows[rowidx]
+
+    if not getattr(sheet, 'defer', False):
```
anjakefala: Is this the neatest way to perform this check?

visidata/defermods.py
saulpw: I like the intentionality behind `z Ctrl+S`, though I agree it's a hard one to communicate to newbs.  We could put an explicit statement in the docs telling people about the design decision and how to set the binding for SqliteSheet (or other specific sheets) in their visidatarc.

saulpw (approved):
Looks good!  Thanks so much for doing this work, @anjakefala.

saulpw:
Merge defermods back into core.

Issue #282, opened 2019-05-15
Select starting table in postgres from command-line


bknowles:
Related to #214 , I'd like to be able to specify a starting table when opening a particular postgres database.  The database I'm trying to work with has dozens of tables, some of which have millions of rows.  So, it takes so long to list all the tables and how many rows are in each table that the database query times out and I can't ever select the particular table that I want to open inside of the database.

Note that Postico doesn't have any problems with opening the database and getting a list of tables, but at the top level it doesn't really need to know about how many rows are in each table.  So, I know I can get a list of tables quickly, and that would at least let me specify in advance from the command line which particular table I want to open with visidata.

Please let me know if you want to see any examples of the kind of thing I'm looking for.

saulpw:
Thanks for the bug report, @bknowles. I thought `COUNT(*)` would be a fast operation, and it's weird that the queries are taking so long.  Would you be able to check out the source code, and comment out the nrows column?  https://github.com/saulpw/visidata/blob/stable/visidata/loaders/postgres.py#L73
We still want to provide a generic way to open straight to a particular sheet, but that might be a short-term workaround.  Let me know if it works, and we might provide an option to disable it.

bknowles:
@saulpw -- I did a git clone on the repo at https://github.com/saulpw/visidata.git, made your suggested one line change, and then did a "pip3 install ." in the visidata directory.

I can confirm that by omitting the count operation, we can load the list of tables almost instantaneously, and then I can select the table in question and pull up the contents.

So, this seems like a good solution to me!

bknowles:
Note that, in our case, this database has 41 tables, and several of those tables have several million rows.  So, doing a count of the number of rows in each table will take quite a while.  And doing all those count operations in parallel in multiple asynchronous threads will probably make that something close to 41 times worse.

Fortunately, this is just a read replica server and no one else is using it right now.

saulpw:
Thanks for running that experiment, @bknowles, and glad it will work for you.  Though, a database really should not require a linear scan to count the number of rows in a table.  I'm wondering if it doesn't have more to do with the number of concurrent threads as you suggest, maybe some kind of locking contention.  If you are able, it would be great to try the same test but with 41 tables that each have only a few hundred rows.  If that is slow, we can throttle the number of concurrent threads vd can have (probably a good idea anyway).  But if it turns out that the number of rows really does impact `COUNT(*)`, we should look into that, because someone must be doing something wrong.

bknowles:
[...]

PR #549, opened 2020-05-08
Change the postgres loader to load an estimate of row count for improved performance

danielcynerio:
In Postgres (and other DBs) doing a regular `SELECT COUNT(*) FROM tablename` [0], is rather slow. For Postgres there is different way to estimate the row list that is much quicker[1]. This pull request implements that for the sheet table on connection to a DB which, at least on the DB I'm working against, brought it down from 30s to instantly.

Solves #282 as well.

[0] https://wiki.postgresql.org/wiki/Slow_Counting
[1] https://wiki.postgresql.org/wiki/Count_estimate

### DIFF
--- a/visidata/loaders/postgres.py
+++ b/visidata/loaders/postgres.py
@@ -82,7 +82,13 @@ class PgTablesSheet(Sheet):
     rowtype = 'tables'
 
     def reload(self):
-        qstr = "SELECT table_name, COUNT(column_name) AS ncols FROM information_schema.columns WHERE table_schema = 'public' GROUP BY table_name"
+        qstr = '''
+            SELECT relname table_name, column_count.ncols, reltuples::bigint est_nrows
+                FROM pg_class, pg_namespace, (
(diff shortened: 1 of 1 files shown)


danielcynerio:
BTW I changes the name of the column from ncols to est_ncols (for estimated). I could change to the original ncols, but wanted it to be clearer that it is an estimation and may be slightly off.

IMHO, if you want to exact row count, just go into the DB itself.

saulpw (approved):
Hey @danielcynerio, thanks for putting this together.  I like the idea, it's very clean.  I can't test it myself right now, but we'll merge it on your word that it works.  Have you been using the postgres loader a lot with VisiData?

danielcynerio:
Yeah, that is my primary use case. I'm tying to use it instead of a workflow of "data from product" -> "export to excel" -> "import to google sheets", and instead to do the pivot tables / searches I want in visidata directly.

I still have a few issues with the JSON column expansion that I am trying to figure out to make it fully work for me. I will open bug reports on those later.

PR #552, opened 2020-05-10
Fix expand column to work on a json column in postgres

danielcynerio:
Currently when expanding a JSON column in a Postgres sheet you get an AttributeError:
`
File "/home/daniel/.local/pipx/venvs/visidata/lib64/python3.8/site-packages/visidata/basesheet.py", line 123, in execCommand
  File "expand-col", line 1, in <module>
    import math
  File "/home/daniel/.local/pipx/venvs/visidata/lib64/python3.8/site-packages/visidata/pyobj.py", line 19, in expand_cols_deep
    newcols = _addExpandedColumns(col, rows, sheet.columns.index(col))
  File "/home/daniel/.local/pipx/venvs/visidata/lib64/python3.8/site-packages/visidata/pyobj.py", line 77, in _addExpandedColumns
    col.sheet.addColumn(c, idx+i+1)
AttributeError: 'NoneType' object has no attribute 'addColumn'
`

This fixes the error. I wasn't sure if this is how you would fix it, but if not just point me in the right direction.

### DIFF
--- a/visidata/loaders/postgres.py
+++ b/visidata/loaders/postgres.py
@@ -69,12 +69,9 @@ class SQL:
 
 
 def cursorToColumns(cur, sheet):
-    cols = []
+    sheet.columns = []
     for i, coldesc in enumerate(cur.description):
-        c = ColumnItem(coldesc.name, i, type=codeToType(coldesc.type_code, coldesc.name), sheet=sheet)
-        cols.append(c)
-    return cols
-
+        sheet.addColumn(ColumnItem(coldesc.name, i, type=codeToType(coldesc.type_code, coldesc.name)))
 
 
 # rowdef: (table_name, ncols)
@@ -98,7 +95,7 @@ class PgTablesSheet(Sheet):
             r = cur.fetchone()
             if r:
                 self.addRow(r)
-            self.columns = cursorToColumns(cur, self)
+            cursorToColumns(cur, self)
             self.setKeys(self.columns[0:1])  # table_name is the key
 
             for r in cur:
@@ -113,7 +110,7 @@ class PgTable(Sheet):
             r = cur.fetchone()
             if r:
                 self.addRow(r)
-            self.columns = cursorToColumns(cur, self)
+            cursorToColumns(cur, self)
             for r in cur:
                 self.addRow(r)
 
--- a/visidata/loaders/postgres.py
+++ b/visidata/loaders/postgres.py
@@ -68,10 +68,10 @@ class SQL:
             cur.close()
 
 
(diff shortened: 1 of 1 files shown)


visidata/loaders/postgres.py
```diff
@@ -68,10 +68,10 @@ def query_async(self, qstr, callback=None):
             cur.close()
 
 
-def cursorToColumns(cur):
+def cursorToColumns(cur, sheet):
     cols = []
     for i, coldesc in enumerate(cur.description):
-        c = ColumnItem(coldesc.name, i, type=codeToType(coldesc.type_code, coldesc.name))
+        c = ColumnItem(coldesc.name, i, type=codeToType(coldesc.type_code, coldesc.name), sheet=sheet)
```
saulpw: Ah, yes, this may fix the immediate issue but leaves the door open for other bugs.  The proper fix is to always use `addColumn` to add a column to a sheet.  That, or calling `sheet.recalc()` after all columns have been added manually.  But I prefer the addColumn method; recalc may be deprecated at some point.

visidata/loaders/postgres.py
```diff
@@ -92,7 +92,7 @@ def reload(self):
             r = cur.fetchone()
             if r:
                 self.addRow(r)
-            self.columns = cursorToColumns(cur)
+            self.columns = cursorToColumns(cur, self)
```
saulpw: This is where we should iterate through the cursorToColumns result and call `addColumn` for each one.

danielcynerio:
This fixes it using addColumns

visidata/loaders/postgres.py
```diff
@@ -98,7 +95,7 @@ def reload(self):
             r = cur.fetchone()
             if r:
                 self.addRow(r)
-            self.columns = cursorToColumns(cur)
+            cursorToColumns(cur, self)
```
saulpw, line 98: sure, that'll work.  Whenever we pass 'self' into a function we should consider if it should be a member function (i.e. called as `self.cursorToColumns(cur))`.  Probably in this case it would be neater but since this is isolated in an optional loader I'm less picky about style things like that.

saulpw (approved):

saulpw:
Thanks, @danielcynerio !

visidata/loaders/postgres.py
danielcynerio: I was considering just removing the function, as it is basically two lines and doesn't really remove complexity from the sheets. Just didn't feel comfortable to do a whole refactor of the code there yet.

danielcynerio:
Happily!

PR #555, opened 2020-05-11
[postgres] Support diving into database objects (such as JSON) as sheets

danielcynerio:
This is very useful to allow to dive into db objects, such as JSON columns, as sub sheets.

I implemented this only for postgres, but the code is pretty generic. Not sure if the solution shouldn't just be to subclass from a sheet that would provide this generically?

### DIFF
--- a/visidata/loaders/postgres.py
+++ b/visidata/loaders/postgres.py
@@ -113,5 +113,6 @@ class PgTable(Sheet):
             cursorToColumns(cur, self)
             for r in cur:
                 self.addRow(r)
-
+    
 PgTablesSheet.addCommand(ENTER, 'dive-row', 'vd.push(PgTable(name+"."+cursorRow[0], source=cursorRow[0], sql=sql))', 'open postgres table in current row')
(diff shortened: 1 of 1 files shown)


danielcynerio:
BTW, was thinking on this more, and I think it might make sense to do this on all "anytype" columns or add a new type "obj" or something like that.

saulpw:
You know, I think we can put nearly this exact command on Sheet itself.  The only thing that's off is the name of the sheet, which should not be the str() of the row, but use the keystr or rownum.  If you want to put this together in this PR, you can, otherwise I can do it.  Thanks for your initiative on this, I know some people will be pleased with this decision.

danielcynerio:
What if the sheet doesn't have a keystr?

saulpw:
That's when we use the rownum :)

danielcynerio:
I am OK with you working on the PR :)
I will probably not be able to work on it for a while and this will require me diving into parts I haven't looked at yet

anjakefala:
@saulpw Did end up doing that work, so we can close this. =)

PR #566, opened 2020-05-21
[loader-html] minor fix, string formatting & option to exclude table title

geekscrapy:
Fixed string formatting issue for the html table name when saving
Added the option to exclude the sheetname when saving sheet as html table

### DIFF
--- a/visidata/loaders/html.py
+++ b/visidata/loaders/html.py
@@ -1,7 +1,7 @@
 import html
 from visidata import *
 
-option(name='include_html_title', default=True, helpstr='include sheet name as the table header when saving as HTML')
+option('include_html_title', True, 'include sheet name as the table header when saving as HTML')
 
 class HtmlTablesSheet(IndexSheet):
     rowtype = 'sheets'  # rowdef: HtmlTableSheet (sheet.html = lxml.html.HtmlElement)
--- a/visidata/loaders/html.py
+++ b/visidata/loaders/html.py
@@ -1,6 +1,7 @@
 import html
 from visidata import *
 
+option(name='include_html_title', default=True, helpstr='include sheet name as the table header when saving as HTML')
 
 class HtmlTablesSheet(IndexSheet):
     rowtype = 'sheets'  # rowdef: HtmlTableSheet (sheet.html = lxml.html.HtmlElement)
@@ -104,7 +105,8 @@ def save_html(vd, p, *vsheets):
     with open(p, 'w', encoding='ascii', errors='xmlcharrefreplace') as fp:
         for sheet in vsheets:
 
-            fp.write('<h2 class="sheetname">%s</h2>\n'.format(sheetname=html.escape(sheet.name)))
+            if options.include_html_title:
+                fp.write('<h2 class="sheetname">{sheetname}</h2>\n'.format(sheetname=html.escape(sheet.name)))
 
             fp.write('<table id="{sheetname}">\n'.format(sheetname=html.escape(sheet.name)))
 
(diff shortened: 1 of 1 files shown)


anjakefala:
Hey @geekscrapy! Thanks for fixing the html formatting.

Would you mind formatting the `option()` call so that it has a similar style to other `option()` calls?

For example: https://github.com/saulpw/visidata/blob/develop/visidata/loaders/graphviz.py#L4

Feel free to push force to update this commit with that change. Once it has that, I am okay for merging this in.

geekscrapy:
I think I have added it to the existing commit!

saulpw (commented):
Let's rename the option to `html_title`, as loader-specific options should start with the filetype.  Also I'm wondering it if shouldn't be a string, see separate code comment.

visidata/loaders/html.py
```diff
@@ -104,7 +105,8 @@ def save_html(vd, p, *vsheets):
     with open(p, 'w', encoding='ascii', errors='xmlcharrefreplace') as fp:
         for sheet in vsheets:
 
-            fp.write('<h2 class="sheetname">%s</h2>\n'.format(sheetname=html.escape(sheet.name)))
+            if options.include_html_title:
```
saulpw: What if this became:
```
hdr = html.escape(options.html_title.format(sheet=self, vd=vd))
fp.write('<h2>'+hdr+'</h2>')
```

Then the default would be an empty string (empty h2 seems fine to me), and you could set it to `{sheet.name}`, and someone else could set it to `"TPS Report"`.

geekscrapy:
Personally, I don't agree with the approach of having `h2` there by default. It's extra stuff that isn't expected - and it's easier to remove now than later.

Also, I think I get what you are going for with just having it as a string, however this seems to be overly static if it is set in options. I usually just rename the sheet in `Sheets Sheet`. Something I can control per sheet and quickly changeable - what happens if I use `saveSheets` (disclosure, I've not used/tested this!)? It's quicker and easier to rename in `Sheets Sheet` than diving into each sheet and editing it in options, plus it's quickly apparent what the title will be from within the sheet view.

I get this would be helpful in batch processing, but it has a lot more limitations like having to dive into options if `dup-rows` is used to get a subset of the rows, the same table name will be used, but the data is for a different purpose.

My personal inclination is to have a boolean option to include or not include the title, and the title be set to the sheet name. If a title is included, then we add it within an `h2`.

I can do it whichever way you'd prefer though :D

saulpw:
Did you see that using `options.html_title="<h2>{sheet.name}</h2>"` would accomplish exactly what you want?  (if we remove the tags also from the code).   This format approach is infinitely more flexible and is already in use by the statusbar, and has no downsides as far as I can tell.

geekscrapy:
I guess the issue I have is that it's not directly obvious to the user where this header value should be set. Personally, I wouldn't look in options to change the header name, I'd assume it was taken from the sheets title.

I get that it's more flexible, however I think it'd be a bit hidden and a bit more static, especially if the user sets the option in their .rc. But then again, you could include all kinds of data in the header then (# of rows etc.). My two cents 😇 I can do it either way you want though as you have z future thoughts! 👍

saulpw:
Yes, I would like it to be the more flexible way, please.  Boolean options are a design smell, in my opinion. I hope we can document and advertise the feature to overcome its less-than-ideal placement.

geekscrapy:
I _think_ it's as you wanted 👍 This now allows complete HTML formatting and inclusion of `sheet` or `vd` objects in the html header.
I left out `html.escape` otherwise the user wouldn't be able to use html in their options definition.

saulpw (approved):
Looks good!  I'm not sure if the default should be with or without header, but we can leave it like this for now.  It serves as a very useful example, in any case.  (though maybe it should be `html_header`..ah well)

PR #574, opened 2020-05-27
Added save_xls and save_xlsx

geekscrapy:
Thought I'd add it as I'll need it soon!

### DIFF
--- a/visidata/loaders/xlsx.py
+++ b/visidata/loaders/xlsx.py
@@ -55,7 +55,7 @@ class XlsSheet(SequenceSheet):
 
             
 @VisiData.api
-def save_xlsx(vd, p, vs, *vsheets):
+def save_xlsx(vd, p, *sheets):
     import openpyxl
 
     wb = openpyxl.Workbook()
@@ -85,7 +85,7 @@ def save_xlsx(vd, p, vs, *vsheets):
 
 
 @VisiData.api
-def save_xls(vd, p, vs, *vsheets):
+def save_xls(vd, p, *sheets):
     import xlwt
 
     wb = xlwt.Workbook()
--- a/visidata/loaders/xlsx.py
+++ b/visidata/loaders/xlsx.py
@@ -53,6 +53,61 @@ class XlsSheet(SequenceSheet):
         for rownum in Progress(range(worksheet.nrows)):
             yield list(worksheet.cell(rownum, colnum).value for colnum in range(worksheet.ncols))
 
+            
+@VisiData.api
+def save_xlsx(vd, p, vs, *vsheets):
+    import openpyxl
+
+    wb = openpyxl.Workbook()
+    wb.remove_sheet(wb['Sheet'])
+
+    for vs in vsheets or [vs]:
+        ws = wb.create_sheet(title=vs.name)
+
+        headers = [col.name for col in vs.visibleCols]
+        ws.append(headers)
+
+        for dispvals in vs.iterdispvals(format=False):
+
+            row = []
+            for v in dispvals.values():
+                if type(v) == date:
+                    v = datetime.datetime.fromtimestamp(int(v.timestamp()))
+                elif not type(v) in [int,float,str]:
+                    v = str(v)
+                row.append(v)
+            ws.append(row)
+
+    wb.active = ws
+
+    wb.save(filename=p)
+    status(f'{p} save finished')
+
+
+@VisiData.api
+def save_xls(vd, p, vs, *vsheets):
+    import xlwt
+
+    wb = xlwt.Workbook()
+
+    for vs in vsheets or [vs]:
+
+        ws1 = wb.add_sheet(vs.name[:32]) # sheet name can not be longer than 32 characters
+
+        for col_i, col in enumerate(vs.visibleCols):
+            ws1.write(0, col_i, col.name)
+
+        for r_i, dispvals in enumerate(vs.iterdispvals(format=False)):
+            r_i += 1
+            for c_i, v in enumerate(dispvals.values()):
+
+                if not type(v) in [int,float,str]:
+                    v = str(v)
+
+                ws1.write(r_i, c_i, v)
+
+    wb.save(p)
+    status(f'{p} save finished')
 
 vd.filetype('xlsx', XlsxIndexSheet)
 vd.filetype('xls', XlsIndexSheet)
--- a/visidata/loaders/xlsx.py
+++ b/visidata/loaders/xlsx.py
@@ -70,12 +70,13 @@ def save_xlsx(vd, p, *sheets):
         for dispvals in vs.iterdispvals(format=False):
 
             row = []
-            for v in dispvals.values():
-                if type(v) == date:
+            for col, v in dispvals.items():
+                if col.type == date:
                     v = datetime.datetime.fromtimestamp(int(v.timestamp()))
-                elif not type(v) in [int,float,str]:
+                elif not isNumeric(col):
                     v = str(v)
                 row.append(v)
+
             ws.append(row)
 
     wb.active = ws
@@ -97,13 +98,9 @@ def save_xls(vd, p, *sheets):
         for col_i, col in enumerate(vs.visibleCols):
             ws1.write(0, col_i, col.name)
 
-        for r_i, dispvals in enumerate(vs.iterdispvals(format=False)):
+        for r_i, dispvals in enumerate(vs.iterdispvals(format=True)):
             r_i += 1
             for c_i, v in enumerate(dispvals.values()):
-
-                if not type(v) in [int,float,str]:
-                    v = str(v)
-
                 ws1.write(r_i, c_i, v)
 
     wb.save(p)
--- a/visidata/loaders/xlsx.py
+++ b/visidata/loaders/xlsx.py
@@ -61,7 +61,7 @@ def save_xlsx(vd, p, *sheets):
     wb = openpyxl.Workbook()
     wb.remove_sheet(wb['Sheet'])
 
-    for vs in vsheets or [vs]:
+    for vs in sheets:
         ws = wb.create_sheet(title=vs.name)
 
         headers = [col.name for col in vs.visibleCols]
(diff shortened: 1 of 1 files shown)


visidata/loaders/xlsx.py
```diff
@@ -53,6 +53,61 @@ def iterload(self):
         for rownum in Progress(range(worksheet.nrows)):
             yield list(worksheet.cell(rownum, colnum).value for colnum in range(worksheet.ncols))
 
+            
+@VisiData.api
+def save_xlsx(vd, p, vs, *vsheets):
```
anjakefala: This is not a full review, but please note that the api for this should have the form

```
def save_foo(vd, p, *sheets)
```

anjakefala:
Thanks for sharing this with us @geekscrapy!

I am not Saul, but I can make comments on things I notice when I have time to review. =)

visidata/loaders/xlsx.py
geekscrapy: Yup, done!

geekscrapy:
By the way, I couldn't figure out how to get a datetime object from vd.date, hence the fromtimestamp hack. Anyone got a better way?
I did this because the openpyxl only accepts datetime objects as a date field...

visidata/loaders/xlsx.py
```diff
@@ -53,6 +53,61 @@ def iterload(self):
...
 
+            
+@VisiData.api
+def save_xlsx(vd, p, *sheets):
+    import openpyxl
+
+    wb = openpyxl.Workbook()
+    wb.remove_sheet(wb['Sheet'])
+
+    for vs in vsheets or [vs]:
```
anjakefala: Remove the `or [vs]` :)

anjakefala:
@geekscrapy to answer that question, can you walk me through why you chose `iterdispvals(format=False)` as your iterator? What kinds of objects do you want to save to the excel sheet?

visidata/loaders/xlsx.py
```diff
@@ -53,6 +53,61 @@ def iterload(self):
...
+    status(f'{p} save finished')
+
+
+@VisiData.api
+def save_xls(vd, p, *sheets):
+    import xlwt
+
+    wb = xlwt.Workbook()
+
+    for vs in vsheets or [vs]:
```
anjakefala: Here, too!

visidata/loaders/xlsx.py
```diff
@@ -53,6 +53,61 @@ def iterload(self):
...
+
+        ws1 = wb.add_sheet(vs.name[:32]) # sheet name can not be longer than 32 characters
```
anjakefala, line 96: Thanks for the comment.

visidata/loaders/xlsx.py
geekscrapy: I thought I changed these!! Done now

visidata/loaders/xlsx.py
geekscrapy: Np

geekscrapy:
> RE `iterdispvals(format=False)`

If you provide base types like int, float, and datetime objects to `openpyxl` instead of strings it sets the values as the given type in the Excel file. So a py int will be rendered as a number in Excel, the same will be done with python datetime (avoids the user having to convert the column to a date type - which can be filtered as a date once in a table). Of course, it would have been easier just to have all values passed to Excel as strings, but if the functionality is there to make the data more useable from Excel, I thought it was worth it. For me I export a lot of date columns to Excel so by having this I didn't have to do any post processing to convert the column to a date type in Excel.

geekscrapy:
For reference, I did have a version which also included using the formatted date from `options.disp_date_fmt` which worked nicely, however it required converting that date format (strftime % format) to an Excel/openpyxl friendly version (`YYYY` == `%Y`) and I couldn't find a quick way to do this. I tried to ask the question but I didn't get a relevant answer: https://stackoverflow.com/questions/62042852/convert-the-string-y-m-d-to-yyyy-mm-dd-for-use-in-openpyxl-namedstyle-num?noredirect=1#comment109734131_62042852

I eventually got it working with a crude lookup dict for conversion between the two, but then I wanted to do the same thing for the float format but that was a *lot* more complicated so I just left it all out...

If you have any thoughts on this I'd love to hear how you think we can get the formatting into the xls+x files!

visidata/loaders/xlsx.py
```diff
@@ -53,6 +53,61 @@ def iterload(self):
...
+
+        for col_i, col in enumerate(vs.visibleCols):
+            ws1.write(0, col_i, col.name)
+
+        for r_i, dispvals in enumerate(vs.iterdispvals(format=False)):
+            r_i += 1
+            for c_i, v in enumerate(dispvals.values()):
+
+                if not type(v) in [int,float,str]:
```
saulpw: I think this should be `isNumeric(col)`, which would also detect currency, date, and vlen types.  `visidata.date` inherits from `datetime.datetime` btw, so did you try using it directly instead of doing any conversions?

saulpw:
This looks good, @geekscrapy, thanks for putting this together.  I'm sure other people will find it useful (and I even have need to save a .xlsx from time to time).   It would be great to have formatting and other metadata saveable, but as I consider xls/x as more of an input format than either an output or a working format, it's fine to just save the values. Though we will need to add the caveat to our docs that .xls/.xlsx aren't "roundtrip-able"; you can't load up a .xlsx, change a single number, and save it back out without losing all of the formatting etc.

So, assuming this works (I haven't tested it), aside from the typing issue I mentioned separately, we can merge it.

visidata/loaders/xlsx.py
geekscrapy: Hey, yea, I tried giving it `date` directly but unfortunately openpyxl looks at the `type` when deciding if it is a object it can consume and convert to an Excel type. In this case it's looking for datetime.datetime.

I'll have a look and try to test with `isNumeric`. I have a feeling it'll break currency formatting....!

geekscrapy:
Given what comes from my `isNumeric` tests, I may have to change it to just using `format=True` to be able to cover all use cases, including any future types that maybe made.

geekscrapy:
Ok, so I did a quick test and I may have found a related bug, Path object not having a `.type`.

```
  File "/Users/Documents/pyv/py3/lib/python3.7/site-packages/visidata/threads.py", line 201, in _toplevelTryFunc
  File "/Users/.visidata/save_excel.py", line 50, in save_xls
    if not isNumeric(v):
  File "/Users/Documents/pyv/py3/lib/python3.7/site-packages/visidata/_types.py", line 72, in isNumeric
    return col.type in (int,vlen,float,currency,date)
  File "/Users/Documents/pyv/py3/lib/python3.7/site-packages/visidata/path.py", line 78, in __getattr__
    r = getattr(self._path, k)
AttributeError: 'PosixPath' object has no attribute 'type'
```

Without a fix to give `Path` a `.type` the following works, but I would assume that you'll want the Path.type fixed instead of hacking it 🙂:

```
try: v = v if isNumeric(v) else str(v)
except: v = str(v)
```

If this is fixed, we can use isNumeric.

geekscrapy:
> Ok, so I did a quick test and I may have found a related bug, Path object not having a `.type`.
> 
> ```
> Traceback (most recent call last):
>   File "/Users/Documents/pyv/py3/lib/python3.7/site-packages/visidata/threads.py", line 201, in _toplevelTryFunc
>     t.status = func(*args, **kwargs)
>   File "/Users/.visidata/save_excel.py", line 50, in save_xls
>     if not isNumeric(v):
>   File "/Users/Documents/pyv/py3/lib/python3.7/site-packages/visidata/_types.py", line 72, in isNumeric
>     return col.type in (int,vlen,float,currency,date)
>   File "/Users/Documents/pyv/py3/lib/python3.7/site-packages/visidata/path.py", line 78, in __getattr__
>     r = getattr(self._path, k)
> AttributeError: 'PosixPath' object has no attribute 'type'
> ```
> 
> Without a fix to give `Path` a `.type` the following works, but I would assume that you'll want the Path.type fixed instead of hacking it 🙂:
> 
> ```
> try: v = v if isNumeric(v) else str(v)
> except: v = str(v)
> ```
> 
> If this is fixed, we can use isNumeric.

This also happens with `str`:

```
  File "/Users/matthew.holley/Documents/pyv/py3/lib/python3.7/site-packages/visidata/threads.py", line 201, in _toplevelTryFunc
  File "/Users/matthew.holley/.visidata/save_excel.py", line 22, in save_xlsx
    elif not isNumeric(v):
  File "/Users/matthew.holley/Documents/pyv/py3/lib/python3.7/site-packages/visidata/_types.py", line 72, in isNumeric
    return col.type in (int,vlen,float,currency,date)
AttributeError: 'str' object has no attribute 'type'
```

saulpw:
`isNumeric` takes a column, not a value :)

anjakefala:
@geekscrapy What is the status of this PR?

geekscrapy:
I've just got to take another look at using isNumeric correctly 🙄

geekscrapy:
Added the column checks to use isNumeric. I also modified `save_xls` to use `format=True` as it doesn't care to use python types anyway as far as I've seen.
Round up is, `save_xlsx` can provide numeric data which is set as a numeric field by excel standards, same for date, and everything else is converted to a string for safety.

anjakefala:
Thanks @geekscrapy! =)

geekscrapy:
A note that Excel (and probably some other software) doesn't seem to like sheet names which are longer than 32 characters, or duplicate sheet names. I thought this was just a limitation imposed by xls, but it seems to affect xlsx too.

Any thoughts on how to intelligently truncate the sheet names to 32 characters? I can obviously do a `[:32]` but Excel also doesn't like 2 sheets with the same name...

Potential solutions:
- Refuse to create a malformed xls file. This would show a message to the user describing why the file hasn't been saved (please change the sheet name to less than 32 chars
- Truncate the name to 32 chars and add an auto increment int at the end of the sheet name if there is a sheet with the same name to limit duplicate sheet names

anjakefala:
Which would you prefer personally as a user @geekscrapy?

I would personally prefer the truncation + appending, along with a status informing me, but I am not the one who might be saving a lot of xls/xlsx files. Especially, considering that most people are not intentionally naming sheets within VD.

geekscrapy:
Ok, I'll go along with your solution:

- truncate
- append an auto-increment
- message to user upon full save advising them

I'll do this for both xls and xlsx

Issue #577, opened 2020-05-29
Error writing rstatus when repeating a command with no keystrokes


ajkerrigan:
**Small description**
If the last command in the cmdlog has an empty keystrokes column, `repeat-input` throws `TypeError`s trying to update the status line.

**Expected result**
Commands are repeatable whether they were issued with keystrokes or long names.

**Actual result with screenshot**
If you get an unexpected error, please include the full stack trace that you get with `Ctrl-E`.

```
 Traceback (most recent call last):
   File "/Users/akerrigan/code/visidata/visidata/statusbar.py", line 183, in drawRightStatus
     rstatus = ' '+rstatus
 TypeError: can only concatenate str (not "NoneType") to str
```

**Steps to reproduce with sample data and a .vd**
Please attach the commandlog (saved with `Ctrl-D`) to show the steps that led to the issue.
See [here](http://visidata.org/docs/save-restore/) for more details.

```
vd sample_data/benchmark.csv
<space> select-after
<space> repeat-input
```

Trying to replay a .vd with these commands _won't_ show the error. Inside VisiData, the `keystrokes` value for `select-after` is `None`. That gets converted to an empty string when we save the command log. So we can only see the error in cases like repeating or redoing a longname command interactively.

**Additional context**
Please include the version of VisiData.

2.-4dev (70e837dbca12f4df56485a7a5da0a0483e78439b)

saulpw:
Nice find, @ajkerrigan.  There is of course an easy fix, but before we do that, I've had the thought many times that maybe the right status should show the longname instead of the keystroke (which it's been doing since before we had longnames).  As long as we're here, what do you (and others if they're listening) think about that?  Which would you prefer to see after each command?  Or maybe both?  Or neither?

anjakefala:
@saulpw Keystrokes take up less space which may matter with smaller terminal sizes ... that is one thing to check out.

Beyond that, that might be a really good thing to do. It gives a much clearer idea of "this is the most recent action that happened". Not every command has a keystroke, but *every* command has a longname.

ajkerrigan:
Ooh interesting idea :). My gut reaction is to want the long name only if there's no keystroke. I'm thinking seeing something like `go-down (j)` would be extra terminal space for little value, but always having _some_ indicator of the last action seems handy.

It could be a configurable option to always display the long name perhaps, but I can't tell if that's a useful option or a copout to avoid making a decision :).

saulpw:
[...]

PR #578, opened 2020-05-29
Have keystrokes default to empty string (#577)

ajkerrigan:
This is one way to address #577. Some other possibilities further down would be using `keystrokes or ''` [here](https://github.com/saulpw/visidata/blob/c281f5685db733100dea32b378485c9461ce4b4e/visidata/cmdlog.py#L158) or `rstatus or ''` [here](https://github.com/saulpw/visidata/blob/c281f5685d/visidata/statusbar.py#L183). Feels like using an empty string default catches edge cases earlier, but it may be too broad a change.

### DIFF
--- a/visidata/basesheet.py
+++ b/visidata/basesheet.py
@@ -102,7 +102,7 @@ class BaseSheet(Extensible):
         'Width of the current sheet, in single-width characters'
         return self._scr.getmaxyx()[1] if self._scr else 80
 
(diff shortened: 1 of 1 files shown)


ajkerrigan:
Reopening this after talking to @anjakefala - while it's probably not the "right" fix it's still forward progress.

saulpw:
I decided to do the default in the status, since the keystrokes staterizer is hardcoded anyway.  Thanks for spurring me on to fix it :)

PR #583, opened 2020-06-03
Allow vfake to generate deterministic values

ajkerrigan:
This adds a couple options to help make vfake's output more reproducible.

I've used this to help with tasks like diffing multiple versions of scrubbed/faked output, or joining multiple faked files that have related values (without needing to go back to the originals).

I initially looked at persisting `fakeMap` but that felt pretty clunky. Using a value-specific seed at generation time works out better from a few different angles.


plugins/vfake.py
```diff
@@ -43,11 +47,16 @@ def setValuesFromFaker(col, faketype, rows):
     fakeMap[options.null_value] = options.null_value
 
     vd.addUndoSetValues([col], rows)
+    deterministic = options.vfake_deterministic
+    salt = options.vfake_seed_salt
```
saulpw: Let's use `vfake_salt` being non-empty to indicate this determinism, instead of having two options.

plugins/vfake.py
```diff
@@ -1,11 +1,15 @@
 # to anonymize a column in vd: do "setcol-fake" with e.g. 'name' 'isbn10' or any of the functions on Faker()
 
+import json
+
 from visidata import vd, Column, Sheet, option, options, asyncthread, Progress
 
-__version__ = '0.9'
+__version__ = '0.10'
```
saulpw: When do we hit 1.0?

plugins/vfake.py
ajkerrigan: Ok. Added a note in the option description and a comment in the logic to (hopefully) clarify things. I like the idea of one option, but I worry about it being a bit obscure and magical.

plugins/vfake.py
ajkerrigan: Now?

anjakefala:
Hi! This looks great. =) Can you please update the [plugins.jsonl](https://github.com/saulpw/visidata/blob/develop/plugins/plugins.jsonl), and then I will be happy to merge it in.

ajkerrigan:
> Hi! This looks great. =) Can you please update the [plugins.jsonl](https://github.com/saulpw/visidata/blob/develop/plugins/plugins.jsonl), and then I will be happy to merge it in.

Sure! I rebased the code changes into a single commit, then used that commit hash to update the `plugins.jsonl` URL in a second commit. There may be a better way to update plugins between releases - if so let me know what works best on your end. Thanks!

anjakefala:
10/10!

plugins/vfake.py
saulpw: I retract my comment, I see you have!  Congrats on the 1.0 release :)

plugins/vfake.py
saulpw: Magical is the VisiData way!  Unless the options interact in some way other than one option enabling another option, it seems like it would also be confusing to set `vfake_seed_salt` and have it not do anything. I've even toyed with changing or removing all boolean options for just this reason.  (Case in point: we have `options.undo` because it may be a performance penalty, so what if it were instead the max number of rows that could be undone?  Instant configurable axis for very little effort.)

But it is your plugin, and so I ultimately defer to your taste.  You're the one that will be using it, after all :)

plugins/vfake.py
ajkerrigan: Ha, your comment motivated a nudge from 0.10 to 1.0! I do believe that makes you a 10x developer :).

🥁🤕

PR #588, opened 2020-06-04
Rownum return new column

geekscrapy:
This allows for quick access to the new rownum col. For my use-case I need to instantly hide the column so it's useful to have it returned upon creation.


saulpw (approved):
Yup, eminently reasonable.  Would love to see your plugin if you can share!

PR #589, opened 2020-06-05
Flesh out MutableMapping behavior for LazyChainMap

ajkerrigan:
During some recent [debugging experiments](https://github.com/ajkerrigan/visidata-plugins#debugging_helpers-integrate-visidata-with-debugging-packages) I ran into issues when a debugger tries to peek at local variables. The issue is that a debugger wants to be able to run `in` checks or iterate over locals, and [LazyChainMap](https://github.com/saulpw/visidata/blob/c281f56/visidata/basesheet.py#L11) doesn't provide all of the magic methods to support that.

This PR adds the required methods for `LazyChainMap` to be an explicit [MutableMapping](https://docs.python.org/3.3/library/collections.abc.html#collections.abc.MutableMapping) implementation similar to the standard library's own [ChainMap](https://docs.python.org/3.3/library/collections.html#collections.ChainMap). I feel confident suggesting that we add these methods, but not confident about this first pass implementation :). I'm assuming we'll want to talk this through and tweak it before merging.

### DIFF
--- a/visidata/basesheet.py
+++ b/visidata/basesheet.py
@@ -1,6 +1,7 @@
 import os
 
 import visidata
+from collections.abc import MutableMapping
 from visidata import Extensible, VisiData, getGlobals, vd, EscapeException
 from unittest import mock
 
@@ -8,7 +9,7 @@ from unittest import mock
 UNLOADED = tuple()  # sentinel for a sheet not yet loaded for the first time
 
 
-class LazyChainMap:
+class LazyChainMap(MutableMapping):
     'provides a lazy mapping to obj attributes.  useful when some attributes are expensive properties.'
     def __init__(self, *objs):
         self.locals = {}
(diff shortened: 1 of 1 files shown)


visidata/basesheet.py
```diff
@@ -24,6 +25,15 @@ def keys(self):
...
         self.locals.clear()
 
+    def __len__(self):
+        return len(set().union(self.objs, self.locals))
+
+    def __contains__(self, key):
+        return key in self.objs or key in self.locals
+
+    def __iter__(self):
+        return iter(set().union(self.objs, self.locals))
```
saulpw, line 35: I think this is incorrect, because `self.objs` is `(k, obj)` instead of `(k, v)`.  Probably this should be an actual generator that loops over `union(self.objs.keys(), self.locals.keys())` and yields `k, self[k]`.

visidata/basesheet.py
```diff
@@ -24,6 +25,15 @@ def keys(self):
     def clear(self):
         self.locals.clear()
 
+    def __len__(self):
+        return len(set().union(self.objs, self.locals))
```
saulpw, line 29: I'm thinking maybe `keys()` above should return the `list(union(...))` mentioned in the iter comment.  Then this can return `len(self.keys())`.

visidata/basesheet.py
```diff
@@ -36,6 +46,9 @@ def __setitem__(self, k, v):
             return setattr(obj, k, v)
         self.locals[k] = v
 
+    def __delitem__(self, k):
+        del locals[k]
```
saulpw, line 50: `self.locals` :)

Issue #587, opened 2020-06-04
[dive-row/pyobj-cell] 


frosencrantz:
**Small description**
Diving into a list of json objects doesn't show keys/cols if they aren't in the first object in the list.

**Expected result**
See all the keys in the data.  Or minimally a hint that data is missing.

**Actual result with screenshot**
Data isn't shown.  This was fixed recently for expand-cols in https://github.com/saulpw/visidata/pull/497 by @ajkerrigan 

**Steps to reproduce with sample data and a .vd**
echo '{"line":1,"data":[{"one":1,"two":2},{"one":1,"two":2,"three":3}]}' | vd -f json

Then pyobj-cell into the cell under the column named `data`
You'll get a table with two columns ('one' and 'two') there is no 'three'
If you dive-row into the row with three keys, Visidata shows all three columns.

**Additional context**
Please include the version of VisiData. latest version on develop.

Maybe this should be listed as a Feature Request.

ajkerrigan:
Good catch @frosencrantz, I'll take a look at this one.

frosencrantz:
Thanks @ajkerrigan !  That's fantastic!

PR #590, opened 2020-06-06
Add columns from all rows in ListOfDictSheet

ajkerrigan:
This PR addresses #587 by checking all rows for new columns.


saulpw (approved):
This could be tighter, but it's fine.  Thanks for fixing this, @ajkerrigan.

PR #593, opened 2020-06-11
Now truncates and "safes" sheet names to comply with Excel standards

geekscrapy:
Truncates long sheet names and adds _1, _2 if there are matching names
Strips characters from sheet names that Excel does not like
self.source does not always have a title


geekscrapy:
Closed as I replaced with a duplicate

PR #594, opened 2020-06-14
[xls/x loader] long sheet names now accommodated

geekscrapy:
Edited `Column('sheet'` as (I found out...!) not all Excel files have a title, so in this case the `source.name` is used which I _believe_ is always set.

We now truncate sheet names that are greater than 31 chars (xlsx can do 32, but it is set to 31 for backwards compatibility).

Added `excel_safename` which provides a safe way to convert sheet names to compatible names.

If any sheet names are truncated, the user is made aware with a `warning` in the status pane


visidata/loaders/xlsx.py
```diff
@@ -5,7 +5,7 @@ class XlsxIndexSheet(IndexSheet):
     'Load XLSX file (in Excel Open XML format).'
     rowtype = 'sheets'  # rowdef: xlsxSheet
     columns = [
-        Column('sheet', getter=lambda col,row: row.source.title),  # xlsx sheet title
+        Column('sheet', getter=lambda col,row: row.source.title or row.source.name),  # xlsx sheet title
```
anjakefala: Is the title a unique aspect of the Excel data format? Is there a relationship between titles and sheet names? Can we just use source names?

visidata/loaders/xlsx.py
```diff
@@ -53,16 +53,39 @@ def iterload(self):
...
+        short_name = name[:excel_limit]
+        used_names.append(short_name)
+        suffix = '_' + str(used_names.count(short_name))
+        name = short_name[:-len(suffix)] + suffix
+
+    # name does not contain any of the following characters:  :  \  /  ?  *  [  or  ]
+    name = ''.join([ '_' if c in '\\/?*[]' else c for c in name])
+
+    # strip leading and trailing _
+    name.strip("_")
```
saulpw: strings are immutable in python, so this should be `name = name.strip("_")`

geekscrapy: Done!

visidata/loaders/xlsx.py
geekscrapy: So as far as I tested `source.title` takes the _Excel sheet name_ from the tab at the bottom of the window. I think it should be kept as-is (without the `or`, as per my recent commit) because this aligns more closely with what is displayed in Excel.

Also `sheet.name` should probably use this logic as well, and if it does, we may as well do away with the `'sheet'` column altogether and show the `'name'` column?

anjakefala: I think it is important to keep the `sheet` column for now because `IndexSheet`s contain rows that represent sheets.

It is also important for every sheet to have a name. Things break in VisiData (e.g. cmdlog) if there are un-named sheets!

anjakefala: So, I think you should return the `title or name`.

anjakefala:
Hey @geekscrapy! Can you please merge develop into your branch, and resolve the merge conflicts? They probably have to do with me replacing `vd.filetype()` with `open_xls()`

anjakefala:
Nevermind, it seems like I can resolve them

anjakefala:
We are going to close this PR for now! @saulpw implemented some changed negotiated off-PR in the #visidata chat.

geekscrapy:
Ok, but I don't think this solves the max limit of 32 chars from what I've seen? I maybe missed something though!

I presume the thought is to mod the excel saver in the future to deal with non cleaned sheet names? As as far as I see, the new col name cleaning option is off by default?

PR #599, opened 2020-06-17
[types] do not group numeric columns by thousands by default.

anjakefala:
- add tutorial for how to group by thousands

### DIFF
--- a/docs/columns.md
+++ b/docs/columns.md
@@ -1,5 +1,5 @@
-- Update: 2018-08-19
-- Version: VisiData 1.3.1
+- Update: 2020-06-17
+- Version: VisiData 2.0
 
 # Columns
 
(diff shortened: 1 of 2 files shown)


docs/columns.md
```diff
@@ -64,6 +64,50 @@ The following example uses the file [sample.tsv](https://raw.githubusercontent.c
...
+---
+
+## How to format numeric columns to contain a thousands separator within a session?
+
+**Note**: tsvs and csvs save as they are displayed.
+
+1. Set a column to a numeric type by pressing `#` (int), `%` (float), or `$` (currency).
+2. Press `C` to open the **Columns Sheet**.
+3. Move to the row referencing the column whose display you wish to format. Move the cursor to the fmtstr column.
+4. Type `e` followed by *{:,.0f}* for an `int` type and *{:,.02f}* for a floating point type.
```
saulpw: style: should we use `{:,.0f}` instead of *{:,.0f}* ?

docs/columns.md
```diff
@@ -64,6 +64,50 @@ The following example uses the file [sample.tsv](https://raw.githubusercontent.c
...
+
+This changes the formatting for a specific column.
+
+or
+
+1. Press `O` to open the **Options Sheet**.
+2. Move the cursor down to the relevant *disp_TYPE_fmt* option where *TYPE* can be *int*, *currency*, or *float*.
```
saulpw: or `date`.

docs/columns.md
```diff
@@ -64,6 +64,50 @@ The following example uses the file [sample.tsv](https://raw.githubusercontent.c
...
+3. Type `e` followed by *{:,.0f}* for an `int` type and *{:,.02f}* for a floating point type.
+
+
+## How to set all float columns always to contain a thousands separator within all of my sessions?
```
saulpw: Let's expand the header to include the related questions we're answering here.  like "change the default date format".  We can make the How to question more general, but probably better to just add more concrete subquestions as people might be searching for.

docs/columns.md
```diff
@@ -64,6 +64,50 @@ The following example uses the file [sample.tsv](https://raw.githubusercontent.c
...
+
+**Note**: tsvs and csvs save as they are displayed.
+
+Some types have an option for their default display formatting.
+
+Type      Option            Default
+--------- ----------------  --------
+int       disp_int_fmt      {:.0f}
+float     disp_float_fmt    {:.02f}
+currency  disp_currency_fmt %0.2f
```
saulpw, line 79: We should explicitly say somewhere that both `%` and `{}` format strings are supported, and the minor difference(s) between them.

saulpw:
Thank you for putting this together, @anjakefala !  You are anja-on-the-spot with these docs.

docs/columns.md
anjakefala: Oh, I have been using *blah* for arguments to distinguish them from commands.

E.g. `z+` *sum*

docs/columns.md
anjakefala: Intentionally excluded because this was very specifically about thousands separator, but since we are generalising it, I will include it!

docs/columns.md
```diff
@@ -66,6 +66,60 @@ The following example uses the file [sample.tsv](https://raw.githubusercontent.c
...
+date      disp_date_fmt     %Y-%m-%d
+
+Ways to adjust the display formatting:
+* The `fmtstr` column on the **Columns Sheet** allows you to specify the formatting for specific columns within that session, without affecting the default for the others.
+* The `disp_TYPE_fmt` option can be changed on the **Options Sheet** to set the formatting for all columns of type `TYPE` in that session.
+* The `--disp-type-fmt` argument can be passed through the commandline to set the formatting for all columns of type `TYPE` in that session.
```
saulpw: `TYPE` should be capitalized

docs/columns.md
```diff
@@ -66,6 +66,60 @@ The following example uses the file [sample.tsv](https://raw.githubusercontent.c
...
+* The `options.disp_TYPE_fmt` can be set in the `~/.visidatarc` to change the default formatting for all columns of type `TYPE` for all sessions.
+
+There are several formatting styles offered:
+* Formatting that starts with **%** (e.g. *%0.2f*) will use [locale.format_string()](https://docs.python.org/3.6/library/locale.html#locale.format_string).
```
saulpw: This seems like it should be `%`

saulpw (approved):
Looks good pending those minor changes. Thanks for documenting this, @anjakefala !

docs/columns.md
anjakefala: I want to distinguish it from float type keytsroke. :X Maybe, I will try `'%'`

PR #600, opened 2020-06-17
[yaml] Use the default safe YAML loader

tsibley:
The full loader is unsafe because serialized files can be constructed
which run arbitrary code during their deserialization.  The safe loader
supports a very large subset of YAML and supports the most common uses
of YAML.


anjakefala:
Thanks @tsibley! :blush:

saulpw:
Thanks @tsibley, this is definitely safer.

PR #601, opened 2020-06-17
[yaml] Support files containing multiple documents

tsibley:
Files with multiple documents are treated the same way as a file with a
single list-valued document.

---

Builds off #600 so if this is merged that will be too.

### DIFF
--- a/visidata/loaders/yaml.py
+++ b/visidata/loaders/yaml.py
@@ -1,19 +1,39 @@
 from visidata import *
+from itertools import chain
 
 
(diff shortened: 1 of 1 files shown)


tsibley:
Yep! I updated the commit message to better describe the change re: existing behaviour:

> The loader previously threw an exception when given a YAML file with
> multiple documents.  Now it successfully loads the file, with each
> document presented as a row.  All other existing behaviour remains the
> same.

(and rebased onto develop to make the safe loader change disappear from this PR since it's now merged)

PR #602, opened 2020-06-17
Fix error-raising typo in z+

jsvine:
Typo seems to have been introduced in https://github.com/saulpw/visidata/commit/e364cc957aacc9e8cb665681e7f235b71ab88504.

(On `z+`, it was raising `NameError: name 'aggregators_choices' is not defined`.)


anjakefala:
Nice catch @jsvine! Thanks!

jsvine:
Thanks for the quick merge, @anjakefala!

PR #604, opened 2020-06-20
Fix bug in z+

jsvine:
Previously, pressing z+ and then making a/any choice raised this error (using "sum" as an example):

```
  File "[...]/visidata/visidata/basesheet.py", line 127, in execCommand
  File "show-aggregate", line 1, in <module>
    import math
TypeError: string indices must be integers
```

And this left-status: `TypeError: string indices must be integers | "sum" | ['sum']`

This change appears to fix the problem for me, but I acknowledge that there may be other/better solutions. Thanks!


saulpw:
Looks good to me!  Thanks for taking the time to fix it.

PR #611, opened 2020-06-26
Add an option for number of rows to scan when expanding columns

ajkerrigan:
#497 started scanning `visibleRows` instead of `cursorRow`, but that was too shortsighted on my part. Default to 1000 rows instead - still pretty arbitrary, but seems reasonable enough based on a quick IRC chat.

Make it configurable to help people optimize for speed or thoroughness as needed.


visidata/pyobj.py
```diff
@@ -3,16 +3,18 @@
...
 class PythonSheet(Sheet):
     def openRow(self, row):
         return load_pyobj("%s[%s]" % (self.name, self.keystr(row)), row)
 
 
-def expand_cols_deep(sheet, cols, rows, depth=0):  # depth == 0 means drill all the way
+def expand_cols_deep(sheet, cols, rows=None, depth=0):  # depth == 0 means drill all the way
     'expand all visible columns of containers to the given depth (0=fully)'
     ret = []
+    rows = rows or sheet.rows[:options.expand_col_scanrows] or sheet.rows
```
saulpw: One issue with this, is that if you find a larger list far down a large dataset, you won't be able to expand to use it, because it'll always use the same first 1000 rows.  I don't know if it's worth it to implement, but I guess the ideal would be "1000 rows around the cursor row" or something like that.

geekscrapy:
My two cents: I will usually remember that it does a row sample, so I will often change the col to a z# (can't remember the name 🤪, zlen?) type then do a sort descending, followed by the col-expand.
Helps me get around the sampling, and means I always fully expand

visidata/pyobj.py
ajkerrigan: Ah that's an interesting case. I like that suggestion! I considered and rejected using a random sample of rows to avoid overly weighting the top rows, but that seemed really evil :). Using the cursor row as guidance seems like a smarter approach, thanks!

ajkerrigan:
> My two cents: I will usually remember that it does a row sample, so I will often change the col to a z# (can't remember the name 🤪, zlen?) type then do a sort descending, followed by the col-expand.
> Helps me get around the sampling, and means I always fully expand

Ah nice, that's a handy trick :). The thing that surfaced this issue for me was looking at a bunch of rows with a dict in one column. The dicts all had one key, but that key changed a few times over thousands of rows. I don't know how common that case is for other folks, but it's tough to feel confident about an expansion there without checking everything.

saulpw:
It's definitely a good idea to have a mechanism to "check all rows".  It's the best behavior for a small/medium dataset.  We just need to make sure there's a way to "just use this row(s)" for performance reasons in a large dataset.

visidata/pyobj.py
ajkerrigan: Ok, I made a little change so instead of "first n rows" it will use "n rows centered around the cursor row" as you suggested. So for datasets larger than 1k rows, the default behavior (assuming you expand before moving) would look at the first and last 500.

At this point I'm still not sure if it would be safer to default to 1,000 and have people tweak to be more thorough, or default to 0 (all rows) and have people tweak for speed. I left 1,000 for now but would be fine flipping it.

saulpw: Cool, thanks AJ!  I think I like the default to be 1000 since it covers 99% of cases, including the big and the small and the usual.  It only doesn't cover the "extra field not found in 1000 rows around cursor" case, which we should mention in the docs for these commands.

ajkerrigan: Sounds good to me. When you say "the docs for these commands", where would that be? There's a mention of expanding columns in the json section of the man page, but I'm not sure if there's a separate command help somewhere that I'm missing.

ajkerrigan:
Asciicasts are loading properly now for the column doc page in my local tests, and the layout for the expansion piece seems to match other sections. It may still need some content/wording tweaks, but doesn't seem jarring or fundamentally broken anymore :).

saulpw (approved):
:ship:

Issue #612, opened 2020-06-27
How to access non-public schema in Postgres DB?


p3k:
**Small description**

I maintain a Postgres database having a non-public schema, i.e. I either need to explicitely provide the schema name or set the `search_path`:

```sql
# either one of these works
select * from foo.bar;

set search_path = foo;
select * from bar;

alter database foobar set search_path to foo; # this makes it persistent
select * from bar;
```

Trying to connect to this database with Visidata works but returns an empty set of tables – even with the persistent `search_path` using `alter database`:

```plain
$ vd postgres://user:pw@localhost/foobar
 table_name         ‖ ncols             #| nrows             #‖
```

**Expected result**

Something along the lines of this output:

```plain
 table_name   ‖ ncols  #| nrows  #‖
 tag_hub      ‖       5 |        …‖
 log          ‖       7 |        …‖
 metadata     ‖       6 |        …‖
 skin         ‖       9 |        …‖
 layout       ‖       8 |        …‖
 content      ‖      16 |        …‖
 choice       ‖       5 |        …‖
 site         ‖      10 |        …‖
 poll         ‖       9 |        …‖
 account      ‖       7 |        …‖
 vote         ‖       7 |        …‖
 membership   ‖       8 |        …‖
 image        ‖      10 |        …‖
 file         ‖      12 |        …‖
 tag          ‖       4 |        …‖
```

I achieved this output by temporarily changing [line 61 of `loaders/postgres.py`]( https://github.com/saulpw/visidata/blob/stable/visidata/loaders/postgres.py#L61) to the desired schema name:

```diff
64c64
<         qstr = "SELECT table_name, COUNT(column_name) AS ncols FROM information_schema.columns WHERE table_schema = 'public' GROUP BY table_name"
---
>         qstr = "SELECT table_name, COUNT(column_name) AS ncols FROM information_schema.columns WHERE table_schema = 'antville' GROUP BY table_name"
```

Although this shows the correct list of tables, it still wouldn’t seem sufficient to somehow inject the correct schema name dynamically because of the missing `nrows` values (`…`) – entering one of the rows shows an empty output.

**Actual result with screenshot**

![image](https://user-images.githubusercontent.com/16631/85919337-88d62080-b86a-11ea-9150-4780cb9cbbba.png)

**Additional context**

```
→ vd -v
saul.pw/VisiData v1.0
```

p3k:
Also tested this with v2 and here I generally get a full-blown error:

```plain
[...]

PR #615, opened 2020-06-28
Add support for custom Postgres schema

p3k:
Schema defaults to `public` but can be overridden using the `--postgres-schema` flag:

```plain
vd --postgres-schema=foo postgres://user:pw@localhost/foobar
```

### DIFF
--- a/visidata/loaders/postgres.py
+++ b/visidata/loaders/postgres.py
@@ -2,6 +2,8 @@ from visidata import *
 
 __all__ = ['openurl_postgres', 'openurl_rds', 'PgTable', 'PgTablesSheet']
 
+option('postgres_schema', 'public', 'The desired schema for the Postgres database')
+
 def codeToType(type_code, colname):
     import psycopg2
     try:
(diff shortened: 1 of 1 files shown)


visidata/loaders/postgres.py
```diff
@@ -45,7 +47,9 @@ def openurl_postgres(url, filetype=None):
                 dbname=dbname,
                 host=url.hostname,
                 port=url.port,
-                password=url.password)
+                password=url.password,
+                # Not sure if this is necessary
```
saulpw: If it's not necessary, let's remove it; otherwise the comment should be removed.

saulpw (changes requested):
Looks good! Just one small change and then we can merge it.  Thanks for adding this, @p3k.

visidata/loaders/postgres.py
p3k: Alright, removed in f60501428b6a5aafc73d213f5b6c351c67ceed12.

saulpw:
Ah cool, did you make sure it still works to load the individual tables, etc?

p3k:
> Ah cool, did you make sure it still works to load the individual tables, etc?

Yes, at least with my little database I can go down on the table level and see the expectable table rows.

I also can go further down and see the sheet about “denominator” et al.

saulpw:
Sounds good, thanks @p3k!

saulpw:
Ah and let's reference it in #612 so we can close that too.

PR #617, opened 2020-06-29
Add loader for MySQL database

p3k:
From the experience I gained making the Postgres loader work again in v2 I decided to give a try to create a MySQL loader. Without further ado, here it is. (This time branched off of develop.)

### DIFF
--- a/visidata/loaders/mysql.py
+++ b/visidata/loaders/mysql.py
@@ -36,7 +36,6 @@ class SQL:
         self.conn = conn
 
     def cur(self, qstr):
-        # FIXME: The Postgres loader is using a named cursor but afaik MySQL does not know named cursors… problem?
         cur = self.conn.cursor()
         cur.execute(qstr)
         return cur
--- a/visidata/loaders/mysql.py
+++ b/visidata/loaders/mysql.py
@@ -83,8 +83,6 @@ class MyTablesSheet(Sheet):
         '''
 
         with self.sql.cur(qstr) as cur:
-            self.nrowsPerTable = {}
-
             self.rows = []
             # try to get first row to make cur.description available
             r = cur.fetchone()
--- a/visidata/loaders/mysql.py
+++ b/visidata/loaders/mysql.py
@@ -5,8 +5,13 @@ __all__ = ['openurl_mysql', 'MyTable', 'MyTablesSheet']
 def codeToType(type_code, colname):
     import MySQLdb as mysql
 
-    if type_code == mysql.NUMBER:
+    types = mysql.constants.FIELD_TYPE
+
+    if type_code in (types.TINY, types.SHORT, types.LONG, types.LONGLONG, types.INT24,):
         return int
+
+    if type_code in (types.FLOAT, types.DOUBLE, types.DECIMAL, types.NEWDECIMAL,):
+        return float
 
     if type_code == mysql.STRING:
         return str
--- a/visidata/loaders/mysql.py
+++ b/visidata/loaders/mysql.py
@@ -1,0 +1,114 @@
+from visidata import *
+
+__all__ = ['openurl_mysql', 'MyTable', 'MyTablesSheet']
+
+def codeToType(type_code, colname):
+    import MySQLdb as mysql
+
+    if type_code == mysql.NUMBER:
+        return int
+
+    if type_code == mysql.STRING:
+        return str
+
+    return anytype
+
+
+def openurl_mysql(url, filetype=None):
+    import MySQLdb as mysql
+
+    url = urlparse(url.given)
+    dbname = url.path[1:]
+    conn = mysql.connect(
+                user=url.username,
+                database=dbname,
+                host=url.hostname,
+                port=url.port or 3306,
+                password=url.password,
+                use_unicode=True,
+                charset='utf8')
+
+    return MyTablesSheet(dbname+"_tables", sql=SQL(conn), schema=dbname)
+
+
+class SQL:
+    def __init__(self, conn):
+        self.conn = conn
+
+    def cur(self, qstr):
+        # FIXME: The Postgres loader is using a named cursor but afaik MySQL does not know named cursors… problem?
+        cur = self.conn.cursor()
+        cur.execute(qstr)
+        return cur
+
+    @asyncthread
+    def query_async(self, qstr, callback=None):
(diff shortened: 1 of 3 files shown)


visidata/loaders/mysql.py
```diff
@@ -0,0 +1,114 @@
+from visidata import *
+
+__all__ = ['openurl_mysql', 'MyTable', 'MyTablesSheet']
+
+def codeToType(type_code, colname):
+    import MySQLdb as mysql
+
+    if type_code == mysql.NUMBER:
```
saulpw: What does mysql use for decimal values?

visidata/loaders/mysql.py
```diff
@@ -0,0 +1,114 @@
...
+
+    return MyTablesSheet(dbname+"_tables", sql=SQL(conn), schema=dbname)
+
+
+class SQL:
+    def __init__(self, conn):
+        self.conn = conn
+
+    def cur(self, qstr):
+        # FIXME: The Postgres loader is using a named cursor but afaik MySQL does not know named cursors… problem?
```
saulpw: It's not a problem if it works :)

visidata/loaders/mysql.py
```diff
@@ -0,0 +1,114 @@
...
+                        table_schema = '{self.schema}'
+                    group by
+                        table_name
+                ) as column_count
+            where
+                t.table_name = column_count.table_name;
+        '''
+
+        with self.sql.cur(qstr) as cur:
+            self.nrowsPerTable = {}
```
saulpw: Is this variable used anywhere?

visidata/loaders/mysql.py
```diff
@@ -0,0 +1,114 @@
+from visidata import *
+
+__all__ = ['openurl_mysql', 'MyTable', 'MyTablesSheet']
+
+def codeToType(type_code, colname):
+    import MySQLdb as mysql
```
saulpw, line 6: Please add this to `requirements.txt`.

saulpw (changes requested):
Thanks for putting this together, @p3k!  I had a few questions in the code but overall it looks good.  Do you use mysql a lot?  Do you have any interest in getting other features like database editing working (delete rows, edit values, like the sqlite loader can do)?

visidata/loaders/mysql.py
p3k: Um, you mean I rather should check for some of these types?

https://github.com/PyMySQL/mysqlclient-python/blob/4c1950148ba10e23017bd5190f9da6bdb347e6b3/MySQLdb/constants/FIELD_TYPE.py#L7-L40

visidata/loaders/mysql.py
p3k: Just to be sure, because I don’t know enough about async Python: could a named cursor be necessary for the methods decorated with `@asyncthread`? If it was introduced because of those in the Postgres loader, it might be necessary to rewrite the MySQL loader synchronously…?

visidata/loaders/mysql.py
p3k: Nope, seems to be superfluous already in the Postgres loader…

visidata/loaders/mysql.py
p3k: OK

p3k:
> Do you use mysql a lot? 

I use it (well, actually the compatible MariaDB) as much as Postgres in a long-running [pet project](/antville/antville), a blogging platform that works with both and even some more database products.

> Do you have any interest in getting other features like database editing working (delete rows, edit values, like the sqlite loader can do)?

Actually, I like VisiData as a quick and easy analysis tool of choice. I guess I still would run the mysql binary if I need to modify database values, even if vd was capable of editing…

visidata/loaders/mysql.py
p3k: Fixed in cf3966827e005b373de5f3875aab4f056e9cd0a8.

visidata/loaders/mysql.py
p3k: Fixed in 909ab520c0b5188cd912f89a4e42d4f719a9bb37.

visidata/loaders/mysql.py
p3k: I checked again, and I think `mysql.NUMBER` is the correct way to evaluate `type_code`; it actually is a set of type codes looking like this:

```py
DBAPISet({0, 1, 3, 4, 5, 8, 9, 13, 246})
```

saulpw: Yes, sorry, I meant that some numbers are `int` and some numbers are `float` and we should distinguish them like MySQL does.  (so that 54.8434 can be a float but 54 can be an int).

visidata/loaders/mysql.py
saulpw: I guess it might be true, but that seems like a per-loader kind of thing. And we don't have multiple threads accessing the db at the same time; they may be different threads, but they are completely isolated in time.  So as long as it works, we don't need it yet.

saulpw: (So let's just remove the comment).

saulpw:
Almost there!  Thanks for your patience.

visidata/loaders/mysql.py
p3k: I am still not sure how to proceed: should I check and return `int` for all of those types, then?

* DECIMAL = 0
* TINY = 1
* SHORT = 2
* LONG = 3
* LONGLONG = 8
* INT24 = 9
* NEWDECIMAL = 246

saulpw: Yes, though it seems like we should return `float` for FLOAT, DOUBLE, DECIMAL, and NEWDECIMAL?

p3k: Aha! I just went along with the Postgres loader as template, which only returns `INTEGER` and `STRING`, too… let me see.

visidata/loaders/mysql.py
p3k: Fixed in 5151db12db9ba71de0807219bbf14930192df408.

visidata/loaders/mysql.py
p3k: Like this? 56bbee88f2147c3b47ba772d8d1329e37199b347

PR #624, opened 2020-07-05
[config] load config file + plugins before setting cli options

anjakefala:


saulpw (approved):
Thank you, @anjakefala!

PR #626, opened 2020-07-06
[warnings] ignore deprecation + resource warnings

anjakefala:
Pushing this for us to think about holistically.

The problem is that sometimes result in Deprecation + Resource Warnings, that we cannot do anything about. These warnings result in a lot of noise, and I think we might be better off ignoring them.

This does not feel like the obvious ideal solution, and I would like to think through with you what the best approach should be.


saulpw (approved):
Looks good, thanks!

PR #632, opened 2020-07-15
[plugins] bump vds3 to 0.4

ajkerrigan:
2.0 compatibility updates:

- Don't break defermods
- Ditch vd.filetypes references

[full plugin diff](https://github.com/ajkerrigan/visidata-plugins/compare/v0.3...vds3-0.4?diff=split#diff-17b27942510ee5891097de5843083a7b) for v0.3 --> v0.4


saulpw (approved):
We should probably establish a practice of linking to a diff of the changes from the previous version.  I trust you but someone unscrupulous could use this as an attack vector.

ajkerrigan:
That's probably a good call 👍 . [Here's](https://github.com/ajkerrigan/visidata-plugins/compare/v0.3...vds3-0.4?diff=split#diff-17b27942510ee5891097de5843083a7b) the diff from v0.3 to v0.4. The non-vds3 sideshow bits make the changes look more substantial than they are 😳 .

Issue #516, opened 2020-04-21
[build] Package into VisiData a set up of unit tests + basic functionality tests


anjakefala:
We want a set of tests to help catch packaging errors. These are part of the requirements for packaging authorities like Debian, GNU Guix, and Homebrew.

Basics:
* Test that `Ctrl+H` works
* package in the unit tests
* benchmark.csv + some test .vd to make sure it loads and maybe saves

ryanprior:
If you can ping me when this lands I'd appreciate it!

anjakefala:
Hi @ryanprior! I packaged a basic set of unit tests into VisiData.

Tests are run by executing `pytest visidata/tests/` wherever the VisiData root
directory is installed. Sadly, `setup.py` has deprecated `setup.py tests`.

This will be part of VisiData 2.0.

I hope that helps!

PR #636, opened 2020-07-20
[tests] package unit tests

anjakefala:
Addresses #516


saulpw:
Ah, okay, so this makes it possible for packagers to include the running of unit tests after the package is installed?  Let's include a simple comment in this commit message about how to run the unit tests, and maybe on visidata.org/dev too.

anjakefala:
> Ah, okay, so this makes it possible for packagers to include the running of unit tests after the package is installed?

Yup! This would package the unit tests in, so they would be located in the same place as the rest of the VisiData code. Packagers could then run them.

Sure!

PR #639, opened 2020-07-22
[unfurl] don't unfurl string values

ajkerrigan:
Treat a string as a non-iterable, so it will be left intact when
unfurling a column.

**Sample Data**

```json
[
    {
        "key": "names",
        "value": [
            "us-east-1a",
            "us-east-1b",
            "us-east-1c",
            "us-east-1d",
            "us-east-1e",
            "us-east-1f"
        ]
    },
    {
        "key": "state",
        "value": "available"
    }
]
```

**Furled?**

```
 key   | value     ║
 names | [6]       ║
 state | available ║
```

**Unfurled (without this change)**

```
 key   | value_key | value_value ║
 names | 0        #| us-east-1a  ║
 names | 1        #| us-east-1b  ║
 names | 2        #| us-east-1c  ║
 names | 3        #| us-east-1d  ║
 names | 4        #| us-east-1e  ║
 names | 5        #| us-east-1f  ║
 state | 0        #| a           ║
 state | 1        #| v           ║
 state | 2        #| a           ║
 state | 3        #| i           ║
 state | 4        #| l           ║
 state | 5        #| a           ║
 state | 6        #| b           ║
 state | 7        #| l           ║
 state | 8        #| e           ║
```

**Unfurled (with this change)**

```
 key   | value_key | value_value ║
 names | 0        #| us-east-1a  ║
 names | 1        #| us-east-1b  ║
 names | 2        #| us-east-1c  ║
 names | 3        #| us-east-1d  ║
 names | 4        #| us-east-1e  ║
 names | 5        #| us-east-1f  ║
 state | 0        #| available   ║
```

### DIFF
--- a/visidata/unfurl.py
+++ b/visidata/unfurl.py
@@ -28,7 +28,7 @@ class UnfurledSheet(Sheet):
         for row in Progress(self.source.rows):
             val = self.source_col.getValue(row)
 
-            if not isinstance(val, Iterable):
+            if not isinstance(val, Iterable) or isinstance(val, str):
                 val = [ val ]
 
             if isinstance(val, Mapping):


jsvine:
FWIW: I think that’s a reasonable change 👍. When I first developed `unfurl`, I vaguely recall *intentionally* letting strings be unfurled ... but now I’m struggling to remember why. I think treating them as non-iterable is, while perhaps Pythonically unorthodox, a more user-friendly behavior.

ajkerrigan:
Ah thanks for weighing in @jsvine! I would have felt bad pinging you but it's good to get your thoughts on the change :).

The only other question I had on unfurl was how to handle 0-length iterables. Currently unfurl removes the row, which works fine for the cases where I've used it so far. But I was wondering if it would be preferable to leave a single row with a null value in the unfurled column instead. I didn't have a clear opinion one way or the other, just noticed enough to raise the question.

jsvine:
Hmmm, that's a good question. I also don't have a strong opinion — I can see it going either way. My gut instinct, though, is to keep the current logic. That would match my most frequent use-case for `unfurl`, which is "please give me one row per value in this column of lists/dicts." (I am very polite when speaking to my computer.)

Some other points in favor of that:

- If someone wanted to keep info from the rows with zero-length values, they could still join the unfurled table to the original table.

- Reduces the ambiguity of an unfurled value of `None`: Did it come from `[ None ]` or from `[ ]`?

geekscrapy:
Personally, I'd really like to keep rows with a None value.
- it's easy enough to delete the None rows manually afterwards
- if rows are just deleted and the user doesn't know, this could really mess up later calculations on the data (freq table counts etc). Yes, you know when you know, but until that time...
- finally, it's harder to re insert a deleted row that you wanted, than to delete the row

If you do stick with deleting the row, it might be worth a status message saying "X empty rows removed" or something similar.

Or another mitigating thing would be to make it work on selected rows only?
 
Either way, if there was an option to enable/disable the behaviour that'd be awesome, I can see a use for it!

saulpw (approved):
Yes, I agree, this is the right thing to do.  If @jsvine has a use case for unfurling strings we can revisit whether to add an option, or a new "list" type, or whatever.

saulpw:
As for 0-length lists, there definitely should be an option, and it should apply to Melt as well.  I think I agree with @geekscrapy, the default should be to a create a row with a null/None in both key and value columns, so they're distinct and easily selectable.  (Too bad if you have a `None` key in your dict, sorry.)  Would you be up for putting that together in a separate PR?

jsvine:
@saulpw Ah, you’re right, it’s only the `{ None: None }` scenario that would create ambiguity, not `[ None ]`, which would be distinguishable by its `key` being `0`. Fine by me :)

PR #649, opened 2020-07-26
Add iota_base option to set the start value for incremental columns

### LINKED ISSUE #647: [wishlist] Number rows starting at 1
VisiData has gone from [not having a convenient way](https://github.com/saulpw/visidata/issues/215) to number rows to having two different ways to do it:

- using <kbd>i</kbd> (iota) and family
- the [rownum plugin](https://github.com/saulpw/visidata/blob/19dcd50da53361527c2272636e0de0eea8c306fe/plugins/rownum.py)*

\* side note: is the plugin still needed now that support for this is built in?

These both work well :tada:, but the 0-based index makes it inconvenient to do basic things like find an item's rank or quickly count the number of items in a category. An extra calculation has to be done in each case to translate the 0-based index into the 1-based rank favored by [Excel](https://exceljet.net/formula/increment-a-calculation-with-row-or-column), [Postgres](https://www.postgresql.org/docs/10/functions-window.html)/[SQLite](https://www.sqlite.org/windowfunctions.html#built_in_window_functions) etc.

1-based numbering could be supported by adding a parameter to the existing commands, or adding new commands, but I'd be happy if it was just a config option, e.g. `options.iota_base` (default: 0), though I think the default [should be 1](https://ux.stackexchange.com/a/84298).

chocolateboy:
Add `options.iota_base` to set the start value for incremental columns + change the default start value from 0 to 1.

Closes #647.

### DIFF
--- a/visidata/data.py
+++ b/visidata/data.py
@@ -7,6 +7,7 @@ from visidata import Sheet, Column, asyncthread, Progress, status, error
 from visidata import *
 
 option('filetype', '', 'specify file type', replay=True)
+option('iota_base', 1, 'start value for incremental columns', replay=True)
 
 
 def _default_colnames():
@@ -19,8 +20,8 @@ def _default_colnames():
 
 vd.default_colnames = _default_colnames()
 
-def frange(n, step):
-    yield from (x*step for x in range(n))
+def frange(start, stop, step):
+    yield from (x*step for x in range(start, stop))
 
 def num(s):
     try:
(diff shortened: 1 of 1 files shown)


saulpw:
This is a good idea, @chocolateboy.  I found myself wanting it to be `1` by default too.  The only thing I'm wondering about, is what the option should be called.  What do you think about [`index_origin`](https://aplwiki.com/wiki/Index_origin) or `increment_origin`?  I feel like `iota` might be a bit too obscure and jargon-y, even if it is the keybinding's namesake.

chocolateboy:
Any name's fine by me. Happy for it to be changed (or to change it) to whatever you prefer.

For the record, I chose "base" because it is (or was) used for this purpose by [Perl](https://metacpan.org/pod/Array::Base). It's also the default way of referring to this topic both conversationally and formally, i.e. [0-based](https://en.wikipedia.org/wiki/Zero-based_numbering), 1-based etc.

I think the option name should match the feature name. IMO, "index" is either too vague (too many search engine results) or too closely tied to 0-based (array) indices. "Increment" is OK but it's a verb...

Personally, I think "iota" is fine: it's easy to search for in the help page, easy to Google, and less of a mouthful than, say, "auto-increment" or "auto-number".

saulpw:
Thanks for your keen opinion, @chocolateboy, I appreciate it.  I went ahead and made it `incr_base`, and I renamed the commands to be `incr` as well.  The float case was a little trickier than it looked due to Python's `range` requiring `int`.

PR #653, opened 2020-08-03
[pivot] if single aggr, place in sheet name

anjakefala:


visidata/pivot.py
```diff
@@ -139,8 +139,12 @@ def addAggregateCols(self):
 
                         if len(aggregatorlist) > 1 or len(aggcols) > 1:
                             colname = '%s_%s' % (aggname, valname)
+                            if not self.name:
```
saulpw, line 142: `if A: B C else: D C` -> `if A: B else: D; C`

saulpw (approved):

PR #664, opened 2020-08-17
[slide] add ability to slide keycolumns within key column space

### LINKED ISSUE #640: [v2.-4dev] Issues with sliding columns
From @anjakefala 

> Two issues:
- [x] Key columns cannot be re-arranged.
- [x] When a regular column reaches into the key column space, with `slide-left`, it can snap back to its starting position. It should just remain to the right of the key column space.
>
> This issue will be closed after both of these are investigated. Attempting a fix is preferable, but if a fix is too tricky, then this will be documented in /limits before closure.

anjakefala:
Closes #640


saulpw (approved):
Looks good, thank you @anjakefala for working to get this right for all the edge case!

PR #666, opened 2020-08-20
[push] remove non-precious sheets from sheets-stack upon dive

anjakefala:
I am definitely unsure about this, but making the PR to foster a communication.

By design, every time a new sheet is pushed, if the current sheet at the top of the stack is not-precious, it should be dropped from the stack.

However, I cannot see this happening anywhere in the code?

`git grep precious`

```
visidata/basesheet.py:    precious = True      # False for a few discardable metasheets
visidata/basesheet.py:        ret.precious = True  # copies can be precious even if originals aren't
visidata/choose.py:    precious = False
visidata/cmdlog.py:    return CommandLog(p.name, source=p, precious=True)
visidata/cmdlog.py:    return CommandLogJsonl(p.name, source=p, precious=True)
visidata/cmdlog.py:    precious = False
visidata/describe.py:    precious = True
visidata/help.py:    precious = False
visidata/metasheets.py:    precious = False
visidata/metasheets.py:    precious = False
visidata/metasheets.py:    precious = False
visidata/sheets.py:    precious = False
visidata/sheets.py:    if vs.precious and vs not in vd.allSheets:
visidata/sheets.py:# when diving into a sheet, remove the index unless it is precious
visidata/statusbar.py:    precious = False
visidata/textsheet.py:    precious = False
visidata/threads.py:    precious = False
visidata/vdobj.py:        self.allSheets = []  # list of all non-precious sheets ever pushed
```
Deep-diving onto the only one that might be:

```
                                                                                                                                                                                                                
     if vs.precious and vs not in vd.allSheets:                                                                                                                                                                 
         vd.allSheets.append(vs)  
```

It seems like preciousness just determines whether it gets added to `allSheets`. `allSheets` is referred to in `cmdlog` and in `ALT-jump` which is why preciousness matters with those features, but not with `jump-prev`.

### DIFF
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -1035,6 +1035,7 @@ SheetsSheet.addCommand('gI', 'describe-selected', 'vd.push(DescribeSheet("descri
 SheetsSheet.addCommand('z^C', 'cancel-row', 'cancelThread(*cursorRow.currentThreads)', 'abort async thread for current sheet')
 SheetsSheet.addCommand('gz^C', 'cancel-rows', 'for vs in selectedRows: cancelThread(*vs.currentThreads)', 'abort async threads for selected sheets')
 IndexSheet.addCommand('g^S', 'save-selected', 'vd.saveSheets(inputPath("save %d sheets to: " % nSelected), *selectedRows, confirm_overwrite=options.confirm_overwrite)', 'save all selected sheets to given file or directory')
+SheetsSheet.addCommand(ENTER, 'open-row', 'dest=cursorRow; vd.sheets.remove(sheet) if not sheet.precious else None; vd.push(openRow(dest))', 'open sheet referenced in current row')
 
 BaseSheet.addCommand('q', 'quit-sheet',  'vd.quit(sheet)', 'quit current sheet')
 globalCommand('gq', 'quit-all', 'vd.quit(*vd.sheets)', 'quit all sheets (clean exit)')


anjakefala:
This change breaks joining tests, so this is not ready to review. =)

The simplest change may be that `jump-prev` looks for the most recent precious sheet.

anjakefala:
The tests are failing because re-opening the `sheets-stack` was not manually added to the cmdlog, and `join`s would occur from the sheets-stack.

anjakefala:
Found the origin of the bug!

```
SheetsSheet.addCommand(ENTER, 'open-row', 'dest=cursorRow; vd.sheets.remove(sheet) if not sheet.precious else None; vd.push(dest)')
```


This used to be the design for ENTER on a `SheetsSheet`.

This got lost when `open-row` became more generalised:

```
visidata/sheets.py:Sheet.addCommand(ENTER, 'open-row', 'vd.push(openRow(cursorRow))', 'open sheet with copies of rows referenced in current row')
```

ajkerrigan:
YES! Nice digging @anjakefala. I could have sworn there was some behavior like that at one point :). Although for the record I feel like your approach of handling the precious drop right in `vd.push()` feels more intuitive.

anjakefala:
@ajkerrigan It turns out things are a bit more complicated!

Example, if you open the vdmenu, and then PluginsSheet, and then press `q`, you close the vd session. Which feels unintuitive. For now, I am going to go with `stable` behaviour.

PR #669, opened 2020-08-23
[evalexpr] allow evalexpr on any descendant of BaseSheet

anjakefala:


visidata/sheets.py
```diff
@@ -328,7 +328,7 @@ def evalexpr(self, expr, row=None):
         if row:
             contexts = vd._evalcontexts.setdefault((self, self.rowid(row)), LazyComputeRow(self, row))
         else:
-            contexts = None
+            return super().evalexpr(expr)
```
saulpw: I would say let's leave this code alone; adding the BaseSheet.evalexpr should be sufficient.

saulpw (approved):

Issue #668, opened 2020-08-22
[expand-cols undo] undo does not consistently work for expand-cols


frosencrantz:
**Small description**
Undo doesn't always work with expand-cols, and can leave columns behind.

**Expected result**
Expanded columns are contracted.

**Actual result with screenshot**
https://asciinema.org/a/AoxbEyqboRGYNTfcx5MfsoSm6

**Steps to reproduce with sample data and a .vd**

The issue seems to happen when many columns are created:
```
jo a=$(jo -a {1..35}) b=$(jo -a {1..100}) | vd -f json
```
expand-cols g(
then undo

contract-col seems to work much better.

**Additional context**
Please include the version of VisiData. Latest version of develop.

anjakefala:
This should be fixed now on `develop`. The problem was that with complicated async commands, sometimes `vd.activeCommand` would reset before the Undo was fully set. Undo requires a `vd.activeCommand` to set on.

We moved where `vd.activeCommand` is reset, and so this issue should not be occurring anymore. This is a risky change, and it is possible it would need to be reverted.

anjakefala:
This bug came up https://github.com/saulpw/visidata/issues/700

We decided to revert the fix, for now. This is being actively worked on.

anjakefala:
This change should result in undos being more consistently set for expand-cols: https://github.com/saulpw/visidata/pull/704

Please update, and let me know if you run into any issues!

PR #670, opened 2020-08-24
[cmdlog] move activeCommand reset to beforeExecHook

anjakefala:
WARNING: risky commit; activeCommand reset has been in afterExecSheet
since early cmdlog

For complicated async commands like expand-cols, sometimes
afterExecSheet ran before their undos could be added.

addUndo adds undos to the activeCommand.

Tying the activeCommand reset to the execution of a fresh command will
fix that.


anjakefala:
Addresses https://github.com/saulpw/visidata/issues/668

PR #672, opened 2020-08-29
[api] Convert openSource to vd.open

anjakefala:


visidata/data.py
```diff
@@ -9,9 +9,11 @@
 option('filetype', '', 'specify file type', replay=True)
 option('incr_base', 1.0, 'start value for column increments', replay=True)
 
+__all__ = ['SettableColumn', 'inputFilename', 'numrange', 'updateColNames', 'num', 'fillNullValues', 'inputPath', 'loadInternalSheet', 'open_txt']
+
 
 def _default_colnames():
-    'A B C .. Z AA AB .. ZZ AAA .. to infinity'
+    'a b c .. z aa ab .. zz aaa .. to infinity'
```
saulpw: why was this changed? The names are actually uppercase.

anjakefala: I have no idea, I noticed that in the diff, and assumed it was you. I should have asked, I will revert it!

visidata/data.py
```diff
@@ -160,7 +162,7 @@ def openPath(vd, p, filetype=None):
     return openfunc(p)
 
 
-@VisiData.global_api
+@VisiData.api
```
saulpw: keep as `global_api` for now, we'll make that be the deprecation mechanism and clean it up later.  (this way other plugins/etc using it as a global should still work).

PR #685, opened 2020-09-08
Add reload-selected to IndexSheet

geekscrapy:
Useful to have, not only on SheetsSheet

### DIFF
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -1052,6 +1052,7 @@ Sheet.addCommand('gz'+ENTER, 'dive-selected-cells', 'for r in selectedRows: vd.p
 
 # when diving into a sheet, remove the index unless it is precious
 SheetsSheet.addCommand('g^R', 'reload-selected', 'for vs in selectedRows or rows: vs.reload()', 'reload all selected sheets')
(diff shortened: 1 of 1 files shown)


visidata/sheets.py
```diff
@@ -1052,6 +1052,7 @@ def updateColNames(sheet, rows, cols, overwrite=False):
 
 # when diving into a sheet, remove the index unless it is precious
 SheetsSheet.addCommand('g^R', 'reload-selected', 'for vs in selectedRows or rows: vs.reload()', 'reload all selected sheets')
+IndexSheet.addCommand('g^R', 'reload-selected', 'for vs in selectedRows or rows: vs.reload()', 'reload all selected sheets')
```
saulpw, line 1054: The SheetsSheet is also an IndexSheet, so we probably just want to change SheetsSheet to IndexSheet.  Please then verify that g^R still works on the SheetsSheet.

geekscrapy: Agreed, just done this and tested with SheetsSheet 👍

geekscrapy:
Updated to bind command to IndexSheet

anjakefala:
Thanks @geekscrapy! =) Merged.

PR #694, opened 2020-09-13
Move PyobjSheet related commands and class functions in vdcore to pyobj module

anjakefala:
Additionally, deprecates push_pyobj


saulpw (approved):
Thanks for cleaning this up, @anjakefala !

PR #696, opened 2020-09-14
Rework sparline for new 2.0 api

layertwo:
The sparkline plugin bit rot and needed to be updated to support the new api. I've converted it to that.

As a side note, I've also removed the feature of adding a sparkline from the ColumnSheet.

### DIFF
--- a/plugins/sparkline.py
+++ b/plugins/sparkline.py
@@ -1,37 +1,48 @@
-from visidata import Column, ColumnsSheet, Sheet, option, options, BaseException
+"""
+Generate sparkline column for numeric columns
+"""
+from visidata import Column, Sheet, option, options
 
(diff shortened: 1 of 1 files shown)


plugins/sparkline.py
```diff
@@ -1,37 +1,48 @@
...
+            if val < b:
+                output += lines[bounds.index(b) - 1]
+                break
+        else:
+            output += max(lines)
+    return output
+
+
+@Sheet.api
+def sparkline_col(sheet, source):
```
saulpw, line 39: Maybe rename `source` to `sourcecols`.

saulpw (approved):

saulpw:
Thanks for doing this, @layertwo !

PR #698, opened 2020-09-15
[syspaste] create a cmdlog entry for syspaste

anjakefala:
This allows us to Undo syspaste.


saulpw (approved):
This is a good idea, @anjakefala. I'm not 100% sure about recording paste values in the cmdlog, but without it syspaste would have to remain unreplayable.  So I think overall it's better to take it.

PR #704, opened 2020-09-21
[cmdlog risky] addUndo directly modifies cmdlog row if activeCommand …

### LINKED ISSUE #700: [develop] There should be one CommandLog entry per command executed
**Small description**
As a result of this commit: https://github.com/saulpw/visidata/commit/7b01e98c923b8bca2c297e85cc0f892b92f84b02 Using `exec-longname` results in the command executed having multiple entries in the **CommandLog**.

The problem is understood, but the approach needs to be figured out. We do not want complex commands to have undos missing, and we want only one entry in the **CommandLog** per command.

**Additional context**
The develop branch

anjakefala:
…has already been logged

WARNING: risky commit;

It is still possible to find race conditions if
the user presses commands fast enough, however, this will allow some grace for complex Undos.
If we want to further reduce race conditions, we can track sheet threads
and block more than one modification, but that is a much bigger change.

Closes #700, #668

### DIFF
--- a/visidata/undo.py
+++ b/visidata/undo.py
@@ -11,8 +11,12 @@ option('undo', True, 'enable undo/redo')
 def addUndo(vd, undofunc, *args, **kwargs):
     'On undo of latest command, call undofunc()'
     if options.undo:
-        r = vd.activeCommand
-        if not r:
+        # occurs when VisiData is just starting up
+        if not vd.sheet:
+            return
+        r = vd.modifyCommand
+        # some special commands, like open-file, do not have an undofuncs set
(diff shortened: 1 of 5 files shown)


visidata/undo.py
```diff
@@ -11,8 +11,10 @@
 def addUndo(vd, undofunc, *args, **kwargs):
     'On undo of latest command, call undofunc()'
     if options.undo:
-        r = vd.activeCommand
-        if not r:
+        if not vd.sheet:
```
saulpw, line 15: Let's add a comment about why we need each of these

Issue #708, opened 2020-09-28
[v2.-5] Cannot save a list of **dates** or dict of **dates** to json


anjakefala:
**Steps to reproduce with sample data and a .vd**
Open `benchmark.csv`.
Use `@` to make the `Date` column a **date** type.
Use `=` followed by *[Date]* to place it in a list.
Try to save to *benchmark.json*.

**Additional context**
v2.-5

anjakefala:
@ajkerrigan fixed this! :tada:

PR #709, opened 2020-09-28
Stringify inner objects directly when saving nested JSON

ajkerrigan:
For cells containing nested JSON, only the outermost value is a `Cell`
wrapper type. Inner objects can be stringified directly during
encoding, to avoid errors like #708.

Not sure if `str(obj)` is the preferred way to handle nested objects, but the general approach of "unwrap/encode the outer cell, and directly encode inner bits" seems reasonable.


saulpw (approved):
Thanks for fixing this bug, @ajkerrigan!

PR #710, opened 2020-09-28
Don't lose data when updating a pandas index

ajkerrigan:
TL;DR: Oops.

This partially reverts commit fc63620802cd5526f3828970b4a17bf20bbd6998.

If we have an Int64Index with valuable data, converting it to a `RangeIndex` steamrolls over that data. It is safer to use `reset_index()`, so the previous index becomes a column and we don't lose anything. The way it used to work in the first place!

😊


saulpw (approved):
Okay, I trust that this is better :)

PR #724, opened 2020-10-13
Add traceback warnings for deprecated calls

ajkerrigan:
During a VisiData 2.0 update pass on some of my plugins I ran into a helpful deprecation message about the `load_pyobj` function. I peeked at how VisiData generates deprecation notices, and saw a comment that said "ideally would include a stacktrace". And hey, that does sound useful! So here's one way we could provide a stack trace for deprecated calls. It produces this if I try to use `load_pyobj({})` in an `exec-python` string:

image

Useful? Not useful? Needs tweaks? I don't know, worth a shot though :).

### DIFF
--- a/visidata/deprecated.py
+++ b/visidata/deprecated.py
@@ -1,4 +1,4 @@
-from visidata import Column, PyobjSheet, VisiData, vd
+from visidata import VisiData, vd
 import visidata
 
 alias = visidata.BaseSheet.bindkey
@@ -42,7 +42,7 @@ def SubrowColumn(*args, **kwargs):
 
 @deprecated('1.6')
 def DeferredSetColumn(*args, **kwargs):
-    return Column(*args, defer=True, **kwargs)
+    return visidata.Column(*args, defer=True, **kwargs)
 
 @deprecated('2.0')
 def bindkey_override(keystrokes, longname):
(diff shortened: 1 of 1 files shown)


visidata/deprecated.py
```diff
@@ -1,16 +1,20 @@
-from visidata import VisiData
+from visidata import Column, PyobjSheet, VisiData, vd
```
saulpw: For consistency with how it is in this file currently, and Reasons, let's keep this importing only `VisiData` and `vd`; reference the rest like `visidata.Column`.

visidata/deprecated.py
```diff
@@ -1,16 +1,20 @@
...
 alias = visidata.BaseSheet.bindkey
 
 def deprecated(ver, instead=''):
     def decorator(func):
         def wrapper(*args, **kwargs):
-            # ideally would include a stacktrace
+            import traceback
+
+            for line in reversed(traceback.extract_stack(limit=6)[:-1]):
+                vd.warning(f'    file {line.filename} at line {line.lineno} in {line.name}')
```
saulpw, line 12: What do you think about putting the entire traceback in a single multiline string?

ajkerrigan: I'd like that if you could help me sort out how. I tried a couple ways of sending a multi-line string to `vd.warning`. The message history seemed to have trouble with `\t` or `\n` characters:

image

If there's a nice way to avoid calling `vd.warning` repeatedly I'm all for it.

visidata/deprecated.py
ajkerrigan: Sure thing, I support Reasons 😁 !

visidata/deprecated.py
saulpw: If you go multiline with `v`, the newlines are handled better (not sure about the tabs or overall workability though).  If we can get this to work reasonably, I prefer it somewhat, but if it's too much of a pain, we can stick with the original stream of vd.warning.

Actually, now that I think about it, in the deprecated usage case, do we actually want the full stack trace, or just the file/line where the deprecated function is used directly?

PR #728, opened 2020-10-14
Fix HDF5 loader

amotl:
Hi there,

after trying to load a HDF5 file at hand, we found different problems when drilling down into leaf nodes. These updates improve things slightly.

With kind regards,
Andreas.

### DIFF
--- a/visidata/loaders/hdf5.py
+++ b/visidata/loaders/hdf5.py
@@ -32,7 +32,7 @@ class Hdf5ObjSheet(Sheet):
             elif len(source.shape) == 2:  # matrix
                 ncols = source.shape[1]
                 for i in range(ncols):
-                    self.addColumns(ColumnItem('', i, width=8), index=i)
+                    self.addColumn(ColumnItem('', i, width=8), index=i)
                 self.recalc()
                 yield from source  # copy
             else:
@@ -42,10 +42,16 @@ class Hdf5ObjSheet(Sheet):
 
 
     def openRow(self, row):
+        import h5py
         if isinstance(row, BaseSheet):
             return row
-        if isinstance(row, h5py.Object):
-            return H5ObjSheet(row)
+        if isinstance(row, h5py.HLObject):
+            return Hdf5ObjSheet(row)
+
+        import numpy
+        from .npy import NpySheet
+        if isinstance(row, numpy.ndarray):
+            return NpySheet(None, npy=row)
 
 
 Hdf5ObjSheet.addCommand('A', 'dive-metadata', 'vd.push(SheetDict(cursorRow.name + "_attrs", source=cursorRow.attrs))', 'open metadata sheet for object referenced in current row')


anjakefala:
Hi @amotl!

I use the hdf5 loader, and I am intrigued by your suggested changes. Thanks for contributing. I am going to take on reviewing this. I do have one question, which I will leave in a comment.

requirements.txt
```diff
@@ -6,7 +6,7 @@ requests            # http
 lxml                # html/xml
 openpyxl            # xlsx
 xlrd                # xls
-h5py                # hdf5
+h5py==2.10.0        # hdf5
```
anjakefala: We do not generally like to version lock, unless it is necessary. This is because it might result in the user having to navigate dependency conflicts.

Why is 2.10.0 the preferred version?

amotl: No worries. I've just removed the version pinning on h5py. Thanks!

visidata/loaders/hdf5.py
```diff
@@ -42,10 +42,16 @@ def iterload(self):
...
+        import h5py
         if isinstance(row, BaseSheet):
             return row
-        if isinstance(row, h5py.Object):
-            return H5ObjSheet(row)
+        if isinstance(row, h5py.HLObject):
+            return Hdf5ObjSheet(row)
+
+        import numpy
+        from .npy import NpySheet
```
anjakefala, line 52: Is the `from .npy` needed? `from visidata import *` fails to import it?

visidata/loaders/hdf5.py
```diff
@@ -42,10 +42,16 @@ def iterload(self):
 
 
     def openRow(self, row):
+        import h5py
         if isinstance(row, BaseSheet):
             return row
-        if isinstance(row, h5py.Object):
-            return H5ObjSheet(row)
+        if isinstance(row, h5py.HLObject):
```
anjakefala, line 48: What is the difference between `h5py.Object` and `h5py.HLObject`?

anjakefala (commented):
Thanks for finding and fixing these bugs! ^^

I have a couple of questions about the thought behind some changes. Once I understand those, I would be happy to merge this.

visidata/loaders/hdf5.py
amotl: While I recognized `import *` is used all over the place throughout the code base, it is considered bad practice in general (no offense!). So, my mind didn't actually lean towards not having to import specific things ;].

However, I just tried to remove that line based on your suggestion. But then, VisiData croaks:

image

visidata/loaders/hdf5.py
amotl: It looks like, with the most recent version of `h5py`, `h5py.Object` is no more:
```
>>> import h5py
>>> dir(h5py)
['AttributeManager', 'Dataset', 'Datatype', 'Empty', 'ExternalLink', 'File', 'Group', 'HLObject', 'HardLink', 'Reference', 'RegionReference', 'SoftLink', 'VirtualLayout', 'VirtualSource', '_ModuleWrapper', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', '__version__', '_conv', '_errors', '_hl', '_objects', '_proxy', '_register_converters', '_register_lzf', '_warn', 'absolute_import', 'check_dtype', 'check_enum_dtype', 'check_ref_dtype', 'check_string_dtype', 'check_vlen_dtype', 'defs', 'enable_ipython_completer', 'enum_dtype', 'filters', 'get_config', 'get_enum', 'get_vlen', 'h5', 'h5a', 'h5ac', 'h5d', 'h5ds', 'h5f', 'h5fd', 'h5g', 'h5i', 'h5l', 'h5o', 'h5p', 'h5pl', 'h5py_warnings', 'h5r', 'h5s', 'h5t', 'h5z', 'highlevel', 'is_hdf5', 'new_enum', 'new_vlen', 'ref_dtype', 'regionref_dtype', 'register_driver', 'registered_drivers', 'run_tests', 'special_dtype', 'string_dtype', 'unregister_driver', 'utils', 'version', 'vlen_dtype']
```

Using `h5py.HLObject` instead was just my best guess and it worked! ;]

visidata/loaders/hdf5.py
anjakefala: This makes sense!

visidata/loaders/hdf5.py
anjakefala: Gotcha. So, this will croak with earlier versions of `h5py`?

amotl: It looks like there has never been a `h5py.Object`.
```
git clone https://github.com/h5py/h5py
cd h5py
git log -p | grep 'class Object' | grep -v ObjectID | grep -v ObjectHeader
```

image

-- https://geekandpoke.typepad.com/geekandpoke/2010/10/geekpoke-looks-behind-the-scenes-of-coders.html

saulpw: Nice find, @amotl, and thanks for digging into it.  That may have just been dead code.  I'm glad the hdf5 loader is getting some attention!

anjakefala: Yeah, thanks! =)

PR #754, opened 2020-10-26
Add mode and stdev to aggregator options

jsvine:
Perhaps `mode` and `stdev` aren't options for a particular reason; if so, my apologies. But it seemed logical to me to add them to the aggregator options, especially since (a) they can be drawn from the Python standard library and require no dependencies, and (b) they're already defaults on the Describe Sheet.

Two additional notes / judgment calls:

- Running stdev on n=1 groupings produces the following `ValueError` (captured gracefully by the aggregator architecture): `variance requires at least two data points`. I'd understand if you wanted a different default behavior for that.

- I've slotted the new aggregators in the order that made the most sense for me (`mode` after `median`, and `stdev` before the quantiles) but totally understand if you prefer a different ordering.

Thanks, and thanks for all your work on VisiData!

### DIFF
--- a/visidata/aggregators.py
+++ b/visidata/aggregators.py
@@ -1,6 +1,7 @@
 import math
 import functools
 import collections
+from statistics import mode, stdev
 
 from visidata import Progress, Column
 from visidata import *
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Sounds good to me.  Now that we have the fancy chooser, I'm game for adding any aggregators that only use Python stdlib.

@anjakefala Let's put together a checklist for what to do when adding an aggregator (like update the documentation, describe sheet, etc) and then we can do those things in this PR and merge all at once.

geekscrapy:
@jsvine you beat me too it 👍  I made this a while ago and forgot to share: https://github.com/geekscrapy/visidata_plugins/blob/master/agg_stdev.py

I like the fact it gives 3 columns (like the quartile agg) so you can see how many are in each sector. See if you like it, if you do feel free to include, if not that's fine too!!

anjakefala:
Thanks @jsvine :blush:

Issue #769, opened 2020-11-07
[loader] CoNLL Loader


polm:
Hello, I thought it'd be handy to have a loader for CoNLL data. I actually went ahead and implemented it in [my fork][]; I didn't run into any issues but it might be possible to make it easier to use.

CoNLL is a data format used for training data for Natural Language Processing. It's basically TSV and each line represents a word, but there are also special comment lines for each sentence. There are also some tricky details to how fields are formatted. Thankfully these are handled by the [pyconll][] library, so writing sheet to wrap it for Visidata isn't that hard.

Having access to this data in Visidata makes it easy to answer questions like "how common is X part of speech?" or "what is the longest sentence in our data? is it an error?". For training data that has to be checked and sometimes adjusted to match changing specs this is very helpful.

You can see the short spec for the format [here][conllu], and below is a relatively short sample of raw CoNLL data, two sentences from the [Japanese UD GSD corpus][ja-gsd]:

[my fork]: https://github.com/polm/visidata/tree/conll
[pyconll]: https://github.com/pyconll/pyconll
[ja-gsd]: https://github.com/UniversalDependencies/UD_Japanese-GSD/blob/master/ja_gsd-ud-dev.conllu
[conllu]: https://universaldependencies.org/format.html

<details markdown=1>

```
# sent_id = dev-s1
# text = ただし、50周年ソングに変更後は、EDも歌つきのものが使われた。
1	ただし	ただし	CCONJ	接続詞	_	18	cc	_	BunsetuPositionType=SEM_HEAD|LUWBILabel=B|LUWPOS=接続詞|SpaceAfter=No|UniDicLemma=但し
2	、	、	PUNCT	補助記号-読点	_	1	punct	_	BunsetuPositionType=CONT|LUWBILabel=B|LUWPOS=補助記号-読点|SpaceAfter=No|UniDicLemma=、
3	50	50	NUM	名詞-数詞	_	5	nummod	_	BunsetuPositionType=CONT|LUWBILabel=B|LUWPOS=名詞-普通名詞-一般|SpaceAfter=No|UniDicLemma=五十
4	周年	周年	NOUN	名詞-普通名詞-助数詞可能	_	5	compound	_	BunsetuPositionType=CONT|LUWBILabel=I|LUWPOS=名詞-普通名詞-一般|SpaceAfter=No|UniDicLemma=周年
[...]

PR #772, opened 2020-11-09
Add conll loader to plugins file

polm:
This is following up on #769. The plugin appears to be working so I think it's good for an initial release, let me know if anything else is necessary.


anjakefala:
Thanks! :blush: 

Could you please place a small sample conll dataset in your plugin's repo? =)

Additionally, it would be great if you ran `vd -b sample_data/sample.conll -o output.jsonl` (where output.jsonl can be any of the default formats, like tsv), and also added `output.jsonl` to the repo, so that we have an idea of what the loaded `sample.conll` is expected to look like!

polm:
Got it, I added a sample data file and `output.jsonl`. Let me know if there's anything else I should do.

plugins/plugins.jsonl
```diff
@@ -11,3 +11,4 @@
...
 {"name": "diff", "description": "adds command to create diff sheets", "maintainer": "Anja Kefala @anjakefala", "latest_release": "2020-10-09", "url": "https://raw.githubusercontent.com/saulpw/visidata/492d558ff4aba55a192965a27ae6499f8d219072/plugins/diff.py", "latest_ver": "0.9", "visidata_ver": "2.0", "sha256": "0085bf8afdc9abf74b7dd52c251f7b7f7befc507aae262bbb656c2ab7379ddd1"}
 {"name": "marks", "description": "adds commands for marking selected rows, and selecting + viewing marked rows", "maintainer": "Saul Pwanson @saulpw", "latest_release": "2020-10-09", "url": "https://raw.githubusercontent.com/saulpw/visidata/54e49feee97a607cce5355bf0e3a79b2466c3d8a/plugins/marks.py", "latest_ver": "0.1", "visidata_ver": "2.0", "sha256": "f15977f327ccc387be208922b66afe4b65666912e8066d02352c53b1aaff941f"}
+{"name":"conll", "description":"CoNLL data loader", "maintainer": "Paul McCann <polm@dampfkraft.com>", "latest_release": "2020-11-09", "latest_ver": "0.1.0", "url": "https://github.com/polm/visidata-conll", "visidata_ver": "v2.0", "pydeps": "pyconll", "sha256": "a757ba4907baf5f9c001d817e1cefe38f6bbebed31f1e79783a94091e39810c1"}
```
anjakefala: `url` needs to point to the raw text on a specific commit! This url is what is used to download the file.

E.g. https://raw.githubusercontent.com/polm/visidata-conll/a11ffa99fe497b83c522bef86d89b246b2a8d6e0/conll.py

anjakefala (changes requested):
The `url` in the jsonl needs to point to a specific commit of `conll.py` (see my comment).

Once that is taken care of, I will merge this in, and deploy it. =) This is really cool, thank you for contributing!

**Edit**: I originally had a comment that `conll.py` needs a `vd.addGlobals(globals())`. This is in motion to be deprecated. But, I just checked, and the alternative (using `@VisiData.api`) does not neatly work for plugins, yet. 

**It would probably be best if you add `vd.addGlobals(globals())` for now, so that your plugin works for VisiData v2.0.**

If you do not add it, your plugin would still be loaded if someone manually wrote `from plugins.conll import *`, but would not if someone used `import plugins.conll`.

polm:
Thanks for the comments, I pushed the changes as described. 

Also thanks for the clarification about `addGlobals`, I had thought that wasn't needed if the decorator was used.

anjakefala (approved):
Awesome! Thank you! =)

`addGlobals()` will not be needed in v2.1 ^^

PR #785, opened 2020-11-15
Support running as `python -m visidata`

abitrolly:


ajkerrigan:
This seems like a useful change (personally, it would help me clean [this](https://github.com/ajkerrigan/visidata-oxide/blob/ff96245/pyoxidizer.bzl#L162) up). What's the motivation for you @abitrolly?

abitrolly:
@ajkerrigan I use a version 1.5.2 from Fedora 33 and I wanted to play with latest version.

Actually, I created `.vd` file and wanted to clean it up from non-deterministic operations (such as those dependent on value under cursor), then add a header to list the command that provides input data.

I attach example which is non-cleaned up.
[lxd-minimal-image.zip](https://github.com/saulpw/visidata/files/5542799/lxd-minimal-image.zip)

PR #807, opened 2020-11-24
[tests] Run tests with an empty config file and visidata dir

tsibley:
Otherwise user-preferences and plugins affect test success/failure.  For
example, this bit me because I have a custom disp_date_fmt.

---

I first ran into this issue a while ago and thought I'd submitted a PR for it, but I couldn't for the life of me find it. So I guess I didn't!


saulpw (approved):
Thanks for fixing it, @tsibley!  Sorry it bit you.

tsibley:
@saulpw Ah, no worries at all! Thanks for the merge!

Issue #535, opened 2020-04-26
[wishlist] ';' to utilise named capturing groups


geekscrapy:
it would be very useful to be able to use named capturing groups to automatically name columns based on the capture group name. This would significantly decrease the time to rename all the columns when a lot of regex groups are captured.

Another very useful aspect is this: I am building up a library of regex matches for different log formats. when I open one of these log files currently I have to run the match command and then rename multiple columns manually.

I think that completely user driven, deterministic (although I appreciate they are deterministic now!) columns names would be a great addition for ';'

anjakefala:
Hi @geekscrapy! Thanks for making a feature request.

SInce we are really trying to get 2.0 out, our current procedure is to kondo wishlist items that are not on the 2.0 roadmap. @saulpw may revert this decision, but I am going to close for now.

If you could please add an example (i.e. these are the columns, this was my capture-group regex, these should be the resulting column names), that will help me when I pick this up in the future.

It is cool to see people using `;`. :blush:

geekscrapy:
Np about closing the issue, just wanted it to be logged!

Use cases for this would be to auto parse logs (which could be an awesome plugin that's floating around in my head) to vd rows.

A simple example would be this:
`127.0.0.1 - frank [10/Oct/2000:13:55:36 -0700] "GET /apache_pb.gif HTTP/1.0" 200 2326`

The regex for this with named groups would be:
`(?P<src_ip>.*?)\s(?P<ident>.*?)\s(?P<user>.*?)\s\[(?P<finish_processing>.*?)\]\s"(?P<req_method>\w+)\s(?P<req_uri>.*?)\s(?P<req_protocol>.*)"\s(?P<resp_status>\d+)\s(?P<resp_bsize>\d+)`

In a small py programe:

```
>>> import re
>>> log = '127.0.0.1 - frank [10/Oct/2000:13:55:36 -0700] "GET /apache_pb.gif HTTP/1.0" 200 2326'
>>> p = re.compile(r'(?P<src_ip>.*?)\s(?P<ident>.*?)\s(?P<user>.*?)\s\[(?P<finish_processing>.*?)\]\s"(?P<req_method>\w+)\s(?P<req_uri>.*?)\s(?P<req_protocol>.*)"\s(?P<resp_status>\d+)\s(?P<resp_bsize>\d+)')
>>> p.search(log).groupdict()
[...]

PR #808, opened 2020-11-24
[regex] Use capture names for column names, if available, in capture-col

tsibley:
Allows for predetermining friendlier column names, saving a renaming
step later.

### DIFF
--- a/tests/capture-col-named.vd
+++ b/tests/capture-col-named.vd
@@ -1,0 +1,3 @@
+sheet	col	row	longname	input	keystrokes	comment
+			open-file	sample_data/benchmark.csv	o	
+benchmark	Date		capture-col	(?P<month>\d{1,2})/(?P<day>\d{1,2})/(?P<year>\d{4}) (?P<hour>\d{1,2}):(?P<min>\d{2})(?P<meridiem>[ap])	;	add new column from capture groups of regex; requires example row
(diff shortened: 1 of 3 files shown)


geekscrapy:
FYI, addresses #535 👍

tsibley:
@geekscrapy Ah, thanks for that pointer. Glad I'm not the only one who wants this feature! :-)

I notice that in your plugin version the capture group names are used as-is. In this PR version, they're prefixed with the original column name. I went with the prefix because it paralleled the existing numbering behaviour (e.g. `col_re0`, `col_re1`, etc), but I know there will be times when I really don't want the original column name crufting things up. I didn't have any good ideas for how to parameterize that behaviour (other than with an additional command, which I didn't want to do), but maybe you or someone else does?

One idea I had but that I didn't like was to reverse the default (so no prefix by default) and then flag that you want a prefix by naming your captures with a leading underscore. But again, I think this is probably worse than just defaulting to a prefix with no way to change it.

geekscrapy:
Yes that does keep inline I guess. Tbh, when I created that it was to scratch an itch I had so didn't think about it much.

My feeling is that it would be great to have the option of both (group names and extended col names) as you could potentially be creating a lot of columns with this.

On the flip side, if you know a way to mass rename columns this maybe isn't so much of an issue?

saulpw:
Let's keep the prefix by default, especially as it's easier to take out than insert after-the-fact.  As @geekscrapy mentioned, columns can be mass renamed in the ColumnsSheet with e.g. `g*` which makes it easy enough to remove a prefix (though you do have to input the column name or fiddle with regex, so we could streamline that at some point--maybe with formally segmented column names).  So this is great @tsibley, thanks a lot for putting this together.

Where should we make a note of this in the user docs, @anjakefala?

tsibley:
@saulpw Thanks! I've pushed [another PR](https://github.com/saulpw/visidata/pull/810) to note this in the docs. I should have done that the first time around! :-)

PR #812, opened 2020-11-29
[MissingAttrFormatter] handle missing attributes in status formatting

### LINKED ISSUE #764: [dev 2.0+] disp_rstatus_fmt causes halting exception on GraphSheet
**Small description**
I modified the `disp_rstatus_fmt` to provide information I require when working with sheets, however when using GraphSheet vd crashes with the following exception. I currently have to use `^g` to see the same information as I would be able to see using this `disp_rstatus_fmt`.

`options.disp_rstatus_fmt = ' {sheet.cursorRowIndex}/{sheet.nRows}[{sheet.nSelectedRows}] {sheet.rowtype} '`

`AttributeError: 'GraphSheet' object has no attribute 'cursorRowIndex'`

**Expected result**
GraphSheet to load

FWIW, I was also attempting to make a % counter in `disp_rstatus_fmt` however I get some odd other exceptions (this isn't really an issue so I haven't opened a separate bug, however this was my original use-case).

anjakefala:
Closes #764

This PR adds `MissingAttrFormatter`, a subclass of [string.Formatter](https://docs.python.org/3.8/library/string.html#string.Formatter) which handles `AttributeErrors` and `KeyErrors` in format strings, substituting them with `''`.
This was then used to format the `rightStatus`. 
This allows users to safely customise the `disp_rstatus_fmt` with relevant attributes that not every sheet contains.

### DIFF
--- a/visidata/statusbar.py
+++ b/visidata/statusbar.py
@@ -1,7 +1,7 @@
 import collections
 import curses
 
-from visidata import vd, VisiData, BaseSheet, Sheet, ColumnItem, Column, RowColorizer, options, colors, wrmap, clipdraw, ExpectedException, update_attr, theme
+from visidata import vd, VisiData, BaseSheet, Sheet, ColumnItem, Column, RowColorizer, options, colors, wrmap, clipdraw, ExpectedException, update_attr, theme, MissingAttrFormatter
 
 
 __all__ = ['StatusSheet', 'status', 'error', 'fail', 'warning', 'debug']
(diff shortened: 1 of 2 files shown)


visidata/statusbar.py
```diff
@@ -153,7 +153,7 @@ def drawLeftStatus(vd, scr, vs):
 @VisiData.api
 def rightStatus(vd, sheet):
     'Return right side of status bar.  Overrideable.'
-    return options.disp_rstatus_fmt.format(sheet=sheet, vd=vd)
+    return MissingAttrFormatter().format(options.disp_rstatus_fmt, sheet=sheet, vd=vd)
```
saulpw: Also let's use `sheet.options.disp_rstatus_fmt`, which will allow different sheet classes to have different options.

Issue #787, opened 2020-11-15
[macros] Recording a macros fails


frosencrantz:
**Small description**
Recording a macro fails.

**Expected result**
Macro is saved.

**Actual result with screenshot**
https://asciinema.org/a/7T0YgyrSQFhHyWfdok17oi6Qj

```
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/visidata/basesheet.py", line 136, in execCommand
    exec(code, vdglobals, LazyChainMap(vd, self))
  File "macro-record", line 1, in <module>
  File "/usr/lib/python3.9/site-packages/visidata/macros.py", line 52, in startMacro
    vd.cmdlog.saveMacro(vd.macroMode.rows, ks)
  File "/usr/lib/python3.9/site-packages/visidata/macros.py", line 37, in saveMacro
    append_tsv_row(vd.macrosheet, (ks, macropath))
  File "/usr/lib/python3.9/site-packages/visidata/loaders/tsv.py", line 81, in append_tsv_row
    if not vs.source.exists():
AttributeError: 'NoneType' object has no attribute 'exists'
```

**Steps to reproduce with sample data and a .vd**
Record a macro, provide a filename when asked, and vd throws an exception.

**Additional context**
Please include the version of VisiData. Latest develop version.

anjakefala:
This should now be fixed on `develop`!

frosencrantz:
Hi @anjakefala! I was trying to confirm the fix.  But I'm not sure how to run a macro.  It looks like the macro name becomes the name of a new command, but when I try to run it, I get a KeyError: ' '  It is getting late, I don't have time now to look further.  Will have to take a look when I have time.

frosencrantz:
Hi @anjakefala I think I understand my issue.  

When it is time to save the macro, I wasn't giving a keystroke, but giving it longer name, `delete-ten-rows` rather than `zx`.  If I then try to do a long-command: `delete-ten-rows`, I get the KeyError.

https://asciinema.org/a/mUuTYelsvQ6gduB4qMKQpaWmM

```
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/visidata/basesheet.py", line 136, in execCommand
    exec(code, vdglobals, LazyChainMap(vd, self))
  File "cmdlog", line 1, in <module>
KeyError: ' '
```

anjakefala:
@saulpw, should we allow longnames, as well as keystrokes? Hmmmmm.

[...]

PR #813, opened 2020-11-29
[macros] fixes for #787

anjakefala:
- pass original macrosheet to append_tsv_row
    - vd.macrosheet returns a real_macrosheet IndexSheet
    - real_macrosheet is an IndexSheet with a slightly different
    structure than the actual file on disk recording the macros
    - that original macros file is the one we actually want to
    update
- perform cmdlog.afterExecSheet after internal macro cmdlog is updated
    - internal cmdlog macro must be updated before the activeCommand is
    reset

### DIFF
--- a/visidata/macros.py
+++ b/visidata/macros.py
@@ -9,7 +9,7 @@ def macrosheet(vd):
     macrospath = Path(os.path.join(options.visidata_dir, 'macros.tsv'))
     macrosheet = vd.loadInternalSheet(TsvSheet, macrospath, columns=(ColumnItem('command', 0), ColumnItem('filename', 1))) or vd.error('error loading macros')
 
-    real_macrosheet = IndexSheet('user_macros', rows=[])
+    real_macrosheet = IndexSheet('user_macros', rows=[], source=macrosheet)
     for ks, fn in macrosheet.rows:
         vs = vd.loadInternalSheet(CommandLog, Path(fn))
         vd.status(f"setting {ks}")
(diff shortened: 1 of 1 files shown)


visidata/macros.py
```diff
@@ -10,6 +10,7 @@ def macrosheet(vd):
     macrosheet = vd.loadInternalSheet(TsvSheet, macrospath, columns=(ColumnItem('command', 0), ColumnItem('filename', 1))) or vd.error('error loading macros')
 
     real_macrosheet = IndexSheet('user_macros', rows=[])
+    real_macrosheet.origin = macrosheet
```
saulpw: maybe we could use 'source' instead of origin, and pass it to the constructor.

visidata/macros.py
```diff
@@ -34,16 +35,18 @@ def saveMacro(self, rows, ks):
...
 
 @CommandLog.api
 @wraps(CommandLog.afterExecSheet)
 def afterExecSheet(cmdlog, sheet, escaped, err):
-    cmdlog.afterExecSheet.__wrapped__(cmdlog, sheet, escaped, err)
     if vd.macroMode and (vd.activeCommand is not None) and (vd.activeCommand is not UNLOADED):
         cmd = copy(vd.activeCommand)
         cmd.row = cmd.col = cmd.sheet = ''
         vd.macroMode.addRow(cmd)
+    # happens after, bc
```
saulpw: I didn't understand this comment, what does it mean?

anjakefala: This invocation of the original `afterExecSheet` needs to happen after the command is added to the `macro.vd`, because once `afterExecSheet` runs, then `activeCommand` will be set to `None`.

anjakefala: I tried to make it clearer

saulpw (approved):

PR #815, opened 2020-11-29
Some selected rows

### LINKED ISSUE #767: [wishlist option] Have commands (like g*) work on all rows if none selected
**Small description**
Since rev2.0.1 g* does not work if no rows are selected.
**Expected result**
In 1.5.2 g* would work (do regex replacement in place) without selecting any rows. I did not expect this to change on 2.0.1, I'm not sure if this is intentional but now I have to do gs g* instead.
"no rows selected" is shown in the status line (bottom).

Load any datasheet go to a column and type g*

**Additional context**
Worked on 1.5.2
Stopped working on 2.0.1

anjakefala:
Renames `someSelectedRows` to `onlySelectedRows`. `onlySelectedRows` will return selected rows or `fail()` if none are selected.

Adds `someSelectedRows`. `someSelectedRows` will return selected rows, or all rows if none are selected when `options.selected_or_rows` is True. In batch mode or when `options.selected_or_rows` is False, it behaves like `selectedRows`. 

So far, transitioned `g*` and `gz*` to use `someSelectedRows`. The rest remained as they were. Commands will be transferred on a case-by-case basis.

Closes #767.

### DIFF
--- a/docs/api/data.rst
+++ b/docs/api/data.rst
@@ -4,6 +4,7 @@ Selected Rows
 Each TableSheet has a set of *selected rows*, which is a strict subset of the rows on the sheet.
 
 .. autoattribute:: visidata.TableSheet.selectedRows
+.. autoattribute:: visidata.TableSheet.someSelectedRows
 .. autoattribute:: visidata.TableSheet.onlySelectedRows
 .. autoattribute:: visidata.TableSheet.nSelectedRows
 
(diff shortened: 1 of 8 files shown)


saulpw (commented):
Thanks @anjakefala, this is great.  The only thing I'm not sure about is the actual option name.   Any suggestions on other names for this option?  @maufdez?

anjakefala:
Set the option to `some_selected_rows` until we come up with a better name.

Some ideas:

* `empty_select_rows`
* `empty_select_fail`
* `empty_select_all_rows`
* `none_select_*`
* `selected_or_rows`
* `some_selected_rows`

Issue #802, opened 2020-11-21
SettingWithCopyWarning: A value is trying to be set on a copy of a slice from a DataFrame.


s-celles:
Hello,

This line

https://github.com/saulpw/visidata/blob/6691869060c8c3d0fcb4fb3d528f87af2dd9ad07/visidata/loaders/pandas_freqtbl.py#L108

raises

`SettingWithCopyWarning: A value is trying to be set on a copy of a slice from a DataFrame.`

(I was using `visidata.view_pandas` method in an IPython console)

This kind of warning can generally be fixed using `df.loc[...]`
See https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy for more informations

Steps to reproduce

```
$ ipython
Python 3.7.6 | packaged by conda-forge | (default, Mar 23 2020, 22:45:16)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.13.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import pandas as pd

In [2]: import visidata

In [3]: df = pd.read_csv("https://jsvine.github.io/intro-to-visidata/_downloads/83e70cf67e909f3ac177575439e5f3c5/faa-wildlife-strikes.csv")

In [4]: visidata.view_pandas(df[["STATE"]])
```

Press Shift + F

![Capture d’écran 2020-11-21 à 22 54 27](https://user-images.githubusercontent.com/109167/99888551-081b2500-2c4e-11eb-9b80-23c0f4827a23.png)

Kind regards

ajkerrigan:
Hi there @scls19fr - you can give the changes in #816 a shot when you get a chance, they should resolve this issue.

s-celles:
LGTM
Thanks @ajkerrigan

Issue #800, opened 2020-11-21
view_pandas(df.transpose()) raises AttributeError: 'int' object has no attribute 'startswith'


s-celles:
Hello,

The following code:

```python
import visidata
import pandas as pd
df = pd.read_csv("https://jsvine.github.io/intro-to-visidata/_downloads/83e70cf67e909f3ac177575439e5f3c5/faa-wildlife-strikes.csv")
visidata.view_pandas(df)
df = df.transpose()
visidata.view_pandas(df)
```

raises

```
AttributeError: 'int' object has no attribute 'startswith'
```

A temporary fix is to do

```
df.columns = df.columns.map(str)
```

but using such a transposed version of data is very, very slow.

Any idea?

Kind regards

Related #129, #631 (about transpose on visidata side)

ajkerrigan:
I'm not sure the best way to fix this properly. VisiData seems to assume that column names will be strings, so violating that assumption seems likely to cause odd errors both in the core functionality and any plugins that implicitly inherit that assumption.

Stringifying the colum names isn't an ideal fix, but it's at least more palatable than dying with an `AttributeError`.

saulpw:
Yeah, I think this is reasonable.  The other option I can think of, is to leave the column names blank if transposing a sheet in which the first/key column is numeric.  But I think that's worse in a couple ways, so let's stick with the column-name-stringification approach.

s-celles:
Thanks for this fix.
But watching transposed version of df is still very slow!

saulpw:
Sorry, @scls19fr, we don't know a lot about pandas. Are you interested in trying to make it faster? If you know how to transpose quickly with pandas, it shouldn't be too difficult to make a custom transpose command for PandasSheet that uses the optimized version.  I'm sure it would make a lot of pandas users very happy!

s-celles:
I don't think the problem is on pandas side but visidata code seems to be optimized for data with small number of columns but big number of rows.
When transposing with Pandas you can notice that it's a very quick operation but the problem seems to be when trying to display such a wide data using visidata (ie low number of rows but high number of columns)
My guess is that your are ingesting the whole data into visidata instead of just having a "view" on data (but I may be wrong as I didn't looked intensively to visidata code).
Some years ago @wavexx did a great work on https://github.com/TabViewer/gtabview to provide such a mechanism (unfortunately it wasn't merged into [tabview](https://github.com/TabViewer/tabview)

saulpw:
[...]

Issue #798, opened 2020-11-21
visidata.view_pandas(None) raises UnboundLocalError: local variable 'df' referenced before assignment


s-celles:
Hi @saulpw,

I'm aware that's an odd usage...

```python
In [3]: import visidata

In [4]: visidata.view_pandas(None)
---------------------------------------------------------------------------
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-4-5d0ae9b435d5> in <module>
----> 1 visidata.view_pandas(None)

~/opt/anaconda3/lib/python3.7/site-packages/visidata/loaders/_pandas.py in view_pandas(df)
    318
    319 def view_pandas(df):
--> 320     run(PandasSheet('', source=df))
    321
    322

~/opt/anaconda3/lib/python3.7/site-packages/visidata/mainloop.py in run(*sheetlist)
    223     curses.use_env(True)
    224
--> 225     ret = wrapper(cursesMain, sheetlist)
    226
    227     if ret:

~/opt/anaconda3/lib/python3.7/site-packages/visidata/mainloop.py in wrapper(f, *args)
    213
    214 def wrapper(f, *args):
--> 215     return curses.wrapper(setupcolors, f, *args)
    216
    217 def run(*sheetlist):

~/opt/anaconda3/lib/python3.7/curses/__init__.py in wrapper(*args, **kwds)
    100             pass
    101
--> 102         return func(stdscr, *args, **kwds)
    103     finally:
    104         # Set everything back to normal

~/opt/anaconda3/lib/python3.7/site-packages/visidata/mainloop.py in setupcolors(stdscr, f, *args)
    210             curses.mouseEvents[getattr(curses, k)] = k
    211
--> 212     return f(stdscr, *args)
    213
    214 def wrapper(f, *args):

~/opt/anaconda3/lib/python3.7/site-packages/visidata/mainloop.py in cursesMain(_scr, sheetlist)
    234
    235     for vs in sheetlist:
--> 236         vd.push(vs)
    237
    238     vd.status('Ctrl+H opens help')

~/opt/anaconda3/lib/python3.7/site-packages/visidata/sheets.py in push(vd, vs)
    980         vd.allSheets.append(vs)
    981
--> 982     vs.ensureLoaded()
    983
    984

~/opt/anaconda3/lib/python3.7/site-packages/visidata/basesheet.py in ensureLoaded(self)
    187         if self.rows is UNLOADED:
    188             self.rows = []  # prevent auto-reload from running twice
[...]

Issue #801, opened 2020-11-21
Filtering Pandas DataFrame doesn't work in IPython


s-celles:
Hello,

I was trying to follow https://jsvine.github.io/intro-to-visidata/basics/sorting-and-filtering/#how-to-filter-rows
but in an IPython console... (not using `vd` cli)
and noticed that it doesn't work.

For example : Filtering selected rows with "

```
$ ipython
Python 3.7.6 | packaged by conda-forge | (default, Mar 23 2020, 22:45:16)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.13.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import visidata

In [2]: import pandas as pd

In [3]: df = pd.read_csv("https://jsvine.github.io/intro-to-visidata/_downloads/83e70cf67e909f3ac177575439e5f3c5/faa-wildlife-strikes.csv")

In [4]: visidata.view_pandas(df)
```

Press 3 times on "t" key (to select first 3 rows) and " key (to filter using selected rows)

An empty sheet is shown (with only columns headers) instead of a sheet with selected rows.

Kind regards

s-celles:
Similar issue occurs when trying to select rows from a frequency table.

```
visidata.view_pandas(df)
```

Press 3 times on left arrow to go to STATE column

Press Shift + F to open frequency table

Press down arrow to go to CA row and ENTER

An empty sheet is also shown (with only columns headers) instead of a sheet with rows whom STATE is CA.

ajkerrigan:
Thanks for the report @scls19fr, these are good issues to track down. There is some consistent tension between pandas and VisiData, which @saulpw recaps nicely in [this blog post](https://www.visidata.org/blog/2020/row-major/). That can lead to some whack-a-mole finding and fixing operations just for pandas. I'd like to try to address that at a higher level, but that's a big lift and I don't know pandas well enough!

Based on your reports I'm guessing you know pandas better than VisiData's maintainers or I do. If you'd like to try to team up to improve the pandas loader, it could be pretty useful!

PR #816, opened 2020-11-30
Pandas loader fixes/tweaks

ajkerrigan:
This is an initial response to some pandas issues raised recently. It's not the end of the story, just an incremental step to address:

1. Directly loading Python objects with `view_pandas` ( #798 )
2. [workaround] Errors loading transposed DataFrames ( #800 )
3. [workaround] Errors generating a frequency table from a single-column DataFrame ( #802 )

For items 2 and 3 there, it would be nice to brainstorm long-term solutions with someone better-versed in pandas than I am. @scls19fr might be a good candidate as he seems to be actively using the pandas loader and finding these issues!

### DIFF
--- a/visidata/loaders/pandas_freqtbl.py
+++ b/visidata/loaders/pandas_freqtbl.py
@@ -97,7 +97,7 @@ class PandasFreqTableSheet(PivotSheet):
         # that operates similarly to pd.cut.
         super().initCols()
 
-        df = self.source.rows.df
+        df = self.source.df.copy()
 
         # Implementation (special case): for one row, this degenerates
         # to .value_counts(); however this does not order in a stable manner.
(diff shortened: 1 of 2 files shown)


visidata/loaders/pandas_freqtbl.py
```diff
@@ -97,7 +97,7 @@ def reload(self):
         # that operates similarly to pd.cut.
         super().initCols()
 
-        df = self.source.rows.df
```
saulpw, line 100: So, does opening a freqtbl row still work to get back to a sheet with references to the original rows?  As in, modifications on a freq slice will reflect back to the original sheet?

ajkerrigan: Yes, changes on a freq slice still update the original.

ajkerrigan: > So, does opening a freqtbl row still work to get back to a sheet with references to the original rows? As in, modifications on a freq slice will reflect back to the original sheet?

I can't tell yet, with or without these changes I'm getting empty freqtbl slices with pandas (this has already been reported in #801 but I haven't dug into that piece yet).

PR #822, opened 2020-12-06
Added json mime

geekscrapy:
Tested with https://isc.sans.edu/api/cloudcidrs?json

Also noted there is no sample json file in the `sample_data` dir

### DIFF
--- a/visidata/loaders/http.py
+++ b/visidata/loaders/http.py
@@ -1,7 +1,8 @@
 from visidata import Path, RepeatFile, options, vd
 
 content_filetypes = {
-    'tab-separated-values': 'tsv'  # thanks @lindner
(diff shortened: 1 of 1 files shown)


visidata/loaders/http.py
```diff
@@ -1,7 +1,8 @@
 from visidata import Path, RepeatFile, options, vd
 
 content_filetypes = {
-    'tab-separated-values': 'tsv'  # thanks @lindner
+    'tab-separated-values': 'tsv',  # thanks @lindner
+    'json': 'json'
```
saulpw, line 5: I don't think this should be necessary, as it uses the subtype as the filetype if it's not present in content_filetypes (see line 25).

PR #826, opened 2020-12-06
[isNumeric] move isNumeric to vdobj api

anjakefala:

### DIFF
--- a/visidata/_types.py
+++ b/visidata/_types.py
@@ -117,7 +117,8 @@ vdtype(float, '%')
 vdtype(dict, '')
 vdtype(list, '')
 
-def isNumeric(col):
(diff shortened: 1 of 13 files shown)


visidata/_types.py
```diff
@@ -117,7 +117,8 @@ def getType(vd, typetype):
 vdtype(dict, '')
 vdtype(list, '')
 
-def isNumeric(col):
+@VisiData.api
+def isNumeric(self, col):
```
saulpw: s/self/vd

saulpw (changes requested):
add global `isNumeric` to deprecated so any plugins that use it don't break.

saulpw (approved):

PR #833, opened 2020-12-09
bottomRowIndex fixes

### LINKED ISSUE #832: Cursor jumps upwards from final visible row after moving down a row
Commit c9e8f5ed introduced a jarring cursor movement where the cursor hops between the last and second last row when pressing the key to move downwards (moving upwards is fine).

I think it's just an off by one error. Briefly looking myself, the issue is (at least superficially) solved by subtracting 1 from `self.bottomRowIndex` in https://github.com/saulpw/visidata/blob/c9e8f5edc5044293ca4931de3d976ef10d8e2907/visidata/sheets.py#L551

    self.bottomRowIndex = self.cursorRowIndex - 1

However, this is not tested (other than no crashes so far), and I'm not remotely familiar with the codebase, so this may affect something else I'm unaware of.

anjakefala:
[layout] do not calc topRowIndex offset with single-line rows
Offset required for paging with multiline-rows.
Closes #832

[layout-] fix bottomRowIndex getter
topRowIndex is already counted among nScreenRows, so subtract 1

### DIFF
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -331,7 +331,7 @@ class TableSheet(BaseSheet):
             nrows += h
             i += 1
 
-        self._topRowIndex = newidx-i+2
+        self._topRowIndex = newidx-i+2 if nrows == self.nScreenRows-1 else newidx-self.nScreenRows+1
 
     def __deepcopy__(self, memo):
         'same as __copy__'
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Let's give it a go.  Thanks @anjakefala!

PR #839, opened 2020-12-12
[shortcut] allow shortcut for jump-sheet to be settable

anjakefala:


anjakefala:
Summary of feedback:
* It is not clear how a class-level `_shortcut` works, if its value will get shared among instances. 
* If shortcut stuff is all in cmdlog.py, then `_shortcut` should also be instantiated there. (`BaseSheet.init()`).
* Instead of having an awkward setter ina `Column`, we should expand Extensible so that `@BaseSheet.shortcut.setter` works.

saulpw (approved):

PR #847, opened 2020-12-19
Allow macros to be longnames and add vd.isLongname

anjakefala:

### DIFF
--- a/visidata/basesheet.py
+++ b/visidata/basesheet.py
@@ -226,6 +226,11 @@ def sheet(self):
     'the top sheet on the stack'
     return self.sheets[0] if self.sheets else None
 
+@VisiData.api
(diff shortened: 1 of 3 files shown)


docs/api/commands.rst
```diff
@@ -86,6 +86,7 @@ Other helpful functions
 .. autofunction:: visidata.BaseSheet.execCommand
 .. autofunction:: visidata.addGlobals
 .. autofunction:: visidata.getGlobals
+.. autofunction:: visidata.vd.isLongname
```
saulpw: Let's not add to the API by default.

saulpw (approved):
Remove from API, but otherwise if it works, let's merge it.

Issue #830, opened 2020-12-07
How to use and set http auto-paginate?


aborruso:
Hi,
in the 2.1 blog post (great release, thank you to all) there is this new feature

> [http] auto-paginate by continuing with the given links.next.url from the response.

How to use it?

Thank you

saulpw:
Hi @aborruso, it should happen automatically for APIs that support it. Try, for example, `vd https://api.github.com/repos/django/django/pulls`.  See how it loads them in chunks of 10; that's it following the next.url links.

jsvine:
Ah, very interesting — and clever! Two questions about this behavior:

- Is it possible to override it? (I don't see an option in the implementation logic, but perhaps I've overlooked something.) There are times when I want to examine the *exact* data returned by an HTTP GET request, sans magic.

- What do you think about adding an option for limiting the number of pages to fetch? I can imagine some API endpoints may have nearly limitless (or sufficiently large) pages — more than the user might want to load, or than the server might want to handle.

saulpw:
It's not possible to disable or limit it currently, but I can see how that could be useful, especially in batch mode.  We could make an `options.http_max_next`, with a default of `10` or `50` or something reasonable.

In the meantime, in interactive mode, you can also wait until the first N pages pop up, and then cancel the loader manually with Ctrl+C.

jsvine:
Thanks for the response, and sounds good!

PR #848, opened 2020-12-21
[http-loader] Raise requests.exceptions if http status is not 20X

geekscrapy:
Currently if we try to load a file using the http loader there is no visual difference between these two scenarios:

- 1. The file has loaded, but contains zero rows
- 2. The load has finished, but it ended with a http 404 (or equivalent)

`response.raise_for_status()` will raise an exception if a non 20X status is given

To repro try to load `https://xxxx.githubusercontent.com/saulpw/visidata/develop/sample_data/benchmark.csv`. Without the line above, it will not throw an error or indicate to the user a failure has occurred.

### DIFF
--- a/visidata/loaders/http.py
+++ b/visidata/loaders/http.py
@@ -9,6 +9,7 @@ def openurl_http(path, filetype=None):
     import requests
 
     response = requests.get(path.given, stream=True)
+    response.raise_for_status()
 
     # if filetype not given, auto-detect with hacky mime-type parse
     if not filetype:


anjakefala:
Thanks for getting the ball rolling @geekscrapy!

@jsvine actually mentioned previously being interested in seeing the exact response from an API, and wanting to see the data for the 'failed' status code. This is a good place to talk about how that would look. A sheet with a response, and a reference to the loaded file? Would this be behind an option?

geekscrapy:
Could the response object be added as a http sheet property (so through `g^Y` I think?)? Keeps it clean but still accessible to those that want it

saulpw:
@geekscrapy That's an interesting idea, maybe we will go with that.  I'd like for @jsvine to chime in which probably won't be until after the holidays.

jsvine:
Thanks, and happy new year! In my mind, there are (at least) two separate (but related) aspects re. APIs:

- Being able to override VisiData's default behavior of loading/combining all pages of paginate-able API responses — or, more generally, being able to force VisiData to return just the literal HTTP response body. See comment here for details: https://github.com/saulpw/visidata/issues/830#issuecomment-742213710

- Being able to inspect, in a separate sheet, the details of any VisiData-issued HTTP request. (This seems to be what @geekscrapy is suggesting, and I like that suggested approach.)

The first is more important to me, as it relates to data provenance. The second just seems like it would be a neat feature.

anjakefala:
Thanks @geekscrapy and @jsvine for the feedback!

We are going to merge this as-is. 

@saulpw is taking a look at #830. I will add documentation for getting a raw http response (basically, when loading a `.html` use `-f txt` to get the raw response), and we are going to add a way to get the response headers. It is probably going to be a sheet property.

PR #849, opened 2020-12-21
Gracefully fail unknown URL schema

geekscrapy:
Currently if visidata is given a URL handler it doesn't understand, like `abc://` it will fail in the console like the following:
```
  File "bin/vd", line 6, in <module>
    visidata.main.vd_cli()
  File "/Users/geekscrapy/vd_dev/visidata_upstream/visidata/main.py", line 299, in vd_cli
    rc = main_vd()
  File "/Users/geekscrapy/vd_dev/visidata_upstream/visidata/main.py", line 213, in main_vd
    vs = vd.openSource(p, **opts)
  File "/Users/geekscrapy/vd_dev/visidata_upstream/visidata/_open.py", line 78, in openSource
    vs = vd.openPath(Path(p), filetype=filetype)  # convert to Path and recurse
  File "/Users/geekscrapy/vd_dev/visidata_upstream/visidata/_open.py", line 45, in openPath
    return vd.getGlobals()[openfunc](p, filetype=filetype)
KeyError: 'openurl_abc'
```

My suggestion is to catch this KeyError and returns the following instead:
`Error: no loader for url schema: abc`

### DIFF
--- a/visidata/_open.py
+++ b/visidata/_open.py
@@ -42,7 +42,10 @@ def openPath(vd, p, filetype=None):
     'Call ``open_<filetype>(p)`` or ``openurl_<p.scheme>(p, filetype)``.  Return constructed but unloaded sheet of appropriate type.'
     if p.scheme and not p.fp: # isinstance(p, UrlPath):
         openfunc = 'openurl_' + p.scheme
-        return vd.getGlobals()[openfunc](p, filetype=filetype)
+        try:
+            return vd.getGlobals()[openfunc](p, filetype=filetype)
+        except KeyError:
+            vd.fail(f'no loader for url schema: {p.scheme}')
 
     if not filetype:
         if p.is_dir():


anjakefala:
Thank you! This is a useful change.

I have one request:
> vd.fail(f'no loader for url schema: {p.scheme}') 

Could you write the message as `'no loader for url scheme'` instead of `'no loader for url schema'`? And please rebase so the full PR is on a single commit. =)

Once that is done, we can merge it in!

anjakefala (changes requested):
Thank you! This is a useful change.

I have one request:

    vd.fail(f'no loader for url schema: {p.scheme}')

Could you write the message as 'no loader for url scheme' instead of 'no loader for url schema'? And please rebase so the full PR is on a single commit. =)

Once that is done, we can merge it in!

anjakefala:
Commited here with attribution to @geekscrapy, so it could be part of the next release. =) https://github.com/saulpw/visidata/commit/7de88d5dd9720bd17301058da20e0dc629a36213

Issue #727, opened 2020-10-14
[loaders postgres] Transaction error when viewing table


kbd:
```
$ vd --version
saul.pw/VisiData v2.0.1
```

After executing:

```
$ vd postgres://user@localhost:port/database --postgres-schema=schema
```

I get a correct list of the tables, but when I hit enter to view a table VisiData shows no data and I get the following error:

> psycopg2.errors.InFailedSqlTransaction: current transaction is aborted, commands ignored until end of transaction block

![Screen Shot 2020-10-14 at 3 20 38 PM](https://user-images.githubusercontent.com/728060/96046598-fe252b80-0e30-11eb-9ca4-9c468ee0919a.png)

Here's the traceback:

![Screen Shot 2020-10-14 at 3 23 46 PM](https://user-images.githubusercontent.com/728060/96046884-712ea200-0e31-11eb-99f0-1912db813281.png)

I'm able to view the same tables using DBeaver with no problem.

saulpw:
Hi @kbd, thanks for the bug report. I don't have this problem but my Postgres test rig is very simplistic. Have you been able to investigate this any further?

kbd:
@saulpw got the project set up locally but didn’t start digging into the Postgres loader yet. Will continue looking into it tomorrow.

saulpw:
Closing until there is more information or I have a test setup that exhibits the issue.

jdormit:
I've reproduced this. It looks like the Postgres integration isn't correctly accounting for the `--postgres_schema` when executing SQL to populate the sheet for a table. Here's how I reproduced:

1. Open a Postgres database with some tables within a schema via `vd postgres://user:password@host:5432/database --postgres_schema=my_schema`. The list of tables in the schema is correctly populated when VisiData opens.
2. Navigate to any table and enter it via `<return>`. The table will be blank when it opens.
3. View recent errors via `<C-e>`. You'll see this stacktrace:

```
Traceback (most recent call last):
  File "/Users/jdormit/.local/pipx/venvs/visidata/lib/python3.9/site-packages/visidata/threads.py", line 215, in _toplevelTryFunc
    t.status = func(*args, **kwargs)
  File "/Users/jdormit/.local/pipx/venvs/visidata/lib/python3.9/site-packages/visidata/loaders/postgres.py", line 115, in reload
    with self.sql.cur("SELECT * FROM " + self.source) as cur:
[...]

PR #852, opened 2020-12-27
Account for postgres_schema when rendering Postgres tables

jdormit:
This is a fix for https://github.com/saulpw/visidata/issues/727.

### DIFF
--- a/visidata/loaders/postgres.py
+++ b/visidata/loaders/postgres.py
@@ -112,7 +112,11 @@ class PgTablesSheet(Sheet):
 class PgTable(Sheet):
     @asyncthread
     def reload(self):
-        with self.sql.cur("SELECT * FROM " + self.source) as cur:
+        if self.options.postgres_schema:
+            source = f"{self.options.postgres_schema}.{self.source}"
+        else:
+            source = self.source
+        with self.sql.cur(f"SELECT * FROM {source}") as cur:
             self.rows = []
             r = cur.fetchone()
             if r:


visidata/loaders/postgres.py
```diff
@@ -110,9 +110,13 @@ def openRow(self, row):
 
 # rowdef: tuple of values as returned by fetchone()
 class PgTable(Sheet):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.schema = options.postgres_schema
+
     @asyncthread
     def reload(self):
-        with self.sql.cur("SELECT * FROM " + self.source) as cur:
```
saulpw, line 115: Hi @jdormit, thanks for this PR.  Does this still work if `postgres_schema` is an empty string?  (Is it any different to use a default schema rather than "`public`"?)

Also let's use `self.options.postgres_schema` within reload() itself.  That way, the user can change it as a sheet-specific option and then use Ctrl+R to reload.

jdormit: Ah, good point about `postgres_schema` being blank, I hadn't thought about that.

So I updated the PR to reference `self.options.postgres_schema` within the `reload()` method, but when I do that it appears that `self.options.postgres_schema` is always `public` regardless of the `--postgres_schema` option. Any idea what could be causing that? Maybe the `@asyncthread` decorator?

jdormit: I'll push up my updated code, but it doesn't work right now due to ☝️

jdormit: Hmm, so I am probably revealing my limited VisiData knowledge here. But when I run `vd --postgres_schema=my_schema <postgres url>` and look at the global options via `shift-o`, there is no value set for `postgres_schema`. However, when I am looking at the sheet options (`z shift-o`) for the Postgres tables sheet specifically, `postgres_schema` is set to the expected schema. So that would explain why individual table sheets are ending up with the default value. Do you know why `postgres_schema` is getting set as a sheet-specific option instead of a global option?

saulpw: Yes, options set via the CLI are sheet-specific by default, and will only apply to sheets given as CLI args.  Use the `-g` CLI option to set subsequent CLI options as globals. This behavior is intentional, although as it confuses people (sometimes even me!) we may wind up inverting this behavior at some point.

anjakefala:
Thanks so much for the detailed debugging and fix @jdormit!

visidata/loaders/postgres.py
jdormit: Ah understood. Confirmed that passing `-g --postgres_schema=my_schema` works as expected with this patch!

PR #853, opened 2020-12-30
[cmdlog] blank sheet columns mean perform on file loaded upon start

anjakefala:


docs/save-restore.md
```diff
@@ -23,4 +23,29 @@ To save and restore a session in VisiData:
...
+
+4. Pass the desired file through the cmdline when replaying:
+
+~~~
+vd -p foo.vd data.tsv
+~~~
+
+If the **sheet** column in a **CommandLog** is blank, VisiData will refer to the active sheet. Upon startup, that will be the first sheet passed through the commandline. VisiData will load that entire sheet, before replay progresses.
+
+Important, for this to work, all movements between sheets must be explicit within the .vd. This means you need to use `q` or `Shift+S` or `ALT+SheetNumber` to move between sheets.
```
saulpw: Do not recommend Alt+SheetNumber since the specific shortcut is not guaranteed.  Shift+S is also unnecessary; they should use jump-sheet or quit-sheet (or rely on a previous open).

docs/save-restore.md
```diff
@@ -23,4 +23,29 @@ To save and restore a session in VisiData:
...
+
+1. Load the **CommandLog**:
+
+    a. Either open an already saved .vd with `vd foo.vd`.
+
+    *or*
+
+    b. Load the **CommandLog** for the current session with `Shift+D` (commands executed on current sheet only) or `g Shift+D` (commands executed in the entire session).
+
+2. Ensure your cursor is on the **sheet** column. Press `gs` to select all rows, and `gzd` to empty every cell in the sheet column.
```
saulpw: Emphasize `sheet` as it's a specific column name.

anjakefala:
Tensions:
* `q` and `^^` are not recorded on the CommandLog
* jump-sheet takes the name of a sheet, and so does not help with making generally expandable **CommandLog**s

anjakefala:
I do not think this is a change we will feel comfortable merging anytime soon. If this continues to be a feature request, we can revisit this in the future, with fresh eyes.

PR #865, opened 2021-01-06
Allows columns with Wrapped cells to be expanded

geekscrapy:
When the columns is expanded, this will put `None` into the expanded col


saulpw (approved):
I think this makes sense.  Thanks @geekscrapy!

PR #869, opened 2021-01-15
[mysql-loader] asynchronous data fetch for mysql loader

SuRaMoN:
Unless SSCursor (see https://stackoverflow.com/questions/9395793/mysqldb-and-big-queries-using-cursoruseresultmixin)
is used the mysql driver will first fetch all data and only when all data is
fetched it will pass it to python.
This causes you to wait for all data to be transfered before
it is displayed in visidata. With this fix you see data
immediately and the rest of the data loads in the background.

### DIFF
--- a/visidata/loaders/mysql.py
+++ b/visidata/loaders/mysql.py
@@ -1,4 +1,5 @@
 from visidata import *
+from contextlib import contextmanager
 
 __all__ = ['openurl_mysql', 'MyTable', 'MyTablesSheet']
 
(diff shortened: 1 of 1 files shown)


saulpw:
Hi @SuRaMoN, thanks for putting this together.  Do you use this MySQL loader regularly yourself?  I'm curious how it compares with @aswan89's `genericSQL` plugin (not sure if you've tried that too).

SuRaMoN:
had a lot of trouble with it:
- could not get it to work, hacked my way though to get some performance statistics, general imrpession is that its far from production ready
- loading the sheet that shows all tables goes really slow because it counts all rows in each table. counting in mysql goes slower as the table size grows. the native mysql loader looks at estimates from information_schema and the results are instantaneous. sqlalchemy does not do this (and has no support for it afaics). So afaics the "one loader to rule them all" is kinda a myth for this case
- when opening one table its the same problem as before the fix in this pull request (data is only shown when everything is fetched). If i look here: https://stackoverflow.com/questions/39110864/can-sqlalchemy-use-mysqls-sscursor-for-only-some-queries it seems sqlalchemy can do this, but you need to pass driver specific arguments for each database driver it supports. So once more kinda defeats the purpose

My general impression is: with a lot of extra work and driver specifics optimizations it could be a more elegant solution. but in its current state not at all

anjakefala:
Thanks @SuRaMoN! This will be part of the next release.

PR #872, opened 2021-01-18
[types] add floatlocale

anjakefala:


saulpw (approved):
Thanks @anjakefala, nice work.

PR #873, opened 2021-01-20
Misc fixes for the new options

anjakefala:


visidata/cmdlog.py
```diff
@@ -283,18 +283,20 @@ def replayOne(vd, r):
...
+        if r.sheet and longname not in ['set-option', 'unset-option']:
             vs = vd.getSheet(r.sheet) or vd.error('no sheet named %s' % r.sheet)
         else:
             vs = None
 
         if longname in ['set-option', 'unset-option']:
             try:
                 context = vs if r.sheet and vs else vd
+                option_scope = r.sheet or r.col or 'global'
+                option_scope = 'global' if (option_scope == 'override') else option_scope # override is deprecated
```
saulpw: please simplify to `if option_scope == 'override': option_scope = 'global'`

visidata/settings.py
```diff
@@ -34,27 +34,27 @@ def getobj(self, objname):
...
-    def unset(self, k, obj='global'):
+    def unset(self, k, obj='default'):
         'Remove setting for given key in the given context.'
         objstr = self.objname(obj)
         if objstr in self[k]:
             del self[k][objstr]
 
     def set(self, k, v, obj):
-        'obj is a Sheet instance, or a Sheet [sub]class.  obj="override" means override all; obj="global" means last resort.'
+        'obj is a Sheet instance, or a Sheet [sub]class.  obj="global" means override all; obj="default" means last resort.'
```
saulpw: How about: "global" overrides unless there is a more sheet-specific override

saulpw (commented):
Thanks for tightening this up, @anjakefala. I think this is much clearer internally too.

saulpw:
Does anything in the API docs need to be changed for this?  Also does the manpage options section look right (I haven't checked, just noticed that got changed too).

saulpw (approved):
Great, thanks so much @anjakefala!

PR #874, opened 2021-01-24
Fix map display of Shapefiles.

dracos:
Trying to plot a shapefile only gave an `AttributeError: 'NoneType' object has no attribute 'keyCols'`. Sorry if I've misunderstood something (this code looks to have been this way for some time), but self.source did not appear to exist in the ShapeMap class. My commit passes in keyCols through the plot command, so it is available to use, and this appears to work fine, here is e.g. Northern Ireland health boards from a Shapefile after the change in this PR:

### DIFF
--- a/visidata/loaders/shp.py
+++ b/visidata/loaders/shp.py
@@ -52,7 +52,7 @@ class ShapeMap(InvertedCanvas):
 
         for row in Progress(self.sourceRows):
             # color according to key
-            k = tuple(col.getValue(row) for col in self.source.keyCols)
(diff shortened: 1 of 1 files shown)


saulpw (commented):
Ah, I see what happened here.  ShapeMap used to take a `source` positional arg (in this case, `sheet`), but in 2.0 the API has Sheets taking `*names` as the only positional args, and all others being kwargs.  So the smaller/better fix is to just put `source=sheet` in the command execstr.  Thanks for bringing this to my attention, and doing the work to put together a PR.  I'll close this one and make these changes.

PR #875, opened 2021-01-24
Support more Shapefile geometry types.

dracos:
1x are the same as x, but include Z co-ordinates (ignored), and 2x are the same as x but include a user-defined measurement (ignored). Without this PR, I am unable to process a Shapefile I have, but with it (and with #874 included too), here are the Highways England areas:

Hope that's helpful.


saulpw:
Thanks for putting this PR together and extending the types of Shapefiles vd can render!  Looks good.

PR #876, opened 2021-01-24
Add GeoJSON support

dracos:
This PR hopefully adds GeoJSON support, in much the same way as Shapefiles.

I had to add another save-sheet command to the canvas so that you could save from it, that didn't appear to work on Shapefiles previously.

https://github.com/martinjc/UK-GeoJSON/blob/master/json/administrative/wal/lad.json opened with `vd -f geojson lad.json` and selecting the English name column as a key column gives:
image

And pressing `g.` with the English name column highlighted gives:
image

Hope that's useful.


dev/formats.jsonl
```diff
@@ -18,7 +18,7 @@
...
 {"filetype": "dta", "aliases": "", "requirements": "pandas", "format": "Stata", "VisiData loader": "yes", "VisiData saver": "", "version_added": "1.2", "created": "1985", "creator": "StataCorp", "description": "", "open format": "no", "nestable": "", "plottable": "", "format_url": "https://www.loc.gov/preservation/digital/formats/fdd/fdd000471.shtml"}
-{"filetype": "geojson", "filetype_url": "#shp", "aliases": "", "requirements": "", "format": "Geographic JSON", "VisiData loader": " ", "VisiData saver": "from shp", "version_added": "1.2", "created": "2008", "creator": "", "description": "derivative of JSON", "open format": "yes", "nestable": "", "plottable": "plottable", "format_url": "http://geojson.org/"}
+{"filetype": "geojson", "filetype_url": "#shp", "aliases": "", "requirements": "", "format": "Geographic JSON", "VisiData loader": "yes", "VisiData saver": "", "version_added": "1.2", "created": "2008", "creator": "", "description": "derivative of JSON", "open format": "yes", "nestable": "", "plottable": "plottable", "format_url": "http://geojson.org/"}
```
saulpw: Thanks for remembering this!  Let's also change the version_added to 2.2 (which is the next release).

visidata/loaders/geojson.py
```diff
@@ -0,0 +1,130 @@
...
+                features = [{ 'type': 'Feature', 'geometry': ret }]
+
+            for feature in features:
+                for prop in feature.get('properties', {}).keys():
+                    if prop not in self.colnames:
+                        c = Column(name=prop, getter=getter_factory(prop))
+                        self.colnames[prop] = c
+                        self.addColumn(c)
+
+            yield from Progress(features)
```
saulpw: Let's `yield feature` in the first pass instead of doing two passes over `features` (and move the Progress into that `for` loop).

visidata/loaders/geojson.py
```diff
@@ -0,0 +1,130 @@
...
+                features = [ret]
+            elif ret['type'] == 'GeometryCollection':
+                features = list(map(lambda g: { 'type': 'Feature', 'geometry': g }, ret['geometries']))
+            else: # Some form of geometry
+                features = [{ 'type': 'Feature', 'geometry': ret }]
+
+            for feature in features:
+                for prop in feature.get('properties', {}).keys():
+                    if prop not in self.colnames:
+                        c = Column(name=prop, getter=getter_factory(prop))
```
saulpw, line 36: You should be able to use `ItemColumn` with a nested name: `"properties."+prop`.

visidata/loaders/geojson.py
```diff
@@ -0,0 +1,130 @@
...
+class GeoJSONMap(InvertedCanvas):
+    aspectRatio = 1.0
+    filetype = 'geojson'
+
+    @asyncthread
+    def reload(self):
+        self.reset()
+
+        for row in Progress(self.sourceRows):
+            k = tuple(col.getValue(row) for col in self.source.keyCols)
```
saulpw: Use `self.source.rowkey(row)`.

visidata/loaders/geojson.py
```diff
@@ -0,0 +1,130 @@
...
+    return reduce(
+        lambda a,n: [min(a[0],n[0]), min(a[1],n[1]), max(a[2],n[0]), max(a[3],n[1])],
+        coords, initial)
+
+@VisiData.api
+def save_geojson(vd, p, vs):
+    if (isinstance(vs, ShapeMap)):
+        return save_shp_geojson(vd, p, vs)
+
+    isinstance(vs, Canvas) or vd.fail("must save GeoJSON from map sheet")
```
saulpw: This only works for GeoJSONMap (not all Canvas).  With the change in decorators we can remove this.

visidata/loaders/geojson.py
```diff
@@ -0,0 +1,130 @@
...
+        indent = int(vs.options.json_indent)
+    except Exception:
+        indent = vs.options.json_indent
+
+    with p.open_text(mode='w') as fp:
+        encoder = json.JSONEncoder(indent=indent, sort_keys=vs.options.json_sort_keys)
+        for chunk in encoder.iterencode(featcoll):
+            fp.write(chunk)
+
+GeoJSONSheet.addCommand('.', 'plot-row', 'vd.push(GeoJSONMap(name+"_map", sheet, sourceRows=[cursorRow], textCol=cursorCol, source=sheet))', 'plot geospatial vector in current row')
```
saulpw: You can remove the 'sheet' positional arg in these; `source=sheet` is already setting the `source` attribute, and so here `sheet` is seen as the last part of `*names`.

visidata/loaders/shp.py
```diff
@@ -71,8 +73,7 @@ def reload(self):
 
         self.refresh()
 
-@VisiData.api
-def save_geojson(vd, p, vs):
+def save_shp_geojson(vd, p, vs):
```
saulpw: This should actually be `save_geojson` with `@ShapeMap.api` decorator.

visidata/loaders/shp.py
```diff
@@ -83,7 +84,7 @@ def save_geojson(vd, p, vs):
                 'coordinates': [[x, y] for x, y in coords],
             },
             'properties': {
-                col.name: col.getTypedValue(row) for col in vs.source.visibleCols
+                col.name: col.getTypedValue(row) for col in vs.source.keyCols
```
saulpw: This is the one change I'm not sure of...don't we want all the visibleCols as saved properties?  Does some other GeoJSON tool balk at unneeded columns?

visidata/loaders/geojson.py
```diff
@@ -0,0 +1,130 @@
...
+        return bbox
+
+def reduce_coords(coords, initial):
+    return reduce(
+        lambda a,n: [min(a[0],n[0]), min(a[1],n[1]), max(a[2],n[0]), max(a[3],n[1])],
+        coords, initial)
+
+@VisiData.api
+def save_geojson(vd, p, vs):
+    if (isinstance(vs, ShapeMap)):
```
saulpw: This isn't needed if we use the `@GeoJSONMap.api` decorator (instead of @VisiData.api`).  Then we'd also change the original save_geojson (noted below).

visidata/loaders/geojson.py
```diff
@@ -0,0 +1,130 @@
...
+    except Exception:
+        indent = vs.options.json_indent
+
+    with p.open_text(mode='w') as fp:
+        encoder = json.JSONEncoder(indent=indent, sort_keys=vs.options.json_sort_keys)
+        for chunk in encoder.iterencode(featcoll):
+            fp.write(chunk)
+
+GeoJSONSheet.addCommand('.', 'plot-row', 'vd.push(GeoJSONMap(name+"_map", sheet, sourceRows=[cursorRow], textCol=cursorCol, source=sheet))', 'plot geospatial vector in current row')
+GeoJSONSheet.addCommand('g.', 'plot-rows', 'vd.push(GeoJSONMap(name+"_map", sheet, sourceRows=rows, textCol=cursorCol, source=sheet))', 'plot all geospatial vectors in current sheet')
```
saulpw: What do you think about changing `g.` to be `plot-selected` and have them use `someSelectedRows`?

visidata/loaders/geojson.py
```diff
@@ -0,0 +1,130 @@
...
+InvertedCanvas.addCommand('^S', 'save-sheet', 'vd.saveSheets(inputPath("save to: ", value=getDefaultSaveName(sheet)), sheet, confirm_overwrite=options.confirm_overwrite)', 'save current sheet to filename in format determined by extension (default .tsv)')
```
saulpw: Maybe ultimately this should go on BaseSheet, but for now I think this should be on GeoJSONSheet specifically and not all InvertedCanvas.

saulpw (changes requested):
Hi @dracos, thanks for putting this together!  Overall it looks good, just a few changes and questions.

visidata/loaders/geojson.py
dracos: Hmm, `g.` to me implies global-plot. I'd say if anything `.` on its own should plot current row if no selection, or selected rows if some selected.

visidata/loaders/geojson.py
dracos: AttrColumn handles nested properties, but these are normal dicts, and ItemColumn does not handle that.

visidata/loaders/geojson.py
dracos: I'll update the `shp` loader as well, presumably?

visidata/loaders/geojson.py
dracos: Again, will apply to `shp` loader as well

visidata/loaders/shp.py
dracos: Not sure what I thought here, I guess I confused key with visible. Will revert

visidata/loaders/geojson.py
dracos: I made this change (and similar `shp` one), but I now get "no function to save as type geojson" on both maps.

visidata/loaders/geojson.py
dracos: GeoJSONSheet already has a save-sheet on it, do you mean GeoJSONMap?

visidata/loaders/geojson.py
saulpw: You are right!  It really seems like it should, sorry to give you the runaround.  This again tempts me to make it nestable, even at the cost of having it not work with dict keys that have `.` in them.  But that's not for tonight, so we can keep the getter_factory for now.

visidata/loaders/geojson.py
saulpw: Yes, please!  Thanks for helping to keep it clean.

visidata/loaders/geojson.py
saulpw: I see what you mean. To be honest, plotting only the current row has never been very useful, so I wonder if `.` itself should be someSelectedRows, and `g.` should always be "all rows".  That wouldn't be consistent with most of the TableSheet `g` bindings though. Thoughts for another PR :safety_pin:

visidata/loaders/geojson.py
saulpw: Hm, that should have worked.  Unfortunately I'm out of steam for tonight.  I'll take a look at this tomorrow and we'll hopefully get this merged just in time for 2.2!

anjakefala: @dracos Please update your visidata fork, it should work now with this change! https://github.com/saulpw/visidata/commit/5849b88ac00c1261c3c0d09f0b9e4c479c2ecd08

dracos: Thanks! Yes, that seems to have worked.

saulpw (approved):
Thanks again @dracos!

aborruso:
@dracos thank you very much!

How view a geojson in vd?

anjakefala:
@aborruso Update your installed VisiData to the most recent develop! There will be a release sometime this week. =)

aborruso:
I know it and I have already updated it. 

When I open a geojson I see this. Do I must use only  point layer?

Thank you

image

dracos:
How did you open it, and what is its filename? Its filename needs to end with `.geojson` or you open it with `-f geojson` as in my description in the PR body.

aborruso:
Hi @dracos thank you again. I have opened it using `vd aree.geojson`.

And it's the same if I use `-f geojson`.

I have `saul.pw/VisiData v2.2dev`

anjakefala:
@aborruso I think you are likely referencing an earlier commit of VisiData 2.2dev in your environment. =(

aborruso:
@anjakefala  I'm installing it using `pip3 install --no-warn-script-location git+https://github.com/saulpw/visidata.git@develop`

anjakefala:
@aborruso I would try to completely uninstall all versions of VisiData (`pip3 uninstall visidata`), and use a search to hunt for any stray files in your computer, and then install. If you are still having trouble after doing that, I will take a look!

aborruso:
I have it, I have added `--upgrade`

Now that I have a table, how to render the geoms?

Thank you

image

aborruso:
Using `g .`, it's great!!!

dracos:
As per the instructions I gave in the PR body, (`.` for a single row, or) `g.` for all rows (though see discussion above about how that may change in future). Glad to hear it.

anjakefala:
Yay @aborruso :purple_heart:!! I am so glad.

Issue #878, opened 2021-01-26
[bug] Problems with filtering in pandas


anjakefala:
> The Dataframe displays fine, but when I select rows and try to make a new sheet with them (with "), nothing shows in the new sheet.  Similarly, when I make a Frequency Table and then press enter to drill down into one of the items, nothing shows.

anjakefala:
The basic problems outlined in the report have been fixed here: https://github.com/saulpw/visidata/issues/878.

Pandas is fairly different from the standard loader, and it is going to require hand-writing lots of commands.

PR #881, opened 2021-01-28
[pandas] fix empty subsheets for dup-selected and freqtbl

ajkerrigan:
Opening subsets of a pandas sheet requires different logic than other
typical sheets, because of its dataframe source and custom reload logic.
Tweak the freqtbl behavior and override `dup-selected` so they work as
expected.

Team debugging session w/ @anjakefala woohoo, thanks!

Addresses #878


anjakefala (approved):
Thanks @ajkerrigan! =)

Issue #200, opened 2018-10-29
Detect utf8/utf16 Byte-Order-Marker and set encoding


saulpw:
In `Path.open_text`, autodetect UTF-16 Byte-Order Marker (BOM) and set encoding on file accordingly.

anjakefala:
This issue has been added to our [someday/maybe list](https://github.com/saulpw/visidata/blob/develop/dev/ROADMAP-someday.md). If anyone wants to take it on, they are welcome to reopen the issue.

daviewales:
The someday/maybe link above is broken. Is it perhaps
https://github.com/saulpw/visidata/blob/stable/dev/WISHLIST.md
?

anjakefala:
@daviewales, instead of writing things to the list, we changed strategies to just filtering by the wishlist label! Fewer steps. =)

saulpw:
Now that we have a separate `options.save_encoding`, I changed the default for `options.encoding` to `utf-8-sig` which should detect the BOM and skip it.

Shahin-rmz:
Hi, I am not sure if my issue is here in right place.
an old user of Visidata here.
I am trying to copy some rows in German to clipboard and I get this error.

```
copying 1 rows to system clipboard as tsv                             │
saving 1 sheets to /tmp/tmpo2rldvpn.tsv as tsv                        │
UnicodeEncodeError: 'utf-8' codec can't encode character '\udcf6' in  │
position 294: surrogates not allowed                                  │

```

As far as I see, there is no problem in opening the file, how ever the ö,ä,ü,ß characters can not be shown, which is ok. but I can not copy and paste to the clipboard because of this error.

daviewales:
@Shahin-rmz, you might not get much visibility on a closed issue. Perhaps post a reproducible example as a new issue?

Ideally, attach a file which exhibits the issue.

PR #909, opened 2021-02-22
Automatically handle UTF-8 files with BOM

### LINKED ISSUE #908: JSON Loader doesn't handle UTF-8 BOM
**Small description**
I tried to load a JSON dataset directly from the url as follows

    vd https://data.nsw.gov.au/data/dataset/0a52e6c1-bc0b-48af-8b45-d791a6d8e289/resource/f3a28eed-8c2a-437b-8ac1-2dab3cf760f9/download/covid-case-locations-20210217.json

**Expected result**
I expected it to load as a table.

I get the following error:

``` python
  File "/Users/dwales/.local/pipx/venvs/visidata/lib/python3.9/site-packages/visidata/loaders/json.py", line 31, in iterload
    ret = json.loads(L)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/__init__.py", line 335, in loads
    raise JSONDecodeError("Unexpected UTF-8 BOM (decode using utf-8-sig)",
json.decoder.JSONDecodeError: Unexpected UTF-8 BOM (decode using utf-8-sig): line 1 column 1 (char 0)

During handling of the above exception, another exception occurred:

  File "/Users/dwales/.local/pipx/venvs/visidata/lib/python3.9/site-packages/visidata/threads.py", line 215, in _toplevelTryFunc
  File "/Users/dwales/.local/pipx/venvs/visidata/lib/python3.9/site-packages/visidata/sheets.py", line 284, in reload
    for r in self.iterload():
  File "/Users/dwales/.local/pipx/venvs/visidata/lib/python3.9/site-packages/visidata/loaders/json.py", line 43, in iterload
    ret = json.load(fp)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/__init__.py", line 293, in load
    return loads(fp.read(),
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/__init__.py", line 335, in loads
    raise JSONDecodeError("Unexpected UTF-8 BOM (decode using utf-8-sig)",
json.decoder.JSONDecodeError: Unexpected UTF-8 BOM (decode using utf-8-sig): line 1 column 1 (char 0)
```

`errors.vd`:

```
			open-file	https://data.nsw.gov.au/data/dataset/0a52e6c1-bc0b-48af-8b45-d791a6d8e289/resource/f3a28eed-8c2a-437b-8ac1-2dab3cf760f9/download/covid-case-locations-20210217.json	o	

```

**Additional context**
Please include the version of VisiData.

saul.pw/VisiData v2.2.1

Note: I'm aware that including a BOM in a JSON file is technically invalid. However, I feel that it would be nice if `visidata` automatically removed the BOM before trying to parse the JSON (or perhaps caught the error, and retried after removing the BOM).

daviewales:
Fixes #908.

Replaces the default encoding `utf-8` with the encoding `utf-8-sig`, which automatically detects and removes the UTF-8 BOM, but also works fine with regular UTF-8 files.

### DIFF
--- a/visidata/path.py
+++ b/visidata/path.py
@@ -7,7 +7,7 @@ from functools import wraps
 
 from visidata import *
 
-option('encoding', 'utf-8', 'encoding passed to codecs.open', replay=True)
+option('encoding', 'utf-8-sig', 'encoding passed to codecs.open', replay=True)
 option('encoding_errors', 'surrogateescape', 'encoding_errors passed to codecs.open', replay=True)
 
 @functools.lru_cache()


daviewales:
Incidentally, is this where you would put the logic to detect and set the encoding for #200?

saulpw:
Hi @daviewales, thanks for the suggestion.  I did some research into it and it seems that utf-8-sig always adds a BOM when writing to a file (even when appending) which is a problem since VisiData uses options.encoding when opening files in general. You can of course set this in your own .visidatarc if you don't write data out in any text formats.

saulpw:
and yes, it would probably go in visidata.Path.read_text(), but that seems like it would be part of a larger project to detect the file format in general.

saulpw:
Now that we have a separate `options.save_encoding`, I changed the default for `options.encoding` to `utf-8-sig` which should detect the BOM and skip it. Thanks for the suggestion, @daviewales!

daviewales:
Thanks @saulpw!

PR #919, opened 2021-03-08
Undo commit

anjakefala:
The issue was that when a user `commit-sheet` with sqlite, it came with a massive undo that would erase the contents of each table. That undo was bc of the SqliteSheet.reload()

There are various calls in the reload (`addColumn` and `setKeys` were two that I found), that set those undos.
With my patch, the only undo will be undo-ing deletes (because they do not actually delete until the user commits).

What purpose was the reload serving? Basically, defer keeps some info dicts, and does not actually put the change to the sheet, until you run the `commitBlah()` functions. 
So if you do not have those or a reload, most of your changes wil disappear after you `commit-sheet`. They will be committed to the database, but you will lose them in your VisiData sheet. This is not ideal.
The reload then pulls in the fresh sheet with the changes present.

Those `commitMods()` I replaced it with basically do the work of "putting" and "deleteby"ing the changes to the sheet and they do that without adding extraneous undos and rebuilding the full thing, and so i think they can replace the reload.

A big change is, if something failed to commit to the database, it will now remain in the sheet, even if it was not committed as opposed to just disappearing. You can't quite tell if your changes 'took'.
However, this change allows us to reverse a `commit-sheet`, and I think that is better.

### DIFF
--- a/visidata/modify.py
+++ b/visidata/modify.py
@@ -68,7 +68,8 @@ def rowDeleted(self, row):
     'Mark row as a deferred delete-row'
     self._deferredDels[self.rowid(row)] = row
     def _undoRowDeleted(sheet, row):
-        del sheet._deferredDels[sheet.rowid(row)]
+        if sheet.rowid(row) in sheet._deferredDels:
+            del sheet._deferredDels[sheet.rowid(row)]
     vd.addUndo(_undoRowDeleted, self, row)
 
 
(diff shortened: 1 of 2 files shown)


visidata/modify.py
```diff
@@ -123,7 +125,7 @@ def deleteBy(sheet, func, commit=False):
             sheet.deleteSourceRow(r)
             ndeleted += 1
 
-    if not commit:
+    if undo:
```
anjakefala, line 128: This change is because we want committed deletes to have an undo, at the time of `commit-sheet`. We do not want them to undo before they are actually removed from the sheet.

This existed in the first place because sometimes I want to run deleteBy without setting an undo for it. As an example, if I am using it to undo added rows, then I want to be able to call it without it setting an undo for the undo-ing of those added rows. >_>

anjakefala: So commit=True means "remove these rows from the sheet, do not just colour them". Undo=True means, set an undo for the removal of these rows for the current command.

anjakefala:
We are instead going to do 135aad6e1e20d5e485d9f4950ac7c71784f27330, and not set undos for commit-sheets.

PR #922, opened 2021-03-10
Fix Typo in Deciles Description

cwarden:


saulpw (approved):
Oops!  Thanks for the fix, @cwarden.

PR #930, opened 2021-03-23
fix link to docs on loaders

zachvalenta:


anjakefala:
Thanks for the PR @zachvalenta! And for noticing that problem.

Would you mind changing the link to `/docs/api/loaders`? We support both `visidata.org` and `www.visidata.org`, and I do not want the link to result in a flip flop.

zachvalenta:
sounds good, just updated :)

anjakefala:
Thank you!! =)

Issue #936, opened 2021-03-30
visidata.view_pandas fails when the df index's name is already a column in the DataFrame


thomanq:
**Small description**
```
import pandas as pd
df = pd.DataFrame({"uid": [6,5,4,7,8,9], "a": [29, 0, 3, 34, 6, 9]})
df = df.set_index("uid", drop=False)
import visidata;visidata.view_pandas(df)
```
**Obtained result**
```
ValueError: cannot insert uid, already exists
```
Where it fails:
```
# visidata/loaders/_pandas.py", line 128
        # reset the index here
        if type(df.index) is not pd.RangeIndex:
            df = df.reset_index()
```

**Additional context**
Please include the version of VisiData.
```
pandas                   1.2.1
visidata                 2.2.1
```

thomanq:
PR accepted and merged.

PR #937, opened 2021-03-30
Fix conflict between dropped index and existing column

thomanq:
Fixes issue https://github.com/saulpw/visidata/issues/936


ajkerrigan (approved):
This looks good to me, good catch and thanks for the fix!

PR #951, opened 2021-04-13
[plugins darkdraw] Adds ddw entrypoint script

ryanprior:
This PR adds an entrypoint script for darkdraw, `bin/ddw`. This is almost identical to `bin/vd` but it explicitly imports `darkdraw` and sets the filetype to `ddw`.


saulpw:
Let's add this to the [darkdraw repo](https://github.com/devottys/darkdraw) in a new bin/ folder.  We'll move the Dockerfile there too.

ryanprior:
Makes sense, will do.

PR #957, opened 2021-04-19
[docs] fix bug in vdrc.py on customize page

### LINKED ISSUE #956: docs/customize vdrc.py dict bug
**when running vdrc.py on docs/customize I get the following bug**

```
  File "./vdrc.py", line 5, in <module>
    for optname, val in sorted(options().items()):
TypeError: 'OptionsObject' object is not callable
```

njthomas:
This should fix #956

### DIFF
--- a/docs/customize.md
+++ b/docs/customize.md
@@ -61,7 +61,7 @@ This Python script can be used to generate a sample `~/.visidatarc`, with all of
 ~~~
 from visidata import options
 
-for optname, val in sorted(options().items()):
+for optname, val in sorted(dict(options).items()):
     val = repr(val)
     helpstr = options._get(optname, 'global').helpstr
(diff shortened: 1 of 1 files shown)


docs/customize.md
```diff
@@ -61,7 +61,7 @@ This Python script can be used to generate a sample `~/.visidatarc`, with all of
 ~~~
 from visidata import options
 
-for optname, val in sorted(options().items()):
+for optname, val in sorted(dict(options).items()):
```
saulpw, line 64: This is a clever way of doing it, but I think it should be `sorted(options.get_all().items())`.

njthomas: Yup, your way is better. I was just quickly hacking to make it work.

njthomas:
I'll redo this change.

PR #967, opened 2021-04-27
add postgresql scheme

### LINKED ISSUE #966: no loader for url scheme: postgresql
**Small description**

When attempting to start visidata with
```
vd postgresql:///localdb
```
it fails with
```
Error: no loader for url scheme: postgresql
```

**Expected result**

I would have expected it to work the same way 
```
vd postgres:///localdb
```
works, as [both URL schemes are valid](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING).

P.S.: This is somewhere in between a bug and a feature request. As it's super small and about something existing, I called it a bug. I will raise a PR to fix this shortly.

zormit:
This PR adds the `postgresql://` URL scheme as an alias to `postgres://`

### DIFF
--- a/visidata/loaders/postgres.py
+++ b/visidata/loaders/postgres.py
@@ -1,6 +1,6 @@
 from visidata import *
 
-__all__ = ['openurl_postgres', 'openurl_rds', 'PgTable', 'PgTablesSheet']
+__all__ = ['openurl_postgres', 'openurl_postgresql', 'openurl_rds', 'PgTable', 'PgTablesSheet']
 
 option('postgres_schema', 'public', 'The desired schema for the Postgres database')
 
@@ -50,6 +50,10 @@ def openurl_postgres(url, filetype=None):
                 password=url.password)
 
     return PgTablesSheet(dbname+"_tables", sql=SQL(conn))
+
+
+def openurl_postgresql(url, filetype):
+    return openurl_postgres(url, filetype)
 
 
 class SQL:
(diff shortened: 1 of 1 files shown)


zormit:
note: This is a PR "flying blindly" -- I didn't test this. Still have to work on my dev setup ;)

visidata/loaders/postgres.py
```diff
@@ -52,6 +52,10 @@ def openurl_postgres(url, filetype=None):
     return PgTablesSheet(dbname+"_tables", sql=SQL(conn))
 
 
+def openurl_postgresql(url, filetype):
```
saulpw: Thanks for putting this together, @zormit !  Let's do `openurl_postgresql=openurl_postgres` and save a line of code :)

zormit: haha of course

visidata/loaders/postgres.py
```diff
@@ -52,6 +52,10 @@ def openurl_postgres(url, filetype=None):
...
+    return openurl_postgres(url, filetype)
```
zormit:
```suggestion
openurl_postgresql=openurl_postgres
```

zormit:
@saulpw do you care about a super clean git history? Then I could also squash the two commits again. also i forgot to add a `[tag]` in front of the 2nd commit, whoopsie.

saulpw (approved):
Looks good, @zormit !   Thanks for putting this together.  I'll squash/merge it now.

PR #974, opened 2021-05-07
[loader] Fix editing in final column for fixed-width loader

mwayne:
Returning `None` raises a `TypeError` in `putValue()` when editing cells in the final column.


saulpw (approved):
Hey @mwayne, thanks for fixing this bug!

PR #981, opened 2021-05-16
Replace vls with vsh.

frosencrantz:
Installation is broken otherwise.  This doesn't fix other errors once the tools are install, missing parseArgs.


saulpw:
Thanks, @frosencrantz.  I decided to move `vsh` and other VisiData adjuncts to my new `vdplus` "mono-repo", and have fixed this issue there (by disposing of `vsh`, in fact; it's nothing more than a germ of an idea that didn't become anything).

Issue #975, opened 2021-05-08
Not able to select comparing dates


aborruso:
**Small description**

Hi,
I have this input file

```
id,data
1,2020-12-05
2,2020-07-08
3,2024-12-05
```

I would like to select all rows greater than 2020-08-08, using Python expression.
I set `data` field as date field using `@`, and then `z|`.

My expression is `data > datetime.date(2020,8,8)`.

And I have 0 rows selected

![image](https://user-images.githubusercontent.com/30607/117530546-049e0f00-afde-11eb-9786-a98e5d1dd9cd.png)

**Expected result**

To select 2 rows.

**Additional context**

saul.pw/VisiData v2.5dev

frosencrantz:
I see several issues:
1. `z|` select-expr doesn't show exceptions [code].  It just hides them.
2. If it didn't, you would see an error something like `TypeError: can't compare date to datetime.date`.  Which is confusing, but you would see that the date class is a VisiData class, that subclasses from datetime.datetime.  And Python doesn't allow that comparison.

Some possible things to change would be below. These might not be the right things to do.
1. Don't ignore exceptions from select-expr (e.g. add a vd.exceptionCaught)
2. Rename the Visidata class date to make it clearer it is basically a datetime.
3. Change the Visidata date class to also allow comparisons with datetime.date.

@aborruso,  you could just use `data > datetime.datetime(2020, 8, 8)`


[code]: https://github.com/saulpw/visidata/blob/develop/visidata/selection.py#L88

aborruso:
> @aborruso, you could just use `data > datetime.datetime(2020, 8, 8)`

Great, it works, thank you very much!

My real goal was to select date in the future, and now I know that I can write ` data > datetime.datetime.now()`

saulpw:
Thanks @aborruso for filing this issue, and @frosencrantz for the quick workaround.  I've added specialized comparators for this situation (honestly I don't know why Python datetime.datetime itself doesn't do this automatically).  Note that `visidata.date` is a bit of a misnomer, as @frosencrantz says; it always includes a time as well.  Ideally it would correctly compare partially-specified dates and/or times, such that e.g. `visidata.date("2016-10") == visidata.date("2016-10-29")`, but that is a feature request outside the scope of this bug.

aborruso:
Thank you very much

PR #982, opened 2021-05-16
gatherBy() reports exceptions

frosencrantz:
Copying change from https://github.com/saulpw/visidata/commit/ad02b59b486ba2f2d4cf51c032d760a0ad661666

Could have helped with https://github.com/saulpw/visidata/issues/975

- [ ] If contributing a core loader, [the loader checklist](https://visidata.org/docs/contributing-to-visidata#loader) was referenced.
- [ ] If registering an external plugin, [the plugin checklist](https://visidata.org/docs/contributing-to-visidata#plugins) was referenced.

### DIFF
--- a/visidata/selection.py
+++ b/visidata/selection.py
@@ -84,8 +84,8 @@ def gatherBy(self, func, gerund='gathering'):
             r = self.rows[i]
             if func(r):
                 yield r
-        except Exception:
-            pass
+        except Exception as e:
+            vd.exceptionCaught(e)
 
 @Sheet.property
 def selectedRows(self):


saulpw (commented):
Hi @frosencrantz, did this help you debug something you were working on?  Generally I start with these except clauses calling `exceptionCaught`, and then remove them as they clutter the status line with even 'normal' errors.  So maybe there should be a `debug=` kwarg on exceptionCaught, which only displays a status message when options.debug is True, and/or a "silent" status message type which is not reported on the status line but does show up on the previous statuses sheet.

Again, I'm not opposed to the content of the change, but want to have a general solution to this class of problem.

frosencrantz:
Hi @saulpw , I used it to debug https://github.com/saulpw/visidata/issues/975#issuecomment-835900045

I like your idea of providing a way to control the noise level of errors.  I do like seeing the most recent errors in the status message,  otherwise there would be issues I completely miss.

saulpw:
Okay, with 6a56ee5, errors in this context will be shown on the status history sheet, but not on the status line.  Thanks for the impetus, @frosencrantz !

frosencrantz:
Brilliant, @saulpw!  Thank you!  I think that would have made it easier to debug.

PR #1010, opened 2021-06-15
path/RepeatFile: implement line-seek operations

### LINKED ISSUE #1007: vds: can't load a .vds file from stdin
**Small description**

`cat test.vds | vd --filetype vds` results in: `AttributeError: 'RepeatFile' object has no attribute 'readline'`.

It should instead just work(tm).

**Additional context**

Problem is present on v2.4 and on the develop branch (commit 3350d9fd8c9e64ebf409deae4b31085d12efeb7f)

pacien:
See commit and their messages for more details.
This fixes #1007.

### DIFF
--- a/visidata/loaders/http.py
+++ b/visidata/loaders/http.py
@@ -62,7 +62,7 @@ def openurl_http(path, filetype=None):
             response = requests.get(src, stream=True)
 
     # add resettable iterator over contents as an already-open fp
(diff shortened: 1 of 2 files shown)


saulpw (approved):
Thanks @pacien!  This intersects with another issue I've been looking at, so there may be some more churn in RepeatFile, but this looks better overall so we'll merge this proactively and deal with issues/regressions if/as they arise.

PR #1020, opened 2021-06-20
Add `view_sf()` loader of StaticFrame containers

flexatone:
Greetings, and thank you for this excellent tool.

I am the primary author of StaticFrame, a Pandas-like DataFrame library that employs an immutable data model and defines a variety of multi-DataFrame containers (https://github.com/InvestmentSystems/static-frame).

We use StaticFrame widely in our organization; having integration with VisiData would accelerate our workflows. Our most common use case is in debugging or interactive contexts where we have StaticFrame containers in-memory and want to explore their contents. For this VisiData provides an excellent resource. 

For this reason, this PR adds a `view_sf()` function that handles all StaticFrame containers. If accepted, I would then add `to_visidata()` exporter methods to all StaticFrame containers, permitting opening a VisiData session with just one method call. I look forward to being able to promote VisiData through these new interfaces on StaticFrame.

The implementation is based on `view_pandas()`, but is extended and refined in a number of ways. A few are isolated here:

- As StaticFrame uses an immutable data model, numerous methods were updated to not use in-place mutation of internal data.
- As StaticFrame only uses NumPy types, type mapping is simpler.
- As all StaticFrame containers have `name` attributes, those names when set are always passed on to the Sheet.
- StaticFrame introduces novel multi-DataFrame containers, called the Bus and Batch. These naturally correspond to IndexSheet objects, and for that reason I have implemented a derived IndexSheet that supports those containers.
- I have added a "random-columns" command, similar to the "random-rows" command.

I consider this PR a draft and look forward to feedback and suggestions; I am happy to make refinements and changes as needed. I have a few initial questions:

1. I have reviewed the "loader checklist" but I am unsure if (a) I need to define a `rowtype` string or `# rowdef` comment and (b) if adding to `sample_data` related file-based tests is appropriate for this sort of in-memory loader.
2. I implemented this off of the "stable" branch; if that is a problem I can recast the PR off the "develop" branch.
2. I can add other tests as appropriate, but I am unsure how to test an in-memory loader such as this. Any direction would be appreciated.
3. I can change any formatting or extend doc strings, comments, or type hints as needed.
4. I have manually tested nearly all commands and believe to have good compatibility. I have identified two commands that do not work yet, and would benefit from some assistance.
   a. `^`: Column renaming is not yet working. This also does not work in `view_pandas()`.
   b. `&`: While some join types work, inner, outer, diff, extend, and merge do not. I have some idea of what is going wrong but could use some assistance.

Here is a small example of usage for ad-hoc testing. I can provide more extended examples as needed.

```python
>>> import static_frame as sf
>>> import visidata as vd
>>> f1 = sf.Frame.from_records([[True, 1.2, 'red'], [False, 3.5, 'blue']], columns=('x', 'y', 'z'), name='A')
>>> f2 = sf.Frame.from_records([[True, 1.2, 'red'], [True, -20, 'green']], columns=('x', 'y', 'z'), name='B')
>>> bus = sf.Bus.from_frames((f1, f2))
>>> vd.view_sf(bus)
```
Thank you for your consideration.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,6 +25,6 @@ tabula              # pdf tables
 vobject             # vcf
 tabulate            # tabulate saver
 wcwidth             # tabulate saver with unicode
-static-frame>=0.8.13
+static-frame>=0.8.14
 
 -e git+https://github.com/saulpw/pyxlsb.git@master#egg=pyxlsb # xlsb
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,5 +25,6 @@ tabula              # pdf tables
 vobject             # vcf
 tabulate            # tabulate saver
 wcwidth             # tabulate saver with unicode
+static-frame>=0.8.13
 
 -e git+https://github.com/saulpw/pyxlsb.git@master#egg=pyxlsb # xlsb
--- a/visidata/__init__.py
+++ b/visidata/__init__.py
@@ -122,6 +122,7 @@ from .loaders.spss import *
 from .loaders.xml import *
 from .loaders.yaml import *
 from .loaders._pandas import *
+from .loaders.sf import *
 from .loaders.graphviz import *
 from .loaders.npy import *
 from .loaders.usv import *
--- a/visidata/loaders/sf.py
+++ b/visidata/loaders/sf.py
@@ -5,8 +5,6 @@ from visidata import *
 # Commands known to not work
 # ^ Rename current column (and related commands; fails with view_pandas)
 
-# Unsure
-# cannot have the name of the Frame set as the name of the sheet.
 
 class StaticFrameAdapter:
 
@@ -77,9 +75,11 @@ class StaticFrameSheet(Sheet):
     @frame.setter
     def frame(self, val):
         if isinstance(getattr(self, 'rows', None), StaticFrameAdapter):
+            # If we already have a rows attribute and it is a SFA, then we assume val us a Frame and inject it intot the SFA
             self.rows.frame = val
         else:
             self.rows = StaticFrameAdapter(val)
+        self.name = '' if val.name is None else val.name
 
     def getValue(self, col, row):
         '''Look up column values in the underlying Frame.'''
@@ -123,7 +123,7 @@ class StaticFrameSheet(Sheet):
                 vd.fail('error building Frame from source data: %s' % err)
 
         # If the index is not an IndexAutoFactory, try to move it onto the Frame. If this fails it might mean we are trying to unset an auto index post selection
-        if frame.index._map: # if it is not an IndexAutoFactory
+        if frame.index.depth > 1 or frame.index._map: # if it is not an IndexAutoFactory
             frame = frame.unset_index()
 
         # VisiData assumes string column names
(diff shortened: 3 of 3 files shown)


saulpw:
Hey @flexatone, this is really cool!  I've not heard of StaticFrame before but it seems really useful.

I think what you've got here would be a perfect candidate for a plugin.  You did a nice job keeping it self-contained, following VisiData patterns, and thinking through some of the trickier nuances, so it should be pretty easy to take what you've got and pluginify it.
This will make it easier for you to update and maintain the plugin; in particular you won't be tied to the VisiData release "schedule" (which is pretty erratic), and so you can update it whenever you want and with a lot less friction (a PR to update a line in our plugins.jsonl, which we release independently of VisiData itself).

I'd love to help you with this, in particular so we can update/organize our plugin docs so this process is easier for everyone.

Here's some answers to your questions:

1. rowtype: For a "generic" loader like this (similar to pandas), the rowtype string is probably not necessary; inheriting from Sheet will default to 'rows', and unless your rows are special somehow, this is easiest.  A rowdef comment describing the basic type/structure of a row would still be helpful though.
2. PR branch: we take PRs from the develop branch only, but as a plugin, you'd want to target a released version (and the stable branch is always the most recently released version).
3. tests: we actually haven't designed how to structure plugin tests, but that's a great idea.  I imagine you would include an `vd.open_sf` function and then have a .vd that sets the filetype option to `sf` before opening a StaticFrame supported format, and then testing various commands from there and making sure the data comes out intact.
4. Nice find with pandas `^`; I didn't know that wasn't working.  That's worthy of filing an issue.
5. I'm happy to help you get those join types working.  This one will go much faster if we can have a sync conversation.  Come to our chat sometime (libera.chat #visidata) and we can either talk it through there, and/or set up a pair coding session.

flexatone:
Hi @saulpw, many thanks for your consideration, comments, and answers to my questions.

I am certainly open to implementing this as a plug-in. However, as I understand the plug-in architecture in VisiData, enabling the plug-in cannot be done through a pip install, but instead requires either configuration on the local file system or UI interactions in VisiData. My preferred workflow for StaticFrame users is to be able to simply pip-install VisiData (and any other packages) and have the StaticFrame interaction work without further configuration: I assume this is not possible with VisiData plugins, but please correct me if I am wrong.

I suppose one option is that I could implement the derived `Sheet` and `IndexSheet` classes in StaticFrame itself. I would have to do some slightly awkward things to make importing conditional (as VisiData will be an optional dependency), but it should work. So if you prefer this as a plugin rather than as part of core VisiData, I might try the implementation in StaticFrame first.

That said, I am happy to take you up your offer for help on (4) and (5), above. I am pretty sure that once I see how to solve (4) for StaticFrame, I will know how to solve it for Pandas and I will happily submit a PR for that piece. A pair session would likely be most efficient; I can contact you directly to find a time.

Thanks again for your feedback.

saulpw:
@flexatone I think we might be able to get you the experience you want for your users.  You can either email me or schedule office hours at https://calendly.com/saulpw/visidata-office-hours.  I'd love to help you on this whenever you have time.

flexatone:
Thanks, @saulpw, for your response and offer. I will contact you shortly.

PR #1070, opened 2021-08-08
The inlined GIF is no longer present on visidata.org, embed it into the repository

VxJasonxV:
Repository only change, no effect to the application itself. You can see that referencing a root directory path works in my fork https://github.com/VxJasonxV/visidata#readme


anjakefala:
Thanks so much for bringing this to our attention @VxJasonxV! We had updated the website recently, and there are a few kinks. We ended up just re-adding that file on the website's end, so now that gif is showing. =)

Thanks!

PR #1078, opened 2021-08-17
Added support to use visidata as mysql pager

SuRaMoN:
Loader to use visidata in mysql (or mycli) client. Use the following config to enable
visidata by default in the mysql client:
Edit ~/.my.cnf:
[mysql]
silent
pager=vd -f mysqlclient

- [X] If registering an external plugin, [the plugin checklist](https://visidata.org/docs/contributing#plugins) was referenced.


saulpw (commented):
Hi @SuRaMoN, thanks for submitting this.  I thought about it a bit, and I think this might be better as a plugin.  You do have a few changes to core visidata included but I'm not sure if these are essential.  Can you go into a little bit more depth about why you chose to include these?  Feel free to come on #visidata and chat with me about it too.

SuRaMoN:
When using the mysql client, the output looks like this:
image

For large results this is quite unpleasant to work with.
When configuring visidata as mysql pager the result of the query will be shown in visidata.
Mysql has a builtin option to output data as TSV, so this loader could be used to parse the data from the client.

But the output is slightly different from a normal TSV:
 - a newline (char(10)) is sent as the string \n (so one backslash and the character "n").
 - a backslask is sent as the string \\ (so 2 backslashes).
 - a CR (char(13)) is sent literally, but interpreted by visidata as a row separator as universal newlines are by default enabled when opening files in text mode. this is why i added binary support to the Path class
 - a tab is set as "t" (so one backslash and the character "t")
 
To view the real data (and not encoded data) i parse the data as a TSV in the newly written loader (with only \n as line separator, not \r) and i replace "\\t" -> "\t", "\\n" -> "\n", "\\\\" -> "\\"

If you want i can split up this revision in 2. one for the core changes in the Path class, and one for the loader as an extension.
I would personally not go for a separate extension as the mysql client (or mycli client) is used pretty commonly.
A quick internet search also shows other people having the same problem: https://news.ycombinator.com/item?id=24774947

image

SuRaMoN:
rebased to latest develop branch

visidata/main.py
```diff
@@ -56,7 +56,7 @@ def duptty():
     try:
         fin = open('/dev/tty')
         fout = open('/dev/tty', mode='w')
-        stdin = open(os.dup(0), encoding=vd.options.getonly('encoding', 'global', 'utf-8'))
+        stdin = open(os.dup(0), mode='rb')
```
saulpw: We can't make stdin binary or people won't be able to use text loaders.

visidata/loaders/mysqlclient.py
```diff
@@ -0,0 +1,49 @@
...
+        *rows, buf = buf.split("\n")
+        yield from rows
+
+    yield from buf.split("\n")
+
+
+# rowdef: list
+class MysqlClientSheet(SequenceSheet):
+    def iterload(self):
+        with io.TextIOWrapper(self.source.open("rb"), errors="replace", newline="\n", encoding=self.options.getonly('encoding', 'global', 'utf-8')) as fp:
```
saulpw: can we remove this and just do `with self.source.open_text() as fp:` ?

saulpw (commented):
Hi @SuRaMoN, I'm not sure this merits a core loader.  It looks like this is just tsv with some escape chars?  Maybe we can just add an unescape option to the tsv loader, what do you think?

visidata/main.py
SuRaMoN: i cannot use the open_bytes with stdin. it used the BytesIOWrapper which is really hacky, but it also doesnt work once i use io.TextIOWrapper (read1 method doesnt exists on textiowrapper). i fixed it by opening stdin twice.

can you give an example of a loader that would fail? tsv is a text loader and it doesnt fail. i already use this code change for months on my machine and i have never seen what you describe, see also comment below about refactoring Path

visidata/loaders/mysqlclient.py
SuRaMoN: nope, we need the newline="\n" option
also the open_text method of Path is pretty evil. it contains:
```
        if self.fptext:
            self.rfile = RepeatFile(self.fptext)
            return self.rfile
```

so the newline="\n" option is ignored, as fptext already exists.
i think we should avoid passing fptext to the Path constructor as much as possible, and always try to pass a binary stream. the binary can be easily converted to text, the other way around not.

SuRaMoN:
i also dont like the copy past between the tsv and mysqlclient. i moved it to the tsv loader now as you asked and i added the extra options.
the extra disable_universal_newline and interpret_escaped_chars options feel a bit forced for the tsv loader.
what if we keep the mysqlclient loader but keep the code minimal by refering to the tsv loader? this way we dont have the options, en we also dont have the code copy paste

saulpw:
Hey @SuRaMoN, thanks for your patience.  I've taken another look at this PR at last.

These seem to be the main areas of concern: text formats that aren't compatible with the universal newline, and parsing special characters for mysqlclient, and whether it's better to open stdin in binary instead of text mode.

I'm wondering, what if we added options.newline as a mechanism to set the newline character?  Would that and options.encoding='unicode_escape' work for the mysqlclient use case?  In that case, would we still need the ability to open the stdin as binary?

I think you are [probably right about the open_text method in Path](https://github.com/saulpw/visidata/pull/1078#discussion_r971357607), it is kind of evil. This might be a good chance for us to clean that up.  But if we can get away with the `newline` option and a few minor changes, I'd like to do that first anyway.

If those do work for you, I would still like to merge the tests that are part of your PR, which will help us not regress this functionality in the future.

titosemi:
What I nice idea! :tada: 
I would love to see this feature being integrated :angel:

anjakefala:
@SuRaMoN I am going to close this PR for now as inactive. If you ever want to pick it up again, we can re-open and have a fresh start.

Issue #1058, opened 2021-07-26
[loader-json] Columns with __ are duplicated for each row when --clean-names is True


geekscrapy:
**Small description**
Columns with a double underscore as the key get duplicated

**Expected result**
For columns with a double underscore to only be shown once like other columns

**Actual result with screenshot**
Actual:
![image](https://user-images.githubusercontent.com/11225502/126990172-568c2e8a-dd51-4880-bdde-a642b68c1cd6.png)

Expected:
![image](https://user-images.githubusercontent.com/11225502/126990257-ef650030-5f1b-482a-88e0-d8011d56e864.png)

**Steps to reproduce with sample data and a .vd**
```json
{
    "objects":[
        {
            "col1": "col1a",
            "_col2": "col2 value a",
            "__col3": "valueA"
          },
          {
            "col1": "col1b",
            "_col2": "col2 value b",
            "__col3": "valueB"
          },
          {
            "col1": "col1c",
            "_col2": "col2 value c",
            "__col3": "valueC"
          },
          {
            "col1": "col1d",
            "_col2": "col2 value d",
            "__col3": "valueD"
          },
          {
            "col1": "col1e",
            "_col2": "col2 value e",
            "__col3": "valueE"
          }
    ]
}
```

load with `pbpaste | vd -f json`

**Additional context**
v2.6dev

ajkerrigan:
Hey @geekscrapy, I'm trying to reproduce this locally and seeing the expected results. Any chance you've got something in a plugin or .visidatarc that may be getting tripped up? Do you see the same behavior if you use `pbpaste | vd -f json --config=/dev/null`?

anjakefala:
We should maybe add that to our bug report checklist, @ajkerrigan!

geekscrapy:
Got it: `pbpaste | vd -f json --config=/dev/null --clean_names=True`

geekscrapy:
> We should maybe add that to our bug report checklist, @ajkerrigan!

Can we add this to the issue creation template too?

frosencrantz:
@geekscrapy if you do find the cause of the duplicate columns I would like to know what you find.

I've seen something similar, but haven't had time to track it down.

geekscrapy:
FYI I'm running with this modification for now if anyone wanted to try it out:

visidata/loaders/json.py
```python
        for k in row:
            k = maybe_clean(k, self)        # This is the new line @ 73
            if k not in self._colnames:
[...]

PR #1084, opened 2021-08-22
Stopped cleaned column names from being duplicated

geekscrapy:
Using --clean_names duplicated columns due to the the column name never existing in the current column list.

Addresses #1058

### DIFF
--- a/visidata/loaders/json.py
+++ b/visidata/loaders/json.py
@@ -70,6 +70,7 @@ class JsonSheet(PythonSheet):
         super().addRow(row, index=index)
 
         for k in row:
+            k = maybe_clean(k, self)
             if k not in self._colnames:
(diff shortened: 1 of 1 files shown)


geekscrapy:
Maybe this use of `maybe_clean()` should be added to the loaders doco/checklist?

saulpw (approved):
Thanks @geekscrapy!  This makes sense to me.  I wonder if this fix should maybe go in addColumn to affect all loaders?

saulpw:
Thanks for pulling this together, @geekscrapy.  There were some nuances that needed to be addressed and I went ahead and did them.

geekscrapy:
> Thanks @geekscrapy! This makes sense to me. I wonder if this fix should maybe go in addColumn to affect all loaders?

Yes this is probably the cleanest thing to do 👍

PR #1085, opened 2021-08-22
Fixed mysql loader stopping on first error

SuRaMoN:
- [X] If registering an external plugin, [the plugin checklist](https://visidata.org/docs/contributing#plugins) was referenced.

    Currently the mysql loader stops on the first error.
    In case of e.g. one character encoding issue, all
    consequent rows are no longer fetched.

    This commit fixes this for decoding errors and
    shows a warning of the error


saulpw (approved):
Thanks for putting this together, @SuRaMoN!

PR #1086, opened 2021-08-22
POC connected sheets

SuRaMoN:
Hey, this is a small POC for "connected sheets". It allows you to connect different sheets.
It is similar to how foreign keys work on databases. Once you have connected the sheets, you can dive into a row and you get and view the connected sheet filtered on the value of the current row. 

In the preview below I connect the animal sheet with their food preferences based on the "animal" column. When the sheets are connected, i can view all the preferences of the animal when pressing ENTER

https://asciinema.org/a/yQfuXToHkmwxJHyb3MxtLU8TG

I'm willing to make this POC production ready, but i'm looking for some input.

Things to be implemented:
 - auto connect sheets bases on metadata of e.g. mysql/sqlite, perhaps also auto suggest joins based on the connected sheets
 - how to handle multiple connected sheets? Right now the first one is opened, perhaps a popup that lets you choose?
 - View the metadata of the connections in the sheet sheet
 - Progress bar when opening connected sheet
 - auto connect sheets based on columns name?

### DIFF
--- a/visidata/join.py
+++ b/visidata/join.py
@@ -10,20 +10,27 @@ def ensureLoaded(vd, sheets):
     threads = [vs.ensureLoaded() for vs in sheets]
     vd.status('loading %d sheets' % len([t for t in threads if t]))
 
-def createJoinedSheet(sheets, jointype=''):
+def joinSheets(sheets, jointype=''):
     sheets[1:] or vd.fail("join requires more than 1 sheet")
 
-    if jointype == 'append':
+    if jointype == 'connect':
+        for sourceSheet in sheets:
+            sourceSheet.connectionsSource = sourceSheet.keyCols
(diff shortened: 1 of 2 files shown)


saulpw:
@SuRaMoN this is pretty clever, I see what you're doing here.  Let me sit with this for a bit, as I'd like to think about how to combine this with some other features in the queue.

saulpw:
Hi @SuRaMoN, I know it's been a long time, and we're finally getting around to looking at some of these old PRs.  This is actually a clever interface idea which I think has some merit.  If you're still interested in extending VisiData like this, come find me on IRC or https://visidata.org/chat, I would love to chat with you about how we can get something like this in as a feature, and it will be much easier to do that over sync chat.

PR #1098, opened 2021-09-02
Improved XLSX loader

hoclun-rigsep:
For every column in the worksheet, this loader creates four columns, the
last three of which are hidden by default:

* cell value (as before)
* the openpyxl cell object
* cell font color
* cell fill start_color

See #1088 for prior discussion.

Let me know if you want me to:

* make the extra columns optional
* make each extra column optional
* include some other cell property or properties
* make this a separate class altogether

### DIFF
--- a/visidata/loaders/xlsx.py
+++ b/visidata/loaders/xlsx.py
@@ -28,34 +28,37 @@ class XlsxIndexSheet(IndexSheet):
             src = self.workbook[sheetname]
             yield XlsxSheet(self.name, sheetname, source=src)
 
-
 class XlsxSheet(SequenceSheet):
     # rowdef: AttrDict of column_letter to cell
     def setCols(self, headerrows):
+        from openpyxl.utils.cell import get_column_letter
         self.columns = []
         self._rowtype = AttrDict
 
         if not headerrows:
             return
 
-        headers = [[x.value for x in row.values()] for row in headerrows]
-        column_letters = [x.column_letter for x in headerrows[0].values()]
+        headers = [[cell.value for cell in row.values()] for row in headerrows]
+        column_letters = [
+                x.column_letter if 'column_letter' in dir(x)
+                else get_column_letter(i+1)
+                for i, x in enumerate(headerrows[0].values())]
 
         for i, colnamelines in enumerate(itertools.zip_longest(*headers, fillvalue='')):
             colnamelines = ['' if c is None else c for c in colnamelines]
-            self.addColumn(AttrColumn(''.join(map(str, colnamelines)), column_letters[i] +'.value'))
+            colname = ''.join(map(str, colnamelines))
+            self.addColumn(AttrColumn(colname, column_letters[i] + '.value'))
 
     def addRow(self, row, index=None):
         Sheet.addRow(self, row, index=index)  # skip SequenceSheet
         for column_letter, v in list(row.items())[len(self.columns):len(row)]:  # no-op if already done
-            self.addColumn(AttrColumn('', column_letter+'.value'))
+            self.addColumn(AttrColumn(column_letter, column_letter + '.value'))
 
     def iterload(self):
         from openpyxl.utils.cell import get_column_letter
         worksheet = self.source
         for row in Progress(worksheet.iter_rows(), total=worksheet.max_row or 0):
-            yield AttrDict({get_column_letter(i+1):cell for i, cell in enumerate(row)})
-
+            yield AttrDict({get_column_letter(i+1): cell for i, cell in enumerate(row)})
 
 class XlsIndexSheet(IndexSheet):
     'Load XLS file (in Excel format).'
--- a/visidata/loaders/xlsx.py
+++ b/visidata/loaders/xlsx.py
@@ -1,5 +1,8 @@
 from visidata import *
 
+vd.option('xlsx_cellobject', False, 'include column of cell objects', replay=True)
+vd.option('xlsx_fontcolor', False, 'include cell font colors', replay=True)
+vd.option('xlsx_fillcolor', False, 'include cell fill colors', replay=True)
 
 @VisiData.api
 def open_xls(vd, p):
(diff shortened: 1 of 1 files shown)


geekscrapy:
Awesome work with this @hoclun-rigsep !

Personally I use the unhide all columns command rather a lot, so while the extra features would be nice to have, I think the extra columns would clutter the view. So if we can have a simple class object for the row that has these values as attrs, I think that would work the best 👍

hoclun-rigsep:
I think it's important not to change default behavior here so I think that absent authoritative guidance I will make the new columns off by default. With those columns being opt-in, there's no reason to hide them by default. I'll make the corresponding changes and resubmit.

saulpw:
Hey @hoclun-rigsep, leaving these columns out by default makes sense to me.  I also committed a change yesterday to make the rowdef a list of Cells, which should make this a bit easier; you can use the nested attribute syntax like "A.foo.bar" to get the colors and any other formatting characteristics you want, instead of having to bend over backwards with ExprColumns.

hoclun-rigsep:
https://github.com/saulpw/visidata/commit/53c9d7cdf82094145cc817b8e078e8e5f4c22140

visidata/loaders/xlsx.py
```diff
@@ -28,33 +31,60 @@ def iterload(self):
...
+                    AttrColumn(column_letter + '_fillColor',
+                               column_letter + '.fill.start_color.value'))
 
     def iterload(self):
+        from openpyxl.utils.cell import get_column_letter
         worksheet = self.source
         for row in Progress(worksheet.iter_rows(), total=worksheet.max_row or 0):
-            yield AttrDict({cell.column_letter:cell for cell in row})
-
+            # Is it possible for a worksheet to "skip" a column letter?
```
saulpw: This is a good question, but I think even hidden columns will show up in VisiData (though it would be neat if they would stay hidden).

visidata/loaders/xlsx.py
```diff
@@ -28,33 +31,60 @@ def iterload(self):
...
+        headers = [[cell.value for cell in row.values()] for row in headerrows]
+        column_letters = [
+                x.column_letter if 'column_letter' in dir(x)
+                else get_column_letter(i+1)
+                for i, x in enumerate(headerrows[0].values())]
 
         for i, colnamelines in enumerate(itertools.zip_longest(*headers, fillvalue='')):
             colnamelines = ['' if c is None else c for c in colnamelines]
-            self.addColumn(AttrColumn(''.join(map(str, colnamelines)), column_letters[i] +'.value'))
+            colname = ''.join(map(str, colnamelines))
```
saulpw: Let's factor this into a common method (maybe `addXlsxColumns`?).

saulpw (commented):
Okay I think this is pretty reasonable.  If you refactor that common code and remove the comment (assuming letters can't be skipped), then I think we can merge this.

visidata/loaders/xlsx.py
hoclun-rigsep: `addXlsxColumns()` implemented.

visidata/loaders/xlsx.py
```diff
@@ -28,33 +31,52 @@ def iterload(self):
...
-            self.addColumn(AttrColumn(''.join(map(str, colnamelines)), column_letters[i] +'.value'))
+            column_name = ''.join(map(str, colnamelines))
+            self.addColumn(AttrColumn(column_name, column_letters[i] + '.value'))
+            self.addXlsxColumns(column_letters[i], column_name)
 
     def addRow(self, row, index=None):
         Sheet.addRow(self, row, index=index)  # skip SequenceSheet
         for column_letter, v in list(row.items())[len(self.columns):len(row)]:  # no-op if already done
-            self.addColumn(AttrColumn('', column_letter+'.value'))
+            self.addColumn(AttrColumn(column_letter, column_letter + '.value'))
```
saulpw: The reason we used `''` and let it auto-assign the column letter, is that the column name is officially "blank" for purposes of `g^` (which sets blank column names from rows). This might set the column letter different from the actual spreadsheet, but `g^` is important for any table where the headers may be part of the data.

anjakefala:
Can we understand why the tests are failing, and make sure the build is green before we merge?

```
1	1	tests/golden/xlsx-header.tsv
diff --git a/tests/golden/xlsx-header.tsv b/tests/golden/xlsx-header.tsv
index 77657af8..5eb2545c 100644
--- a/tests/golden/xlsx-header.tsv
+++ b/tests/golden/xlsx-header.tsv
@@ -1,4 +1,4 @@
-account number	customer name	sales rep	sku	category	quantity	unit price	ext price	date
+A	B	C	D	E	F	G	H	I
 680916	Mueller and Sons	Loring Predovic	GP-14407	Belt	19	88.49	1681.31	2015-11-17 05:58:34
 680916	Mueller and Sons	Loring Predovic	FI-01804	Shirt	3	78.07	234.21	2016-02-13 04:04:11
 530925	Purdy and Sons	Teagan O'Keefe	EO-54210	Shirt	19	30.21	573.99	2015-08-11 12:44:38
 ```
 
 The header for this test seems to have changed, is this expected?
 
 This is the test:
 
 ```
	override	header	set-option	0		
			open-file	sample_data/sample-sales-reps.xlsx	o	
sample-sales-reps		キsample-salesv4	dive-row		^J	
sample-sales-reps_sample-salesv4		0	rename-cols-row		g^	
sample-sales-reps_sample-salesv4		0	delete-row		d	

```

visidata/loaders/xlsx.py
hoclun-rigsep: Got it. What about the following line? Should I be leaving the names of the optional columns blank too?

saulpw: Yes, I'd leave all the names for the basic value columns blank.  The color/etc additional column should keep their names.

visidata/loaders/xlsx.py
```diff
@@ -28,33 +31,52 @@ def iterload(self):
             src = self.workbook[sheetname]
             yield XlsxSheet(self.name, sheetname, source=src)
 
-
 class XlsxSheet(SequenceSheet):
     # rowdef: AttrDict of column_letter to cell
     def setCols(self, headerrows):
+        from openpyxl.utils.cell import get_column_letter
```
anjakefala, line 37: > The function get_column_letter has been relocated in Openpyxl version 2.4 from openpyxl.cell to openpyxl.utils.

So, we will need to minimum bound the openpyxl version.

visidata/loaders/xlsx.py
```diff
@@ -28,33 +31,52 @@ def iterload(self):
...
         if not headerrows:
             return
 
-        headers = [[x.value for x in row.values()] for row in headerrows]
-        column_letters = [x.column_letter for x in headerrows[0].values()]
+        headers = [[cell.value for cell in row.values()] for row in headerrows]
+        column_letters = [
+                x.column_letter if 'column_letter' in dir(x)
+                else get_column_letter(i+1)
+                for i, x in enumerate(headerrows[0].values())]
```
anjakefala, line 48: This section is actually not necessary for MergedCells in the header to work!

visidata/loaders/xlsx.py
```diff
@@ -28,33 +31,52 @@ def iterload(self):
...
+            self.addColumn(AttrColumn('', column_letter + '.value'))
+            self.addXlsxColumns(column_letter, column_letter)
 
     def iterload(self):
+        from openpyxl.utils.cell import get_column_letter
         worksheet = self.source
         for row in Progress(worksheet.iter_rows(), total=worksheet.max_row or 0):
-            yield AttrDict({cell.column_letter:cell for cell in row})
-
+            yield AttrDict({get_column_letter(i+1): cell for i, cell in enumerate(row)})
```
anjakefala, line 66: Ported this part here: https://github.com/saulpw/visidata/commit/106ad501384758c380d5cec9de4ee7a49eca7aa8

visidata/loaders/xlsx.py
anjakefala: It might still be reasonable to keep it, in case `.column_letter` does not work for another reason.

anjakefala:
I made a PR that handles the merge conflict, here! https://github.com/hoclun-rigsep/visidata/pull/1

Once you take care of that, we will accept the PR. =) If you merge within the next week or so, this can be part of VisiData's next release.

anjakefala:
Thanks so much for all the work @hoclun-rigsep!

We are going to change the 3 options to a single option `xlsx_meta_columns`. Setting that one to True will add all the extra columns. =)

frosencrantz:
@hoclun-rigsep this looks really amazing!  Thank you for adding it!

Are you planning adding any follow up features?  It seems like adding a colorizer that can format the data cells would be very useful.

hoclun-rigsep:
Thanks @frosencrantz. Adding a colorizer: are you talking about doing work on the xlsx writer? If not, what?

frosencrantz:
Hi @hoclun-rigsep   Now that you’ve shown vd the colors for XLS files, it would be great if the cells could be colored based on that color value. It looks like XLS is using a 24 bit color, and curses might only allow up to 256 colors. So the colors would need to be mapped, but that would be helpful.

hoclun-rigsep:
Okay, so you mean coloring cells in curses, not .xlsx output. I can try.

PR #1108, opened 2021-09-14
First pass at helpstrings

stuartellis:

This is a first pass at the helpstrings.

There are a few decision points in here:

- "Empty" as a synonym for "null". I am offering it to see what you think, but suspect that it might not be sufficiently accurate.
- "This locks the cell values" instead of "cells will be evaluated". Again, I am trying to express the result more plainly and avoid the technical term, but it might be at the expense of accuracy.
- "Show" instead of "unhide"
- "Data source" instead of "source". I think that using this phrase improves clarity for a small size increase.
- "Supported" plugins vs. "curated" - I don't think that "supported" necessarily captures your intent, but I don't like the word "curated" in general. It might be sufficiently common to be OK for an international audience, but it always feels ambigious to me.

### DIFF
--- a/visidata/fill.py
+++ b/visidata/fill.py
@@ -33,4 +33,4 @@ def fillNullValues(vd, col, rows):
     vd.status("filled %d values" % n)
 
 
-Sheet.addCommand('f', 'setcol-fill', 'fillNullValues(cursorCol, someSelectedRows)', 'fills null cells in selected rows of current column with contents of non-null cells up the current column')
+Sheet.addCommand('f', 'setcol-fill', 'fillNullValues(cursorCol, someSelectedRows)', 'Fills empty cells in selected rows of current column. This uses the contents of non-null cells up the current column.')
--- a/visidata/freeze.py
+++ b/visidata/freeze.py
@@ -51,7 +51,7 @@ class StaticSheet(Sheet):
                     row.append(val)
 
 
-Sheet.addCommand("'", 'freeze-col', 'sheet.addColumnAtCursor(StaticColumn(cursorCol))', 'add a frozen copy of current column with all cells evaluated')
+Sheet.addCommand("'", 'freeze-col', 'sheet.addColumnAtCursor(StaticColumn(cursorCol))', 'Add a frozen copy of current column. This locks the cell values.')
 Sheet.addCommand("g'", 'freeze-sheet', 'vd.push(StaticSheet(sheet)); status("pushed frozen copy of "+name)', 'open a frozen copy of current sheet with all visible columns evaluated')
 Sheet.addCommand("z'", 'cache-col', 'cursorCol.resetCache()', 'add/reset cache for current column')
 Sheet.addCommand("gz'", 'cache-cols', 'for c in visibleCols: c.resetCache()', 'add/reset cache for all visible columns')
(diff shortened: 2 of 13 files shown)


visidata/pyobj.py
```diff
@@ -337,7 +337,7 @@ def launch_repl(v, i):
 Sheet.addCommand('z(', 'expand-col-depth', 'expand_cols_deep(sheet, [cursorCol], depth=int(input("expand depth=", value=1)))', 'expand current column of containers to given depth (0=fully)')
 Sheet.addCommand('gz(', 'expand-cols-depth', 'expand_cols_deep(sheet, visibleCols, depth=int(input("expand depth=", value=1)))', 'expand all visible columns of containers to given depth (0=fully)')
 
-Sheet.addCommand(')', 'contract-col', 'closeColumn(sheet, cursorCol)', 'unexpand current column; restore original column and remove other columns at this level')
+Sheet.addCommand(')', 'contract-col', 'closeColumn(sheet, cursorCol)', 'Restore original size of current column. This removes other columns at the same level.')
```
saulpw, line 340: "contract" isn't about size, but nested columns; it's the opposite of expand.

visidata/modify.py
```diff
@@ -284,4 +284,4 @@ def new_rows(sheet, n):
 Sheet.addCommand('za', 'addcol-new', 'addColumnAtCursor(SettableColumn(input("column name: "))); cursorRight(1)', 'append an empty column')
 Sheet.addCommand('gza', 'addcol-bulk', 'addColumnAtCursor(*(SettableColumn() for c in range(int(input("add columns: "))))); cursorRight(1)', 'append N empty columns')
 
-Sheet.addCommand('z^S', 'commit-sheet', 'commit()', 'commit changes back to source.  not undoable!')
+Sheet.addCommand('z^S', 'commit-sheet', 'commit()', 'Commit changes back to data source. You cannot undo these changes.')
```
saulpw, line 287: How do we feel about having "You" in the helpstrings?  I think this would be the first usage, and we should make a decision about how technical vs. conversational the helpstrings (and other docs) are.

visidata/freeze.py
```diff
@@ -51,7 +51,7 @@ def reload(self):
                     row.append(val)
 
 
-Sheet.addCommand("'", 'freeze-col', 'sheet.addColumnAtCursor(StaticColumn(cursorCol))', 'add a frozen copy of current column with all cells evaluated')
+Sheet.addCommand("'", 'freeze-col', 'sheet.addColumnAtCursor(StaticColumn(cursorCol))', 'Add a frozen copy of current column. This locks the cell values.')
```
saulpw, line 54: "Locked" isn't quite right.  It's more like, a completely evaluated copy.  Maybe 'frozen' is the wrong nomenclature too; 'snapshot' maybe?

visidata/fill.py
```diff
@@ -33,4 +33,4 @@ def _undo():
     vd.status("filled %d values" % n)
 
 
-Sheet.addCommand('f', 'setcol-fill', 'fillNullValues(cursorCol, someSelectedRows)', 'fills null cells in selected rows of current column with contents of non-null cells up the current column')
+Sheet.addCommand('f', 'setcol-fill', 'fillNullValues(cursorCol, someSelectedRows)', 'Fills empty cells in selected rows of current column. This uses the contents of non-null cells up the current column.')
```
saulpw, line 36: I see how "empty" is more international English than "null", but "null" has a specific meaning in VisiData: None, or the value of `options.null_value`.  And if this option isn't set to the empty string, then `fill` won't actually fill empty cells.

saulpw (commented):
We merged the helpstring changes that worked well, and I've left comments here on the ones we didn't merge.  I'm going to close this PR; let's start another PR when we do another round of helpstrings.

visidata/modify.py
stuartellis: We've previously discussed writing some guidelines - I will create a PR with a strawman set as a starting point.

visidata/freeze.py
stuartellis: Snapshot feels much better to me.

visidata/fill.py
stuartellis: OK. My instinct is to minimise technical terms in documentation, but we can't sacrifice accuracy.

PR #1119, opened 2021-09-23
[plugins] bump vds3 to version 0.8

ajkerrigan:
- Support VisiData 2.6
  - Add S3 context menus
  - Use `vd` methods over globals where possible

- Only support the 'append' join type for `join-rows`

https://github.com/ajkerrigan/visidata-plugins/releases/tag/vds3-0.8


anjakefala:
Thanks @ajkerrigan!

PR #1139, opened 2021-10-04
Add documentation guidelines to CONTRIBUTING.md

stuartellis:

This adds a concise set of guidelines for documentation work to the CONTRIBUTING.md file.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,6 +73,17 @@ VisiData has two main branches:
 
 All pull requests should be submitted against `develop`. Submitters will need to e-sign a Copyright Assignment Agreement (CAA) before a pull request will be accepted.
 
+# Documentation and Help Text
+
+Follow these guidelines when writing documentation for VisiData, such as the help text for menu items:
+
+1. Use International English
+2. Avoid using pronouns. Use "you" or "they" where appropriate.
+3. Ensure that explanations are clear and precise. VisiData enables users to quickly query and change data in powerful ways, and our documentation should support that goal.
+4. Avoid assuming that users are familiar with command-line shells
+5. Avoid assuming that users are familiar with the Python programming language
+6. The name of the product is "VisiData", capitalized as shown.
+
 # Open Source License and Copyright
 
 VisiData is an open-source utility that can be installed and used for free (under the terms of the [GPL3](https://www.gnu.org/licenses/gpl-3.0.en.html)).


CONTRIBUTING.md
```diff
@@ -73,6 +73,17 @@ VisiData has two main branches:
 
 All pull requests should be submitted against `develop`. Submitters will need to e-sign a Copyright Assignment Agreement (CAA) before a pull request will be accepted.
 
+# Documentation and Help Text
+
+Follow these guidelines when writing documentation for VisiData, such as the help text for menu items:
+
+1. Use International English
```
saulpw, line 80: What's a good link/reference for learning how to use good International English?

CONTRIBUTING.md
```diff
@@ -73,6 +73,17 @@ VisiData has two main branches:
...
+2. Avoid using pronouns. Use "you" or "they" where appropriate.
+3. Ensure that explanations are clear and precise. VisiData enables users to quickly query and change data in powerful ways, and our documentation should support that goal.
+4. Avoid assuming that users are familiar with command-line shells
+5. Avoid assuming that users are familiar with the Python programming language
```
anjakefala, line 84: I do not actually agree with 4 and 5. I do agree that there should be a part of the documentation that caters to users that are not familiar with command-line shells, and the Python programming language, but I think it is very fair to assume in the remainder of the documentation a baseline of technical knowledge. For one thing, it will result in a ton of repetition throughout the documentation, and in helpstrings we cannot reall afford that. For another thing, I find that sometimes the use of simpler language, is less technically precise, and for me personally leads to more confusion. (See the example when a change was proposed to reword `null` to a less technically precise word (`empty cell`) that was subtly wrong.) Sacrificing comprehension for a theoretical accessibility can make the writing less accessible and more tedious to read. I would rather we include a glossary, and educational materials, than assume that readers still have no background when they are 10 pages into the documentation. VisiData can be the gateway for someone gaining a comfort with the commandline, and Python. Users are smarter than given credit for here.

anjakefala: This is a post that I would recommend us reading: http://jvns.ca/blog/confusing-explanations/#pattern-2-having-inconsistent-expectations-of-the-reader-s-knowledge

Inconsistent expectations are a huge problem, and so we should be clear at each set of documentation what our expectations are for the baseline reader's knowledge.

saulpw: I largely agree with @anjakefala on this.  I think there's room for us to provide beginner terminal and Python resources (whenever I look through @jsvine's tutorial I notice that the external link to basic Python expressions is not good for beginners, and I haven't been able to find a better one myself).  But it's outside the scope of the tool itself, as VisiData in general is optimized for terminal users and people with basic math/programming experience (if not Python specifically).

anjakefala: I think we would definitely benefit from including more examples of the sorts of Python expressions you can do with VisiData! Some examples for `;` and `=` in particular would go really far.

CONTRIBUTING.md
stuartellis: I think that this page of the Google developer documentation style guide is the best summary:

https://developers.google.com/style/translation

Basically, US English spelling and grammar, written in plain language, avoiding abbreviated words and local terms.

The US Government provide official guidance on what "plain language" means, but technical style guides like the Google guide tend to include similar advice.

https://www.plainlanguage.gov/guidelines/

CONTRIBUTING.md
stuartellis: Thank you. I definitely agree with the point that it's probably more useful to state what existing knowledge is assumed for a specific set of documentation, rather than trying to just exclude areas. I also take the points that using VisiData to the full requires you to have some level of knowledge of the terminal (and desire to use it), and a little Python, and that VisiData may be a gateway for some people to increase their knowledge in those areas as they work with it.

I will have another shot at this, with specific reference to helpstrings.

WRT the idea of a glossary: I also thought this might be a way to address the fact that we have to use specific terms within helpstrings, but we can't explain them there. A glossary Web page might provides something that search engines can pick up, as well as functioning as part of the documentation set.

saulpw: I like the idea of a glossary, and it should ideally be in the manpage as well as on the web.

anjakefala:
Cleaning up PRa that have not had follow-up activity.

Issue #1131, opened 2021-09-30
[v2.6.1] Warning on startup, if no .visidata/plugins/


anjakefala:
```
   File "/usr/local/lib/python3.9/dist-packages/visidata/settings.py", line 364, in loadConfigAndPlugins
     vd.addGlobals(importlib.import_module(modname).__dict__)                
   File "/usr/lib/python3.9/importlib/__init__.py", line 127, in import_module   
     return _bootstrap._gcd_import(name[level:], package, level)       
   File "<frozen importlib._bootstrap>", line 1030, in _gcd_import      
   File "<frozen importlib._bootstrap>", line 1007, in _find_and_load      
   File "<frozen importlib._bootstrap>", line 984, in _find_and_load_unlocked                            
 ModuleNotFoundError: No module named 'plugins'     
 ```

PR #1152, opened 2021-10-14
[start-up] create plugins dir, if does not exist

anjakefala:
Not sure about this change, so opening PR to chat about it.

The plugins directory is where VisiData plugins are imported from. These are imported by default, resulting in a warning if the directory itself is not present. The discussed approach is to create the directory, if it does not exist.

Related to #1131.

**Edit**: We decided to instead catch the specific exception, and not log a warning for it.

- [x] Change URL to # 1131
- [x] Test PR on VisiData docker image

### DIFF
--- a/visidata/settings.py
+++ b/visidata/settings.py
@@ -363,6 +363,11 @@ def loadConfigAndPlugins(vd, args=AttrDict()):
     for modname in (args.imports or vd.options.imports or '').split():
         try:
             vd.addGlobals(importlib.import_module(modname).__dict__)
+        except ModuleNotFoundError as e:
+            # issue #1131
+            if 'plugins' in e.args[0]:
+                continue
+            vd.exceptionCaught(e)
         except Exception as e:
             vd.exceptionCaught(e)
             continue


saulpw (commented):
This was my first idea too, but on second thought we should just catch the specific exception during import, instead of modifying the filesystem.

visidata/settings.py
```diff
@@ -363,6 +363,10 @@ def loadConfigAndPlugins(vd, args=AttrDict()):
     for modname in (args.imports or vd.options.imports or '').split():
         try:
             vd.addGlobals(importlib.import_module(modname).__dict__)
+        except ModuleNotFoundError as e:
+            # https://github.com/saulpw/visidata/issues/1131
```
saulpw: what do you think about putting `#1131` or `# issue #1131` on the line above, instead of the full URL?

anjakefala: Ah yeah, that makes more sense! I just wanted to leave a reference for why this specific exception for future Us. =)

saulpw (approved):
This special case means that we should probably import `plugins` explicitly instead of using options.imports. But this is fine for now.

PR #1157, opened 2021-10-18
Updated sqlite handler to wrap column names in quotes

davidskeck:
Any column names which included spaces would cause the following error on line 208: sqlite3.OperationalError: near "test": syntax error
This fix wraps the column names in quotes, which is what the sqlite documentation suggests to do when columns have spaces.


saulpw (approved):
Thanks for putting this together, @davidskeck!  This is a good fix.

PR #1158, opened 2021-10-19
[sqlite-] allow encoding/decoding of sqlite db with invalid unicode

### LINKED ISSUE #1156: SQLite databases with invalid unicode not supported
**Small description**

Opening a SQLite database will fail when a cell in a TEXT column has invalid unicode characters

**Expected result**

Either the specific cell that has invalid unicode is not imported, or it is imported but the invalid characters are stripped or slash-escaped.

error screenshot

error screenshot

(those are errors from two different DBs that I've seen this problem on)

Here's a SQLite file that has this problem: https://hack.wesleyac.com/test.sqlite

**Additional context**

`vd --version`: `saul.pw/VisiData v2.4`

dbcli/litecli#89 is a similar issue in a different python project, which may have useful context for fixing this

anjakefala:
Closes #1156


visidata/loaders/sqlite.py
```diff
@@ -21,7 +27,9 @@ def resolve(self):
 
     def conn(self):
         import sqlite3
-        return sqlite3.connect(str(self.resolve()))
+        sq = sqlite3.connect(str(self.resolve()))
```
saulpw: Use the standard `con` varname.

visidata/loaders/sqlite.py
```diff
@@ -2,13 +2,19 @@
...
+vd.option('sqlite_encoding', 'utf-8', 'decoder used by sqlite3', replay=True)
+
 @VisiData.api
 def open_sqlite(vd, p):
     return SqliteIndexSheet(p.name, source=p)
 
 VisiData.open_sqlite3 = VisiData.open_sqlite
 VisiData.open_db = VisiData.open_sqlite
 
+def sqlite_encoder(s):
```
saulpw: This is a decoder, so let's name it `sqlite_decoder` instead.

visidata/loaders/sqlite.py
```diff
@@ -2,13 +2,19 @@
 
 from visidata import *
 
+vd.option('sqlite_encoding', 'utf-8', 'decoder used by sqlite3', replay=True)
```
saulpw: Let's just use the existing `options.encoding`.

visidata/loaders/sqlite.py
saulpw: Also let's move this onto the sheet object and use `sheet.options.encoding`.  Also we want to only get the option once instead of every string decoding (options are alas quite slow), so we could make it a `@cached_property` or just turn it into a one-liner:

         sq.text_factory = lambda s, enc=self.options.encoding: s.decode(enc)

saulpw (changes requested):
Good find with the `text_factory`, let's clean it up and we can merge it!

PR #1164, opened 2021-11-02
Move urlcache to vd object

geekscrapy:
This is a change which may have unintended consequences so feel free to decline!

But this would allow urlcache to be overridden within .visidatarc. This is mainly so that additional headers etc can be added to allow auth

### DIFF
--- a/plugins/geocoding.py
+++ b/plugins/geocoding.py
@@ -10,7 +10,7 @@ __version__='0.1'
 import urllib.parse
 import json
 
-from visidata import vd, urlcache, AttrDict, Sheet, Column, ColumnExpr
+from visidata import vd, AttrDict, Sheet, Column, ColumnExpr
 
 vd.option('ntk_key', '', 'API Key for nettoolkit.com')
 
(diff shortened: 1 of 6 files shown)


saulpw:
We should also expose it as a global in case another plugin is depending on it that way.  Use either `@VisiData.global_api` or add `visidata.addGlobals({'urlcache': vd.urlcache})` to deprecated.py (not sure if this latter will work for imports).

geekscrapy:
> We should also expose it as a global in case another plugin is depending on it that way. Use either `@VisiData.global_api` or add `visidata.addGlobals({'urlcache': vd.urlcache})` to deprecated.py (not sure if this latter will work for imports).

Done 👍  Added is as global_api

Used this for testing in my `.visidatarc`
```
old_urlcache = vd.urlcache
def new_urlcache(url, *args, **kwargs):
   vd.status(f'Downloading: {url}')
   return old_urlcache(url, *args, **kwargs)
vd.urlcache = new_urlcache
```

saulpw (approved):
Looks good to me, thanks for pulling this together!

PR #1174, opened 2021-11-06
Fix async deepcopy async-deepcopy-error

pstuifzand:
Changes after `gz"` should not be reflected on the source sheet, because
it is a deepcopy. Fixes the error when maken a copy with that command.

### DIFF
--- a/docs/rows.md
+++ b/docs/rows.md
@@ -49,7 +49,7 @@ An example usage follows.
 
     **or**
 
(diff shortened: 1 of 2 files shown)


anjakefala:
Thanks @pstuifzand! This was very appreciated. I picked the commit here onto develop: https://github.com/saulpw/visidata/commit/34592ba7a711c2a18a8c0f6b6067aba79035da18 with a small adjustment.

Issue #1189, opened 2021-11-13
[load_lazy] expand options.load_lazy to work for the loading of multiple files


anjakefala:
Currently, `options.load_lazy` being set to False, results in the loading of all subsheets for embedded sheet formats like `xlsx` or `hdf5`. It would be great if it would also load all files users seperately pass to VisiData.

e.g. `vd sample.tsv benchmarks.csv` with `options.load_lazy` set to `False` should mean both `sample.tsv` and `benchmarks.csv` are loaded upon start-up.

geekscrapy:
It would be great if you could also explicitly set this on a sheet itself. I have some sheets which are part of an indexSheet that are processor intensive and so want these loaded only on user interaction

I briefly tried `MySheet.class_options.lazy_load = True` and it didn't seem to work

saulpw:
That should work, @geekscrapy !  Maybe try with `load_lazy` (I made the same mistake myself :)

geekscrapy:
> That should work, @geekscrapy ! Maybe try with `load_lazy` (I made the same mistake myself :)

Hmm doesn't seem to work for my sheets in the index sheet, and it doesn't look like it should:
![image](https://user-images.githubusercontent.com/11225502/141655605-07780222-4d8a-4576-8d60-9149d44f4c48.png)

Maybe it should be:

![image](https://user-images.githubusercontent.com/11225502/141655734-9a7c6f19-3d0d-44c7-896d-654b4075f138.png)

geekscrapy:
PR: #1193 

> > That should work, @geekscrapy ! Maybe try with `load_lazy` (I made the same mistake myself :)
> 
> Hmm doesn't seem to work for my sheets in the index sheet, and it doesn't look like it should: ![image](https://user-images.githubusercontent.com/11225502/141655605-07780222-4d8a-4576-8d60-9149d44f4c48.png)
> 
> Maybe it should be:
> 
> ![image](https://user-images.githubusercontent.com/11225502/141655734-9a7c6f19-3d0d-44c7-896d-654b4075f138.png)

saulpw:
It should have worked if you set it on the index sheet.  Did you fix the mispelling of the option when you set it? (from lazy_load to load_lazy)

anjakefala:
I have tested `MySheet.class_options.load_lazy = True` and it does work!

PR #1193, opened 2021-11-13
Do not auto load if sheet.load_lazy is True or global load_lazy is True

geekscrapy:
I have done limited testing for this, but it seemed to work well for me

### DIFF
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -963,7 +963,7 @@ class IndexSheet(Sheet):
 
     def addRow(self, sheet, **kwargs):
         super().addRow(sheet, **kwargs)
-        if not self.options.load_lazy:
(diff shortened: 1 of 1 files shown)


saulpw (approved):

visidata/sheets.py
```diff
@@ -963,7 +963,7 @@ def getSheet(self, k):
 
     def addRow(self, sheet, **kwargs):
         super().addRow(sheet, **kwargs)
-        if not self.options.load_lazy:
+        if not self.options.load_lazy and not sheet.options.load_lazy:
```
anjakefala, line 966: `and` was chosen to fit @geekscrapy's particular use-case here: https://github.com/saulpw/visidata/issues/1189

`options.load_lazy` would be `False` by default, and sub-sheets that geekscrapy did not want loaded would be set to `True`.

Issue #1200, opened 2021-11-17
[infosec] cmdlog_histfile and input_history save location discrepency


geekscrapy:
**Small description**
`options.cmdlog_histfile` saves to the current directory where as `options.input_history` saves to the users `~/.visidata` directory

**Expected result**
Both same to the same location if no path is provided.

**Actual result with screenshot**
given the below settings in my `.visdatarc`, `.input_history`will save to `~/.visidata`, but `.cmdlog_histfile` will save to the users current directory
```
options.cmdlog_histfile = '.cmdlog_histfile'
options.input_history = '.input_history'
```

Additionally, this may actually have other ramifications such as security of the `.cmdlog_histfile` if vd is used within a path that others have access to. Additionally, the user may not have write permissions to the current directory.

For the security reason alone I would suggest this is patched ASAP, if indeed this is a bug ;)

geekscrapy:
Looks fixed to me 👍

PR #1202, opened 2021-11-17
removed concern of #1200 but must be saved in ~

geekscrapy:
"Fixes" issue in #1200 - unsure what the `date()` part was originally for

### DIFF
--- a/visidata/cmdlog.py
+++ b/visidata/cmdlog.py
@@ -202,7 +202,7 @@ class _CommandLog:
             sheet.cmdlog_sheet.addRow(vd.activeCommand)  # add to sheet-specific cmdlog
             if options.cmdlog_histfile:
                 if not getattr(vd, 'sessionlog', None):
-                    vd.sessionlog = vd.loadInternalSheet(CommandLog, Path(date().strftime(options.cmdlog_histfile)))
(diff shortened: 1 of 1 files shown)


saulpw (commented):
Let's keep the `strftime`, but otherwise this is a good fix.

geekscrapy:
As agreed, we can close this PR

Summary:
cmdlog option needs to be an absolute path, otherwise the file is saved relative to the users CWD. strfmt should also be used.

From vd chat as an example of other tools saving history: "Both bash and zsh [history files] are relative [to the CWD] unless an absolute path is given [in HISTFILE]". This said, you usually start the bash (shell) from your $HOME dir so it would be saved in a standard location. The only time this would change (if you have a relative path set as HISTFILE) is if you open the shell from another location (which doesn't occur that often).

Bash default HISTFILE value is: ~/.bash_history . So an absolute path

So while this PR is being closed, I feel that input_history and cmdlog_histfile should act the same, and should mirror the behaviour of shell logging features.

I'll try whip something up if you agree with the above concepts

saulpw:
I agree they should act the same...is there a way we can do that, without breaking existing behavior?  or gracefully deprecating one of the options?

PR #1215, opened 2021-11-23
use returncode to determine if pip install failed

geekscrapy:


saulpw (commented):
I agree with the basic premise here.  But if there is information on stderr, we should show it as a warning, whether or not the overall command succeeded.

geekscrapy:
Yea, I did think this however even if the install is successful it outputs on stderr which threw me off a number of times not thinking it was installed.

In addition, I found that it outputs to stderr even when pip is out of date...

Maybe if debug flag is set it is shown?

Or as you say, show it as a warning and only vd.fail if the retcode != 0   ?

S'up to you, just a couple of thoughts while fudging my way though failed pip installs 😁

saulpw (approved):

PR #1228, opened 2021-12-03
Sheet types addGlobals and adjusted imports

geekscrapy:
The initial reason for this PR was adding CsvSheet to global.

Also modified other loaders to not use import * syntax, and add similar other sheet types to global using addGlobal.

I can continue to do this for the other loaders if you want it uniform across all loaders, but I thought I'd just do a few to give an example of what it might look like, to see if you want it done across them all.

Personally, I see it as beneficial as all Sheet types should be as accessible as the next.

If it doesn't suit you, then that's fine, it was a good exercise!

### DIFF
--- a/visidata/loaders/eml.py
+++ b/visidata/loaders/eml.py
@@ -1,4 +1,6 @@
-from visidata import *
+import os
+
+from visidata import vd, VisiData, TableSheet, Column, vlen
 
 @VisiData.api
 def open_eml(vd, p):
@@ -41,9 +43,13 @@ def extract_parts(sheet, givenpath, *parts):
         for part in parts:
             vd.execAsync(sheet.extract_part, givenpath / part.get_filename(), part)
     elif len(parts) == 1:
-        vd.execAsync(sheet.extract_part, givenpath, part)
+        vd.execAsync(sheet.extract_part, givenpath, parts[0])
     else:
         vd.fail('cannot save multiple parts to non-dir')
 
 EmailSheet.addCommand('x', 'extract-part', 'extract_part(inputPath("save part as: ", value=cursorRow.get_filename()), cursorRow)')
 EmailSheet.addCommand('gx', 'extract-part-selected', 'extract_parts(inputPath("save %d parts in: " % nSelectedRows), *selectedRows)')
+
+vd.addGlobals({
+    'EmailSheet': EmailSheet
+})
--- a/visidata/loaders/fixed_width.py
+++ b/visidata/loaders/fixed_width.py
@@ -1,5 +1,4 @@
-
-from visidata import *
+from visidata import vd, VisiData, Column, SequenceSheet, Sheet, Progress
 
 vd.option('fixed_rows', 1000, 'number of rows to check for fixed width columns')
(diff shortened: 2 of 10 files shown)


visidata/loaders/fixed_width.py
```diff
@@ -72,13 +70,12 @@ def iterload(self):
...
         self.headerlines = headerlines
 
-
 @VisiData.api
 def save_fixed(vd, p, *vsheets):
     with p.open_text(mode='w', encoding=vsheets[0].options.encoding) as fp:
         for sheet in vsheets:
             if len(vsheets) > 1:
-                fp.write('%s\n\n' % vs.name)
+                fp.write('%s\n\n' % sheet.name)
```
saulpw, line 78: Nice find!  Let's put this in a separate PR.

visidata/loaders/eml.py
```diff
@@ -41,9 +43,13 @@ def extract_parts(sheet, givenpath, *parts):
         for part in parts:
             vd.execAsync(sheet.extract_part, givenpath / part.get_filename(), part)
     elif len(parts) == 1:
-        vd.execAsync(sheet.extract_part, givenpath, part)
+        vd.execAsync(sheet.extract_part, givenpath, parts[0])
```
saulpw, line 46: Another good catch!  separate commit at least though.

saulpw (changes requested):
Thanks for putting this together, @geekscrapy!  It's great to clean up the imports, and expose some of these Sheet types.  I have a few notes:
- I don't want to have to adhere to the internals of the loaders, so I'm hesitant to expose more than we need to.  If a sheet type is useful (either to inherit from, or instantiate in some other way), then let's definitely expose it, but I'm not sure if we need to expose every sheet type automatically.  I could be convinced though.
- be careful when converting global `options` to `vd.options`.  The usages may be equivalent in most cases, but sometimes the global should actually be a sheet-specific option.  `vd.options` indicates that there's no way it could be a sheet-specific option, either because it doesn't make sense, or there is no relevant sheet in that context.  This may or may not be true; for example, `options.html_title` should probably be `sheet.options` instead of `vd.options`.  Or maybe `vsheets[0]`, as we might be converging on a rule of "multi-sheet save takes options from the first sheet".  So they each have to be considered.
- Keep the final newline at the end of the file (probably an editor or git setting).
- Don't add/remove independent blocks of whitespace unless we're adhering to a standard (which we haven't set yet).

So we should at least fix the two bugs you found though (in separate commits), and I'd like to at least keep the import cleanups you've made.  Let's think about each of the options and globals separately, though we can add CsvSheet to globals too (and could you make a separate commit from the import changes?).

geekscrapy:
No problem with all the above.

I'll close this and do some separate PRs based on the themes you set above 👍

In regards to the points above (in order)

- I presume you are referring to only exposing what is required? Yes, totally understandable, however my opinions would be that 1. When you want it, it's not there 2. All sheet types should be exposed the same?
- 👍 I actually tried to do `from visidata import options` but it gave me a recursion error regarding imports. I'll make a separate PR for this and each can be taken on their own merits.

PR #1234, opened 2021-12-09
Add CsvSheet, ZipSheet, TarSheet to global

geekscrapy:
Added CsvSheet, ZipSheet, TarSheet to globals and resolved to absolute imports.
Also added a required import of TextSheet to shell.py which was not present


geekscrapy:
Ugh really sorry about this...

PR #1237, opened 2021-12-09
fixed_width saver_fixed var name bug

geekscrapy:
save_fixed did not have the correct variable name


geekscrapy:
I **will** get this...

PR #1241, opened 2021-12-09
getSheet fix when sheetname is "options"

geekscrapy:
When sheetname is 'options' the call to `self.optionsSheet` should be `vd.globalOptionsSheet` as it called in method


saulpw (approved):
Totally right, @geekscrapy, thanks!

PR #1265, opened 2022-01-12
Revert default VisiData behaviour to choose saver from extension

anjakefala:
Additionally, fallback to `.vds` as a default save filetype for sheets that lack a `Path`.

Needed since ea3a924.


anjakefala:
Related: https://github.com/saulpw/visidata/discussions/1266

anjakefala:
Will re-open fresh OAFA, when we work out all of the important saving considerations. I will want to open a fresh PR when that day comes.

PR #1296, opened 2022-02-20
[delete-row] clear deleted rows from selectedRows

### LINKED ISSUE #1284: [2.9dev] using "d" to delete a selected row doesn't remove from sheet.selectedRows
**Small description**
When a row is deleted with `d`, it is not removed from `sheet.selectedRows`.

**Expected result**
For the row to be removed like when `gd` is used.

1. open with `vd --config/=/dev/null file.csv`
2. select a row with "`t`" 
3. delete that row with "`d`"
4. `=vd.sheet.selectedRows` < should be 0

anjakefala:
Closes #1284

On deferred sheets, deleted rows are unselected immediately.

### DIFF
--- a/visidata/clipboard.py
+++ b/visidata/clipboard.py
@@ -106,6 +106,10 @@ def delete_row(sheet, rowidx):
     if not sheet.defer:
         oldrow = sheet.rows.pop(rowidx)
         vd.addUndo(sheet.rows.insert, rowidx, oldrow)
+        # clear the deleted row from selected rows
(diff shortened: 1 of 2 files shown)


geekscrapy:
> On deferred sheets, should deleted rows be unselected immediately, or only after committed?

As selection does not affect the underlying data, I'd say immediately?

saulpw (approved):
Thanks, @anjakefala, great job.

PR #1305, opened 2022-02-26
[confirm] fix display of yes/no in confirm

anjakefala:
## Before
Confirm Dialog develop

## After
Confirm Dialog fix-form

I tried a lot of different things before settling on manually specifying which character to underline. =( It was really hard to have both relative positioning, and a specific character underlined.

### DIFF
--- a/visidata/form.py
+++ b/visidata/form.py
@@ -10,7 +10,7 @@ def open_mnu(vd, p):
 vd.save_mnu=vd.save_tsv
 
 class FormSheet(VisiDataMetaSheet):
-    rowtype='labels' # rowdef: { .x .y .text .color .command .input }
+    rowtype='labels' # rowdef: { .x .y .text .color .command .input .underline}
 
 @VisiData.api
 def replayCommand(vd, longname, sheet=None, col='', row=''):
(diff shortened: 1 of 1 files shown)


visidata/form.py
```diff
@@ -33,12 +33,15 @@ def draw(self, scr):
                 continue
             x, y = r.x, r.y
             if isinstance(y, float) and (0 < y < 1) or (-1 < y < 0): y = h*y
-            if isinstance(x, float) and (0 < x < 1) or (-1 < x < 0): x = w*(x-len(r.text)*w)
+            if isinstance(x, float) and (0 < x < 1) or (-1 < x < 0): x = w*x-(len(r.text)/2)
```
anjakefala, line 36: This is the main fix for the display. Even if we do not go with the underlying logic, this change will fix the general display. I can definitely seperate them into two seperate commits, or just include this change, if desired.

anjakefala: Without this, we were getting x-coordinates that were negative, and large.

geekscrapy:
Just a quick q based on the screenshots above, as I've not seen the intended dialog box yet: how do you know which option is selected? Or do you currently select with the underlined characters only?

anjakefala:
@geekscrapy Underlined characters only, afaik!

saulpw (commented):
Let's make `underline` have the substring to be underlined.

saulpw (approved):
Okay, looks good!

Issue #1315, opened 2022-03-07
"NameError: name 'PandasSheet' is not defined" when creating a new sheet from selected rows.


jasonwirth:
**Small description**
Reading a Parquet file uses the pandas loader. 
Selecting rows and pressing the quote `"` to duplicate the sheet with the selected rows throws a `NameError`.

**Expected result**
New sheet with the selected rows

**Actual result with screenshot**
[error_PandasSheet_not_defined.txt](https://github.com/saulpw/visidata/files/8194318/error_PandasSheet_not_defined.txt)

```python
Traceback (most recent call last):
  File "/Users/jasonwirth/code/visidata/visidata/basesheet.py", line 170, in execCommand
    escaped = super().execCommand2(cmd, vdglobals=vdglobals)
  File "/Users/jasonwirth/code/visidata/visidata/basesheet.py", line 72, in execCommand2
    exec(code, vdglobals, LazyChainMap(vd, self))
  File "dup-selected", line 1, in <module>
NameError: name 'PandasSheet' is not defined

```

**Steps to reproduce with sample data and a .vd**
[select_parquet.vd.txt](https://github.com/saulpw/visidata/files/8194355/select_parquet.vd.txt)
```
sheet	col	row	longname	input	keystrokes	comment
			open-file	sample_data/sample.parquet	o	
sample		0	select-row		s	select current row
sample		1	select-row		s	select current row
sample		2	select-row		s	select current row
sample			dup-selected		"	open duplicate sheet with only selected rows

```

**Additional context**
Please include the version of VisiData.
```
❯ vd --version
saul.pw/VisiData v2.9dev
```

anjakefala:
This was fixed by @jasonwirth! The fix is now on `develop`.

jasonwirth:
Thanks! Glad to contribute. VD is awesome. Thanks for everyone's hard work!

PR #1316, opened 2022-03-07
Fix dup-selected for pandas loaders and add test

jasonwirth:
- [ Y ] If contributing a core loader, [the loader checklist](https://visidata.org/docs/contributing#loader) was referenced.

Fix for [#1315 ](https://github.com/saulpw/visidata/issues/1315)

I added a sample .vd file as well as a sample .parquet file in `/sample_data`


anjakefala:
The tests are failing because we are missing a golden result for `sample_parquet.vd`!

Could you please `vd -p tests/sample_parquet.vd`, check if the output looks as expected, and then `Ctrl+S` *tests/golden/sample_parquet.tsv*, and commit the saved file?

tests/select_parquet.vd
```diff
@@ -0,0 +1,6 @@
+sheet	col	row	longname	input	keystrokes	comment
```
anjakefala: I really appreciate you taking the time to add a test for the bug fix!

jasonwirth:
I added the golden tests file but the CI fails with this [error](https://app.circleci.com/pipelines/github/saulpw/visidata/1162/workflows/812dbc16-f9be-451d-b82b-6f310f11e6cc/jobs/3969?invite=true#step-110-1323) which is really about needing a parquet library:


```
 - Missing optional dependency 'fastparquet'. fastparquet is required for parquet support. Use pip or conda to install fastparquet.
```

I can reproduce the issue without using parquet, it's just that I discovered the bug while reading parquet files. Since parquet isn't natively supported by VD (unlike csv) it'll fallback to pandas `read_parquet` method. I have to manually specify the loader to make it error again.

Let me create another tests uisng a .tsv.

anjakefala (approved):

anjakefala:
Thanks so much @jasonwirth! For the fix, and the tests. =)

PR #1337, opened 2022-03-21
Remove use of nonexistent prettykeys `kDN`/`kUP`

### LINKED ISSUE #1336: `kDN` and `kUP` prettykeys don't actually exist
**Small description** [PR incoming]

The pretty keys `kDN` and `kUP` are used in two places in the code, but they are not defined in the translation dictionary.

**Expected result**
Default keybindings should include "Shift+Up", "Shift+Down", "gShift+Up", and "gShift+Down".

Bogus keybindings:
2022-03-21-184238_958x528_scrot

**Additional context**
Running visidata-2.8 on Arch Linux.

djpohly:
Replaces kDN and kUP with the correct versions (KEY_SF and KEY_SR) if they are not already present.  Also makes the order of the Shifted bindings in slide.py consistent with the rest of the file (non-functional change).

Fixes #1336.

### DIFF
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -389,9 +389,7 @@ def editCell(self, vcolidx=None, rowidx=None, value=None, **kwargs):
         value = value or col.getDisplayValue(self.rows[self.cursorRowIndex])
 
     bindings={
-        'kUP':        acceptThenFunc('go-up', 'rename-col' if rowidx < 0 else 'edit-cell'),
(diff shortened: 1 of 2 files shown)


saulpw:
Hey @djpohly, thanks for taking the initiative on this issue!  Unfortunately we can't merge this PR, as some systems/terminals report different values for these extended keystrokes, and this change would break [Chesterton's keybindings](https://www.chesterton.org/taking-a-fence-down/) on those systems. If the displayed prettykeys in the commands list are wrong, let's fix that instead. Would you be willing to submit another PR trying that approach?

PR #1340, opened 2022-03-28
[man] have vd --help open the .txt manpage by default

### LINKED ISSUE #1332: vd --help doesn't play well with pipes and non-ttys
**Small description**

    vd --help | col -b > ~/tmp/vd-help.txt

has garbled codes in the resulting file.

**Expected result**

A nice plain text file containing the help text to vd.

**Additional context**

This happens in version 2.6.1 and version 2.8

See also [issue 1089](https://github.com/saulpw/visidata/issues/1089).

_Side note:_ I personally think `vd --help` should not use _any_ control characters  and be rather brief, while a full man page could be produced with `--man` - but that's obviously subjective.

anjakefala:
Control characters in vd.1 groff are incompatible with non-tty pipes,
e.g. `col -b`.

Closes #1332

The appearance is bc we format the .txt to 80 characters.

### DIFF
--- a/visidata/main.py
+++ b/visidata/main.py
@@ -13,6 +13,7 @@ import locale
 import datetime
 import signal
 import warnings
+from pkg_resources import resource_filename
 
 from visidata import vd, options, run, BaseSheet, AttrDict
 from visidata import Path
@@ -81,6 +82,14 @@ optalias('force_valid_colnames', 'clean_names')  # deprecated
 
 def main_vd():
     'Open the given sources using the VisiData interface.'
+    if '-v' in sys.argv or '--version' in sys.argv:
+        print(vd.version_info)
+        return 0
+    if '-h' in sys.argv or '--help' in sys.argv:
+        with open(resource_filename(__name__, 'man/vd.txt'), 'r') as fp:
+            print(fp.read())
+        return 0
+
     try:
         locale.setlocale(locale.LC_ALL, '')
     except locale.Error as e:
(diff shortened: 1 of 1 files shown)


visidata/help.py
```diff
@@ -134,7 +134,13 @@ def openManPage(vd):
     import os
     with SuspendCurses():
         if os.system(' '.join(['man', resource_filename(__name__, 'man/vd.1')])) != 0:
-            vd.push(TextSheet('man_vd', source=Path(resource_filename(__name__, 'man/vd.txt'))))
+            os.system(' '.join(['man', resource_filename(__name__, 'man/vd.txt')]))
```
anjakefala: This is the change (l137), that I think merits further discussion.

geekscrapy:
My two pence 😉:

I've not tested, but I presume doing openManPage would instanciate the .visidatarc (thereby executing custom, potentially buggy code)? I wouldn't (normally) expect this. What I _would_ generally expect would be minimal output of the tools features, like the txt file, and very limited loading of any vd code.

Q: Do other tools read their config files when "--help" is used? If they did and there was an error in the config, it may prevent the help being shown. Could this happen with vd?
Additionally, I presume things like input history etc, would be logged when showing this man page? Is this expected behaviour when using --help?

Given the above I'd prefer just to dump the txt file to stdout. It's greppable, lessable etc., simple, and doesn't need to rely on anything vd related

saulpw:
Thanks for the thoughts, @geekscrapy, I think I agree with you.  `--help` should not read config or do anything else, and should output plain ascii.  Probably it should be a very abridged CLI usage + options output, though for the moment our standard vd.txt will work fine.

visidata/main.py
```diff
@@ -81,6 +82,14 @@ def optalias(abbr, name, val=None):
 
 def main_vd():
     'Open the given sources using the VisiData interface.'
+    if any(i in ['-v', '--version'] for i in sys.argv):
```
saulpw: Let's invert it: `any(x in sys.argv for x in ['-v', '--version']`.   Or since there are only two, probably cleaner to use `'-v' in sys.argv or '--version' in sys.argv`.

anjakefala: My heart for a "if (x or y) in z".

saulpw (approved):

PR #1342, opened 2022-03-30
Add shell completion

Freed-Wu:

Imitate <https://github.com/soimort/you-get/blob/master/README.md?plain=1#L118-L120>. Add a shell completion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,10 @@ To install the cutting edge `develop` branch (no warranty expressed or implied):
 
 See [visidata.org/install](https://visidata.org/install) for detailed instructions for all available platforms and package managers.
 
+#### Shell completion
+
+Completion definitions for Zsh can be found in [`contrib/completion`](https://github.com/saulpw/visidata/tree/develop/contrib/completion). Please consult your shell's manual for how to take advantage of them.
+
 ### Usage
 
     $ vd <input>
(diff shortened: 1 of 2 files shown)


saulpw:
Hi @Freed-Wu, thanks for making this!  I don't think we can take this into core visidata (and definitely not into `stable`), as it won't be updated to reflect new options, etc. Would you be up for putting it into a repo of your own, or a gist, or something?  Be sure to include which version you made it for.  We can link to it from the website.

Freed-Wu:
> as it won't be updated to reflect new options, etc

Hey, I think <https://github.com/ytdl-org/youtube-dl/blob/master/devscripts/zsh-completion.py> maybe satisfy your requirements? It uses a python script to generate zsh shell completion script. Also, <https://github.com/ytdl-org/youtube-dl/blob/master/devscripts/fish-completion.py> and <https://github.com/ytdl-org/youtube-dl/blob/master/devscripts/bash-completion.py>.

For library, I only know <https://pypi.org/project/completion-utils/> can help developer to generate bash completion script.

saulpw:
Hi @Freed-Wu, yes, if you port these to VisiData, we can include them in our build process.  It should not be too hard but will require testing on the various shells, which I don't have time for at the moment.

PR #1343, opened 2022-04-03
Fix renaming and saving of HtmlTableSheets from HtmlTablesSheets

### LINKED ISSUE #1339: Impossible to set current column for selected rows to incremental values and to bulk save all sheets
**Small description**

Hi, I want to open and bulk save all the html tables I have [here](https://lab24.ilsole24ore.com/ecosistema-urbano/indexT.php). All the tables have the same name, then I need to rename them using `g+i`, and then save them in bulk `g+CTRL+s`.

When I run `g+i` I have `NameError: name 'setitem' is not defined`.

If I change the name of the tables, in the saved HTML files, when I run `g+CTRL+s`, I have `AttributeError: 'str' object has no attribute 'name'`

I start from 

```
 vd https://lab24.ilsole24ore.com/ecosistema-urbano/indexT.php --save-filetype=csv
```

**Expected result**

- to have the tables renamed;
- to be able to save the tables in multiple CSV files:

image

image

image

image

Please attach the commandlog (saved with `Ctrl-D`) to show the steps that led to the issue.

```
sheet   col     row     longname        input   keystrokes      comment
global          null_value      set-option
global          save_filetype   set-option      csv
                        open-file       https://lab24.ilsole24ore.com/ecosistema-urbano/indexT.php      o
indexT                  select-rows             gs      select all rows
indexT  name            setcol-incr             gi      set current column for selected rows to incremental values
```

**Additional context**
Please include the version of VisiData: the latest dev saul.pw/VisiData v2.9dev

anjakefala:
Closes #1339

### DIFF
--- a/visidata/save.py
+++ b/visidata/save.py
@@ -163,6 +163,6 @@ def save_txt(vd, p, *vsheets):
 
 BaseSheet.addCommand('^S', 'save-sheet', 'vd.saveSheets(inputPath("save to: ", value=getDefaultSaveName()), sheet, confirm_overwrite=options.confirm_overwrite)', 'save current sheet to filename in format determined by extension (default .tsv)')
 BaseSheet.addCommand('g^S', 'save-all', 'vd.saveSheets(inputPath("save all sheets to: "), *vd.stackedSheets, confirm_overwrite=options.confirm_overwrite)', 'save all sheets to given file or directory)')
-IndexSheet.addCommand('g^S', 'save-selected', 'vd.saveSheets(inputPath("save %d sheets to: " % nSelectedRows, value="_".join(vs.name or "blank" for vs in source)), *selectedRows, confirm_overwrite=options.confirm_overwrite)', 'save all selected sheets to given file or directory')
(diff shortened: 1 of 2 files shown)


visidata/basesheet.py
```diff
@@ -91,7 +91,7 @@ def options(self):
 
     def __init__(self, *names, **kwargs):
         self._name = None   # initial cache value necessary for self.options
-        self.names = names
+        self.names = list(names)
```
frosencrantz: I was looking at this report and found the same bugs.  I do not know the code very well, and was wondering does the getter/setter from visidata/sheets line 949  work?  Will changing the name from the Index sheet change the name of the file that is saved?  It updates names[] but it doesn't update name.  I don't know how the two fields names and name are supposed to interact. It seems like one is calculated from the other looking at this line and the next.

https://github.com/saulpw/visidata/blob/fb0a6ad18247586b9c0a6b38e708abfebd14dc44/visidata/sheets.py#L949

anjakefala: You are absolutely right! It changes the sheet name through the `SheetsSheet`, but not for html tables, at least.

anjakefala: Thank you! I think I got a much better fix now.

saulpw (approved):

visidata/basesheet.py
frosencrantz: Glad I could help!  Yes, I like this fix better, and I learned something.

PR #1355, opened 2022-04-18
Notify when plugin update available

geekscrapy:
This will show a status message if an update to a plugin is available

I often find I am out of date, but it is not common for me to actually check the plugin sheet...!

The prevalence (vd.status/warning etc.) should be evaluated, is this the correct level? I suspect so as it allows people to ignore if they want to.

My only concern is that on load, the status bar is becoming cluttered... Thinking about it, this might be an ideal candidate to have an icon/message etc. in the upper right corner to show a plugin has an update? Or maybe something in the menu showing the number of updates available? Many options! I think this is a good start though

### DIFF
--- a/visidata/plugins.py
+++ b/visidata/plugins.py
@@ -12,6 +12,7 @@ from visidata import VisiData, vd, Path, CellColorizer, JsonLinesSheet, AttrDict
 
 
 vd.option('plugins_url', 'https://visidata.org/plugins/plugins.jsonl', 'source of plugins sheet')
+vd.option('plugins_check', True, 'notify if plugin version is out of date')
 
 
 @VisiData.lazy_property
@@ -107,6 +108,10 @@ class PluginsSheet(JsonLinesSheet):
                 vd.addGlobals({funcname: func})
                 setattr(vd, funcname, func)
 
+            if vd.options.plugins_check: # show status if plugin update is found
+                ver = _loadedVersion(r)
+                if ver and ver != r.latest_ver:
+                    vd.status(f'update available for plugin "{r.name}"')
 
     def installPlugin(self, plugin):
         # pip3 install requirements
--- a/visidata/plugins.py
+++ b/visidata/plugins.py
@@ -12,7 +12,6 @@ from visidata import VisiData, vd, Path, CellColorizer, JsonLinesSheet, AttrDict
 
 
 vd.option('plugins_url', 'https://visidata.org/plugins/plugins.jsonl', 'source of plugins sheet')
-vd.option('plugins_check', True, 'notify if plugin version is out of date')
 
 
 @VisiData.lazy_property
@@ -108,10 +107,10 @@ class PluginsSheet(JsonLinesSheet):
                 vd.addGlobals({funcname: func})
                 setattr(vd, funcname, func)
 
-            if vd.options.plugins_check: # show status if plugin update is found
-                ver = _loadedVersion(r)
-                if ver and ver != r.latest_ver:
-                    vd.status(f'update available for plugin "{r.name}"')
+        # check for stale_plugins (out of date)
+        stale_plugins = list(filter(lambda r: _loadedVersion(r) != r.latest_ver, self.rows))
+        if len(stale_plugins) > 0:
+            vd.warning(f'update available for {len(stale_plugins)} plugin{"s"[:len(stale_plugins)^1]}')
 
     def installPlugin(self, plugin):
         # pip3 install requirements
--- a/visidata/plugins.py
+++ b/visidata/plugins.py
@@ -107,10 +107,14 @@ class PluginsSheet(JsonLinesSheet):
                 vd.addGlobals({funcname: func})
                 setattr(vd, funcname, func)
 
-        # check for stale_plugins (out of date)
-        stale_plugins = list(filter(lambda r: _loadedVersion(r) != r.latest_ver, self.rows))
+        # check for plugins with newer versions
+        def is_stale(r):
+            v = _loadedVersion(r)
+            return v and v != r.latest_ver
+        
+        stale_plugins = list(filter(is_stale, self.rows))
         if len(stale_plugins) > 0:
-            vd.warning(f'update available for {len(stale_plugins)} plugin{"s"[:len(stale_plugins)^1]}')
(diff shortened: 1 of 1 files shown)


anjakefala:
I can confirm that the change works, and that it will load upon startup. 

If versions are guaranteed to be numeric and incrementing (e.g. 0.1, then 0.2), then probably better to change this to if `ver < r.latest_version`?

geekscrapy:
> If versions are guaranteed to be numeric and incrementing (e.g. 0.1, then 0.2), then probably better to change this to if `ver < r.latest_version`?

If we could guarantee float values as integers then I would agree.

Additionally I took inspiration from the plugin colouriser which uses the same check. This was done so that the two checks are aligned

saulpw (commented):
hey @geekscrapy, thanks for the idea.  Rather than creating a new option, I'm thinking that we could populate a variable with the list of stale plugins, add a `vd.helpstatus` (or something) property that returns either "N plugin updates available" or the standard "Ctrl+H for help" message if N=0, and then set `options.disp_menu_fmt="{vd.helpstatus}"`.  This assumes that users with plugins know how to access help but that seems reasonable to me.  What do you think?

saulpw:
> If versions are guaranteed to be numeric and incrementing (e.g. 0.1, then 0.2), then probably better to change this to if `ver < r.latest_version`?

Alas, versions are not floating point numbers.  I think `!=` is reasonable here.

geekscrapy:
> > If versions are guaranteed to be numeric and incrementing (e.g. 0.1, then 0.2), then probably better to change this to if `ver < r.latest_version`?
> 
> Alas, versions are not floating point numbers. I think `!=` is reasonable here.

Do we want to change the colorizer check to != so they are aligned? I can see a situation where one says it is out of date and the other does not

geekscrapy:
> hey @geekscrapy, thanks for the idea. Rather than creating a new option, I'm thinking that we could populate a variable with the list of stale plugins, add a `vd.helpstatus` (or something) property that returns either "N plugin updates available" or the standard "Ctrl+H for help" message if N=0, and then set `options.disp_menu_fmt="{vd.helpstatus}"`. This assumes that users with plugins know how to access help but that seems reasonable to me. What do you think?

So, helpstatus is new to me 😅 (so I'll try it out) but I do like the idea of just counting the number of out-of-date plugins. Personally, I would avoid overwriting a user customisable option (disp_menu_fmt) for the obvious reason (a user may have set it specifically), and also I actually use that option to display info to the user for a plugin 🙂 but I always append, never overwrite the value

saulpw:
> So, helpstatus is new to me sweat_smile (so I'll try it out) but I do like the idea of just counting the number of out-of-date plugins. Personally, I would avoid overwriting a user customisable option (disp_menu_fmt) for the obvious reason (a user may have set it specifically), and also I actually use that option to display info to the user for a plugin slightly_smiling_face but I always append, never overwrite the value

Yes, I agree, we won't overwrite it, it should work like this:

```
vd.stale_plugins = []

@VisiData.property
def plugin_status(vd):
    if vd.stale_plugins:
        return f'{len(vd.stale_plugins} plugins can be updated'
    return ''

@VisiData.property
def help_status(vd):
    return 'Ctrl+H for help'

vd.option('disp_menu_fmt', '{vd.plugin_status or vd.help_status}', ...)
```

and then you can re-use `vd.plugin_status` yourself and/or append to `options.disp_menu_fmt` as you have been.

saulpw:
> Do we want to change the colorizer check to != so they are aligned? I can see a situation where one says it is out of date and the other does not

It looks like `_pluginColorizer` already uses `!=`, does it not?

geekscrapy:
> It looks like `_pluginColorizer` already uses `!=`, does it not?

Yes, I think we got wires crossed here:
Both the colorizer and this PR use `!=`

geekscrapy:
> > So, helpstatus is new to me sweat_smile (so I'll try it out) but I do like the idea of just counting the number of out-of-date plugins. Personally, I would avoid overwriting a user customisable option (disp_menu_fmt) for the obvious reason (a user may have set it specifically), and also I actually use that option to display info to the user for a plugin slightly_smiling_face but I always append, never overwrite the value
> 
> Yes, I agree, we won't overwrite it, it should work like this:
> 
> ```
> vd.stale_plugins = []
> 
> @VisiData.property
> def plugin_status(vd):
>     if vd.stale_plugins:
>         return f'{len(vd.stale_plugins} plugins can be updated'
>     return ''
> 
> @VisiData.property
> def help_status(vd):
>     return 'Ctrl+H for help'
> 
> vd.option('disp_menu_fmt', '{vd.plugin_status or vd.help_status}', ...)
> ```
> 
> and then you can re-use `vd.plugin_status` yourself and/or append to `options.disp_menu_fmt` as you have been.

If I'm completely honest, the complexity of extra variables to `vd` is lost on me, unless you have an idea to use these in the future. By placing the plugin detail at the top, you may be eating into the users message real estate. Also, if the user has `options.disp_menu_fmt` customised, they might not re-add the `vd.plugin_status` back to the `options.disp_menu_fmt` and therefore they are in the same situation where they don't know things are out of date.
Also, the user cannot dismiss the "out of date" message without a change (if we take hints from other apps, they show a popup to say something is out of date and the user can instantly dismiss if required).

I agree that it is probably OTT to display a message for each out of date plugin, so I would prefer to replace the many messages with one, showing the total number which is out-of-date. Also the simplicity is that the user can just ignore it (as it is just a status message) if they want to without making any changes to their vdconf.

saulpw:
The main things I was balking at, were the additional option and one message per plugin.  So if we ditch the option, and provide a tally instead with `vd.warning()`, then I think we can use your original idea.  The proliferation of startup messages is something we'll have to deal with separately.

geekscrapy:
Ok, to keep the PR neat I just added the check at the end

geekscrapy:
I'm thinking that if there is scope, we may want to provide a method to the sheet to compare versions. `!=` and `>` dont catch things. Also, when the installed column is blank the plugin is considered out of date

anjakefala:
I have tested it on a case where there were no plugins that needed an update, and one in which there were.

In both cases, it behaved as expected.

PR #1361, opened 2022-04-25
Index columns

### LINKED ISSUE #1349: Hiding column cancels replay
Hello,

I like the `Visidata` replay feature and hope to be able to use it with my data.

I've noticed that if I hide a column in a MS Excel spreadsheet, the replay is cancelled.

1. Open vd with a MS Excel file that has multiple columns
2. Hide a column with `-`
3. Save the playback as test.vd
4. Quit vd
5. `vd --play test.vd`
6. Notice `replay canceled` will appear

I would expect the replay to hide the column I selected.

This is what's in the playback:
```
myfile  0               hide-col                -       Hide current column
```

$ vd -v
saul.pw/VisiData v2.9dev

anjakefala:
* set .vdj to be the default CmdLog save format
* in the *CmdLog* type **col** indices as `int`, and **col** names as `str`.

since 6e1334167f740a78832054e63020215acc6d5a88, the type of the *col* attribute matters.
If replaying, and *col* is an `int`, the CmdLog will index by position.
If *col* is a `str` it will index by column name.

PR also adds a test for the replay indexing features supported by .vd, and .vdj.

Closes #1349

### DIFF
--- a/visidata/cmdlog.py
+++ b/visidata/cmdlog.py
@@ -190,7 +190,7 @@ class _CommandLog:
 
         comment = vd.currentReplayRow.comment if vd.currentReplayRow else cmd.helpstr
         vd.activeCommand = self.newRow(sheet=sheetname,
-                                            col=str(colname),
+                                            col=colname,
                                             row=str(rowname),
                                             keystrokes=keystrokes,
                                             input=args,
--- a/visidata/cmdlog.py
+++ b/visidata/cmdlog.py
@@ -48,10 +48,10 @@ def checkVersion(vd, desired_version):
 @VisiData.api
 def fnSuffix(vd, prefix):
     i = 0
-    fn = prefix + '.vd'
+    fn = prefix + '.vdj'
     while Path(fn).exists():
         i += 1
-        fn = f'{prefix}-{i}.vd'
+        fn = f'{prefix}-{i}.vdj'
 
     return fn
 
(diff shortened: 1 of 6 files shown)


visidata/cmdlog.py
```diff
@@ -147,7 +147,7 @@ def commandCursor(sheet, execstr):
         rowname = keystr(k) if k else sheet.cursorRowIndex
 
     if contains(execstr, 'cursorTypedValue', 'cursorDisplay', 'cursorValue', 'cursorCell', 'cursorCol', 'cursorVisibleCol'):
-        colname = sheet.cursorCol.name or sheet.visibleCols.index(sheet.cursorCol)
+        colname = str(sheet.cursorCol.name) or int(sheet.visibleCols.index(sheet.cursorCol))
```
saulpw: These type coercions should not be necessary; `index()` should return an int, and the column name should be a string already.

saulpw (changes requested):
One small change and then should be good to go!  Thanks @anjakefala.

frosencrantz:
This looks really great @anjakefala! 

I have existing .vd files.  What is the best way to convert these files to the new .vdj format?

anjakefala:
`vd -b foo.vd -o foo.vdj`!

Or `vd -p foo.vd`.
Then `Ctrl+D` *foo.vdj*. =) I would use the cutting edge on `develop`.

frosencrantz:
Great!  Will try that.  And thanks for making this change.

PR #1363, opened 2022-05-03
[cmdlog] log 'open-source .' when launching with 'vd'

anjakefala:


visidata/main.py
```diff
@@ -251,6 +252,10 @@ def main_vd():
         else:
             vd.push(vd.currentDirSheet)
 
+            # log source to cmdlog
+            vd.cmdlog.openHook(vd.currentDirSheet, vd.currentDirSheet.source)
+            sources.append(vd.currentDirSheet)
```
saulpw: I don't think we want this, esp because it's pushed just above.  (Note that below, `sources` is used to push the first sheet, causing the reload, and then for the `+:pos::` syntax.)

anjakefala: That makes sense!

I removed it.

saulpw (approved):

PR #1368, opened 2022-05-05
Use newline='' for csv.writer

### LINKED ISSUE #1362: Inconsistent CRLF behaviour between Windows and Linux
**Small description**

If I open a CSV with `LF` line endings in VisiData on Windows, make a change, then save the file, I get `CRCRLF` line endings.
Note that this is not just `CRLF` line endings, which is what you get on Linux (see #1076). Note that there are somehow two `CR` characters before the `LF`!

**Expected result**

The line endings should either be `LF` (preserve original) or `CRLF` (Python and VisiData default for CSV writer on Linux).
The line endings should not be `CRCRLF`...

If you get an unexpected error, please include the full stack trace that you get with `Ctrl-E`.

**Original file** (note the LF (`0a`) line ending:
[lf.csv](https://github.com/saulpw/visidata/files/8599982/lf.csv)

```
$ xxd lf.csv
00000000: 612c 620a 312c 320a                      a,b.1,2.
```

**New file** (note the CRCRLF (`0d 0d 0a`) line ending!)
[lf.csv](https://github.com/saulpw/visidata/files/8599988/lf.csv)

```
xxd lf.csv
00000000: 612c 620d 0d0a 332c 340d 0d0a            a,b...3,4...
```

**CommandLog (zipped)**:
[lf.zip](https://github.com/saulpw/visidata/files/8599990/lf.zip)

**Additional context**
Please include the version of VisiData: saul.pw/VisiData v2.7.1

daviewales:
Fixes #1362

According to the docs for csv.writer:

> If csvfile is a file object, it should be opened with newline=''
> https://docs.python.org/3/library/csv.html#csv.writer

Additionally:

> If newline='' is not specified, newlines embedded inside quoted fields will
> not be interpreted correctly, and on platforms that use \r\n linendings on
> write an extra \r will be added. It should always be safe to specify
> newline='', since the csv module does its own (universal) newline handling.
> https://docs.python.org/3/library/csv.html#id3

This commit adds `newline=None` as a keyword argument for `open_text` in
`path.py`. It then uses this new keyword argument when opening a file
for writing in `loaders/csv.py`

### DIFF
--- a/visidata/loaders/csv.py
+++ b/visidata/loaders/csv.py
@@ -47,7 +47,7 @@ class CsvSheet(SequenceSheet):
 @VisiData.api
 def save_csv(vd, p, sheet):
     'Save as single CSV file, handling column names as first line.'
-    with p.open_text(mode='w', encoding=sheet.options.encoding) as fp:
+    with p.open_text(mode='w', encoding=sheet.options.encoding, newline='') as fp:
         cw = csv.writer(fp, **options.getall('csv_'))
         colnames = [col.name for col in sheet.visibleCols]
         if ''.join(colnames):
--- a/visidata/path.py
+++ b/visidata/path.py
@@ -159,7 +159,7 @@ class Path(os.PathLike):
     def __truediv__(self, a):
         return Path(self._path.__truediv__(a))
 
-    def open_text(self, mode='rt', encoding=None):
+    def open_text(self, mode='rt', encoding=None, newline=None):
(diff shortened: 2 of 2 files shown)


daviewales:
Hmm... This fixes the `CRCRLF` issue on Windows, and doesn't break simple CSVs on *nix. But I'm not quite sure why it's breaking the tests.

EDIT: It's because I changed the wrong function... 😮 Fixed now.

saulpw:
Thanks so much for tracking this down and fixing it, @daviewales!  Much appreciated.

PR #1374, opened 2022-05-09
[input-] handle history for non-str input values

### LINKED ISSUE #1371: Title: [input] Up arrow stack traces and exits input
**Small description**
Up arrow causes stack trace, and exits input

**Expected result**
No stack trace, and doesn't exit.

https://asciinema.org/a/DW4cydUhUFBBAFRC3QufhBxQb

```
  File "/home/felix/git/visidata/env/lib/python3.9/site-packages/visidata/basesheet.py", line 167, in execCommand
  File "/home/felix/git/visidata/env/lib/python3.9/site-packages/visidata/basesheet.py", line 69, in execCommand2
  File "aggregate-col", line 1, in <module>
  File "/home/felix/git/visidata/env/lib/python3.9/site-packages/visidata/choose.py", line 63, in chooseMany
    chosenstr = vd.input(prompt+': ', completer=CompleteKey(choice_keys), bindings={'^X': throw_fancy})
  File "/home/felix/git/visidata/env/lib/python3.9/site-packages/visidata/_input.py", line 325, in input
    ret = self.editText(sheet.windowHeight-1, promptlen, sheet.windowWidth-promptlen-rstatuslen-2,
  File "/home/felix/git/visidata/env/lib/python3.9/site-packages/visidata/_input.py", line 264, in editText
  File "/home/felix/git/visidata/env/lib/python3.9/site-packages/visidata/_input.py", line 233, in editline
    elif history and ch == 'KEY_UP':           v, i = history_state.up(v, i)
  File "/home/felix/git/visidata/env/lib/python3.9/site-packages/visidata/_input.py", line 120, in up
    i = len(v)
TypeError: object of type 'int' has no len()
```

1. Set column width Enter
2. Start to add aggregator, use up arrow.

**Additional context**
Please include the version of VisiData. Lasest develop branch.

anjakefala:
Needed since 46686d3

Closes #1371

### DIFF
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -117,7 +117,7 @@ class HistoryState:
         if self.hist_idx > 0:
             self.hist_idx -= 1
             v = self.history[self.hist_idx]
-        i = len(v)
+        i = len(str(v))
         return v, i
 
     def down(self, v, i):
@@ -129,7 +129,7 @@ class HistoryState:
         else:
             v = self.prev_val
             self.hist_idx = None
-        i = len(v)
+        i = len(str(v))
         return v, i
 
 
@@ -245,6 +245,8 @@ def editline(vd, scr, y, x, w, i=0, attr=curses.A_NORMAL, value='', fillchar=' '
             i += 1
 
         if i < 0: i = 0
+        # v may have a non-str type with no len()
+        v = str(v)
         if i > len(v): i = len(v)
         first_action = False
         complete_state.reset()


visidata/_input.py
```diff
@@ -245,6 +245,8 @@ def find_nonword(s, a, b, incr):
             i += 1
 
         if i < 0: i = 0
+        # v may have a non-str type with no len()
+        v = str(v)
```
anjakefala, line 249: Without this, we need `len(str(v))` in a lot more places!

VisiData casts `v` into a `str` in line 150, but then when `v` changes through the use of any of the other `editline()` commands, it can be changed to a non-str'. So, it needs to be re-cast. This is the cleanest way to not have `str()` peppered in a lot of other places.

saulpw: Do we need the rest of the `str(v)` changes then too?

anjakefala: Yes, unfortunately. Because they involve a change in the value, before the value is `str()`. Actually, we do not need line 98, but we do need the `up()` and `down()`.

saulpw (approved):

PR #1396, opened 2022-06-14
[setcol-format-enum-] fix enum formatting; add tests

anjakefala:


visidata/column.py
```diff
@@ -202,16 +209,19 @@ def fmtstr(self, v):
 
     @property
     def formatter(self):
-        return self._formatter.__name__[7:]  # remove leading 'format_'
+        return self._formatter  # remove leading 'format_'
```
anjakefala: For some reason, this was not needed?

visidata/column.py
```diff
@@ -202,16 +209,19 @@ def fmtstr(self, v):
...
 
     @formatter.setter
     def formatter(self, v: str):
-        self._formatMaker = getattr(col, 'format_'+(v or 'generic'))
-        self._formatter = col._formatMaker(self._format)
+        self._formatMaker = getattr(self, 'format_'+(v or 'generic'))
+        self._formatter = self._formatMaker(self._formatdict)
```
anjakefala: Renamed `_format` to a clearer name `_formatdict`.

visidata/column.py
```diff
@@ -202,16 +209,19 @@ def fmtstr(self, v):
 
     @property
     def formatter(self):
-        return self._formatter.__name__[7:]  # remove leading 'format_'
+        return self._formatter  # remove leading 'format_'
 
     @formatter.setter
     def formatter(self, v: str):
-        self._formatMaker = getattr(col, 'format_'+(v or 'generic'))
```
anjakefala, line 209: `col` was `self`

visidata/column.py
```diff
@@ -202,16 +209,19 @@ def fmtstr(self, v):
...
-        self._formatter = col._formatMaker(self._format)
+        self._formatMaker = getattr(self, 'format_'+(v or 'generic'))
+        self._formatter = self._formatMaker(self._formatdict)
 
     def format_generic(self, fmtstr):
         return self.formatValue
 
+    def format(self, *args, **kwargs):
+        return self.formatter(*args, **kwargs)
```
anjakefala: `cursorCol.format` not `cursorCol.formatter` was what was called everywhere. I think we want `cursorCol.format` to be returning `formatter` not `formatValue`. Otherwise, we always use `formatValue` which uses the type's formatting, and `cursorCol.formatter` is never called.

Kept the default `formatter` to `formatValue`.

visidata/sheets.py
```diff
@@ -1210,13 +1210,8 @@ def _async_deepcopy(newlist, oldlist):
 BaseSheet.addCommand(None, 'rename-sheet', 'sheet.name = input("rename sheet to: ", value=sheet.name)', 'Rename current sheet')
 
 
-@Column.cached_property
-def _format(col):
-    return dict(val.split('=', maxsplit=1) for val in col.fmtstr.split())
```
anjakefala, line 1215: I can return this here. This got moved to `_formatdict` in `columns.py`. I removed the `cached_property`, but I can return that.

visidata/sheets.py
```diff
@@ -1210,13 +1210,8 @@ def _async_deepcopy(newlist, oldlist):
...
-
-
 @Column.api
-def format_enum(col, val):
-    return lambda val, vals=col._format: vals.__getitem__(val)
+def format_enum(col, fmtdict):
+    return lambda val, fmtdict=fmtdict: fmtdict.__getitem__(val)
```
anjakefala, line 1215: Renamed `vals` to the clearer `fmtdict`.

visidata/tests/test_commands.py
```diff
@@ -50,6 +50,7 @@ def isTestableCommand(longname, cmdlist):
                  'addcol-expr': 'Unit',          # just copy the column
                  'addcol-incr-step': '2',
                  'setcol-incr-step': '2',
+                 'setcol-format-enum': '1=cat',
```
anjakefala, line 53: Each time we add a new `input` command, we need to add it to `test_commands.py`.

visidata/column.py
```diff
@@ -119,7 +120,7 @@ def __deepcopy__(self, memo):
         return self.__copy__()  # no separate deepcopy
 
     def __getstate__(self):
-        d = {k:getattr(self, k) for k in 'name width height expr keycol formatter fmtstr voffset hoffset aggstr'.split() if hasattr(self, k)}
```
anjakefala: `_formatter` is the raw value

visidata/column.py
```diff
@@ -101,6 +101,7 @@ def __init__(self, name=None, *, type=anytype, cache=False, **kwargs):
         self.height = 1       # max height, None/0 to auto-compute for each row
         self.keycol = 0       # keycol index (or 0 if not key column)
         self.expr = None      # Column-type-dependent parameter
+        self.formatter = 'generic'
```
anjakefala, line 104: Maybe this should be `_formatter`.

tests/format-enum-freeze.vd
```diff
@@ -0,0 +1,4 @@
+sheet	col	row	longname	input	keystrokes	comment
+			open-file	sample_data/sample.tsv	o	
+sample	Item	0	setcol-format-enum	Pencil=1		add secondary type translator to current column from input enum (space-separated)
```
anjakefala, line 3: `freeze-col` will carry over enum and typed formatting

tests/golden/format-enum-freeze.tsv
```diff
@@ -0,0 +1,44 @@
+OrderDate	Region	Rep	Item	Item_frozen	Units	Unit_Cost	Total
+2016-01-06	East	Jones	1	1	95	1.99	189.05
```
anjakefala, line 2: tsv saves the displayed enums

tests/golden/format-enum.json
```diff
@@ -0,0 +1,45 @@
+[
+{"OrderDate": "2016", "Region": "East", "Rep": "Jones", "Item": "Pencil", "Units": 95, "Unit_Cost": 1.99, "Total": "189.05"},
```
anjakefala, line 2: json saves the underlying typed value

visidata/tests/test_commands.py
saulpw: Yes, and let's do `'1=cat 2=dog'` to test multiple, as long as we're here.

visidata/column.py
saulpw: So the idea is that there's a `formatter` property which is viewable/editable by string name on the columns sheet, but `_formatter` is the actual formatter generating function itself.  Same with `fmtstr` and `_format` (now `_formatdict`).  Then `fmtstr` would become `format` that could be either string (most cases) or dict for our enum formatter, and be viewable/editable in the same way.

visidata/column.py
```diff
@@ -227,7 +229,6 @@ def formatValue(self, typedval):
             typedval = typedval.decode(options.encoding, options.encoding_errors)
 
         return vd.getType(self.type).formatter(self.fmtstr, typedval)
-    format=formatValue
```
anjakefala, line 230: As long as this is here, `setcol-format-enum` will never work.

visidata/column.py
```diff
@@ -200,18 +207,13 @@ def fmtstr(self):
...
-    def formatter(self, v: str):
-        self._formatMaker = getattr(col, 'format_'+(v or 'generic'))
-        self._formatter = col._formatMaker(self._format)
-
     def format_generic(self, fmtstr):
         return self.formatValue
 
+    def format(self, *args, **kwargs):
+        self._formatMaker = getattr(self, 'format_'+(self.formatter or 'generic'))
+        return self._formatMaker(self._formatdict)(*args, **kwargs)
```
anjakefala, line 215: Now, `col.format` will build the `_formatMaker`, and `col.formatter` is a simple string.

visidata/column.py
```diff
@@ -191,6 +192,12 @@ def width(self, w):
                 vd.addUndo(setattr, self, '_width', self.width)
             self._width = w
 
+    @property
+    def _formatdict(col):
```
anjakefala, line 196: Was `_format` in `sheets.py`.

PR #1420, opened 2022-07-04
[wip] xdg support

jck:
- [x] vendor https://github.com/ActiveState/appdirs/blob/master/appdirs.py
- [x] prefer `appdirs.user_config_dir('visidata') / config.py` over `~/.visidatarc`
- [x] use `appdirs.user_cache_dir('visidata')` 
- [ ] unit tests
- [ ] update docs

### DIFF
--- a/visidata/settings.py
+++ b/visidata/settings.py
@@ -7,6 +7,7 @@ import os
 
 import visidata
 from visidata import VisiData, BaseSheet, vd, AttrDict
+from visidata.vendor.appdirs import user_config_dir
 
 
 # [settingname] -> { objname(Sheet-instance/Sheet-type/'global'/'default'): Option/Command/longname }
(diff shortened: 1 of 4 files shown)


anjakefala:
(Not a reviewer, just helping with why the tests are failing.)

Do you have the latest `develop`? https://github.com/saulpw/visidata/commit/e7f096909cf411d9dc76cbd68cabfa4cb0df8367

I believe the tests are failing because you are missing this commit. Can you update your branch?

saulpw:
@jck We do need to keep compatibility with `~/.visidatarc` for config, but for the cache directory, we don't need to maintain compatibility with `~/.visidata/cache`.

visidata/settings.py
```diff
@@ -349,6 +350,18 @@ def addOptions(parser):
             pass
 
 
+def _get_config_file():
+    xdg_config_file = visidata.Path(user_config_dir('visidata')) / 'config.py'
```
anjakefala, line 354: Is `config.py` the standard way python XDG config files are named?

anjakefala:
I have tested this code. It does not seem to break the existing configuration functionality, macros, or plugins. I am not very comfortable with XDG, but it does seem to follow the environment variable spec. 

Unsure about the naming of the XDG config file as `config.py`. That should definitely get documented.

saulpw:
Thanks @jck, we much appreciate the contribution!

PR #1421, opened 2022-07-06
Fix curses mousemask for pypy3

LaPingvino:
curses mousemask is unsigned and using a signed negative number makes it crash for pypy3 (undefined behavior)


saulpw:
Thanks, @LaPingvino !  This change got made separately.

PR #1432, opened 2022-07-20
Prepare for v2.9

anjakefala:


CHANGELOG.md
```diff
@@ -1,5 +1,46 @@
 # VisiData version history
 
+# 2.9 (2022-04-XX)
+
+- add XDG support (thanks @jck for the PR #1420)
+    - `options.config` default is now `"$XDG_CONFIG_HOME"/visidata/config.py` if `$XDG_CONFIG_HOME` is set and `config.py` exists. If not, falls back to the standard `/home/.visidatarc`.
```
saulpw: change `/home` to `$HOME`

visidata/__init__.py
```diff
@@ -144,6 +145,10 @@ def getGlobals():
 import visidata.loaders.vds
 import visidata.loaders.odf
 import visidata.loaders.lsv
+import visidata.loaders.arrow
+import visidata.loaders.parquet
```
saulpw, line 149: also add `parquet` to formats list

visidata/_input.py
```diff
@@ -42,7 +42,11 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, tb):
         with suppress(curses.error):
             curses.curs_set(0)
-            curses.mousemask(-1)
+            if options.mouse_interval:
```
saulpw, line 45: this `-1` needs to be `0xffffffff`

visidata/deprecated.py
```diff
@@ -138,6 +139,49 @@ def load_tsv(fn):
     vs = open_tsv(Path(fn))
     yield from vs.iterload()
 
+class VisiDataSheet(visidata.IndexSheet):
```
saulpw: We need to get rid of this.

visidata/mainloop.py
```diff
@@ -295,11 +325,11 @@ def initCurses():
     colors.setup()
 
     curses.noecho()
-    curses.cbreak()
 
     curses.raw()    # get control keys instead of signals
     curses.meta(1)  # allow "8-bit chars"
-    curses.mousemask(-1 if options.mouse_interval else 0)
+    curses.MOUSE_ALL = 0xffffffff
```
saulpw, line 331: notice that we put this on the curses module for use later, so it's weird that it's not available in the other context handler.

saulpw (approved):
pending CHANGELOG

visidata/mainloop.py
anjakefala: test_commands.py does not go through `initCurses`!

saulpw: Then let's just use `0xffffffff` as an inline constant in both places with the comment `# curses.MOUSE_ALL`.

anjakefala:
We are going to move to a release branch!

PR #1447, opened 2022-07-29
Add HTTP request argument option

### LINKED ISSUE #1446: Enable setting HTTP User-Agent header
I'm trying to explore this API, but it has aggressive 'security' controls which block the default requests user agent.
This doesn't make much sense for an API, but I can work around it in requests by using the Firefox User-Agent. However, I'd really like to do this in visidata so I can hit the endpoint directly, rather than downloading the JSON, then loading it with visidata.

This is what I'm doing in requests at the moment:

``` python
import requests as rq
headers = {'User-Agent': 'Mozilla/5.0 (Windows NT x.y; Win64; x64; rv:10.0) Gecko/20100101 Firefox/10.0'}
r = rq.get('https://myhospitalsapi.aihw.gov.au/api/v1/datasets/8980', headers=headers)
r.json()
```

But I'm not sure how to get visidata to do that when I try to load a URL.

Not sure if it makes more sense to add this as a CLI flag or a config file setting?
I'm guessing one would start by adding a headers option here?
https://github.com/saulpw/visidata/blob/5e476e1bce454373f1253697dc9a4460839fd77d/visidata/loaders/http.py#L11

daviewales:
Add option to add header argument to http requests.

Fixes #1446

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # VisiData version history
 
+# v2.9.1 (2022-07-21)
+
+- [unzip-http] move urllib3 to optional dependencies
+
 # 2.9 (2022-07-20)
 
 - [ux] add confirming modal dialog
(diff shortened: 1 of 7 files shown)


daviewales:
The changeset is a mess because it's based on v2.9.1. I'll rebase on develop when [this](https://github.com/saulpw/visidata/issues/1446#issuecomment-1199537564) is fixed.

This is the actual change:
https://github.com/saulpw/visidata/pull/1447/commits/f6d78b2e80aaacb6bfc9e0125c21ea35d059e5d9

daviewales:
I've explicitly added an option for `http_req_headers`, so it appears in the list of global options. Should I add options for the other optional arguments for `requests.get`?

saulpw:
Nah, I think that one is just fine.  We do it in a generic way so that if someone comes along and needs to pass insome kwarg, that capability is there--it flows through to the underlying requests call.  If other kwargs prove to be commonly useful then we can add them later.  Thanks for putting this together!

saulpw:
Thanks for putting this together, @daviewales !  I made the changes myself and have pushed them up.  Please let me know if this works for your case.

PR #1472, opened 2022-08-11
[print-] fix usage of original builtins.print

anjakefala:
needed since bdf2cc58168e9611366b7d01223e87164c19108d

Fixes usage of visidata in batch mode. Essentially this line: https://github.com/saulpw/visidata/blob/02f1b80c5ea9d27d019d4cc07d0bf7c71765f5c7/visidata/main.py#L222 on `develop` results in a **RecursiveException** when you run the tests.

### DIFF
--- a/visidata/loaders/graphviz.py
+++ b/visidata/loaders/graphviz.py
@@ -20,7 +20,8 @@ def save_dot(vd, p, vs):
     srccol = vs.keyCols[0]
     dstcol = vs.keyCols[1]
     with p.open_text(mode='w', encoding='utf-8') as fp:
-        print('graph { concentrate=true;', file=fp)
+        pfp = lambda *args: vd.printout(*args, file=fp)
+        pfp('graph { concentrate=true;')
         for row in Progress(vs.rows, 'saving'):
             src = srccol.getTypedValue(row)
             dst = dstcol.getTypedValue(row)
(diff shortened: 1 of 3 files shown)


visidata/loaders/graphviz.py
```diff
@@ -20,7 +20,8 @@ def save_dot(vd, p, vs):
     srccol = vs.keyCols[0]
     dstcol = vs.keyCols[1]
     with p.open_text(mode='w', encoding='utf-8') as fp:
-        print('graph { concentrate=true;', file=fp)
+        print_wrap = print.__wrapped__
```
saulpw: Let's make this `p = lambda *args: print.__wrapped__(*args, file=fp)` and then adjust the following code accordingly.

visidata/main.py
```diff
@@ -103,7 +104,7 @@ def main_vd():
         vd.warning(e)
 
     warnings.showwarning = vd.warning
-    builtins.print = lambda *args, **kwargs: vd.status(*args)
+    builtins.print = functools.wraps(builtins.print)(lambda *args, **kwargs: vd.status(*args))
```
saulpw, line 107: Let's add functions `vd.printout` and `vd.printerr` that do this (or `print_stdout` and `print_stderr`) and then use those in preference to `print.__wrapped__` in the actual code.

visidata/loaders/graphviz.py
anjakefala: I am going to use `pfp` instead of `p` because `p` is the **Path** variable up-top!

saulpw (approved):
Thanks for noticing this and doing the PR, @anjakefala !

PR #1473, opened 2022-08-11
[menu-] upon menu keypress move to item

### LINKED ISSUE #1470: Confusing model and keystroke consumption in menu bar
From [a discussion](https://github.com/saulpw/visidata/discussions/1464#discussioncomment-3368730):

I would think that once the menubar is active (ctrl+h) I can press those letters to go to their sub-menus.

Additionally, if I hit alt+c for the column, and then change my mind and want to go to the row menu:

1. `r` does nothing;
2. `alt+r` calls a `row key by regex` command; I suspect the menu ate the `alt` key and escaped the menu and then executed `r`.

This is confusing because is the menu system eating characters or not?

anjakefala:
Closes #1470

Though, an important part of closing #1470  is that it was confusing for the user in the behaviour of pressing `Alt+` (https://github.com/saulpw/visidata/blob/02f1b80c5ea9d27d019d4cc07d0bf7c71765f5c7/visidata/menu.py#L741), since outside of the menu e.g. `Alt+f` is a way of navigating to the menu. 

I did not want to remove `^[` from the line in list at L741 if it was there for a reason, but my instinct is to remove it.

### DIFF
--- a/visidata/menu.py
+++ b/visidata/menu.py
@@ -776,6 +776,10 @@ def runMenu(vd):
             else:
                 break
 
+        elif k in main_menu.keys():
+            sheet.pressMenu(main_menu[k])
+
+
         sheet.checkMenu()
 
     finally:
@@ -786,6 +790,7 @@ def runMenu(vd):
     vd.draw_all()
     sheet.execCommand(currentItem.longname)
 
+main_menu = {'f': 'File', 'e': 'Edit', 'v': 'View', 'c': 'Column', 'r': 'Row', 'd': 'Data', 'p': 'Plot', 's': 'System', 'h': 'Help'}
 
 BaseSheet.addCommand('^[f', 'menu-file', 'pressMenu("File")', '')
 BaseSheet.addCommand('^[e', 'menu-edit', 'pressMenu("Edit")', '')


visidata/menu.py
```diff
@@ -786,6 +790,7 @@ def runMenu(vd):
     vd.draw_all()
     sheet.execCommand(currentItem.longname)
 
+main_menu = {'f': 'File', 'e': 'Edit', 'v': 'View', 'c': 'Column', 'r': 'Row', 'd': 'Data', 'p': 'Plot', 's': 'System', 'h': 'Help'}
```
saulpw, line 793: Hm, this can work for now, but maybe we should have a more programmatic way of generating this list, esp since plugins can add toplevel menu items.

saulpw (approved):

saulpw:
`^[` is ESC, so removing that means ESC by itself would not close the menu.  That seems wrong to me.  Though note that we don't use ESC in the main program because it can be confused about Alt+ keystrokes for the same reason.  So maybe it shouldn't?

anjakefala:
> ^[ is ESC, so removing that means ESC by itself would not close the menu. That seems wrong to me. Though note that we don't use ESC in the main program because it can be confused about Alt+ keystrokes for the same reason. So maybe it shouldn't?

Oh, that is confusing. =( That explains it.

The current behaviour is if you use `Alt+` while in the menu, the next keystroke is in the **BaseSheet** context. So `Alt+f` will navigate you to `File` when you are outside of the menu, and while in the menu `Alt+f` will activate `fill-cols`. Which feels worse to me than `ESC` not closing the menu, but i also never use `ESC` in VisiData.

saulpw:
Yeah, I agree, that is worse behavior.  Ideally we would parse the Alt+f as a single keystroke in every case (`getkeystroke()` should do this but doesn't), and then we could know when ESC was pressed by itself.  In the meantime we could print a message or flash the screen or something in response to ESC to indicate that you can't exit the menu that way.  I guess the 'best' way to exit the menu is `q`, which is at least consistent with the rest of VisiData, but feels a bit funny and maybe should be called out explicitly in the interface.

saulpw:
I made it so that one ESC gets eaten, and two ESC in a row exits the menu.

PR #1496, opened 2022-08-27
chore: add .direnv to .gitignore

cpcloud:


saulpw (approved):

anjakefala:
How do we feel about removing the `.envrc`? Since I started using `direnv` with my own `.envrc` it being there has added a bit of friction to my development process.

saulpw:
Fine with me, I don't use it myself.

PR #1510, opened 2022-09-07
Update usd plugin api 2022 09

hanfried:
### Update data.fixer.io using apilayer with requests

The original currency conversion API data.fixer.io
redirects already to apilayer.com and I couldn't find
a way to create API keys for the former data.fixer.io API,
so I updated the url and passing of parameters.
Especially the api key is passed now as header instead as a url param
what's certainly reasonable for security.

Another problem was that apilayer blocks urllib requests as bot requests
(for at least for the free account) and following
https://www.pythonpool.com/urllib-error-httperror-http-error-403-forbidden/
I changed the method to requests library as also this is officially
supported by apilayer as can be seen in:
https://apilayer.com/marketplace/fixer-api?live_demo=show

It would be probably a good idea to change to requests in general,
but I'm not sure how to test it accordingly, that's why a parameter to
the urlcache function was added, so you can switch between urllib and
requests.

### Allow currency caching days to be configurable

Found it convenient and don't see any reason why not or why it might disturb a user 
(unless of course a clean, small options list is preferred when pressing Shift-O)

### Add requests library to requirements

Added them manually to setup.py and moved it also manually out of optional part in requirements.txt

If these files are created by a tool like pip-tools or similar, I ignored it, but would be fine to use such one, too.
Disclaimer: I did not test the installation process.

### Unit tests

I didn't wrote or changed any unit tests here (should I?),
for it least the current unit tests are running, but show some deprecation warnings who should not come from this PR:

```
➜  visidata git:(update_usd_plugin_api_2022_09) pytest                  
=============================================================================================== test session starts ===============================================================================================
platform linux -- Python 3.8.10, pytest-5.4.2, py-1.8.1, pluggy-0.13.1
rootdir: /home/janek/Repositories/visidata
collected 21 items                                                                                                                                                                                                

visidata/tests/test_commands.py .                                                                                                                                                                           [  4%]
visidata/tests/test_edittext.py ...................                                                                                                                                                         [ 95%]
visidata/tests/test_path.py .                                                                                                                                                                               [100%]

================================================================================================ warnings summary =================================================================================================
visidata/tests/test_commands.py: 396 tests with warnings
  /usr/lib/python3/dist-packages/pkg_resources/__init__.py:1145: DeprecationWarning: Use of .. or absolute path in a resource path is not allowed and will raise exceptions in a future release.
    return get_provider(package_or_requirement).get_resource_filename(

-- Docs: https://docs.pytest.org/en/latest/warnings.html
======================================================================================== 21 passed, 396 warnings in 20.47s ========================================================================================
```
### Support for former data.fixer.io API

I think, technical it still would work, but new API keys can't really created and I guess, this will be deprecated anyway (but don't have hard information about this assumption).
Not sure, whether both old and new API should be supported, this PR only supports the new one so far.
There is enough legacy in a corporate, so I didn't it to write it backward compatible only for the USD plugin, unless someone thinks, we really should.
But the _urlcache is kept backwards compatible to be sure not breaking anything that relies on urllib implementation.


saulpw (changes requested):
Hi @hanfried, thanks for putting this together.  Can we send the proper user-agent in `headers` instead of using requests for this?  I don't want to make requests a required dependency (even though we have recently decided to install it by default in most distributions).  This should limit the changes to just the `plugins/usd.py` file, and merging that should be no problem.

hanfried:
@saulpw Just adding proper user-agents didn't worked (that was even my first attempt before switching to requests library), but the solution is to add more headers instead following https://stackoverflow.com/questions/13303449/urllib2-httperror-http-error-403-forbidden (it was easier to stackoverflow than to reverse engineer the requests library code :-))

Also removes the need for any special casing, so just minor changes now to the plugin and nowhere else, what's probably a good idea for any stability concerns :-)

OT: Anyway, this has underlined IMHO why I always use requests whenever possible (so close to always unless I'm in corporate setting with an old python2 running and no possibility to install something). But I agree, it's a good idea to reduce dependencies as much as possible to make the usage (and probably packaging, too) easier.

saulpw:
Thanks for making the changes, @hanfried!  I agree, I use requests whenever possible myself.  We may eventually do the same for VisiData, but not right now.

anjakefala:
Hi @hanfried!

I am going to be merging `plugins/usd.py` into core VisiData, and I had a few questions.

Are you still using this plugin? What is the value that you set for `options.fixer_currency_cache_days`?

I am currently getting `403 Forbidden`, but going to go through all of the documentation you linked to see if I can update it, a second time. If you have been using it, have you been running into those? Do you know why it might be getting blocked again?

hanfried:
Hi @anjakefala!

Yes, for me, it still seems to work, just tried a minimal example:
image
And looking to some other references, the exchange rate seems to be correct.

At the moment I have set `options.fixer_currency_cache_days = 0` (so deactivating the cache).

Note: any old API key from fixer.data.io is invalid.
As written in the beginning of the PR:

> The original currency conversion API data.fixer.io
> redirects already to apilayer.com and I couldn't find
> a way to create API keys for the former data.fixer.io API

So, I had create a key from apilayer.com - but I still can use the same key as when I created this PR.

And I think, still using a deprecated API key would also explain the 403 Forbidden message :-)

I personal hate it if I get an error without an explication what to do,
so maybe we should add a little hint in visidata to explain that you might need a new key and also from apilayer.com
Also, the description for the API key is confusing as it still explicitly directs to fixer.io instead of apilayer.com (I didn't changed it in this PR): https://github.com/saulpw/visidata/pull/1510/files#diff-cc53d282149685c1809177faad79d5992271aa66be0731494ca84795b69aaca6R9 
So, I have the feeling we should change that, too.

Do you know what's the way to send an error message from a plugin to visidata in case?
I could prepare another PR (even though it might take a week as I'm a bit time constrained due to a bank project as income source at the end of a quarter :-o).

anjakefala:
Thank you so much @hanfried!

I am going to try to update it based on your feedback, and I will link you to where the code got moved. =) Then in a week, if you want to make more improvements, you are more than welcome to!

> Do you know what's the way to send an error message from a plugin to visidata in case?

Yeah! I would use a `vd.error('error message')`.

anjakefala:
Oh, nevermind. There is not a good way to write a customised `vd.error` (unless @saulpw knows how to get around this). 

Because `USD` might be used in an `expr-col`, it will get re-calculated at some frequency. This means any `vd.error` we put will get posted to the status-line as long as that **ExpressionColumn** exists. 

Maybe I should create a custom Exception that I then wrap into a `TypedExceptionWrapper`?

anjakefala:
Hi @hanfried!

We are in the process of moving our custom plugins into Core VisiData.

I wanted to update you that the code is now here on the `develop` branch: https://github.com/saulpw/visidata/blob/develop/visidata/features/currency_to_usd.py. It will be imported by default into VisiData. The option `fixer_currency_cache_days` is now `fixer_cache_days`. =) Everything else should be pretty similar.

If you want to make any additional updates or changes, that is the place to do it!

PR #1517, opened 2022-09-09
Unguard sheet command

hanfried:
### unguard-sheet command

IMHO a guarding of a sheet should be reverseable,
so added this simple command.

Not sure if we should show a status,
when guarded it's obviously we show,
but I decided against showing a status,
as unguarded is the normal behaviour and see no reason to differentiate
in the status between never guarded or once guarded and then unguarded.

### Menu option

There was a File > Guard menu item so far.
Didn't want to add a 2nd File > Unguard item for a probably very seldom
used functionality.

So changed it to an on/off selection as

    File > Guard > on
                   off

what I think is still convenient and very self explainary.

### Documentation

I didn't find any mentioning of guard-sheet in docs folder, so I didn't update anything in docs folder about unguard-sheet, too.

In the intro tutorial https://jsvine.github.io/intro-to-visidata/basics/understanding-sheets/ it is mentioned and I've already prepared a fork+branch with a short mentioning of unguard-sheet in case https://github.com/jsvine/intro-to-visidata/compare/master...hanfried:intro-to-visidata:unguard_sheet_command and I would open a PR with it once this one is merged with the remark that it should be only merged after the next release.

### DIFF
--- a/visidata/menu.py
+++ b/visidata/menu.py
@@ -116,7 +116,10 @@ vd.menus = [
         Menu('New', 'open-new'),
         Menu('Open file/url', 'open-file'),
         Menu('Rename', 'rename-sheet'),
-        Menu('Guard', 'guard-sheet'),
+        Menu('Guard', 
+            Menu('on', 'guard-sheet'),
+            Menu('off', 'unguard-sheet')
+        ),
         Menu('Duplicate',
             Menu('selected rows by ref', 'dup-selected'),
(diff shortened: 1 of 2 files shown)


saulpw (approved):

anjakefala:
Thanks @hanfried! This is a great addition. We renamed it to `guard-sheet-off` to go with our general longname structure. =)

PR #1518, opened 2022-09-10
Fix clipboard on Windows

### LINKED ISSUE #1431: Unable to copy to Windows clipboard
**Small description**

Trying to copy rows to system clipboard results in a permission error.

**Expected result**

Should be able to copy data to system clipboard.

image

image

Select rows with `gs`, copy with `gY`.

**Additional context**

VisiData v2.8

daviewales:
Windows doesn't allow files to be opened for reading if they are already open. Previously, we created and opened a temporary file, then opened it two more times while it was already open.

However, NamedTemporaryFile creates AND opens a file, so now we open it with the read-write mode 'w+' and use seek(0) so we can reuse a single open file handle for everything.

Fixes #1431 

Note: This is a conservative change which still creates a temporary file. If I get time, I'll propose a more experimental version which skips the temporary file completely, and writes directly to stdin.

**EDIT**: Experimental alternative version which pipes directly to stdin for `syscopyValue` (but still uses a file for `syscopyCells`) is #1519.

### DIFF
--- a/visidata/clipboard.py
+++ b/visidata/clipboard.py
@@ -73,16 +73,13 @@ def syscopyCells_async(sheet, cols, rows, filetype):
     vd.status(f'copying {vs.nRows} {vs.rowtype} to system clipboard as {filetype}')
 
     # use NTF to generate filename and delete file on context exit
-    with tempfile.NamedTemporaryFile(suffix='.'+filetype) as temp:
-        # Windows won't open a file which is already open.
-        # I'd prefer to reuse the open file handle like we do for syscopyValue
-        # above, but vd.SaveSheets expects to open a file from a file path, and
-        # won't accept an open file handle without refactoring.
-        temp.close()
-        vd.sync(vd.saveSheets(Path(temp.name), vs))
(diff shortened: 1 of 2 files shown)


visidata/clipboard.py
```diff
@@ -74,6 +74,11 @@ def syscopyCells_async(sheet, cols, rows, filetype):
 
     # use NTF to generate filename and delete file on context exit
     with tempfile.NamedTemporaryFile(suffix='.'+filetype) as temp:
+        # Windows won't open a file which is already open.
+        # I'd prefer to reuse the open file handle like we do for syscopyValue
+        # above, but vd.SaveSheets expects to open a file from a file path, and
```
saulpw: visidata.Path can take an `fptext` argument which might/should work for this case.

daviewales: Oh fun! I'll see if I can get that working.

daviewales: I had to add a `write` method to `RepeatFile`, but it appears to work now with `fptext`. I've only tested it on Mac though.

daviewales:
I _think_ the failing tests are not from these changes. I get the same error on `develop`.

PR #1519, opened 2022-09-10
Fix clipboard on Windows (pipe to clip cmd stdin)

### LINKED ISSUE #1431: Unable to copy to Windows clipboard

daviewales:
This is a more experimental version of #1518 which tries to pipe data directly to the clipboard command's stdin.

Windows doesn't allow files to be opened for reading if they are already open.

For syscopyValues:
Previously, we created and opened a temporary file, then opened it two more times while it was already open.

Now we just pipe the data directly to the clipboard command's stdin.

For syscopyCells_async: (This is the same as #1518)
We can't reuse the open file handle here without refactoring vd.saveSheets, as it relies on being passed a file path with an extension to determine which saver to use. So, we take the pragmatic option, and just close the file as soon as we create it to make Windows happy, then open and close two more times, once to save the data, and once to read it.

Fixes #1431

### DIFF
--- a/visidata/clipboard.py
+++ b/visidata/clipboard.py
@@ -42,17 +42,13 @@ def copyCells(sheet, col, rows):
 
 @Sheet.api
 def syscopyValue(sheet, val):
-    # use NTF to generate filename and delete file on context exit
-    with tempfile.NamedTemporaryFile(suffix='.txt') as temp:
-        with open(temp.name, "w", encoding=sheet.options.encoding) as fp:
-            fp.write(val)
+    # pipe val to stdin of clipboard command
 
-        p = subprocess.Popen(
-            sheet.options.clipboard_copy_cmd.split(),
-            stdin=open(temp.name, 'r', encoding=sheet.options.encoding),
-            stdout=subprocess.DEVNULL,
-            close_fds=True)
-        p.communicate()
+    p = subprocess.run(
+        sheet.options.clipboard_copy_cmd.split(),
+        input=val,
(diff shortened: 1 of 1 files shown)


saulpw:
Hi @daviewales, thanks for putting these together!  I also prefer this stdin approach if it works just the same.  I'll try it out for a few days and make sure it works on Linux in my case.

daviewales:
Thanks @saulpw. Any thoughts on how we could pipe directly to stdin for the whole-sheet clipboard?
https://github.com/saulpw/visidata/pull/1519/files#diff-77b4076ed011ef457d45fdd06437022968e28468e93cae619016ed75f61f14adR72-R80

At the moment, it just reuses the code from `vd.saveSheets`.
I'm imagining that if we had a version of `saveSheets` which returned a generator we could iterate through it here, and send the output of the generator into `stdin` here, using `Popen.communicate`.

saulpw:
Hi @daviewales, I got it to work with io.StringIO, and squashed it into your commit and pushed them directly.  Thanks for doing the legwork and submitting the PRs, this will definitely be an improvement for Windows users.  Send me a note if you want some VisiData stickers and I'm happy to send you a few!

daviewales:
Thanks @saulpw. The StringIO version looks pretty neat!

PR #1515, opened 2022-09-08
Document `-d` option

abitrolly:
Fixes #1491

- [z] If contributing a core loader, [the loader checklist](https://visidata.org/docs/contributing#loader) was referenced.
- [z] If registering an external plugin, [the plugin checklist](https://visidata.org/docs/contributing#plugins) was referenced.

PR #1520, opened 2022-09-11
CI/CD check for man pages build

abitrolly:

Split from #1515.


anjakefala:
Thanks @abitrolly! It builds green, and will be a great addition to our testing pipeline. =)

PR #1522, opened 2022-09-12
Set short aliases

abitrolly:

* [ ] parse new syntax
* [ ] store short alias somewhere

### DIFF
--- a/visidata/loaders/tsv.py
+++ b/visidata/loaders/tsv.py
@@ -6,7 +6,7 @@ import collections
 from visidata import vd, asyncthread, options, Progress, ColumnItem, SequenceSheet, Sheet, FileExistsError, getType, VisiData
 from visidata import namedlist, filesize
 
-vd.option('delimiter', '\t', 'field delimiter to use for tsv/usv filetype', replay=True)
+vd.option('-d --delimiter', '\t', 'field delimiter to use for tsv/usv filetype', replay=True)
 vd.option('row_delimiter', '\n', 'row delimiter to use for tsv/usv filetype', replay=True)
 vd.option('tsv_safe_newline', '\u001e', 'replacement for newline character when saving to tsv', replay=True)
 vd.option('tsv_safe_tab', '\u001f', 'replacement for tab character when saving to tsv', replay=True)


abitrolly:
I am trying to trace where `vd.option` comes from, and the `AddGlobals` function is puzzling me.

https://github.com/saulpw/visidata/blob/4263c8396399f5cab3b44155d69921db0e563a5f/visidata/__init__.py#L166

Where does it inject the symbols? Into the module it is defined in (`visidata/__main__.py` which means `visidata`), or into the module where its object is defined (which is `visidata/vdobj.py`)?

anjakefala:
Hey @abitrolly!

You would probably be interested in the api docs: https://www.visidata.org/docs/api/options

Saul created an `Extensible` class that allows api expansions: https://www.visidata.org/docs/api/extensible. `@VisiData.api` decorators can then make a function referencable as `vd.foo`. Extensible allows us to remove entire `.py` files, the code implementing a feature is completely located in its `.py` file.

Can you add your [question](https://github.com/saulpw/visidata/pull/1515#issuecomment-1243483189) to your description? To make it clear that this PR is more of a discussion than an actual proposed change. 

We are currently going through a release, and it might take us a bit of time to get to your PRs that require more presence.

abitrolly:
@anjakefala API docs don't help much. I need to find a place to add parsing logic and store short option, and options seem to be injected by `AddGlobals`.

saulpw:
Hi @abitrolly, the aliases are set with `optalias()`: https://github.com/saulpw/visidata/blob/develop/visidata/main.py#L85

I think it makes more sense without the `--`:

```
vd.option('d delimiter tsv_delimiter', '\t', 'delimiter for tsv files')
```

I'm not sure it's necessary, but if you really want to get this done, I'd see about moving `option_aliases` (and probably `optalias()` too) in main.py onto the vd object, and then adding any aliases from `vd.option()` into it.  

The `addGlobals()` function is defined in `__init__.py` and adds the given globals to the whole package's globals, so they will be available to exec strings and toplevel imports.

Closing the PR for now until there's more code to review.

abitrolly:
> I think it makes more sense without the `--`:

With `--` it is easier to find the options in codebase with grep, because it is a common `optparse` and `argparse` conventions for option strings.

saulpw:
> > I think it makes more sense without the `--`:
> 
> With `--` it is easier to find the options in codebase with grep, because it is a common `optparse` and `argparse` conventions for option strings.

I don't think we should change all options to use this syntax, and I don't think it accomplishes much to only use it for some/new/updated options.  So let's skip this for now.  Can we better document options so that people know to do e.g. `grep delimiter | grep vd.option` ?

abitrolly:
> Can we better document options so that people know to do e.g. `grep delimiter | grep vd.option` ?

It is possible to write code walkthrough that describes how options are handled in comparison with `argparse`, i.e. explaining why `argparse` is not enough.

saulpw:
> > Can we better document options so that people know to do e.g. `grep delimiter | grep vd.option` ?
> 
> It is possible to write code walkthrough that describes how options are handled in comparison with `argparse`, i.e. explaining why `argparse` is not enough.

Please read this page from the Options API docs: https://www.visidata.org/docs/api/options

Basically there are several ways options can be set besides the CLI, including from within a visidatarc config file, on subclasses, and from the Options Sheet in a running instance.  So simply using argparse is not sufficient.  Also VisiData supports things like the `+sheetname:col:row` syntax for the starting position, and sheet-specific options on the CLI (like `vd --encoding cp437 file1.csv --encoding utf-8 file2.csv` should apply the different encodings to the sheets following them).  There are a perhaps surprising number of subtle ways people expect/want CLI options parsing to work in VisiData and argparse is not powerful enough to handle them.

PR #1538, opened 2022-09-24
[csv] Suggest CSV dialect

SuRaMoN:
In many EU countries the CSV dialect is different. Quote from wiki:
`Semicolons are often used instead of commas in many European [locales](https://en.wikipedia.org/wiki/Locale_(computer_software)) in order to use the comma as the decimal separator and, possibly, the period as a decimal grouping character.`

This pull auto detects this, and suggest the user to reload with the correct delimiter

https://asciinema.org/a/MZvpalPPUY72KhF58EsVllqf8


saulpw:
Hey @SuRaMoN, thanks for putting this together.  I think there's probably a simpler way to go about doing it; maybe a `csv_sniff` option that, if set to True, just uses the dialect directly from the sniffer?

saulpw:
Okay @SuRaMoN, it's taken us awhile to get this framework in place, so thanks for your patience.

If you're still interested in making this happen, I think it would be pretty reasonable to do it now.  See the `guess_csv` function here: https://github.com/saulpw/visidata/commit/4743f92bb855cf931d896e65845c549ce6027e2f

Basically these `guess_` functions return a dictionary with:
- `filetype` being the filetype they detect (corresponding to the `open_<filetype>`)
- `_likelihood` being a number from 0-10, 10 being most likely and 0 meaning a last ditch effort if nothing else will take it.
- any other key/values are set as options on the sheet the open_ function returns

So if you want to move your code into this `guess_csv` function such that the appropriate options are set, that would be really useful, and then we can add some simple tests+data to make sure it works.

Thanks for bringing the European perspective on data formats to us.  I definitely owe you a foil contributor VisiData sticker once we get this merged!

anjakefala:
Hi @SuRaMoN!

I added @saulpw's guidance on adding to `guess_csv` into here: https://github.com/saulpw/visidata/blob/fd74d4c1fef44e4afb582a617d5c594365fe3d6a/docs/api/loaders.rst#L268

Since this would require a fresh PR, I am going to close this draft. A newly opened PR with your code moved into `guess_csv` would be very welcome.

PR #1545, opened 2022-09-30
[loadPlugins-] catch exception if plugin fails to load

anjakefala:


visidata/settings.py
```diff
@@ -381,9 +381,14 @@ def loadConfigAndPlugins(vd, args=AttrDict()):
             vd.warning('plugin autoload failed; see issue #1529')
 
         for ep in eps:
-            plug = ep.load()
-            sys.modules[f'visidata.plugins.{ep.name}'] = plug
-            vd.status(f'Plugin {ep.name} loaded')
+            try:
+                plug = ep.load()
+                sys.modules[f'visidata.plugins.{ep.name}'] = plug
+                vd.status(f'Plugin {ep.name} loaded')
```
saulpw: Let's convert this from `status()` to `debug()`.

saulpw (approved):

PR #1563, opened 2022-10-12
Write STATA files

raffaelemancuso:

### DIFF
--- a/visidata/loaders/_pandas.py
+++ b/visidata/loaders/_pandas.py
@@ -17,6 +17,40 @@ for ft in 'feather gbq orc pickle sas stata'.split():
     if not getattr(VisiData, funcname, None):
         setattr(VisiData, funcname, lambda vd,p,ft=ft: PandasSheet(p.name, source=p, filetype=ft))
 
+@VisiData.api
+@asyncthread
+def save_dta(vd, p, *sheets):
+    import pandas as pd
+    import numpy as np
+
+    # STATA is a one-sheet software
+    # Save only the first sheet
+    vs = sheets[0]
+
+    columns = [col.name for col in vs.visibleCols]
+    
+    # Get data types
+    types = list()
+    dispvals = next(vs.iterdispvals(format=True))
+    for col,_ in dispvals.items():
+        if col.type in [bool, int, float]:
+            types.append(col.type)
+        elif vd.isNumeric(col):
+            types.append(float)
+        else:
+            types.append(str)
+
+    # Populate numpy array
+    data = np.empty((vs.nRows, len(columns)), dtype=object)
+    for r_i, dispvals in enumerate(vs.iterdispvals(format=True)):
+        for c_i, v in enumerate(dispvals.values()):
+            data[r_i, c_i] = v
+
+    # Convert to pandas DataFrame and save
+    dtype = {col:t for col,t in zip(columns, types)}
+    df = pd.DataFrame(data, columns=columns)
+    df = df.astype(dtype)
+    df.to_stata(p, version=118, write_index=False)
 
 class DataFrameAdapter:
     def __init__(self, df):


raffaelemancuso:
It contains debug code (the `fh.write`). It's just to get feedback.

visidata/loaders/_pandas.py
```diff
@@ -17,6 +17,57 @@ def open_dta(vd, p):
     if not getattr(VisiData, funcname, None):
         setattr(VisiData, funcname, lambda vd,p,ft=ft: PandasSheet(p.name, source=p, filetype=ft))
 
+@VisiData.api
+def save_dta(vd, p, *sheets):
```
saulpw, line 22: Save functions should be decorated with `@asyncthread`, as they could take an indefinite amount of time to save a large sheet.

visidata/loaders/_pandas.py
```diff
@@ -17,6 +17,57 @@ def open_dta(vd, p):
...
+
+    #for vs in sheets:
+    vs = sheets[0]
+
+    columns = [col.name for col in vs.visibleCols]
+    fh.write(f"columns={columns}\n")
+    fh.flush()
+
+    types = list()
+    for r_i, dispvals in enumerate(vs.iterdispvals(format=True)):
```
saulpw: This loop unnecessarily goes over all items when all you want is the first item.  You could use a `break` instead of a `continue`, but really it should just be `dispvals = next(vs.iterdispvals())`.  If it were a passed-in iterator instead of `iterdispvals()` (which is restartable), then we'd want to reuse the first value instead of starting the loop over again below.

visidata/loaders/_pandas.py
```diff
@@ -17,6 +17,57 @@ def open_dta(vd, p):
...
+    fh.write(f"Dataframe created\n")
+
+    df = df.astype(dtype)
+    fh.write(f"Data types written")
+
+    df.to_stata(p, version=118, write_index=False)
+    fh.write("DataFrame saved\n")
+
+    fh.close()
+    vd.status(f'{p} save finished')
```
saulpw: Individual savers should omit a finishing status.  I'd rather put one in `saveSheets()` so it applies to all savers.

saulpw (changes requested):
Thanks for putting this together, @raffaem!  I have a few comments, and if you address those and remove the debug messages, I think we could test this and get it merged.

visidata/loaders/_pandas.py
raffaelemancuso: I need to compute the number of rows, and when the loop written this way ends `r_i` will contain the number of rows - 1 :) 

What's the right way to get the number of rows?

raffaelemancuso:
@saulpw just pinging you in case my comments to your review were not delivered

saulpw:
> @saulpw just pinging you in case my comments to your review were not delivered

Hi @raffaem, I don't see any comments or changes yet..

raffaelemancuso:
> > @saulpw just pinging you in case my comments to your review were not delivered
> 
> Hi @raffaem, I don't see any comments or changes yet..

All right I should have committed them.

visidata/loaders/_pandas.py
saulpw: I would use `vs.nRows`.

raffaelemancuso:
The problem is that loading and saving a .dta with this will change the datatype.

Consider the following.

Open STATA and run:
```
webuse nlsw88, clear
save "test.dta"
desc
```

Data types are:
```
-------------------------------------------------------------------------------------------------------------------------------
Variable      Storage   Display    Value
    name         type    format    label      Variable label
-------------------------------------------------------------------------------------------------------------------------------
idcode          int     %8.0g                 NLS ID
age             byte    %8.0g                 Age in current year
race            byte    %8.0g      racelbl    Race
married         byte    %8.0g      marlbl     Married
never_married   byte    %16.0g     nev_mar    Never married
grade           byte    %8.0g                 Current grade completed
collgrad        byte    %16.0g     gradlbl    College graduate
south           byte    %9.0g      southlbl   Lives in the south
smsa            byte    %9.0g      smsalbl    Lives in SMSA
c_city          byte    %16.0g     ccitylbl   Lives in a central city
industry        byte    %23.0g     indlbl     Industry
occupation      byte    %22.0g     occlbl     Occupation
union           byte    %8.0g      unionlbl   Union worker
wage            float   %9.0g                 Hourly wage
hours           byte    %8.0g                 Usual hours worked
ttl_exp         float   %9.0g                 Total work experience (years)
tenure          float   %9.0g                 Job tenure (years)
-------------------------------------------------------------------------------------------------------------------------------
```

Now load and re-save with VisiData.

Data types become:
```
-------------------------------------------------------------------------------------------------------------------------------
Variable      Storage   Display    Value
    name         type    format    label      Variable label
-------------------------------------------------------------------------------------------------------------------------------
idcode          long    %12.0g                
age             long    %12.0g                
race            str5    %5s                   
married         str7    %7s                   
never_married   str16   %16s                  
grade           double  %10.0g                
collgrad        str16   %16s                  
south           str9    %9s                   
smsa            str8    %8s                   
c_city          str16   %16s                  
industry        str23   %23s                  
occupation      str22   %22s                  
union           str8    %8s                   
wage            double  %10.0g                
hours           double  %10.0g                
ttl_exp         double  %10.0g                
tenure          double  %10.0g                
-------------------------------------------------------------------------------------------------------------------------------

```

saulpw:
Hi @raffaem, I've been resistant to merging this, because as you say, the types are not preserved if you load and then save a .dta file.  We need to put together a set of tests for savers so that we can verify some basic functionality (for instance, this saver can't save a VisiData `date` column).  Also the basic functionality of creating a DataFrame and calling `to_foo` on it could be abstracted to support all pandas savers in a generic fashion.  So I apologize for delaying on this PR so long, but I'm getting back to VisiData now and have put this on the list to prioritize.  (Happy new year :)

raffaelemancuso:
Hi and happy new year!

Don't worry. I don't know how to preserve data types.

You are right on both fronts.

Are they preserved on loading in the first place?

saulpw:
The Pandas types are mapped to VisiData types on loading.

saulpw (approved):

saulpw:
Okay, getting back to this, I think we can take this as-is for now, since it offers support for writing Stata .dta files, even if it's not complete.  Thanks again for your patience, @raffaem.  If you send me an email, I'd be happy to send you a VisiData contributor sticker!

PR #1584, opened 2022-11-05
Fix #1583

### LINKED ISSUE #1583: ZSH completion fails with invalid option definition
I'm using v2.10.2 installed via the [latest Arch package](https://archlinux.org/packages/community/any/visidata/). Visidata shell completion in my ZSH is currently broken.

```
$ vd <tab>
_arguments:comparguments:327: invalid option definition: --fmt_expand_list[format str to use for names of columns expanded from list (colname, index) (default: %s[%s])]:str
```

Freed-Wu:

### DIFF
--- a/dev/zsh-completion.py
+++ b/dev/zsh-completion.py
@@ -33,13 +33,14 @@ def generate_completion(opt):
         completion += ":_files"
     elif opt.name in ["plugins_url", "motd_url"]:
         completion += ":_urls"
-    helpstr = opt.helpstr.replace("[", "\\[").replace("]", "\\]")
+    helpstr = opt.helpstr
     selections = pat_select.findall(helpstr)
     if len(selections):
         completion += f":{selections[0].replace('/', ' ')}"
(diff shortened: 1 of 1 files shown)


dev/zsh-completion.py
```diff
@@ -33,13 +33,14 @@ def generate_completion(opt):
         completion += ":_files"
     elif opt.name in ["plugins_url", "motd_url"]:
         completion += ":_urls"
-    helpstr = opt.helpstr.replace("[", "\\[").replace("]", "\\]")
```
anjakefala, line 36: Thanks for pushing a fix so quickly! =)

Can you explain the change?

Freed-Wu: just escape `[]`
```zsh
    --fmt_expand_list'[format str to use for names of columns expanded from list (colname, index) (default: %s\[%s\])]:str' \
```

saulpw: I see, the option default value in line 42 has to be escaped also, and now one of our default values has `[`/`]` in it.

PR #1629, opened 2022-12-20
Store null as empty string in save_xlsx

### LINKED ISSUE #1626: save_xlsx: null values become the string "None"
**Small description**

Setting `--null-value` in either direction doesn't help, so I suspect it isn't just that `options.null_value` is set to `None`.

I found this during the batch conversion. There's code below.

**Expected result**

An empty string (or the `options.null_value`) is more reasonable than `None`, for this conversion. But I can't set an empty string, with `--null-value`.

In lieu of a screenshot, I have console output.

```console
> vd -f json -b --save-filetype=xlsx -o nones.xlsx <<< '[{"foo":"None","bar":null}]'
opening - as json
saving 1 sheets to nones.xlsx as xlsx
Pay attention.
nones.xlsx save finished
> vd -f xlsx -b --save-filetype=json -o - nones.xlsx +:-:: | jq
opening nones.xlsx as xlsx
Let your best be for your friend.
saving 1 sheets to - as json
[
  {
    "foo": "None",
    "bar": "None"
  }
]
```

<details>

<summary>Testing with `--null-value`</summary>

```
> vd -f json -b --save-filetype=xlsx --cmdlog-histfile=vd.log --null-value "None" -o nones.xlsx <<< '[{"foo":"None","bar":null}]'
opening - as json
saving 1 sheets to nones.xlsx as xlsx
Stop this moment, I tell you!
nones.xlsx save finished
> vd -f xlsx -b --save-filetype=json --null-value "" -o - nones.xlsx +:-:: | jq
opening nones.xlsx as xlsx
Listen.
saving 1 sheets to - as json
[
  {
    "foo": "None",
    "bar": "None"
  }
]
> vd -f xlsx -b --save-filetype=json --null-value "None" -o - nones.xlsx +:-:: | jq
opening nones.xlsx as xlsx
Was I the same when I got up this morning?
saving 1 sheets to - as json
[
  {
    "foo": "None",
    "bar": "None"
  }
]
> vd -f json -b --save-filetype=xlsx --cmdlog-histfile=vd.log --null-value "" -o nones.xlsx <<< '[{"foo":"None","bar":null}]'
opening - as json
saving 1 sheets to nones.xlsx as xlsx
Listen.
nones.xlsx save finished
> vd -f xlsx -b --save-filetype=json --null-value "" -o - nones.xlsx +:-:: | jq
opening nones.xlsx as xlsx
I wonder what they'll do next!
saving 1 sheets to - as json
[
  {
    "foo": "None",
    "bar": "None"
  }
]
> vd -f xlsx -b --save-filetype=json --null-value "None" -o - nones.xlsx +:-:: | jq
opening nones.xlsx as xlsx
What are you thinking of?
saving 1 sheets to - as json
[
  {
    "foo": "None",
    "bar": "None"
  }
]
```

</details>

This was all done within `--batch` mode (and setting `--cmdlog-histfile` resulted in no output).

**Additional context**

I'm pretty sure this is due to naive serialization of the python value.

```python
>>> f"{None}"
'None'
```

Version

```
saul.pw/VisiData v2.9.1
```

As it happens, I'm interested in extending the `save_xlsx` functionality to create Tables (there is support in `openpyxl`). If I get round to that sooner rather than later, I'll look to fix this first.

dbaynard:
~~Where neither the global nor the sheet `null_value` options is set,~~ nulls will default to `””`, i.e. the empty string.

Before this change, nulls in the input would result in `None` in the saved xlsx files.

Fixes #1626 

This is a breaking change; ~~however if somebody's workflow depends on `null` being written out as `None` they can preserve the current behaviour by setting `options.null_value` to `None`.~~

~~I tried to follow the coding conventions (`null_value`, like `cleaned_name` in the function above.)~~

I have not checked to see whether this issue affects `xls` files, as I don't write those.

~~(Happy to sign the CAA.)~~

### DIFF
--- a/visidata/loaders/xlsx.py
+++ b/visidata/loaders/xlsx.py
@@ -138,7 +138,9 @@ def save_xlsx(vd, p, *sheets):
 
             row = []
             for col, v in dispvals.items():
-                if col.type == date:
+                if v is None:
+                    v = ""
+                elif col.type == date:
                     v = datetime.datetime.fromtimestamp(int(v.timestamp()))
                 elif not vd.isNumeric(col):
                     v = str(v)


visidata/loaders/xlsx.py
```diff
@@ -131,14 +131,18 @@ def save_xlsx(vd, p, *sheets):
             vd.warning(f'saving {vs.name} as {vs.xls_name}')
         ws = wb.create_sheet(title=vs.xls_name)
 
+        null_value = vs.options.get("null_value", vd.options.get("null_value", ""))
```
saulpw: You should just be able to use `vs.options.null_value`, which will fall back appropriately to the class value, global, etc.  Did it not work for you for some case?

saulpw:
Thanks for putting this together, @dbaynard !  I think this is reasonable, and ideally would apply to other savers in a more generic way.  But I'm happy to take this (pending comments about option usage) for xls to make that saver incrementally better.

visidata/loaders/xlsx.py
dbaynard: I didn't try that — though how do I then default to the empty string, if none of the values is set?

Edit: that works

dbaynard:
Happy to make incremental improvements.

I have only used a handful of formats, and `null` values are handled by coercion to the empty string. I'd assumed this was an oversight, on the part of the xlsx writer. I'm not so sure how many other writers are affected.

I'm not quite sure what to do about testing — I'm happy to add golden tests.

visidata/loaders/xlsx.py
saulpw: Ah I see, the default value of `null_value` is None.  `null_value` is actually for which values should be counted as null during aggregations, and overloading it this way would probably be confusing.  I'm thinking that we can just use `""` specifically in the xls saver for the None case, and if we want to make it configurable, it would be via a new `xlsx_save_null` option.  Though I'm not sure that's necessary at the moment.

It does leave open the question, if `null_value` is set, should that value should be given to the savers as None when it appears?

dbaynard: Right I'm new to visidata and there's only so much complexity one can take in at once — so thanks for your prompt feedback on this PR. I think, as result, there is a good path forward.

Given your comments, it seems visidata is not explicit in how it serializes `null` in formats that require strings. A format-agnostic `save_null` option seems reasonable for this — `None` values can be converted to the text of `save_null` when serialised.

The reverse is trickier: it is probably right not to convert values to `None` all the time, but it can occasionally be useful. This could be a separate feature request.

For this PR, how about just doing what you propose, and setting `None` values to the empty string?

dbaynard:
I've removed the optionality — there's now a null check, instead.

saulpw (approved):

saulpw:
Okay, this is definitely better for xlsx files.  Thanks for submitting it!  Will be in the next release.

PR #1633, opened 2022-12-24
Fix exception causes in utils.py

cool-RR:
More context for this change: https://blog.ram.rachum.com/post/621791438475296768/improving-python-exception-chaining-with


saulpw:
Thanks @cool-RR !  I have wondered how to do this, now I know (and hopefully won't forget next time :)

PR #1637, opened 2022-12-24
Fix exception causes, remove invalid exception

cool-RR:


cool-RR:
Look like the CI failures are not related to my change.

saulpw (approved):

visidata/sheets.py
```diff
@@ -139,12 +139,12 @@ def __getitem__(self, colid):
...
-            except (KeyError, AttributeError):
+            except (KeyError, AttributeError) as e:
                 if colid == 'sheet': return self.sheet
                 elif colid == 'row': c = self.row
                 elif colid == 'col': c = self.col
                 else:
-                    raise KeyError(colid)
+                    raise KeyError(colid) from e
 
         if not isinstance(c, Column):  # columns calc in the context of the row of the cell being calc'ed
```
saulpw, line 149: The build seems to be failing now, stemming from this: https://github.com/saulpw/visidata/actions/runs/3804896050/jobs/6472586534

Do you know why this would be causing a new issue?

cool-RR: I'm looking at the traceback and I can't figure out why this change would cause a problem. Are you sure that this problem wouldn't happen without the change? At least by my naive impression of the code, even without my change a `KeyError` would be raised.
If this is too much work to investigate, I'd be okay with reverting this PR.

PR #1644, opened 2022-12-30
[types] add @vd.numericType() decorator; fix customdate

saulpw:


anjakefala (approved):
The tests are passing, and I tried out `type-customdate` and `type-log10`. I could not find anything objectionable about this PR.

I bless it.

PR #1653, opened 2023-01-03
[sheets] sort all sheets on global Sheets Sheet

### LINKED ISSUE #1620: First row in the All Sheets Sheet is not sorted.
**Small description**
First row is not  sorted with [  or  ]  in columns of Sheets Sheet.

**Expected result**
All rows should be sorted in Sheets Sheet.

**Actual result with screenshot( sort in nRows columns)**

1. open several files.
2. press shift + S to open Sheets Sheet.
3. go to nRows or nCols or other numeric columns, and type [  or  ] to sort.

**Additional context**
Version 2.10.2

anjakefala:
Global Sheets Sheet only has precious sheets, so it is not self-referential.

This means the standard sheet sort applies to it, it does not need to skip the reference to itself like the standard Sheets Sheet does.

Closes #1620

### DIFF
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -1011,6 +1011,10 @@ class SheetsSheet(IndexSheet):
     def sort(self):
         self.rows[1:] = sorted(self.rows[1:], key=self.sortkey)
 
+class GlobalSheetsSheet(SheetsSheet):  #1620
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Thanks for fixing this!

PR #1657, opened 2023-01-07
Speed up fill for sheets with many empty cells

midichef:
During fills, the rows to fill are stored in a list that is searched many times. When there are many rows to fill, the search becomes very slow. Storing them in a dictionary makes the search much faster.

In my testing, filling 100,000 empty cells originally took 30 seconds. After the change, the fill takes less than a second.

### DIFF
--- a/visidata/fill.py
+++ b/visidata/fill.py
@@ -9,7 +9,7 @@ def fillNullValues(vd, col, rows):
     oldvals = [] # for undo
     isNull = col.sheet.isNullFunc()
     n = 0
-    rowsToFill = [id(r) for r in rows]
+    rowsToFill = { id(r):True for r in rows }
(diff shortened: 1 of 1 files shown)


visidata/fill.py
```diff
@@ -9,7 +9,7 @@ def fillNullValues(vd, col, rows):
     oldvals = [] # for undo
     isNull = col.sheet.isNullFunc()
     n = 0
-    rowsToFill = [id(r) for r in rows]
+    rowsToFill = { id(r):True for r in rows }
```
saulpw: I generally use `{id(r):r for r in rows}` but in this case a set would be even better: `set(id(r) for r in rows)`

saulpw:
Thanks, @midichef!

PR #1668, opened 2023-01-08
[xml] allow parsing huge data

midichef:
By default, lxml has built in limits that cause it to fail when parsing input that is large or deep. I want to use visidata to parse data that exceeds the limits.

lxml can turn off the limits with the option huge_tree.
https://lxml.de/parsing.html#parser-options

As I understand it, the limits are in place to prevent denial-of-service attacks when lxml is used for automated parsing of untrusted input. This safety measure isn't necessary in visidata, since most users of visidata are using it interactively. So I believe visidata should default to huge_tree=True. If not, it should have an option to control the huge_tree setting.

Here are some examples of how to trigger the limits.
Long attributes:
`python -c "n=10_000_000; long_str='_'*n; print(f'<xml><elem data=\"{long_str}\" /></xml>')" |vd -f xml`
> lxml.etree.XMLSyntaxError: internal error: Huge input lookup, line 1, column 10000002

Long element names:
`python -c "n=50_001; long_str='_'*n; print(f'<xml><{long_str} /></xml>')" |vd -f xml`
>lxml.etree.XMLSyntaxError: Name too long use XML_PARSE_HUGE option: NCName, line 1, column 50008

Deep trees:
`python -c "depth = 257; s_ = '<span>'; _s = '</span>'; print(f'<xml>{s_ * depth} + {_s* depth}</xml>')" |vd -f xml`
>lxml.etree.XMLSyntaxError: Excessive depth in document: 256 use XML_PARSE_HUGE option, line 1, column 1542

### DIFF
--- a/visidata/loaders/xml.py
+++ b/visidata/loaders/xml.py
@@ -46,7 +46,8 @@ class XmlSheet(Sheet):
     def iterload(self):
         if isinstance(self.source, Path):
             from lxml import etree, objectify
-            self.root = etree.parse(self.source.open_text(encoding=self.options.encoding))
+            p = etree.XMLParser(huge_tree=True)
+            self.root = etree.parse(self.source.open_text(encoding=self.options.encoding), parser=p)
             objectify.deannotate(self.root, cleanup_namespaces=True)
         else: #        elif isinstance(self.source, XmlElement):
             self.root = self.source
--- a/visidata/loaders/xml.py
+++ b/visidata/loaders/xml.py
@@ -1,5 +1,7 @@
 from visidata import VisiData, vd, Sheet, options, Column, Progress, setitem, ColumnAttr, vlen, RowColorizer, Path, copy
 
+vd.option('xml_huge_tree', True, 'allow very deep trees and very long text content')
+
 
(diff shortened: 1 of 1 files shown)


visidata/loaders/xml.py
```diff
@@ -46,7 +46,8 @@ def showColumnsBasedOnRow(self, row):
     def iterload(self):
         if isinstance(self.source, Path):
             from lxml import etree, objectify
-            self.root = etree.parse(self.source.open_text(encoding=self.options.encoding))
+            p = etree.XMLParser(huge_tree=True)
```
saulpw: The VisiData way to do this is:

```
p = etree.XMLParser(**self.options.getall('xml_'))
```

Then you can pass any of the possible kwarg options to XMLParser via standard VisiData options:

`options.xml_huge_tree = True`
`vd --xml-dtd-validation=True`

etc.

We might also want to make `huge_tree=True` the default, for the reasons you suggest, but also allow it to be re-set to False if someone didn't want it:

`vd.option('xml_huge_tree', True, 'explanation for what this option does')`

PR #1672, opened 2023-01-11
[sheets] fix NameError for mincolidx

midichef:

### DIFF
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -607,7 +607,7 @@ class TableSheet(BaseSheet):
                     break
                 mincolidx, maxcolidx = min(self._visibleColLayout.keys()), max(self._visibleColLayout.keys())
                 if self.cursorVisibleColIndex < mincolidx:
-                    self.leftVisibleColIndex -= max((self.cursorVisibleColIndex - mincolid)//2, 1)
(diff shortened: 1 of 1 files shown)


visidata/sheets.py
```diff
@@ -607,7 +607,7 @@ def checkCursor(self):
                     break
                 mincolidx, maxcolidx = min(self._visibleColLayout.keys()), max(self._visibleColLayout.keys())
                 if self.cursorVisibleColIndex < mincolidx:
-                    self.leftVisibleColIndex -= max((self.cursorVisibleColIndex - mincolid)//2, 1)
+                    self.leftVisibleColIndex -= max((self.cursorVisibleColIndex - mincolidx)//2, 1)
```
saulpw, line 610: Wow, this has been there since 2018!  (introduced in 10cbf882fad433a5ef8c313164b475ccec64e6ce).

I've apparently never run into this particular issue (and so must have never run this line of code).  How did you find it?  Is there a repro?

midichef: I was exploring a very artificial situation, replaying command logs that add column expressions that contain time.sleep(). I'll look for a way to consistently reproduce the triggering of this line.

anjakefala:
Thanks so much @midichef !

Issue #1670, opened 2023-01-10
Better filetype detection from Content-Type subtype when opening web pages


Oblomov:
Currently, the filetype detection for pages loaded directly from the web (via the `http` loader) is very simple: it uses the Content-Type subtype with some optional mapping from known subtypes to specific vd types (currently only for `tsv`). The detection could be made a bit smarter in presence of a structured syntax suffix, such as `image/svg+xml`, `application/rss+xml`, `application/activity+json` etc. In this case, it should be possible to use the suffix to determine the filetype in absence of additional specifications. Optionally, both filetypes cold be tried, in case the 'major' type has more specific support (e.g. if there's an `rss` plugin use that, otherwise fall back to `xml` for something that has a `application/rss+xml` content type)

anjakefala:
Proof of concept that will need to be integrated with guessing: https://github.com/saulpw/visidata/pull/1676

saulpw:
Okay, so as of e8bbba1ca89b56f4b7221281d92854b3ed6af5be we're passing the http response object into `guessurl_` functions, and I refactored the mimetype guessing into `guessurl_mimetype`.  So now you should be able to make your own `guessurl_better_mimetype` or something that takes the path and headers and whatever else into account.  Note though that ultimately it'll still have to make a single guess; we can't try opening it 4 different ways and seeing which one doesn't throw an exception, for example.  But at least you can guess based on which `open_` functions are available (see `guessurl_mimetype` for how it does this).

If you want to give your idea another try, please put together a few test cases, and get something working locally.  Then if it seems to work, put together a new PR, either modifying the existing `guessurl_mimetype`, or creating another `guessurl_` function.

PR #1676, opened 2023-01-13
Candidate resolution for #1670

Oblomov:
This is a possible implementation of the idea proposed in #1670. The change has been implemented in two commits.

The first introduces an extension to the `openPath` API function where filetype can be an array instead of a string. In this case on creation or load we verify that the `newfunc`/`openfunc` exists for each candidate filetype, and take the first one that exists.

The second commit uses this from the `http` loader to split structured subtypes, as proposed in #1670.  I have verified that this opens `application/*+knowntype` web document as `knowntype` for at least `xml` and `json`.

### DIFF
--- a/visidata/_open.py
+++ b/visidata/_open.py
@@ -38,6 +38,7 @@ def _completeFilename(val, state):
 @VisiData.api
 def openPath(vd, p, filetype=None, create=False):
     '''Call ``open_<filetype>(p)`` or ``openurl_<p.scheme>(p, filetype)``.  Return constructed but unloaded sheet of appropriate type.
+    filetype may be an array, in which case all types will be tested until a valid one is found.
     If True, *create* will return a new, blank **Sheet** if file does not exist.'''
     if p.scheme and not p.has_fp():
         schemes = p.scheme.split('+')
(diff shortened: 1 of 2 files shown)


saulpw:
Hey @Oblomov, thanks for suggesting this.  I've been giving it some thought.

I would love it if the changes could be localized to the http loader module.  Since `openurl_http` has to return a Sheet anyway, could it just call the `vd.open_` functions itself, instead of putting that same functionality in `vd.openPath` (which is a more core function)?

anjakefala:
Ping @Oblomov!

Oblomov:
(Sorry for the late reply, I've been overloaded with IRL commitments.)

The reason why I think it would be preferrable to have the “multiple candidate” solution in the core openPath function is that I believe other openers may be able to leverage it, again by using sideband metadata to help determine the file type (macOS resource forks may not be relevant anymore, but to give the idea). 

I understand that there is some overlap with the filetype guessing mechanism now. It's possible that the subtype split/parsing could be integrated better with it. However,  I still haven't had time to study how this works yet. (And I'm still somewhat busy, so if anybody wants to pick this up in the mean time I won't be offended, the key change concerning the content type processing is:

        subtype = contenttype.split(';')[0].split('/')[-1]
        subtypes = [ s[2:] if s.startswith('x-') else s for s in subtype.split('+')]


so if you find a good way to adapt this to the guessing mechanism, please do.)

anjakefala:
Since this is linked in issue #1670 as an existing proof-of-concept, and will need to be reworked to be integrated with the guessing mechanic, I am going to close this for now. When someone has time to pick up the gauntlet, we can reference it, and open a new PR.

PR #1709, opened 2023-01-30
Update http.py

er1kb:
Allow for an optional certificate file to be used when opening a http path. Tested with a .pem file, e.g. "vd https://localhost:1234 --certfile=/home/erik/certfile.pem".

### DIFF
--- a/visidata/loaders/http.py
+++ b/visidata/loaders/http.py
@@ -18,9 +18,10 @@ def openurl_http(vd, path, filetype=None):
             vd.fail(f'no vd.openhttp_{sch}')
         return openfunc(Path(schemes[-1]+'://'+path.given.split('://')[1]))
 
+    certfile = vd.options.get('certfile')
     import requests
 
-    response = requests.get(path.given, stream=True, **vd.options.getall('http_req_'))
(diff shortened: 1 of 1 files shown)


visidata/loaders/http.py
```diff
@@ -18,9 +18,10 @@ def openurl_http(vd, path, filetype=None):
             vd.fail(f'no vd.openhttp_{sch}')
         return openfunc(Path(schemes[-1]+'://'+path.given.split('://')[1]))
 
+    certfile = vd.options.get('certfile')
     import requests
 
-    response = requests.get(path.given, stream=True, **vd.options.getall('http_req_'))
+    response = requests.get(path.given, stream=True, **vd.options.getall('http_req_'), verify = certfile)
```
saulpw, line 24: So, that `getall()` means that you can specify `--http-req-verify=certfile.pem` without this change, and it will just work!  We may want to do add some clear documentation of the general mechanism, or an explicit option description, but we shouldn't need to manually add each option as its own kwarg like this.

er1kb: Great, thanks for explaining! I’m still learning :)

PR #1717, opened 2023-02-03
Enable filetype selection for open-syspaste

### LINKED ISSUE #1680: Create New Table from Clipboard
It would be nice to have a command which can create a new table directly from the clipboard.

For example, suppose I have tab separated data in my clipboard.
I would like to be able to execute a visidata command to load a table from the clipboard.
The command might ask me what filetype I am loading (tsv in this case).
Then it would load the data directly from the clipboard.

This would be particularly handy for quick file format conversions. e.g. if I copy a table from OneNote and want to convert it to a Markdown table, I could just import it into Visidata form the clipboard, and copy it back to the clipboard as a markdown table.

daviewales:
This pull request implements filetype selection for `open-syspaste`
The interface attempts to mirror `syscopy-selected` (`gShift+Y`)

Also adds `gShift+P` shortcut for `open-syspaste`.

Closes #1680.


features/open_syspaste.py
```diff
@@ -1,14 +1,19 @@
...
+
 from visidata import vd, BaseSheet, Path
 
 
 @BaseSheet.api
-def open_syspaste(sheet, filetype='tsv'):
+def open_syspaste(sheet, filetype=None):
     import io
 
+    filetype = filetype or vd.input("paste as filetype: ", value='tsv')
```
saulpw: Can we move this `input()` to the execstr?  The menu system looks for the string `input` and that's how it figures out that the command will require input.  (It's cosmetic only, but still how all commands are done).

daviewales: Should this be changed in `clipboard.py` too? (I copied this from there.):
Line 59:

```python
    filetype = filetype or vd.input("copy %d %s as filetype: " % (len(rows), sheet.rowtype), value=sheet.options.save_filetype or 'tsv')
```

saulpw: Ah, probably!  Sorry for the confusion.  Feel free to add to this PR in a separate commit, or I can fix it in a bit.

daviewales: I've pushed a new commit with this change for `open-syspaste`. I might leave the changes to `clipboard.py` for you, as `vd.input` in `syscopyValue()` refers to the sheet row type, and I'm not sure if that's known when the execstr is defined.

saulpw: Ah, okay.  I looked at it, and I'll leave it like it is for the time-being.  If we do an overhaul of the clipboard as mentioned in another thread, we can try to clean that up then.  Maybe we can figure out how to design away the input()s entirely.

saulpw:
I'm ready to merge this, but for some reason which I don't understand, there are merge conflicts.  @anjakefala can you look into resolving these?

Also, where in the menu system do you think this command should go?

I'm also wondering if we shouldn't have a lengthier helpstring that explains the command/feature more completely.  We've written very terse helpstrings in the past, but we're heading towards more verbose messages all around, to make it easier for people, and features like this might be a good place to start flexing in that direction.

anjakefala (approved):
Addressed the merge conflict!

daviewales:
Thanks all!

Issue #1726, opened 2023-02-07
Saving an empty table (heads only) as JSONL produces an empty file


daviewales:
**Small description**

Saving an empty table with headers but no data as JSONL produces an empty file.

**Expected result**

A JSONL file including the headers.
(Compare to the output when saving a TSV)

**Actual result with screenshot**
If you get an unexpected error, please include the full stack trace that you get with `Ctrl-E`.

**Steps to reproduce with sample data and a .vd**
[test.vdj.zip](https://github.com/saulpw/visidata/files/10670604/test.vdj.zip)

Create a table directly in Visidata and save it as `.jsonl`.
The output file is empty.

**Additional context**
Please include the version of VisiData.
v2.11

daviewales:
I'm guessing we need to grab the headers and save them first here?

https://github.com/saulpw/visidata/blob/472122901717ec912659ea09e6d43115a731ea07/visidata/loaders/json.py#L122

saulpw:
Hm, I'm not sure understand what you need.  Each line in a JSONL file is a row, so without any rows, what should the actual file contain?

daviewales:
Oh interesting! I was looking at the [JSONL example page](https://jsonlines.org/examples/), and assumed that Visidata's tabular data would be stored as per the JSONL example. e.g:

``` jsonl
["Name", "Session", "Score", "Completed"]
["Gilbert", "2013", 24, true]
["Alexa", "2013", 29, true]
["May", "2012B", 14, false]
["Deloise", "2012A", 19, true] 
```

It would make sense to be able to save the headers in this case.

But looking at the output from Visidata, I see that it uses rows of dicts, rather than lists.
Interestingly if I try to read the example JSONL data above with Visidata it just comes in as a single unnamed column:

![image](https://user-images.githubusercontent.com/2499066/217149576-aa13aa84-9a6b-47b7-a231-7e25cc0e89e7.png)

Is there a reason why Visidata uses the rows of dicts format rather than the rows of lists format?
Would it make sense to detect and support both styles for reading and writing?

saulpw:
Huh!  I guess I never looked at the examples page, or I completely forgot about it.  For my money, the rows of dicts format is what confers all the advantages of the JSONL format, like being able to add a field to a single row without having to add empty values to all other rows.  It would be reasonable to add support for loading the rows of lists format; it wouldn't be too hard to add a saver too.  Is this rows of lists format in active use?   How are you using JSONL in practice?

[...]

PR #1730, opened 2023-02-09
Improve JSONL handling

daviewales:
Work in progress for #1726


saulpw (approved):

daviewales:
I was thinking of bundling the following into this PR, but also happy if you want to merge this and I'll send another PR when the rest is ready:

- [x] Add warning when saving `jsonl` file with no rows
- [x] Add JSONL arrays loader
- [x] Add JSONL arrays saver
- [x] Add option to save JSONL as arrays
- [x] Add extended warning which hints to use JSONL arrays format to preserve headers
- [x] Add tests

saulpw:
We can sit on it for a bit while you're actively developing, see if we want to tweak the message or anything.

daviewales:
@saulpw, I think I'm ready for you to look at this now. I've made a few potentially controversial choices, so keen to hear your thoughts! A summary of my questionable choices follows:

- JSONL arrays get their own sheet type, inheriting from SequencSheet
- JSONL arrays get their own file extension (`.jsonla`)
- Loading as JSONL arrays automatically falls back to the default JSONL loader if the file doesn't include a JSONL arrays header
- I've split the `json_save_arrays` option in two: `json_save_arrays` and `json_load_arrays`. The first will always use the arrays format for `.jsonl` files, and the second will always attempt use the arrays format for `.jsonl` files, falling back to the default if the header is not detected.

From a user perspective, there is _no change_ if they save as `.jsonl` and have the new options unset.

I haven't added any tests yet, so I'll look into that.

daviewales:
**EDIT**: Sorry, I think I understand now...

~~Perhaps I don't understand how testing works.
But it looks to me that the existing `save-json.vd` test doesn't actually save anything?
https://github.com/saulpw/visidata/blob/develop/tests/save-json.vd~~

daviewales:
Hmm... Testing has revealed a bug when saving as `.jsonla`.

daviewales:
I think we're good now.

**Edit**: Helps to commit the sample data...

**Edit 2**: Removed some vestigial code and force pushed

tests/load-jsonla.vdj
```diff
@@ -0,0 +1,5 @@
+#!vd -p
+{"longname": "open-file", "input": "sample_data/officials.jsonla", "keystrokes": "o"}
+{"sheet": "officials", "col": "", "row": "", "longname": "columns-sheet", "input": "", "keystrokes": "Shift+C", "comment": "open Columns Sheet: edit column properties for current sheet"}
+{"sheet": "officials_columns", "col": "name", "row": "", "longname": "sort-asc", "input": "", "keystrokes": "[", "comment": "sort ascending by current column; replace any existing sort criteria"}
```
saulpw: Let's remove the column sorting; we should be checking that the columns come in, in the order of the array.

tests/save-jsonla.vdj
```diff
@@ -0,0 +1,3 @@
+#!vd -p
+{"longname": "open-file", "input": "sample_data/sample.tsv", "keystrokes": "o"}
```
saulpw, line 2: Let's type the columns at least as `int` and `float`, so we can make sure they're being saved as such since .jsonla supports numbers.

visidata/loaders/json.py
```diff
@@ -1,22 +1,64 @@
...
+    ["A", "B", "C"]
+    [1, "blue", true]
+    [2, "yellow", false]
+
+    The header array must be a flat array of strings
+
+    If no suitable header is found, fall back to generic JSON load.
+    '''
+
+    with p.open_text(encoding=JsonlaSheet.options.encoding) as fp:
```
saulpw: Use `vd.options` here; only use `<SheetClass>.options` for overriding (setting) defaults on SheetClass.

visidata/loaders/json.py
```diff
@@ -126,13 +168,49 @@ def write_jsonl(vs, fp):
...
+            else:
+                vd.status("Saving JSONL dicts")
+                vs.write_jsonl(fp)
+
+
+@VisiData.api
+def save_jsonla(vd, p, *vsheets):
+    with p.open_text(mode='w', encoding=vsheets[0].options.encoding) as fp:
+        for vs in vsheets:
+            vd.status("Saving JSONL arrays")
```
saulpw: Let's remove these extra `vd.status()` in this function

visidata/loaders/json.py
```diff
@@ -1,22 +1,64 @@
 import json
 
-from visidata import vd, date, VisiData, PyobjSheet, AttrDict, stacktrace, TypedExceptionWrapper, options, visidata, ColumnItem, wrapply, TypedWrapper, Progress, Sheet, InferColumnsSheet
+from visidata import vd, date, VisiData, PyobjSheet, AttrDict, stacktrace, TypedExceptionWrapper, options, visidata, ColumnItem, wrapply, TypedWrapper, Progress, Sheet, InferColumnsSheet, deduceType, SequenceSheet
 
 vd.option('json_indent', None, 'indent to use when saving json')
 vd.option('json_sort_keys', False, 'sort object keys when saving to json')
 vd.option('default_colname', '', 'column name to use for non-dict rows')
+vd.option('json_save_arrays', False, 'save table as rows of JSON arrays instead of JSONL dicts')
```
saulpw: We don't need these options, I think.  For saving, using the filetype `jsonla` should take care of it.  For loading, `open_jsonla` already checks the first line for an array type and falls back to `open_json` if it's not.  We'll make a better structure for guessing filetypes ultimately, but until then it should be workable without extra options.

saulpw (commented):
Thanks for pulling this together, @daviewales !  I had a few comments, and if we address those I think we should be able to merge it.

tests/save-jsonla.vdj
daviewales: Hmm... I just tried to play this command log and this happened:
Screen Shot 2023-02-15 at 9 35 14 pm

From what I can tell, it glitches when asking whether to overwrite an existing file.
Otherwise the replay works without issue.

visidata/loaders/json.py
daviewales: Removed.

visidata/loaders/json.py
daviewales: Done.

visidata/loaders/json.py
daviewales: Done.

tests/save-jsonla.vdj
daviewales: I've updated the test to type the columns. It seems to work in batch mode, just not in play mode.

tests/load-jsonla.vdj
daviewales: Done.

daviewales:
Thanks @saulpw, I've pushed some commits with the requested changes.

tests/load-jsonla.vdj
```diff
@@ -0,0 +1,3 @@
+#!vd -p
+{"longname": "open-file", "input": "sample_data/officials.jsonla", "keystrokes": "o"}
+{"sheet": "officials", "col": "", "row": "", "longname": "columns-sheet", "input": "", "keystrokes": "Shift+C", "comment": "open Columns Sheet: edit column properties for current sheet"}
```
saulpw, line 3: ah, here I meant also to not open the columns sheet :)  I'll fix this after merging.

Issue #1728, opened 2023-02-08
getMaxWidth() is slow for for columns of large collections


midichef:
**Small description**
For JSON files with long lists or dictionaries, visidata is slow to open, and `go-rightmost` is slow, because `getMaxWidth()` is slow.

`getMaxWidth()` has a few problems:
1) It can return widths larger than the current window.
2) When there are no rows, the header column width does not get +2 added.
3) It's slow when the column holds many lists or dictionaries that are written as a wide string.
To see this, run
`python -c 'nrows=1000;ncols=100;cell_len=300; cell = ["_"*cell_len];  data = [ { f"col{i}":cell for i in range(ncols) } for _ in range(nrows) ]; import json; print(json.dumps(data))' |vd -f json`
Visidata is slow to start, because it runs `getMaxWidth()` for the columns on screen.
Then execute `go-rightmost`, which runs `getMaxWidth()` once for every column that is offscreen.

I've got a chain of fixes that addresses these problems. https://github.com/midichef/visidata/compare/169d22a7c55f77691b37ac6b3b68e5dde6853a93...4fa24ed4154b7f6f43f01d4d3a4319cf95f6d771

For very wide columns of data, the running time is O(nrows\*ncols\*cell_len). With this fix it drops to O(1*ncols\*cell_len). Visidata becomes noticeably faster at opening JSON files with large dictionaries.

**Additional context**
saul.pw/VisiData v2.11

midichef:
The runtime could be lowered further to O(ncols), by capping the width in other places. Ultimately, the slowness ultimately comes from `clipstr()` being called with a width of 0. Perhaps `getCell()` or `formatValue()` could set a max width of `windowWidth`?

saulpw:
Hi @midichef, thanks for looking into this!  I think your changes are strictly better, so if you put together a PR, I'd be happy to merge it.

As to this:

> Ultimately, the slowness ultimately comes from clipstr() being called with a width of 0. Perhaps getCell() or formatValue() could set a max width of windowWidth?

That's a reasonable idea; however, those functions are also used for non-display purposes (e.g. saving), where we'd definitely want the entire value.  Is there a way to speed it up in the common case, without breaking the saving case?

anjakefala:
It seems like you addressed the main core of this issue @midichef! Please re-open if there is more work to be done. =)

PR #1747, opened 2023-02-20
speed up getMaxWidth() for wide columns, and correct some edge cases

midichef:
Addresses https://github.com/saulpw/visidata/issues/1728
The motivation is to speed up display of JSON files with columns that hold wide dictionaries.


saulpw (approved):
Thanks, @midichef!  The previous code was a mouthful, for sure :)

Issue #1772, opened 2023-02-28
save json files in utf-8


joaosousa1:
When i open and save a JSON file with utf-8 encode, the Visitata insists on saving the file to us-ascii encode.

I check the "default option values applying to every sheet"
The default encoding is utf-8

When I open the file in another editor, words like "Olá" result in ascii encode like "Ol\u00e1".  Visidata save the file in us-ascii encode.

I tested the "visidata.vd.options.set('encoding', 'utf-8', obj='global')" option in .visidatarc and have the same results. 

Visidata v2.11 (installed via pip3)
Python v3.10.6
Ubuntu 22.04 LTS
My system locale is LANG=pt_PT.UTF-8

obs: I tested the version v2.2.1 from ubuntu repo and have same results.

PR #1776, opened 2023-03-01
Add option to ensure_ascii

joaosousa1:
- [#1772 by @joaosousa1](https://github.com/saulpw/visidata/issues/1772)

Add option to non-ASCII characters be saved to the JSON . If set to False, JSON file will be encoded to utf-8.

### DIFF
--- a/visidata/loaders/json.py
+++ b/visidata/loaders/json.py
@@ -4,6 +4,7 @@ from visidata import vd, date, VisiData, PyobjSheet, AttrDict, stacktrace, Typed
 
 vd.option('json_indent', None, 'indent to use when saving json')
 vd.option('json_sort_keys', False, 'sort object keys when saving to json')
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Thanks, @joaosousa1 !  I had thought JSON was ASCII-only, I didn't realize this was even an option.

saulpw:
I'd also be willing to entertain arguments that the default value for this option in VisiData should be `False`.

anjakefala:
For now, I set the default to `True` which is the pre-existing behaviour!

joaosousa1:
@saulpw I'm happy to contribute.
@anjakefala you are right. The default value is ```True``` in `JSONEncoder` Class.
Now i have the option to set ```options.json_ensure_ascii = False``` in my .visidatarc. :smiley:

PR #1784, opened 2023-03-04
[regex] perform checks for capture group and if there are columns to add

### LINKED ISSUE #1778: [capture-col] When nothing is captured, "IndexError: list index out of range" is thrown
_Minor user exp bug_

**Small description**
When using `capture-col` on a column and the regex does not capture any values, the below error is thrown. No new column is created (as expected)

**Expected result**
Just to show the "no columns to add" error. Preferably, change the error to something more informative such as "no matches found, didn't add column"

```
  File "/Users/geekscrapy7/lib/python3.7/site-packages/visidata/threads.py", line 200, in _toplevelTryFunc
  File "/Users/geekscrapy7/lib/python3.7/site-packages/visidata/features/regex.py", line 77, in addRegexColumns
    vs.addColumnAtCursor(*cols.values())
  File "/Users/geekscrapy7/lib/python3.7/site-packages/visidata/sheets.py", line 473, in addColumnAtCursor
    firstnewcol = [c for c in cols if not c.hidden][0]
```
I presume this error is from an insufficient catch of an exception and it is attempting to create the column even though there are no groups to show.

`# echo -e 'helloooo\nthere' | vd -f txt --config=/dev/null --visidata_dir=/dev/null`

```tsv
-	text		capture-col	(ABC)	;	add new column from capture groups of regex; requires example row
```

**Additional context**
v2.12dev

anjakefala:
Closes #1778

### DIFF
--- a/visidata/features/regex.py
+++ b/visidata/features/regex.py
@@ -24,6 +24,8 @@ def makeRegexSplitter(vd, regex, origcol):
 
 @VisiData.api
 def makeRegexMatcher(vd, regex, origcol):
+    if not regex.groups:
+        vd.fail('specify a capture group')  #1778
     def _regexMatcher(row):
         m = regex.search(origcol.getDisplayValue(row))
         if m:
--- a/visidata/features/regex.py
+++ b/visidata/features/regex.py
@@ -74,6 +74,10 @@ def addRegexColumns(vs, regexMaker, origcol, regexstr):
         else:
             raise TypeError("addRegexColumns() expects a dict, list, or tuple from regexMaker, but got a "+type(m).__name__)
 
+    if not cols:
+        vd.warning("no regex matches found, didn't add column")
+        return
+
     vs.addColumnAtCursor(*cols.values())
 
 


visidata/features/regex.py
```diff
@@ -120,11 +124,15 @@ def regex_flags(sheet):
     'Return flags to pass to regex functions from options'
     return sum(getattr(re, f.upper()) for f in sheet.options.regex_flags)
 
+@BaseSheet.api
+def verify_regex_capture(sheet, regexstr):
```
saulpw: Let's move this check to the existing `makeRegexMatcher` (outside the wrapper).

visidata/features/regex.py
```diff
@@ -120,11 +124,15 @@ def regex_flags(sheet):
...
+    return bool(re.search('\(.*\)', regexstr))  #1778
```
saulpw: Doesn't this need to be `r'`?

visidata/features/regex.py
anjakefala: That ended up making it a lot easier and less hackier!

Issue #1765, opened 2023-02-27
Improve User Experience of installing loader dependencies


snooppr:
The readme informs the user about a simple installation:
`pip3 install visidata`

The package is installed normally.  But when using for example (site that contains tables):
`vd url`  
the utility starts and displays one default tabular line instead of a table, confusing the user (this is not an error, as if the site simply does not have tables, or another table or something else).

I had to check the vd on several url(s) to understand that the vd was not working.

Since I'm using vd on an Android gadget/termux, the zoom didn't allow me to see the full message: a message that the "lxml" package is missing.
After installing the package "lxml" vd started working.
`pip3 install lxml`

Why doesn't vd install the dependencies it needs by default?  (lxml).  
The end user is a user, not a developer, he can't figure out why vd doesn't work out of the box.

**Update vd dependencies, add lxml package.**

anjakefala:
Duplicate of https://github.com/saulpw/visidata/issues/1739

geekscrapy:
I would like to generally suggest (as this has caught me out multiple times) that all deps for _any_ core feature (esp. if it is advertised on the website) is added to the requirements.txt.

A new user will click lots of things, open lots of files (this is a new tool to them, they should go wild) and it's discouraging for them to try and open something and it not work (esp. if they aren't terminal users and this is their first foray into the transition into gui+terminal tools), this is exponential if multiple things fail as they try to challenge vd more and more (they might ask the question, as old timers still do: what **_can't_** it open). This is probably also true for long term users (inc. myself). And this "it can open loads of formats" is one of the core draws to vd.

I'd love to just be able to say to a new user who may want to open XML "yup, just install visidata, and you're good to go". At the moment we have a multi step process: install visidata, try and open XML, it fails, try figure out why (which may not be obvious for a new py/terminal/pip user), install lxml, try again.

We have deps auto installed with plugins, so this should be the standard. I install vd and things that are included should just work. I understand the reason for keeping it light, but I feel there's a fine line between keeping it light and setting expectations. This expectation is set extremely high when there are loads of file formats shown (in the drop down) on the visidata home page, most of which can only be used once the file formats "process of elimination" has had its way with you 😅. I feel this would turn people off quite fast as they will find a tool that would work out of the box.

_(I appreciate this next comment refers to the dev branch, which may not automatically be part of a release, but it provides another example)_
The latest thing that caught me out was the inclusion of the menu item "zulip" (which if I'm honest is confusing why it is included by default)

[...]

PR #1787, opened 2023-03-05
[import] add importExternal for most loaders and features

### LINKED ISSUE #1739: Feature request: Show nicer error messages instead of ModuleNotFoundError
As I'm working on the VisiData talk, I'm trying to open different file formats with VisiData. Some of these require modules such as `psycopg2` or `xlrd`. When you don't have these modules, you get a message like `ModuleNotFoundError: No module name 'xlrd'`.

It's usually easy for me to understand that I should just run `pip install xlrd`, but in some cases it isn't trivial, like `pdfminer` which actually requires `pdfminer.six`, and `png` which actually requires `pypng`. Also, VisiData should be usable for people who know nothing about Python.

Therefore, I think that we should have nice error messages like "Please run in the shell: pip install pdfminer.txt". If we'll ever want to go the extra mile, we could show a dialog that allows calling Pip automatically from within VisiData.

anjakefala:
In draft until I can get a passing build, and am sure that `vd` is imported everywhere where it is used.

 `importExternal(module_name, python_package_name)`
when they are both the same, `module_name` is sufficient.
It is possible some were missed. This was a first pass.

Closes #1739
Address #1765

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,8 +7,10 @@ requests            # http
 lxml                # html/xml
 openpyxl>= 2.4      # xlsx
 xlrd                # xls
+xlwt                # xls
 h5py                # hdf5
 psycopg2-binary     # postgres
+boto3               # rds
 pyshp               # shapefiles
 mapbox-vector-tile  # mbtiles
 pypng               # png
@@ -29,11 +31,12 @@ wcwidth             # tabulate saver with unicode
 zstandard           # read .zst files
 odfpy               # odt (OpenDocument)
 urllib3             # .zip over http
-pyarrow             # arrow arrows (Apache Arrow)
+pyarrow             # arrow arrows parquet (Apache Arrow)
 seaborn             # plot via seaborn
 matplotlib          # svg saver
 sh                  # ping
 psutil              # procmgr
+numpy               # npy pandas hdf5 arrow
 
 requests_cache      # scraper
 beautifulsoup4      # scraper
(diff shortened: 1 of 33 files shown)


requirements.txt
```diff
@@ -19,21 +21,20 @@ savReaderWriter     # sav (SPSS)
 PyYAML>=5.1         # yaml/yml
 dpkt                # pcap
 dnslib              # pcap
-namestand           # graphviz
 datapackage         # frictionless .json
 pdfminer.six        # pdf
 tabula              # pdf tables
 vobject             # vcf
 tabulate            # tabulate saver
-wcwidth             # tabulate saver with unicode
```
saulpw: Don't we need wcwidth?

visidata/loaders/xlsb.py
```diff
@@ -15,6 +15,7 @@ def open_xlsb(vd, p):
 
 class XlsbIndex(IndexSheet):
     def iterload(self):
+        vd.importExternal('pyxlsb', '-e git+https://github.com/saulpw/pyxlsb.git@visidata#egg=pyxlsb')
```
saulpw, line 18: I'd love to avoid this gnarliness (separate from this change).

saulpw (approved):

requirements.txt
anjakefala: I realised my mistake, and reintroduced!

PR #1793, opened 2023-03-08
[clipboard-] warn when pasting before copying

midichef:
Give a warning when `paste-after` is called without first performing a copy or cut. Otherwise the user sees this error:
```
  File "/home/midichef/.local/lib/python3.10/site-packages/visidata/basesheet.py", line 200, in execCommand
  File "/home/midichef/.local/lib/python3.10/site-packages/visidata/basesheet.py", line 73, in execCommand2
  File "paste-after", line 1, in <module>
    import functools
  File "/home/midichef/.local/lib/python3.10/site-packages/visidata/clipboard.py", line 132, in paste_after
    to_paste = list(deepcopy(r) for r in reversed(vd.memory.cliprows))
TypeError: 'NoneType' object is not reversible
```

### DIFF
--- a/visidata/clipboard.py
+++ b/visidata/clipboard.py
@@ -129,6 +129,9 @@ def delete_row(sheet, rowidx):
 
 @Sheet.api
 def paste_after(sheet, rowidx):
+    if not vd.memory.cliprows:  #1793
+        vd.warning('nothing to paste')
+        return
     to_paste = list(deepcopy(r) for r in reversed(vd.memory.cliprows))
     sheet.addRows(to_paste, index=rowidx)
 


visidata/clipboard.py
```diff
@@ -129,8 +129,11 @@ def delete_row(sheet, rowidx):
 
 @Sheet.api
 def paste_after(sheet, rowidx):
-    to_paste = list(deepcopy(r) for r in reversed(vd.memory.cliprows))
-    sheet.addRows(to_paste, index=rowidx)
+    if vd.memory.cliprows:
+        to_paste = list(deepcopy(r) for r in reversed(vd.memory.cliprows))
+        sheet.addRows(to_paste, index=rowidx)
+    else:
+        vd.warning('nothing to paste')
```
anjakefala, line 133: Would it make sense to change this to

```
if not vd.memory.cliprows:  #1739
     vd.warning('nothing to paste')
     return
  ```
  
  If it does, that would limit the diff to the relevant change, and include a reference to the bug report!

**Review by anjakefala — COMMENTED** [review]
I really appreciate your neat bug fixes @midichef!

File: visidata/clipboard.py (hunk 1 above)
Comment (midichef) [reply visidata/clipboard.py:133]: Sure thing, done!

**Review by anjakefala — APPROVED** [review]
I love that you rebased everything onto a single commit. =))

File: visidata/clipboard.py [hunk 2]
```diff
@@ -129,6 +129,9 @@ def delete_row(sheet, rowidx):
 
 @Sheet.api
 def paste_after(sheet, rowidx):
+    if not vd.memory.cliprows:  #1793
+        vd.warning('nothing to paste')
+        return
```
saulpw, line 134: Typically we use `vd.fail()` which is just a warning + exception.  This has the added benefit of carrying a stacktrace with the error that can be viewed with `Ctrl+E`.  Also, an exception will abort a replay, which is generally desirable with failed commands.

anjakefala: Is aborting a replay and carrying a stacktrace desireable here? I am still developing my sense of taste around when to abort a replay and when to not. When the command is supposed to create rows or create columns, it is better to abort?

saulpw: I'm still developing my taste around that too =)

Let's try to empathize with the user here.  In this case, they have a replay script that selects some number of rows, and then they paste those rows in a different place or on another sheet, and then they go on to do something else with that data.

But in this dataset, for some legitimate reason it selects zero rows.  Should it refuse to proceed with the rest of the replay because of it?

geekscrapy: I would continue to replay. If this is intended by the user, everything is fine. If not intended, they will modify the input.

If we don't continue replay, the user has no choices

geekscrapy: The way I see it: if a replay fails due to bad input, that is on the user (with vd giving some useful troubleshooting output)

PR #1794, opened 2023-03-09
[fill-] allow filling with values that are logically false

midichef:
`setcol-fill` currently tries to fill using a previous value that is logically true. It should be looking more narrowly, for a value that is not None. For example, if the column is numeric, it incorrectly ignores the value 0. In this test file, the nulls below 1 can be filled, but the nulls below 0 cannot:
[fill-fails-for-zero.json.txt](https://github.com/saulpw/visidata/files/10926816/fill-fails-for-zero.json.txt)
This fix changes it so that fill will use any value but None.

### DIFF
--- a/visidata/features/fill.py
+++ b/visidata/features/fill.py
@@ -17,7 +17,7 @@ def fillNullValues(vd, col, rows):
             val = e
 
         if isNull(val):
-            if lastval and (id(r) in rowsToFill):
+            if not isNull(lastval) and (id(r) in rowsToFill):
                 oldvals.append((col,r,val))
                 col.setValue(r, lastval)
                 n += 1


visidata/features/fill.py
```diff
@@ -17,7 +17,7 @@ def fillNullValues(vd, col, rows):
             val = e
 
         if isNull(val):
-            if lastval and (id(r) in rowsToFill):
+            if lastval is not None and (id(r) in rowsToFill):
```
saulpw: Good catch, @midichef!  But let's use `if not isNull(lastval)`, so that the user can define their own null via `options.null_value`.

midichef: Oh, so that is what `isNullFunc` is for! Thanks.

saulpw (approved):

PR #1809, opened 2023-03-18
Change default depth of expand-col(s)-depth to 0

cool-RR:
People who want a depth of 1 can just use `expand-col(s)`


saulpw (approved):
This makes sense to me!

PR #1813, opened 2023-03-20
Add command reopen-last-closed

### LINKED ISSUE #1811: Feature request: Provide command `reopen-closed-sheet` like Chrome's Ctrl-Shift-T
It often happens that I close a sheet by mistake and immediately curse because I've already done a bunch of manipulations on this tab that I'll now have to repeat. (I know I can guard a sheet but that doesn't fit my workflow.)

I wish VisiData had a feature like Chrome's Ctrl-Shift-T. Basically a command that just opens that sheet that was closed most recently, with all the manipulations that I've done to it. 

Of course, this command is challenging because it means VisiData has to keep this information around. Maybe limit it to last 3 sheets.

cool-RR:
Fixes #1811

### DIFF
--- a/visidata/_open.py
+++ b/visidata/_open.py
@@ -185,7 +185,9 @@ def loadInternalSheet(vd, cls, p, **kwargs):
 
 BaseSheet.addCommand('o', 'open-file', 'vd.push(openSource(inputFilename("open: "), create=True))', 'Open file or URL')
 TableSheet.addCommand('zo', 'open-cell-file', 'vd.push(openSource(cursorDisplay) or fail(f"file {cursorDisplay} does not exist"))', 'Open file or URL from path in current cell')
+BaseSheet.addCommand('gU', 'undo-last-quit', 'push(allSheets[-1])', 'reopen most recently closed sheet')
 
 vd.addMenuItems('''
     File > Open file/url > open-file
+    File > Reopen last closed > undo-last-quit
 ''')


visidata/_open.py
```diff
@@ -185,7 +185,9 @@ def loadInternalSheet(vd, cls, p, **kwargs):
 
 BaseSheet.addCommand('o', 'open-file', 'vd.push(openSource(inputFilename("open: "), create=True))', 'Open file or URL')
 TableSheet.addCommand('zo', 'open-cell-file', 'vd.push(openSource(cursorDisplay) or fail(f"file {cursorDisplay} does not exist"))', 'Open file or URL from path in current cell')
+BaseSheet.addCommand('zq', 'reopen-last-closed', 'push(allSheets[-1])', 'reopen most recently closed sheet')
```
saulpw: Does `zq` feel good to you at all??  It makes me feel uneasy to press `q` when I want to do the reverse of quit.

geekscrapy: Yeeeea, I would only ever use the long command in fear of miss typing the z. Does it have to be bound to a key combo?

saulpw: not strictly, but it feels like we want something that makes quitting feel less stressful, like Shift+U does for undo.  I'm kind of leaning towards `g Shift+U` at the moment, since otherwise that would feel like "undo everything" which is just a reload.  Also I think the longname should be `open-last-closed` or maybe `undo-last-quit`, instead of `reopen` which would be a new verb.

geekscrapy: My vote is`undo-last-quit` 👍

cool-RR: `zq` is possibly one of the the most awkward to type key bindings known to man. It doesn't bother me because this really is a "oh shit" command rather than a command that's part of my standard workflow. Logically I think that `zq` makes more sense than `g Shift+U`, but I don't feel strongly about this, so I changed it now to `g Shift+U`.

I disagree about the function name. I've taken the "reopen" teminology from Chrome. The command `undo-last-quit` is awkward enough, but having to change the menu item to "Undo last quit" would be even weirder.

saulpw: The longname and the menu text don't have to be the same.

cool-RR: Changed just the name of the command, kept the menu text.

saulpw:
Thanks, @cool-RR, let's see if this solves the problem!

PR #1824, opened 2023-03-26
Support tuples in YAML files

### LINKED ISSUE #1822: YAML fails reading files with tuples
Example file:

```
foo: !!python/tuple
- 1
- 2
- 3
```

Error:

```
ConstructorError: could not determine a constructor for the tag 'tag:yaml.org,2002:python/tuple'·
```

cool-RR:
Closes #1822

### DIFF
--- a/sample_data/benchmark.yml
+++ b/sample_data/benchmark.yml
@@ -357,4 +357,8 @@
   quantity: '1'
   sku: REPT217
   unit: $99.95
+- foo: !!python/tuple
+    - 1
+    - 2
+    - 3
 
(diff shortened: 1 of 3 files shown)


saulpw:
Well, I don't quite understand it, but if it solves the problem, great.  I'll let @anjakefala take a look and consider if/how we want to test .yml data, and she can merge it.  Thanks for putting it together.

anjakefala:
Thanks so much @cool-RR! I updated the sample yaml to cover your use-case, and everything looks good.

PR #1850, opened 2023-04-13
[graph] fix data on edges being drawn offscreen

midichef:
On any graph, points may appear to be missing from the leftmost, rightmost, top, or bottom edges. The points are present, but they are drawn just a tiny bit offscreen.

The problem exists in various forms in probably every version of Visidata since 2017, I've reproduced it as far back as v1.3. Triggering it is not common, but also not extremely rare. I estimate that if the leftmost data point has an x value near 0, around 8-15% of randomly generated data sets will have a point that is drawn offscreen.  There's another 8-15% chance if the bottommost data point has a y value near 0.

The issue is caused by loss of precision after floating-point division when calculating the coordinates of `self.visibleBox`.

This patch fixes the boundaries so that all points are visible. It does so by getting rid of unnecessary floating point division, by algebraically simplifying the expressions `self.plotviewBox.w / self.xScaler` and `self.plotviewBox.w / self.xScaler`. It preserves the exactness of `self.visibleBox.xmin`, `self.visibleBox.ymin`, `self.visibleBox.w` and `self.visibleBox.h` at the default zoom level.

A fix should be given a high priority, as users may fail to notice the most extreme values in their data, which are often important.

Here is an example of data that exhibits the problem. Neither of these 2 data points will be visible when graphed in visidata versions >= 2.0:
```
x	y
0	121
60	0
```
x=0, y=121 will be offscreen because `self.visibleBox.xmin` will be just a tiny bit greater than 0 (because xmax-xmin = 60, and xmin=0).
x=60, y=0 will be offscreen because of a similar offset for `self.visibleBox.ymin` (because ymax-ymin = 121 and ymin=0).

This data set has the same problem for visidata version 1.3:
```
n	view_count
0	41
93	0
```

### DIFF
--- a/visidata/canvas.py
+++ b/visidata/canvas.py
@@ -492,14 +492,16 @@ class Canvas(Plotter):
                 ymax += 1
             self.canvasBox = BoundingBox(float(xmin), float(ymin), float(xmax), float(ymax))
 
+        w = self.calcVisibleBoxWidth()
+        h = self.calcVisibleBoxHeight()
         if not self.visibleBox:
             # initialize minx/miny, but w/h must be set first to center properly
-            self.visibleBox = Box(0, 0, self.plotviewBox.w/self.xScaler, self.plotviewBox.h/self.yScaler)
-            self.visibleBox.xmin = self.canvasBox.xcenter - self.visibleBox.w/2
-            self.visibleBox.ymin = self.canvasBox.ycenter - self.visibleBox.h/2
+            self.visibleBox = Box(0, 0, w, h)
+            self.visibleBox.xmin = self.canvasBox.xmin + (self.canvasBox.w / 2) * (1 - self.xzoomlevel)
+            self.visibleBox.ymin = self.canvasBox.ymin + (self.canvasBox.h / 2) * (1 - self.yzoomlevel)
         else:
-            self.visibleBox.w = self.plotviewBox.w/self.xScaler
-            self.visibleBox.h = self.plotviewBox.h/self.yScaler
+            self.visibleBox.w = w
+            self.visibleBox.h = h
 
         if not self.cursorBox:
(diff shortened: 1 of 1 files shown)


midichef:
I'm including more detail here for people wondering if they have run into this error in the past.

**What versions of Visidata have this problem?**
All versions since 2017, I think, though I have only tested 1.3 and up.
Versions 2.0 and above will have the problem for some data sets, and versions earlier than 2.0 will affect others.

**How many points can be affected?**
All points that are on the very edges of the data.
As a general rule, the closer your least data points are to zero (xmin near 0 or ymin near 0), the more likely your data were be affected.
When the x axis is a date, the left and right edges probably were not affected. Dates are such large numbers that they usually do not have enough precision to hold the tiny offsets caused by the floating-point rounding in the coordinates of `self.visibleBox`. However, the top and bottom edges may still have been affected.

If there is interest, I can provide code to tell whether any points were drawn offscreen for a given dataset:  it requires knowing the data's extreme values (xmin, xmax, ymin, ymax), the terminal dimensions, and the Visidata version.

saulpw:
Thank you, @midichef.  I imagine this was pretty frustrating to discover.  Thanks so much for diving into it and really understanding the core issue; it makes sense to me and from what I remember of this code, this seems reasonable.  I'll merge it straightaway, and it may be the impetus for a cherry-picked v2.11.1.

Also, great job on your work here.  I appreciate your consistent quality commits, and I'd like to encourage you to feel more confident in making changes, including refactors and style adjustments.  Can we chat about the name change you suggested here?  If you send me an email, I can offer you commit rights to the repo, which I trust would be put to continued good use.

midichef:
Oops, that name change was a note to myself. I didn't mean it to be part of this patch!
But thanks for the encouragement, I will get in touch via email.

PR #1853, opened 2023-04-17
[docs] add guessing filetype tutorial to API loader docs

anjakefala:


docs/api/loaders.rst
```diff
@@ -261,6 +261,35 @@ This is a completely functional loader for the ``sas7bdat`` (SAS dataset file) f
...
+
+`Examples of guess_filetype functions <https://github.com/saulpw/visidata/commit/4743f92bb855cf931d896e65845c549ce6027e2f>`
+
+
+::
+
+        @VisiData.api
+        def guess_foo(vd, p):
+            line = foolib.iterfoo(p.open_text())
+            try:
```
saulpw: I'm not sure if this `try` is needed anymore.

saulpw (approved):
This is good; maybe it would help to be explicit about just using the usual `open_text()` and `read()`, but `seek()` may not work.

docs/api/loaders.rst
anjakefala: I was building off of an earlier `FooSheet` example! I am not attached. I think the `r[0] = "#Foo"` is demonstrated enough (along with the link to actual examples).

anjakefala:
That you could use the standard `read()` to check the first line's contents?

saulpw:
> That you could use the standard `read()` to check the first line's contents?

Yes, or iterate over lines (it's file-like).

PR #1871, opened 2023-04-20
[vgit] skip symbolic refs more explicitly when listing branches

ajkerrigan:
I was wondering how on earth 7ddffcc5c1a4f3142f10e14bb23f6885b83ad594 broke vgit tests :monocle_face: . And then I noticed that vgit was [skipping](https://github.com/saulpw/visidata/blob/7ddffcc5c1a4f3142f10e14bb23f6885b83ad594/visidata/apps/vgit/branch.py#L70) branches if the output from `git branch --list` included `->`. Looks like the intent there is to avoid including symbolic refs in the list, but my commit happened to have a `->` in the description.

I'm not sure this is the best or most complete fix, but I wanted to at least get an explanation up since I broke the thing :laughing: .

### DIFF
--- a/visidata/apps/vgit/branch.py
+++ b/visidata/apps/vgit/branch.py
@@ -61,16 +61,25 @@ class GitBranch(GitSheet):
     nKeys = 1
 
     def iterload(self):
-        branches_lines = self.git_lines('branch',
-                                        '--list',
-                                        '-vv',
-                                        '--no-color',
-                                        *self.git_args)
-        for line in branches_lines:
-            if '->' in line:
-                continue
+        branches_lines = self.git_lines(
+            'branch',
+            '--list',
+            '--format',
+            ' '.join((
+                '%(if)%(symref)%(then)yes%(else)no%(end)',
+                '%(HEAD) %(refname:short) %(objectname:short)',
+                '%(if)%(upstream)%(then)[%(upstream:short)',
+                '%(if)%(upstream:track)%(then): %(upstream:track,nobracket)%(end)]',
+                '%(end)',
+                '%(contents:subject)'
+            )),
+            '-vv',
+            '--no-color',
+            *self.git_args)
 
-            m = re.match(r'''(?P<current>\*?)\s+
+        for line in branches_lines:
+            m = re.match(r'''(?P<is_symref>(yes|no)?)
+                             (?P<current>\*?)\s+
                              (?P<localbranch>\S+)\s+
                              (?P<refid>\w+)\s+
                              (?:\[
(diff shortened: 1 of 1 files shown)


anjakefala:
This seems to look good!

I don't know much about symbolic refs. I want to understand them better before approving the merge. I can confirm that you fixed `vgit branch` behaviour with the existing repo status. =)

ajkerrigan:
A couple things I'm not sure about after fiddling with this:
- Do we actually want to skip symbolic refs? I'm not sure that it actually _was_ skipping them before, I just inferred its intent based on seeing the `->`.
- If we do, and we need to take more direct control of formatting to handle it, would it be worth laying fields out with separators or something so we could `split` them instead of having to parse a gnarly regex?
  - It's also worth mentioning tools like [pygit2](https://pypi.org/project/pygit/) here. It seems like way overkill to bring in a dependency for this specific case, but didn't know if it would be useful in enough other places to be worth keeping in mind.

Yammering about symbolic refs follows...

-----

> I don't know much about symbolic refs. I want to understand them better before approving the merge.

Heh, same here! I understand symbolic refs like `HEAD` in the sense that "instead of pointing to a specific commit hash, it points to another git reference". I didn't realize until [reading the docs](https://git-scm.com/docs/git-symbolic-ref) that git uses symbolic refs as a replacement for symlinks (which were deprecated due to portability issues).

But that left me wondering why we'd be trying to find and skip branches that were symbolic refs in the first place :thinking: . I mean if we have develop at a specific commit:

```console
> cat .git/refs/heads/develop
7ddffcc5c1a4f3142f10e14bb23f6885b83ad594
```

We could create a symbolic ref with:

```console
> git symbolic-ref refs/heads/meh refs/heads/develop

> cat .git/refs/heads/meh
ref: refs/heads/develop

> git branch --list | grep develop
* develop
  meh -> develop
```

And changes to `meh` would affect `develop` going forward (and vice-versa). And if I ran:

```console
> git symbolic-ref refs/heads/meh refs/heads/stable
```

Then `meh` would be in sync with `stable` instead. I haven't needed to do this sort of thing myself, and I can imagine it being either neat or a footgun depending on when/how you use it :grin: .

anjakefala:
>If we do, and we need to take more direct control of formatting to handle it, would it be worth laying fields out with separators or something so we could split them instead of having to parse a gnarly regex? 

You are very much welcome to open a fresh PR with this change, if you are interested! Since your change improves the current status quo, I vote we merge it in. 

This is a discussion on why magit considered silently skipping symbolic refs: https://github.com/magit/magit/issues/2602

This would be the approach that does not involve skipping them: https://github.com/magit/magit/commit/3587e1eff6f47b0fd28d7318d4219171b98edc13

If symbolic refs becomes a desired feature request, we can do something similar. =) Or, if you end up wanting to implement something similar to this (including a column with symbolic refs?), you are welcome to!

PR #1876, opened 2023-04-29
[sort-] show sort arrow for sort columns described by name

midichef:
Right now the DirSheet's modtime column and HelpSheet's sheet and longname columns are used in sorting, but they don't have an arrow showing they are sorted. This patch shows the arrow.


saulpw (approved):
Yup, makes sense to me!

PR #1883, opened 2023-05-01
[path-] create RepeatFile for bytes fp

### LINKED ISSUE #1880: [develop] reading from stdin drops 32 bytes of data
**Small description**
When reading data from standard input with no file type given, once the file is beyond a certain size, 32 bytes of data are lost.

`seq 2000 |vd` will have less than 2000 lines, but
`seq 2000 |vd -f txt` will have 2000 lines as expected.
`seq 1000 |vd` does not have the problem.

**Additional context**
saul.pw/VisiData v2.12dev
Ubuntu 22.04.2

The problem is in `guess_xls()` and `guess_sqlite()`. Each eats 16 bytes.

https://github.com/saulpw/visidata/blob/641bd058a358bb1c20a5167af530b6a85f6e22fc/visidata/loaders/xlsb.py#L7
https://github.com/saulpw/visidata/blob/641bd058a358bb1c20a5167af530b6a85f6e22fc/visidata/loaders/sqlite.py#L22

### LINKED ISSUE #1881: [develop] STDIN redirections are guessed incorrectly as `tar`
```
vd < sample_data/sample.tsv
saul.pw/VisiData v2.12dev
guessed "tar" filetype based on contents
```

This is because [`if tarfile.is_tarfile(p.open_bytes())`](https://github.com/saulpw/visidata/blob/ae6bca7a9b82705f11cf1583f112ae714a323dc9/visidata/loaders/archive.py#L18) is returning `True`.

anjakefala:
Closes #1880

On top of resolving #1880, this would also have resolved #1881 without the need to change the `p.open_bytes()` to `p`. 

My understanding is that the main thing `open_text` does, other than process `-` as a input, is create the `RepeatFile`, and then call `open()`. As long as `open_bytes` calls `open_text` with the specific `rb` mode, it should maintain the same behaviour except that the `RepeatFile` gets created. And the addition of a **RepeatFile** for `open_bytes` was @saulpw recommendation for that bug.

I tested pipes, `.db`, STDIN redirect, and `seq 2000 | vd`.

### DIFF
--- a/visidata/_open.py
+++ b/visidata/_open.py
@@ -164,7 +164,7 @@ def openSource(vd, p, filetype=None, create=False, **kwargs):
 def open_txt(vd, p):
     'Create sheet from `.txt` file at Path `p`, checking whether it is TSV.'
     if p.exists(): #1611
-        with p.open_text(encoding=vd.options.encoding) as fp:
+        with p.open(encoding=vd.options.encoding) as fp:
             delimiter = vd.options.delimiter
             try:
                 if delimiter and delimiter in next(fp):    # peek at the first line
--- a/visidata/_urlcache.py
+++ b/visidata/_urlcache.py
@@ -29,7 +29,7 @@ def urlcache(vd, url, days=1, text=True, headers={}):
         ret = fp.read()
         if text:
             ret = ret.decode('utf-8').strip()
-            with p.open_text(mode='w', encoding='utf-8') as fpout:
+            with p.open(mode='w', encoding='utf-8') as fpout:
                 fpout.write(ret)
         else:
             with p.open_bytes(mode='w') as fpout:
--- a/visidata/cmdlog.py
+++ b/visidata/cmdlog.py
@@ -34,7 +34,7 @@ VisiData.save_vd = VisiData.save_tsv
 
 @VisiData.api
 def save_vdj(vd, p, *vsheets):
-    with p.open_text(mode='w', encoding=vsheets[0].options.save_encoding) as fp:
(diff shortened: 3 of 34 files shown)


saulpw (commented):
Do the arrow/npy savers and the pdf and pcap loaders still work?  If so I guess we should just rename `open_text` to `open` and rename the existing `open` to something else.

anjakefala:
> Do the arrow/npy savers and the pdf and pcap loaders still work?

They do!

anjakefala:
I am not sure if renaming `open` is a good idea. Or at least, I think it would be a pretty complicated change. In VisiData, there are modules with their own implement `open()`, and ones that maybe are intended to fallback to `pathlib.Path.open()`.  

We could rename `open_text` to something else (maybe `open_rp`), while keeping `open` as-is?

anjakefala:
`open_text` -> `open_file`? `open_input`?

saulpw:
> I am not sure if renaming open is a good idea. 

Why is the current `open` the "correct" one?  We would still have an `open` and it would still be overrideable, it would just have minorly different default semantics than currently.

anjakefala:
> Why is the current open the "correct" one? We would still have an open and it would still be overrideable, it would just have minorly different default semantics than currently.

I think that is fair, and I am mostly dreading that it is a complex change. I will default to calling it `popen` for "FileProgressOpen", since that is not already in the codebase, until we come up with a better name.

saulpw (approved):
Okay, this looks fine!

PR #1890, opened 2023-05-07
Don't truncate wide columns with fixed width saver

### LINKED ISSUE #1849: fixed width saver truncates data if columns are not fully expanded
**Small description**

If you save or syscopy a table as `fixed` format, and the visible column width is less than the width of the data in the column, the data is truncated. Also, the resulting file is not a valid fixed width format file as the columns are not aligned with the headers.

**Expected result**

Saving or copying a table when the columns are not fully expanded should yield the same result as when the columns are expanded.

image

[test-vd-fixed.zip](https://github.com/saulpw/visidata/files/11217144/test-vd-fixed.zip)

**Additional context**
saul.pw/VisiData v2.11

daviewales:
Previously the fixed width saver would truncate saved data to match viewed column widths.
Now it saves all the data regardless of the viewed column widths.

Fixes #1849.

### DIFF
--- a/visidata/loaders/fixed_width.py
+++ b/visidata/loaders/fixed_width.py
@@ -84,7 +84,7 @@ def save_fixed(vd, p, *vsheets):
             widths = {}  # Column -> width:int
             # headers
             for col in Progress(sheet.visibleCols, gerund='sizing'):
-                widths[col] = col.getMaxWidth(sheet.rows)
+                widths[col] = col.getMaxWidth(sheet.rows)  #1849 
                 fp.write(('{0:%s} ' % widths[col]).format(col.name))
             fp.write('\n')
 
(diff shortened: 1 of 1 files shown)


visidata/loaders/fixed_width.py
```diff
@@ -84,8 +84,7 @@ def save_fixed(vd, p, *vsheets):
             widths = {}  # Column -> width:int
             # headers
             for col in Progress(sheet.visibleCols, gerund='sizing'):
-                maxWidth = col.getMaxWidth(sheet.rows)
-                widths[col] = col.width if col.width >= maxWidth else sheet.options.default_width or maxWidth
+                widths[col] = col.getMaxWidth(sheet.rows)
```
anjakefala:
```suggestion
                widths[col] = col.getMaxWidth(sheet.rows)  #1849 
```

Hey @daviewales!

You are the first person to come to the repo with preferences about the fixed width saver, so I am complete for just merging in your preference.

Would you mind adding this change that includes a reference towards the issue? This is so if someone in the future comes by with a different preference, we have a reminder of the previous conversation with yours.

Thanks for submitting a PR with the behaviour you wanted!

anjakefala (changes requested):
Once the commit reference comment is added, I will merge this in!

**Edit** NVM, realised you gave me commit privileges =)

daviewales:
Thank you!

PR #1891, opened 2023-05-10
Tests for vdsql on Python 3.11

anjakefala:
Looking into https://github.com/ibis-project/ibis/issues/6180


anjakefala:
It seems like the ibis team is planning on resolving this issue. I will be following https://github.com/ibis-project/ibis/issues/6180 and re-open a PR once this is resolved.

Issue #1587, opened 2022-11-05
Python 3.11 released with builtin toml support.


frosencrantz:
### Discussed in https://github.com/saulpw/visidata/discussions/1580

<div type='discussions-op-text'>

<sup>Originally posted by **frosencrantz** October 29, 2022</sup>
Python 3.11 has been released with speed improvements and builtin toml support: https://docs.python.org/3/whatsnew/3.11.html

- Sounds like VisiData should add support for toml, too.
</div>

Since Python has builtin support for TOML now, it seems like VisiData should also add basic support.

LaPingvino:
Visidata has a relatively low baseline Python version it still supports. It would take some time for 3.11 to be broadly adopted before this can be feasibly considered I think.

frosencrantz:
@LaPingvino  that is true about wide spread 3.11 usage.  Visidata can already handle the case where the base modules isn't available.  Also, the 3.11 support according to  [PEP 680](https://peps.python.org/pep-0680) is actually https://github.com/hukkin/tomli.  So there is a backfill available.

frosencrantz:
I'll mention that @ajkerrigan  has already created this as a plugin: 
 - https://github.com/ajkerrigan/visidata-plugins/blob/master/plugins/vd_toml.py

Wondering if it would be possible to make it standard?

ajkerrigan:
> I'll mention that @ajkerrigan has already created this as a plugin:
> 
>     * https://github.com/ajkerrigan/visidata-plugins/blob/master/plugins/vd_toml.py
> 
> 
> Wondering if it would be possible to make it standard?

 @LaPingvino's point seems reasonable there - making it standard would mean taking a conditional dependency on tomli for Python < 3.11 and that feels like "most people" for the time being. That doesn't feel like too big a deal, but it also doesn't feel like a clear win over installing a plugin :shrug: .

(To give credit where it's due @frosencrantz , I believe I only put that plugin together after you put the idea in my head :laughing: . And for what it's worth, I find myself using TOML support surprisingly often to poke at `pyproject.toml` and `poetry.lock` files!)

frosencrantz:
Thanks for adding the TOML plugin, @ajkerrigan!  Also, happy to hear you use it regularly. 

[...]

PR #1894, opened 2023-05-12
[loaders] add a toml loader

ajkerrigan:
Add a core TOML loader, adapted from [this plugin](https://github.com/ajkerrigan/visidata-plugins/blob/faff40971ceb344f8d3a44150a64bd0a383ee734/plugins/vd_toml.py). Related to discussions in #1580 and #1587.


ajkerrigan:
Ha, the `vd.importExternal` is working as expected :sweat_smile: .

ajkerrigan:
Hmm not sure offhand why that's still failing, expected that `requirements.txt` addition to cause CI to pull in `tomli` [here](https://github.com/saulpw/visidata/blob/183a9ed4c277dbef4b305cb6603a92061bd9f7cb/.github/workflows/main.yml#L45) :thinking: . Timing out for now, will have to have another look later on. Could be a misunderstanding on my part, caching issue, etc... :shrug:

visidata/loaders/toml.py
```diff
@@ -38,6 +31,13 @@ def reload(self):
         self.columns = []
         self.rows = []
 
+        try:
+            # Python 3.11+
+            import tomllib
```
anjakefala, line 46: Oh, so weird! I understand now. `tomllib` is in stdlib in Python 3.11.

@saulpw Would it be better to have a `sys.version_info` check for the Python version, or to have this `try/except` for `tomllib`? My instinct is doing a version info check might be better, in case someone ever puts a dummy package on PyPI for `tomllib`. But maybe that worry is unfounded, they block some names.

anjakefala:
AJ <3 I am so grateful of all the thoughtfulness in this PR.

sample_data/freshwater-mammals.toml
```diff
@@ -0,0 +1,32 @@
+# Adapted lovingly but hackily from:
```
anjakefala, line 1: <3

visidata/loaders/toml.py
bartbroere: I believe `tomli` was (and still is) on PyPI before most of it was included in the standard library and was renamed to `tomllib`. For that reason I think a backport appearing is unlikely. I like the EAFP approach with `try` `except` personally :-)

anjakefala: Yeah, and the thinking is that the `sys.version_info` check would be more likely to be buggy/fragile!

anjakefala:
I will formally review this ASAP! =)

requirements.txt
```diff
@@ -37,6 +37,7 @@ matplotlib          # svg saver
 sh                  # ping
 psutil              # procmgr
 numpy               # npy pandas hdf5 arrow
+tomli               # toml (Python <3.11 only)
```
anjakefala:
```suggestion
tomli; python_version < '3.11'               # toml
```

https://peps.python.org/pep-0508/#environment-markers

=)

anjakefala: (I could not really tell where to indent the `# toml` in the "suggestion" framework.)

visidata/loaders/toml.py
```diff
@@ -0,0 +1,51 @@
+from visidata import (
+    ColumnItem,
+    PythonSheet,
+    VisiData,
+    asyncthread,
+    deduceType,
+    vd,
+)
+
```
anjakefala, line 9:
```suggestion
from visidata import ColumnItem, PythonSheet, VisiData, asyncthread, deduceType, vd
```

visidata/loaders/toml.py
```diff
@@ -0,0 +1,51 @@
+from visidata import (
```
anjakefala, line 1:
```suggestion
from visidata import ColumnItem, PythonSheet, VisiData, asyncthread, deduceType, vd
```

saulpw: [Let's not nitpick formatting, esp outside of visidata core.]

visidata/loaders/toml.py
```diff
@@ -0,0 +1,51 @@
...
+def open_toml(vd, p):
+    return TomlSheet(p.name, source=p)
+
+
+class TomlSheet(PythonSheet):
+    """A Sheet representing the top level of a loaded TOML file.
+
+    This is an intentionally minimal loader with cues taken from
+    VisiData built-in JSON and Python object sheet types.
+    """
```
anjakefala, line 21: @ajkerrigan This is such a thoughtfully put together PR.

I think the only thing that concretely felt like it might be useful to add is some `help` text for the sidebar. I am thinking something like `zEnter` to dive into the Dictionary structures. I am going to put a prototype together!

anjakefala: I added a help string!

@ajkerrigan please review! I found it tricky to try to describe `z Enter` and `expand-col` in non-jargon language.

requirements.txt
ajkerrigan: Thanks for catching this :+1: . It's a long line, so I just went with one space before the comment which seems to match up with other long lines (xlsb, importlib_metadata)

visidata/loaders/toml.py
```diff
@@ -0,0 +1,61 @@
...
+    help = '''# Toml Sheet
+This sheet represents the top level of {sheet.source.name}.{sheet.source.ext}.
+
+Each cell within this sheet can contain dictionaries (representing TOML key:value pairs), lists (representing TOML arrays), or scalars.
+
+Some helpful commands when working with cells of lists and dictionaries:
+
+- `zEnter` on a cell with a dictionary will "dive" into the current cell, creating a seperate sheet with keys as **Columns** and values as **Rows**.
+- `(` (`expand-col`) on a column with lists or dictionaries will "expand" the structures in the cells into new columns within the current sheet.
```
ajkerrigan: Good call, thanks for adding this! You're right, it's hard to avoid jargon when explaining dives/expands. It's also a bit awkward because the row/column layout you get from `zEnter` varies depending on whether you're diving into a list, dict, list of dicts...

It might be cheating to avoid jargon by also being less specific, but maybe we can highlight how both commands zoom into nested data. And the difference is whether that'll happen here or in a new sheet, like...

```suggestion
- `(` (`expand-col`) on a column with lists or dictionaries will "expand" the structures in the cells into new columns within the current sheet.
- `zEnter` on a cell with lists or dictionaries will "dive" into the current cell, expanding its structures into rows and columns in a seperate sheet.
```

ajkerrigan (commented):
Thanks for the thoughtful and careful look, some good catches there :+1:

visidata/loaders/toml.py
anjakefala: Great! Love it! I will probably backport these command descriptions to `pyobj`/`jsonl` sidebars!

PR #1896, opened 2023-05-14
[graph] fix various edge cases

midichef:
This is a series of small fixes to the graph code. Each is short localized fix to some off-by-1 error or edge case. I've split them up to make them individually easy to approve. If they all look acceptable, let me know and I'll bundle them all into 1 commit.

For seeing what these fixes change, it helps to have a graph that is uniformly filled, so I have a script that generates data for that:
[fill_graph.py.txt](https://github.com/saulpw/visidata/files/11471176/fill_graph.py.txt)
It's intended to be run as `fill_graph.py $COLUMNS $LINES |vd -f tsv`. To work properly, it requires a visidata that has already incorporated my first commit edd8fe9ea47088fa18267b5af7beb578521505cf.

edd8fe9ea47088fa18267b5af7beb578521505cf The right margin is drawn as 1 terminal square narrower than it should be.

583a1838e212aa33984ea7f7cc07bc75e1fc624c This commit is a correction to how the graph label coordinates are calculated for the hypothetical situation where the right margin has an odd number of pixels. Right now, the right margin has an even number of pixels, so this does not actually change any behavior.

8c75799e3805a781094bb685f95872119a0c1e4d If you move the cursor far past the left edge of the screen, where the y axis labels are, `rowsWithin()` will select points from the far right of the graph. This commit fixes that.

9a4c7945a489c6e636343b70f758d6fff3441697
If your terminal is not 80x25, until you click the mouse on a graph, the cursor on the graph looks like it is exactly 1x1 terminal characters, but it is actually some fraction different than that. This commit makes sure that the cursor starts out sized exactly 1x1.

54972444df813bf877d2fa78694fcd00fa4c04e4
When `rowsWithin()` selects points in the cursor, it's too greedy, taking a bit more than what is in the cursor.

7950d28a45f3739c7d82136c83fab5706dfb2bc5
This commit changes the behavior of `go-rightmost` and `go-top` and `go-bottom` 
1) `go-top` does not properly handle a cursor that is taller than 1 character.
2) All three commands can move the cursor so it ends up moving a fraction of a terminal square, in some cases moving just a fraction of a square beyond the edges of the graph. For example, if the cursor is exactly 1x1, and starts out exactly aligned with the terminal squares, after `go-rightmost`, the cursor won't be aligned. It will be selecting the right half of 1 square and the left half of the next square to the right. I think that what matches user expectations is that `go-rightmost` should keep the cursor fully within the graph, and aligned as if the user had pressed  go-right N times.
This commit fixes the `go-top`/`go-bottom` handling of tall cursors, and for all three commands, it keeps the cursor within the boundaries of the visible graph. 

6e01879ff1152037283e9da3d46aaa7f606746c0
When the graph is first drawn, the cursor is not quite aligned to the x-axis, it is a fraction of a square a bit above it. This commit makes sure the cursor starts out with its edges touching the x-axis and the y-axis.

### DIFF
--- a/visidata/canvas.py
+++ b/visidata/canvas.py
@@ -209,11 +209,15 @@ class Plotter(BaseSheet):
             self.hiddenAttrs.remove(attr)
         self.plotlegends()
 
-    def rowsWithin(self, bbox):
-        'return list of deduped rows within bbox'
+    def rowsWithin(self, plotter_bbox):
+        'return list of deduped rows within plotter_bbox'
         ret = {}
-        for y in range(bbox.ymin, min(len(self.pixels), bbox.ymax+1)):
-            for x in range(bbox.xmin, min(len(self.pixels[y]), bbox.xmax+1)):
+        x_start = max(0, plotter_bbox.xmin)
+        y_start = max(0, plotter_bbox.ymin)
+        y_end = min(len(self.pixels), plotter_bbox.ymax)
+        for y in range(y_start, y_end):
+            x_end = min(len(self.pixels[y]), plotter_bbox.xmax)
+            for x in range(x_start, x_end):
                 for attr, rows in self.pixels[y][x].items():
                     if attr not in self.hiddenAttrs:
                         for r in rows:
@@ -309,7 +313,7 @@ class Canvas(Plotter):
     rowtype = 'plots'
     leftMarginPixels = 10*2
     rightMarginPixels = 4*2
-    topMarginPixels = 0
+    topMarginPixels = 0*4
     bottomMarginPixels = 1*4  # reserve bottom line for x axis
 
     def __init__(self, *names, **kwargs):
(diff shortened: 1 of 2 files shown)


anjakefala:
Hey @midichef!

I really appreciated how thoroughly you documented all of the pieces of this PR, including providing us with a script to help test your changes with. There is so much attention to detail here.

I talked it over with @saulpw, and verified your fixes: if you mark this ready for review, we will happily merge it in.

I found a bug that actually already existed on `develop`, but that behaves a bit differently on your branch. Before, the Exception would be caught and be a no-op. On your branch the `Exception` is not caught and the canvas is not drawn. Since this is technically not a new bug, we are not going to block the PR on it. 

If you do have the energy to take a look at it, it would be welcome! Do you know why the `Exception` is not being caught anymore? Do you know what is going on?

If you do not have capacity to take a look in the near future, we can go ahead and merge your branch, I will open an issue, and take a look at it later.

This is a basic description:

Do command `_` or `zoom-all` on your `fill_graph.py` with `100` LINES and COLUMNS.

```
  File "/home/anja/git/visidata/visidata/mainloop.py", line 38, in drawSheet
    sheet.draw(scr)
  File "/home/anja/git/visidata/visidata/canvas.py", line 236, in draw
    cursorBBox = self.plotterCursorBox
  File "/home/anja/git/visidata/visidata/canvas.py", line 444, in plotterCursorBox
    return BoundingBox(self.scaleX(self.cursorBox.xmin),
  File "/home/anja/git/visidata/visidata/canvas.py", line 614, in scaleX
    return round(self.plotviewBox.xmin+(x-self.visibleBox.xmin)*self.xScaler)
AttributeError: 'NoneType' object has no attribute 'xmin'
```

midichef:
@anjakefala
Okay, I added 3 more small commits to the original 7.

1) https://github.com/saulpw/visidata/pull/1896/commits/26c3573f8d9b143009c4630d8ea9a3f1facba075
This is just a slight tweak to the naming and decorators in one of the fixes in the original 7: https://github.com/saulpw/visidata/commit/7950d28a45f3739c7d82136c83fab5706dfb2bc5. In the original fix, I made new functions that use the @property decorator, simply because it was present in the code I was copy/pasting. But these new functions are not a natural fit for properties, so this change removes the @property decorator.

2) https://github.com/saulpw/visidata/pull/1896/commits/d67eac8014b927483bf5a052512dcc5c0d3f4d50
This fixes the drawing bug downstream of zoom-all that you pointed out. zoom-all destroys visibleBox, which my patch expected to exist, so I fixed it to check whether visibleBox exists. This removes the extra brokenness my code was adding to zoom-all.

3) https://github.com/saulpw/visidata/pull/1896/commits/499a6d31938791564ae2b3dc734c347ecff0bdd6
This fixes the error that zoom-all was giving on its own. It does so by recreating self.visibleBox before the next call to refresh().

Let me know if this works on your side, and I'll package this up as 1 commit.
Once this is in, I have more changes to the graph code ready, that should be separate commits.

anjakefala:
Thanks so much @midichef! =)) Taking another look at the diff.

Wrt the failing builds, it seems to be because the `develop` build was red when you pushed. I ran the tests on a local checkout of your branch, and they passed fine.

anjakefala:
Hi @midichef, this all seems great! Thank you so much for fixing the found bug on top of it.

When you squash your commit, please rebase it onto current develop! We will merge it right in. =)

anjakefala (approved):

saulpw (approved):

anjakefala:
Thank you! =))

PR #1901, opened 2023-05-15
Don't get the windows-curses release with the bug

### LINKED ISSUE #1841: Keystroke ] not detected on Windows
In Powershell and cmd.exe I encountered that sorting didn't work in both orders. The `[` shortcut was detected and had its effect, but the `]` didn't. I narrowed it down to a problem with `windows-curses`, and in turn with its dependency `PDCurses`: https://github.com/zephyrproject-rtos/windows-curses/issues/41

Here's my plan on how to address it. I hope I'll get around to it somewhere next week.
- [ ] Improve the mapping in `PDCurses` and submit a pull request
- [ ] Bump the git submodule in `windows-curses` to the `PDCurses` version that has the fix and ask/wait for a release of this package
- [ ] Address the issue in this repository, perhaps by pinning `windows-curses` to a version of at least the newly released package.

I'm making this issue here just to document it and track progress. If you're reading this because you have this issue, I would recommend using WSL instead. (WSL is not an option for me unfortunately).

I didn't include the `.vd`-file to reproduce this issue. The simplest way to reproduce it is to get a Windows computer, run `visidata` from Powershell or cmd.exe and sort any column by pressing `]`.

bartbroere:
Fixes #1841.

asottile noticed this was a regression in the last `windows-curses` release in this issue: https://github.com/zephyrproject-rtos/windows-curses/issues/41

Once `windows-curses` releases a new version it will still automatically be fetched. It will just not depend on this one version from now on. As of now it will favour `2.3.0` over `2.3.1` if it all works correctly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ setup(name='visidata',
       py_modules=['visidata'],
       install_requires=[
           'python-dateutil',
-          'windows-curses!=2.3.1; platform_system == "Windows"',
+          'windows-curses<2.3.1; platform_system == "Windows"',  #1841
           'importlib-metadata >= 3.6',
       ],
       packages=['visidata', 'visidata.loaders', 'visidata.vendor', 'visidata.tests', 'visidata.ddw', 'visidata.man', 'visidata.themes', 'visidata.features', 'visidata.experimental', 'visidata.apps', 'visidata.apps.vgit', 'visidata.apps.vdsql', 'visidata.desktop'],
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ setup(name='visidata',
       py_modules=['visidata'],
       install_requires=[
           'python-dateutil',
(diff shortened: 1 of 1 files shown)


ajkerrigan:
Huh, nice bit of troubleshooting here. How fun! (from a certain angle at least :laughing: )

I wonder if it might be sensible to pin <2.3.1 rather than != though. Looks like the issue is still open and the project is looking for maintainers, so we might be cautious about automatically pulling in a new version.

setup.py
```diff
@@ -23,7 +23,7 @@
       py_modules=['visidata'],
       install_requires=[
           'python-dateutil',
-          'windows-curses; platform_system == "Windows"',
+          'windows-curses!=2.3.1; platform_system == "Windows"',
```
anjakefala:
```suggestion
          'windows-curses<2.3.1; platform_system == "Windows"',  #1841
```

PR #1902, opened 2023-05-16
Choose only exactly matching strings in chooser

daviewales:
Given a list of choices such as x, x2, x3, previously `chooseMany()` would select all three if you entered 'x' into the chooser input prompt.

Now it just selects the exact match 'x'. You can still select multiple values by entering them separated by spaces.

See discord conversation for context:
https://discord.com/channels/880915750007750737/880915984481939466/1107804615078072443

### DIFF
--- a/visidata/choose.py
+++ b/visidata/choose.py
@@ -40,7 +40,12 @@ def chooseFancy(vd, choices):
 
 @VisiData.api
 def chooseMany(vd, choices):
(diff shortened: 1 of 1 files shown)


anjakefala (approved):
Thanks @daviewales for opening a PR!

saulpw (approved):

PR #1915, opened 2023-06-08
[graph] fix top margin location and simplify y-coordinate calculation

midichef:
These changes are for the y-coordinate calculations in InvertedCanvas.

They do not properly take `topMarginPixels` into account, though the problem is hidden because the top margin is currently 0. If we increase it, the new margin shows up below the graph points instead of above, and the y-axis labels become misaligned.

Another problem is that the calculation of y-coordinate inversion is done separately at a few too many interfaces (`plotpixel()`, `scaleY()`, `canvasH()`, `canvasMouse()`), and sometimes with slight differences. That makes it extremely difficult to reason about y coordinates. So I have simplified the inversions, making them go through just one pair of functions: `scaleY()` and its inverse `unscaleY()`. The change happens to fix the `topMarginPixels` bug and simplify `calcTopCursorY()`, which is a sign it's a good way to handle the y coordinates.

I've also added a fix to the y coordinates in `zoom-cursor` (`zz`). The zoom used to go incorrectly to the area above the cursor. Now it zooms into it.

### DIFF
--- a/visidata/graph.py
+++ b/visidata/graph.py
@@ -10,37 +10,44 @@ def numericCols(vd, cols):
 
 
 class InvertedCanvas(Canvas):
+    @asyncthread
+    def render_async(self):
+        self.plot_elements(invert_y=True)
+
     def zoomTo(self, bbox):
         super().zoomTo(bbox)
-        self.fixPoint(Point(self.plotviewBox.xmin, self.plotviewBox.ymax), bbox.xymin)
-
-    def plotpixel(self, x, y, attr, row=None):
-        y = self.plotviewBox.ymax-y
-        self.pixels[y][x][attr].append(row)
+        self.fixPoint(Point(self.plotviewBox.xmin, self.plotviewBox.ymax),
+                      Point(bbox.xmin, bbox.ymax + 1/4*self.canvasCharHeight))
 
     def scaleY(self, canvasY):
-        'returns plotter y coordinate, with y-axis inverted'
-        plotterY = super().scaleY(canvasY)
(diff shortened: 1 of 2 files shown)


visidata/graph.py
```diff
@@ -12,35 +12,75 @@ def numericCols(vd, cols):
...
+        return (self.plotviewBox.ymax-plotterY_inverted)/self.yScaler + self.visibleBox.ymin
 
     @property
     def canvasMouse(self):
         p = super().canvasMouse
-        p.y = self.visibleBox.ymin + (self.plotviewBox.ymax-self.plotterMouse.y)/self.yScaler
+        p.y = self.unscaleY(self.plotterMouse.y)
         return p
 
+    def render_sync(self):
```
saulpw: I guess this is almost-but-not-exactly identical to Canvas.render_sync?  Is there any reasonable way to parameterize it for the inverted canvas, without duplicating the code?

visidata/graph.py
```diff
@@ -10,6 +10,10 @@ def numericCols(vd, cols):
 
 
 class InvertedCanvas(Canvas):
+    @asyncthread
+    def render_async(self):
+        self.plot_elements(True)
```
saulpw: This is all great, but we should probably use `invert_y=True` for good measure.

saulpw (approved):
Thanks again, this is really helpful.  I wish you'd email me so I can send you some stickers! =)

midichef:
Oops sorry, I have just rebased the squashed version onto the latest develop commit.

PR #1917, opened 2023-06-09
enable imap

justin2004:
and fix folder name extraction

before this commit running `vd "imap://someone@hotmail.com:pass123@imap-mail.outlook.com:993"`
would produce:
"no loader for url scheme: imap"

now we get:
image


saulpw (approved):
Thanks for putting this together, @justin2004 ! This looks good but I'd like to test it myself before we merge.

justin2004:
sounds good.

after this one i'm working on an OAUTH2 flow so we can do the same for gmail.

PR #1919, opened 2023-06-13
OAuth 2.0 flow for Gmail IMAP

justin2004:
This PR builds upon #1917 

enable OAuth 2.0 for IMAP with gmail
    
- add docs that help users set up a Google API App
- delete "vdplus/api/google/google-creds.json" because that only works for email addresses designated by Saul's Google API App

If you try to use the `google-creds.json` file that this commit deletes you'd get this error:
image


vdplus/api/google/__init__.py
```diff
@@ -6,17 +6,29 @@
 
 def _google_creds_fn():
     from pkg_resources import resource_filename
-    return resource_filename('vdplus.api.google', 'google-creds.json')
+    import os
+    if not os.path.exists(Path(resource_filename('vdplus.api.google', 'google-creds.json'))):
+        vd.error('google-creds.json file does not exist. create it by following this guide: https://github.com/saulpw/visidata/blob/develop/docs/gmail.md')
```
justin2004: I'm not sure how I can reference the URL in a more dynamic (context sensitive) way?

anjakefala (approved):
Thank you so much for describing the flow! It allowed me to clearly test the plugin changes.

Everything works as expected for me.

anjakefala:
My one question is what order do these PRs need to be merged in? This PR seems to have some changes related to #1917. Is it sufficient to merge this one?

visidata/loaders/imap.py
```diff
@@ -1,11 +1,12 @@
-from visidata import vd, TableSheet, asyncthread, ColumnItem, Column, ColumnAttr, Progress
+from visidata import VisiData, vd, TableSheet, asyncthread, ColumnItem, Column, ColumnAttr, Progress
+import vdplus.api.google
```
anjakefala: @saulpw Build is failing because of this line.

We are going to address it by moving the google API to visidata core. I will take care of that.

anjakefala:
This was wonderful, thank you @justin2004 !

vdplus/api/google/__init__.py
saulpw: This should show up at https://visidata.org/docs/google eventually; is that what you mean?

justin2004: Ah, yes. That is probably the best place to reference the docs at. Thanks.

justin2004:
@anjakefala , you are welcome. Happy to contribute to such a useful project.

PR #1921, opened 2023-06-16
[clipboard-] fix paste on Windows

### LINKED ISSUE #1920: Pasting from Clipboard in Windows 10
**Small description**
I cannot get any results pasted from the clipboard onto a visidata cell, selected cell or range of cells. 

On the most basic test:
- Create new sheet
- Create 1 cell
- I copy a short string from another app to the clipboard
- trying:
    - syspaste-cells or 
    - syspaste-cells-selected (while selecting a single cell), or
    - zP
- I get the following message "INFO: Type "CLIP /?" for usage."

This is a message from Windows regarding its CLIP application.

I have tried this, having multiple clipboard history activated, or deactivated. In both cases I get the same result.

Copying from visidata onto clipboard works alright with zY

**Expected result**
Pasted content in cell

If you get an unexpected error, please include the full stack trace that you get with `Ctrl-E`.
image

No error captured (no ctrl-E)

First try reproducing without any user configuration by using the flag `-N`.
e.g. `echo "abc" | vd -f txt -N`

Same result.

Please attach the commandlog (saved with `Ctrl-D`)
[paste_problem_win.zip](https://github.com/saulpw/visidata/files/11748544/paste_problem_win.zip)

 to show the steps that led to the issue.

**Additional context**
Please include the version of VisiData and Python.
Visidata v2.11
Python 3.9.7

midichef:
Closes #1920 .

### DIFF
--- a/visidata/clipboard.py
+++ b/visidata/clipboard.py
@@ -12,7 +12,7 @@ from visidata import Sheet, Path
 
 if sys.platform == 'win32':
     syscopy_cmd_default = 'clip.exe'
-    syspaste_cmd_default = 'clip.exe'
+    syspaste_cmd_default = 'powershell -command Get-Clipboard'
 elif sys.platform == 'darwin':
     syscopy_cmd_default = 'pbcopy w'
     syspaste_cmd_default = 'pbpaste'


anjakefala:
:100:

PR #1922, opened 2023-06-18
fix typo in docs

maxim-uvarov:
's' or 't' keys don't filter rows, but select them.

### DIFF
--- a/docs/rows.md
+++ b/docs/rows.md
@@ -29,7 +29,7 @@ An example usage follows.
 
 ## How to filter rows
 
-1. Press `s` or `t` on the rows to filter.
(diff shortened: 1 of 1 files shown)


docs/rows.md
```diff
@@ -29,7 +29,7 @@ An example usage follows.
 
 ## How to filter rows
 
-1. Press `s` or `t` on the rows to filter.
```
saulpw, line 32: I can see how this would be confusing, thanks for pointing it out. How about `on the rows to be filtered`?

maxim-uvarov: `on the rows to be filtered` even better! 🙏

saulpw (approved):
Thanks, @maxim-uvarov!

Issue #1843, opened 2023-04-07
Merge has surprising result


reagle:
I'm creating toy examples for an Anki deck and find unexpected behavior for merge, which I thought might mean: "keep all rows from first sheet and matching rows from other sheets, updating the first sheet's data with the first non-False value." (The intro says, "Merges differences from other sheets into first sheet"; the quick reference says, "merge  mostly keep all rows from first selected sheet, except prioritise cells with non-null/non-error values.")

```tsv
ID	Name	Age
1	Alice	25
2	Bob	30
3	Carol	
4	David	40
```

```tsv
ID	Name	Age
2		35
3		30
5	Emily	28
```

After setting ID as the key column, when making the first sheet the uppermost (`shift+k`) and merging (`& merge`), I'd expect:

```tsv
ID	Name	Age
1	Alice	25
2	Bob	35
3	Carol	30
4	David	40
```

But I get:

```tsv
ID	Name	Age
1	Alice	25
2	Bob	30
3	Carol	30
4	David	40
5	Emily	28
```

I'm surprised Emily is included and that Bob's age is not 35.

The text files are in this zip. Using v2.11.

[test.zip](https://github.com/saulpw/visidata/files/11179084/test.zip)

reagle:
BTW: the fact that Bob and Carol's name are not removed is intuitive: I imagine because empty/null value isNull or Falsey. So that's good/expected.

yphillip:
> I'm surprised Emily is included

I was looking at the example data set from the original feature request in #405. It seems the implementation matches the example provided the feature request: rows 8, 10, and 11 are **not** present in the initial table, **are** present in the new table, and then **are** present in the merged table.

I'm curious to hear what you think the correct intended behavior should be for new rows. Or maybe the documentation should be updated to clarify this behavior?

> ... and that Bob's age is not 35

I agree and also expected that Bob's age to be 35 instead of 30 in the merged table. I would be interested in taking a stab at this bugfix.

saulpw:
@yphillip If you wanted to take a stab at it, that would be much appreciated!  I'd be happy to review at any point.

reagle:
@yphillip I don't feel strongly what the behavior should be; my concern is that it do what a clear description says it will do. I think it makes sense to follow well-established behaviors/conventions (e.g., from SQL).

reagle:
I’m glad to see this addressed, but I’m not sure what the result is. Does the documentation need to be updated?

saulpw:
[...]

PR #1923, opened 2023-06-18
Join merge bug fix (proposed solution for #1843)

yphillip:
This PR is a proposed solution for the bug described in Issue #1843.

The `join-merge.vd` test is updated so that the `join-merge-1` sheet comes before the `join-merge-2` sheet, and the bug was then able to be reproduced. Before, it was the other way around, which did not accurately reflect the original use case of having the `join-merge-1` sheet be the initial original sheet.

Bug fix was to used `reversed` in the `calcValue()` function of the `MergeColumn` class so that new values are gotten from the last sheet(s) first.

### DIFF
--- a/tests/join-merge.vd
+++ b/tests/join-merge.vd
@@ -1,8 +1,8 @@
 sheet	col	row	longname	input	keystrokes	comment
-			open-file	tests/join-merge-1.jsonl	o	
 			open-file	tests/join-merge-2.jsonl	o	
-join-merge-2	colA		key-col		!	
+			open-file	tests/join-merge-1.jsonl	o	
 join-merge-1	colA		key-col		!	
-join-merge-2			join-sheets-top2	merge	&	
-join-merge-2+join-merge-1	colA		type-int		#	
-join-merge-2+join-merge-1	colA		sort-asc		[	
+join-merge-2	colA		key-col		!	
+join-merge-1			join-sheets-top2	merge	&	
+join-merge-1+join-merge-2	colA		type-int		#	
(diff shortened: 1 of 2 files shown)


yphillip:
I manually tested the data set provided in #1843 and got the expected result (Bob has age of 35 in the merged sheet), but did not explicitly add it as a new test because I figured I could reproduce the bug in the existing `join-merge.vd` test. But if you do want me to explicitly add the new test, please let me know.

saulpw:
Thanks, @yphillip!  Yes, if the existing test has been changed to elicit the bug before the change, then we don't need another test.  This all looks good to me, but I'll let @anjakefala weigh in before we merge it.

saulpw (approved):

anjakefala:
Python 3.7 tests are failing because reversed dict views only became implemented in Python 3.8 :sob: https://github.com/python/cpython/pull/6827

If we want to continue to support 3.7, what is the recommended workaround? To listify it first?

yphillip:
> Python 3.7 tests are failing because reversed dict views only became implemented in Python 3.8 😭 [python/cpython#6827](https://github.com/python/cpython/pull/6827)
> 
> If we want to continue to support 3.7, what is the recommended workaround? To listify it first?

That seems like the most straightforward solution to me too.

anjakefala:
Thanks @yphillip! =)

saulpw:
Thanks again for this, @yphillip!  If you'd like some VisiData stickers, please send me an email with your mailing address and I'd be happy to send you some.

Issue #1808, opened 2023-03-18
[develop] VisiData occassionally gets stuck when importing `requests`


anjakefala:
On `develop` branch, with Python v3.9, occassionally VisiData will fail to progress when loading from a url:

```
anja@kefala:git/visidata ‹develop*›$ vd https://visidata.org/usage.tsv
saul.pw/VisiData v2.12dev
loaded visidatarc
```

And you have to `Ctrl+C` to get the following stacktrace:

```
^CTraceback (most recent call last):
  File "/home/anja/anaconda3/envs/vd39/bin/vd", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/anja/git/visidata/bin/vd", line 6, in <module>
    visidata.main.vd_cli()
  File "/home/anja/git/visidata/visidata/main.py", line 372, in vd_cli
    rc = main_vd()
  File "/home/anja/git/visidata/visidata/main.py", line 294, in main_vd
    vs = vd.openSource(p, create=True, **opts) or vd.fail(f'could not open {p}')
  File "/home/anja/git/visidata/visidata/_open.py", line 145, in openSource
    vs = vd.openPath(Path(p), filetype=filetype)  # convert to Path and recurse
  File "/home/anja/git/visidata/visidata/_open.py", line 82, in openPath
    return openfunc(p, filetype=filetype)
  File "/home/anja/git/visidata/visidata/loaders/http.py", line 21, in openurl_http
    requests = vd.importExternal('requests')
  File "/home/anja/git/visidata/visidata/settings.py", line 459, in importExternal
    m = importlib.import_module(modname)
  File "/home/anja/anaconda3/envs/vd39/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/home/anja/anaconda3/envs/vd39/lib/python3.9/site-packages/requests/__init__.py", line 43, in <module>
    import urllib3
[...]

Issue #1704, opened 2023-01-27
Row of nulls in csv that came from csvbase


calpaterson:
**Small description**

I get a row of unexpected nulls when using a csv from csvbase

**Expected result**

No row between rows 21 and 22

**Actual result with screenshot**

![image](https://user-images.githubusercontent.com/224844/215052189-8cb8a85c-acd4-4028-9d07-0ceddc238a82.png)

**Steps to reproduce with sample data and a .vd**

```shell
vd https://csvbase.com/r-datasets/airquality
```

```
#!vd -p
{"longname": "open-file", "input": "https://csvbase.com/r-datasets/airquality", "keystrokes": "o"}
```

Please attach the commandlog (saved with `Ctrl-D`) to show the steps that led to the issue.
See [here](http://visidata.org/docs/save-restore/) for more details.

**Additional context**

Python 3.10 on debian linux

Is this a bug in visidata?  Or a bug in csvbase?

saulpw:
This is apparently a bug in the `requests` library: psf/requests#5540.

The issue is that for this particular (lucky/unlucky) csv, the `\r\n` between line 21 and 22 straddles bytes 512 and 513, which means that the first chunk ends in `\r` and the second chunk starts with `\n`.  So the iterator splits up the first chunk and yields the final line, and then when it starts to split up the second chunk, it also yields an additional empty line.

I agree with @vitidev from the requests bug, that streaming lines from an HTTP request with `iter_lines()` should act identically to iterating over lines from the same file on disk.  To fix this bug, we'll have to remove the `requests` dependency and rewrite the streaming functionality ourselves.

anjakefala:
We switched over to `urllib` which appears to have fixed this! Please re-open, if you are still running into problems.

calpaterson:
Nice, I was able to reproduce this on v2.11 and not with the latest git.  Many thanks!!!

PR #1924, opened 2023-06-18
[http] fix parsing link header

### LINKED ISSUE #1898: [urllib imports]  Multiple urllib imports are missing
**Small description**
Missing imports for `urllib.error` and `urllib.parse`

**Expected result**
No stack traces from missing imports.

First Trace
```
  File "/usr/lib/python3.9/site-packages/visidata/threads.py", line 200, in _toplevelTryFunc
  File "/usr/lib/python3.9/site-packages/visidata/threads.py", line 219, in _func
    func(*args, **kwargs)
  File "/usr/lib/python3.9/site-packages/visidata/plugins.py", line 97, in reload
    except urllib.error.URLError as e:
AttributeError: module 'urllib' has no attribute 'error'
```

Second trace
```
...
  File "/usr/lib/python3.9/site-packages/visidata/loaders/http.py", line 50, in _iter_lines
    links = urllib.parse.parse_header(linkhdr)
AttributeError: module 'urllib.parse' has no attribute 'parse_header'

```

Reproduce: 
```
vd https://blog.nature.org/2021/04/12/beaver-otter-muskrat-a-field-guide-to-freshwater-mammals/
```

Please attach the commandlog (saved with `Ctrl-D`) to show the steps that led to the issue.

**Additional context**
Please include the version of VisiData. Latest develop version of VisData.  Python 3.9.2

midichef:
Closes #1898.

I could not reproduce the error `AttributeError: module 'urllib' has no attribute 'error'` but I have attempted to patch it anyway by changing the import from urllib to urllib.error.

This commit uses `requests`. I am not sure if I did the import of `requests.utils` properly. I do `import requests.utils`
[but I see there's some reason not to mix](https://github.com/saulpw/visidata/issues/1808#issuecomment-1492221893) `import` and `vd.importExternal`, and I see that it is done using `importExternal` in  https://github.com/saulpw/visidata/blob/fb4352cfcf5ffb16431d22409ef73642e281555a/visidata/loaders/scrape.py#L150

Also I know [we just got rid of most uses of requests because of a bug they refuse to fix](https://github.com/saulpw/visidata/issues/1704#issuecomment-1407250046). I am only using one simple function from it. But maybe we'd rather not use requests at all?

In any case, I have tested that this commit fixes pagination by loading `vd https://api.github.com/repos/django/django/pulls` with `options.http_max_next = 2` and confirming that I see 3*30 = 90 rows.

### DIFF
--- a/visidata/loaders/http.py
+++ b/visidata/loaders/http.py
@@ -20,6 +20,7 @@ def openurl_http(vd, path, filetype=None):
         return openfunc(Path(schemes[-1]+'://'+path.given.split('://')[1]))
 
     import urllib.request
+    import requests.utils
     import urllib.error
     import mimetypes
 
@@ -47,19 +48,24 @@ def openurl_http(vd, path, filetype=None):
             linkhdr = response.getheader('Link')
             src = None
             if linkhdr:
-                links = urllib.parse.parse_header(linkhdr)
-                src = links.get('next', {}).get('url', None)
+                links = requests.utils.parse_header_links(linkhdr)
+                link_data = {}
+                for link in links:
+                    key = link.get('rel') or link.get('url')
+                    link_data[key] = link
+                src = link_data.get('next', {}).get('url', None)
 
             if not src:
                 break
 
             n += 1
             if n > max_next:
-                vd.warning(f'stopping at max {max_next} pages')
+                vd.warning(f'stopping at max next pages: {max_next} pages')
                 break
 
             vd.status(f'fetching next page from {src}')
-            response = requests.get(src, stream=True, **vd.options.getall('http_req_'))
+            req = urllib.request.Request(src, **vd.options.getall('http_req_'))
+            response = urllib.request.urlopen(req)
 
     # add resettable iterator over contents as an already-open fp
     path.fptext = RepeatFile(_iter_lines())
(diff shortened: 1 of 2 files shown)


visidata/loaders/http.py
```diff
@@ -47,19 +48,24 @@ def _iter_lines(path=path, response=response, max_next=vd.options.http_max_next)
...
                 break
 
             n += 1
             if n > max_next:
-                vd.warning(f'stopping at max {max_next} pages')
+                vd.warning(f'stopping at max next pages: {max_next} pages')
                 break
 
             vd.status(f'fetching next page from {src}')
-            response = requests.get(src, stream=True, **vd.options.getall('http_req_'))
```
saulpw, line 62: Hm, this is weird, were we still trying to use the `requests` library without importing it?

saulpw:
You're right, at this point I would prefer not to use `requests` at all.  Can we copy or inline the function you're using?

visidata/loaders/http.py
midichef: Yes, there had previously been two lines with requests.get(), but only the first one had been changed to use urllib.request.urlopen().

saulpw (approved):
Thanks, @midichef, looks good to me.

Issue #1696, opened 2023-01-23
[sheets] cursorDisplay for the sheet's own cursorDisplay crashes in Python 3.8.0 through 3.9.9


midichef:
**Small description**
In SheetsSheet, viewing the cursorDisplay cell for the SheetsSheet leads to infinite recursion, which crashes on versions of Python >= 3.8.0 and <= 3.9.9

**Expected result**
The cursorDisplay should be blank or show a null.

**Actual result with screenshot**
```
Fatal Python error:
_Py_CheckRecursiveCall: Cannot recover from stack overflow.
Python runtime state: initialized

Current thread 0x00007f5f79c5fb80 (most recent call first):
  File "/home/midichef/.pyenv/versions/3.9.9/lib/python3.9/site-packages/visidata/column.py", line 308 in getValue
  File "/home/midichef/.pyenv/versions/3.9.9/lib/python3.9/site-packages/visidata/wrappers.py", line 108 in wrapply
  File "/home/midichef/.pyenv/versions/3.9.9/lib/python3.9/site-packages/visidata/column.py", line 327 in getCell
  File "/home/midichef/.pyenv/versions/3.9.9/lib/python3.9/site-packages/visidata/column.py", line 388 in getDisplayValue
  File "/home/midichef/.pyenv/versions/3.9.9/lib/python3.9/site-packages/visidata/sheets.py", line 453 in cursorDisplay
  File "/home/midichef/.pyenv/versions/3.9.9/lib/python3.9/site-packages/visidata/column.py", line 460 in getattrdeep
  File "/home/midichef/.pyenv/versions/3.9.9/lib/python3.9/site-packages/visidata/column.py", line 497 in <lambda>
  File "/home/midichef/.pyenv/versions/3.9.9/lib/python3.9/site-packages/visidata/column.py", line 270 in calcValue
  File "/home/midichef/.pyenv/versions/3.9.9/lib/python3.9/site-packages/visidata/column.py", line 308 in getValue
  File "/home/midichef/.pyenv/versions/3.9.9/lib/python3.9/site-packages/visidata/wrappers.py", line 108 in wrapply
...
```

**Steps to reproduce with sample data and a .vd**
Open the SheetsSheet and move to the cursorDisplay column, or run:
`vd -p cursordisplay.vdj`
[cursordisplay.vdj.txt](https://github.com/saulpw/visidata/files/10484256/cursordisplay.vdj.txt)

**Additional context**
saul.pw/VisiData v2.11

[...]

PR #1927, opened 2023-06-20
[getattrdeep-] remove unnecessary check

midichef:
In 6b8deca3def66a44ee66a818592198886834c9c2, I saw this note in the log
> - this will make every ItemColumn/AttrColumn slower :(

That slower check has since become unnecessary, as the problem it was attempting to work around, was later fixed directly by
https://github.com/saulpw/visidata/commit/bfdfeb3e139fa4d0603b90eaae8036c359579b7b .
(details of how that worked are here https://github.com/saulpw/visidata/issues/1696#issuecomment-1416659518 )

I've been running Visidata without the check for for a few months. It seems fine. (Though I tested it under python 3.10, not 3.8/3.9 where the recursion errors were seen). The :( can be changed to a :)

### DIFF
--- a/visidata/utils.py
+++ b/visidata/utils.py
@@ -95,8 +95,7 @@ def getattrdeep(obj, attr, *default, getter=getattr):
             return getter(obj, attr, *default)
 
         try:  # if attribute exists, return toplevel value, even if dotted
-            if attr in obj:
-                return getter(obj, attr)
+            return getter(obj, attr)
         except RecursionError:  #1696
             raise
         except Exception as e:


anjakefala:
(@midichef the link to the commit for the recursion crash fix was 404ing. I attempted to update the link. Is that the accurate commit you were referring to?)

anjakefala:
Hi @midichef! =)

The tests seem to be legitimately failing: with the expected "golden output" not matching. 

I will be able to allocate some time to investigate why later, but you could probably get a headstart. Do you need any guidance for understanding how the tests run or work?

midichef:
I was able to figure out tests thanks to `docs/test.md`. I'll take a look at what's going on.

midichef:
I wasn't able to reproduce this error on 3.7. I'll close this for now and reopen when I can explore it.

PR #1931, opened 2023-06-22
[graph] labels: add tick symbol, int precision, right margin

midichef:
One bugfix, and a few improvements to the graph labels.

First is a bugfix to an earlier commit of mine, fixing an error during drawing.

Next, if the graph is graphing integer data, the labels are currently formatted as ints. When zoomed in, seeing axis labels like "1    1    1    1" is not so informative, so this change formats them as floats: "1.0   1.2   1.4  1.6". This only happens when zoomed in, not at the default zoomlevel or when zoomed out.

I also added tick symbols to the x-axis, to clarify exactly which column the label applies to. The symbol defaults to `╵` (Box Drawing Light Up). It's particularly helpful when graphing with dates on the x-axis, since they are wide. It's also helpful at the right edge of the graph, where labels may shift to avoid overrunning the right edge of the window. When that happens, I've changed the behavior so the rightmost label moves entirely to the left side of its tick symbol.

And the last change is to make the right margin expand dynamically, to be as large as necessary to fit the graph legend. This keeps the graph legend from covering up any plotted data points. But I capped the right margin so it takes up at most 1/4th of the window size.

### DIFF
--- a/visidata/canvas.py
+++ b/visidata/canvas.py
@@ -169,7 +169,8 @@ class Plotter(BaseSheet):
         self.labels.append((x, y, text, attr, row))
 
     def plotlegend(self, i, txt, attr=0, width=15):
-        self.plotlabel(self.plotwidth-width*2, i*4, txt, attr)
+        # move it 1 character to the left b/c the rightmost column can't be drawn to
+        self.plotlabel(self.plotwidth-(width+1)*2, i*4, txt, attr)
 
     @property
     def plotterCursorBox(self):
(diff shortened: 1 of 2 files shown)


saulpw:
Thanks for pulling this together, @midichef.  About this:

> When zoomed in, seeing axis labels like "1 1 1 1" is not so informative

I agree with this in principle.  Special-casing this for int columns makes me uneasy though.  How many decimal places should be shown?  If we show a "reasonable" number like 2, then for high zoom we'll just see `1.03 1.03 1.03 1.03` which is also not useful.  But figuring out the right format string would be overly complicated and maybe not even possible for some users or use cases.  So this is why I punted and just used the type/format string of the column itself.  Particularly for dates.

If we're going the heuristic route for ints, though, can we make a function like `isNumeric` (maybe `vd.isIntegerCol(col)` and vd.isNumeric could become `vd.isNumericCol(col)`) that detects integral column types?  For the moment it can just `return col.type is int` but this would allow it to be overridden if necessary.  More importantly it would give us something to search for when looking for these cases.

saulpw (approved):

anjakefala:
Tests are failing due to problems with debian. This PR is passing fine.

PR #1933, opened 2023-06-24
[mysql] unquote password before sending to client

dufferzafar:
so we can %encode passwords with symbols like # : in them


saulpw:
Thanks @dufferzafar!  I didn't know that urlparse didn't decode the components.  We might want to do this for other components as well (and for other loaders that use urlparse), but for now we can merge this as-is for a functional improvement.

Issue #1805, opened 2023-03-13
Add read-only flag


fwdekker:
## The problem
I personally only use VisiData to inspect the contents of SQLite databases used by software on my server, as this is much more secure and requires less maintenance than using a public-facing web interface. These databases may contain logs or user accounts, and I only need to read them; all the writing and updating is handled by the software that created the database. However, I'm not all too familiar with all the shortcuts that VisiData has, and sometimes I press one by accident, and suddenly a table is marked for deletion, at which point I need to make very sure I don't confirm this change, and should immediately quit the application.

## The solution
It would give me significant peace of mind if I could run `vd --read-only ./my_database.db` and then rest assured that no changes are ever written back to disk. An alternative would be to run `sudo -u myspecialuser vd ./my_database.db` and ensure that the user `myspecialuser` has read access but not write access to the file, but unfortunately my databases need a specific user and group assigned to it, and Linux and *BSD do not have fine-grained access control.

## The implementation
I can imagine several ways of implementing a read-only mode for VisiData, and they're all fine with me:
1. The [`nano`](https://en.wikipedia.org/wiki/GNU_nano) way: Similar to when a user runs `nano file.txt` and they have read but not write access to `file.txt`, the editor shows `File file.txt is unwritable` in the bottom, but still allows the user to perform any changes they want inside the editor. However, the editor gives an error when the user tries to save the file.
2. The other `nano` way: Similar to when a user runs `nano -v file.txt` or `nano --view file.txt`, VisiData refuses actions that would edit the database (such as deleting a table, inserting a new row, opening the editor for a record, etc.), showing the message `Key is invalid in view mode`.
3. The cheaty way: VisiData copies the database to `/tmp/visidata-<random>/database.db` and then opens the copy. The user can write changes, but they are stored to the temporary file, which is deleted afterwards, so the changes do not affect the original database. This can potentially be confusing for the user, though, and I'm not sure what the security implications are of copying a database to a directory where other users could potentially see it.

saulpw:
Hey @FWDekker, thanks for the suggestion.  There are only two commands that could conceivably write to a file, `Ctrl+S` and `z Ctrl+S`. It would be pretty straightforward to disable these commands; I think that would be basically your option (1).  Option (2) would be more difficult, since there are so many ways to modify data.  Option (3) is not really an option though, for several reasons.

saulpw:
[...]

PR #1945, opened 2023-07-02
[sqlite-] prevent creation of ./- file when reading from stdin

midichef:
If visidata tries to read a SQLite file from standard input, it will load an empty sheet and create the file `-` in the current directory. This can be seen with `cat sample_data/employees.sqlite |vd -f sqlite3`

That's because the path for stdin gets resolved as the file `-` in the current directory.
As far as I can tell, there's no way to read an SQLite file from standard input, at least not with the Python `sqlite` library. When I try it, using `sqlite.connect()` with `'/dev/stdin'`, I get this error
```
  File "/home/midichef/.local/lib/python3.10/site-packages/visidata/threads.py", line 200, in _toplevelTryFunc
  File "/home/midichef/.local/lib/python3.10/site-packages/visidata/sheets.py", line 235, in reload
    for r in self.iterload():
  File "/home/midichef/.local/lib/python3.10/site-packages/visidata/loaders/sqlite.py", line 179, in iterload
    for row in SqliteSheet.iterload(self):
  File "/home/midichef/.local/lib/python3.10/site-packages/visidata/loaders/sqlite.py", line 84, in iterload
    with self.conn() as conn:
  File "/home/midichef/.local/lib/python3.10/site-packages/visidata/loaders/sqlite.py", line 58, in conn
    con = sqlite3.connect(url, uri=True, **self.options.getall('sqlite_connect_'))
sqlite3.OperationalError: disk I/O error
```

So this commit just disables reading any sqlite file from standard input.

### DIFF
--- a/visidata/loaders/sqlite.py
+++ b/visidata/loaders/sqlite.py
@@ -25,6 +25,8 @@ def guess_sqlite(vd, p):
 
 @VisiData.api
 def open_sqlite(vd, p):
+    if p is vd.stdinSource:
+        vd.fail('cannot read a sqlite database from stdin')
     return SqliteIndexSheet(p.name, source=p)
 
 @VisiData.api


saulpw (approved):
Heh, okay.  How did you find this one?  I never would have thought to pipe a sqlite db into vd.

midichef:
Yeah, it's an odd thing to try! I was trying to find a hack to effectively make the sqlite db read-only, in response to https://github.com/saulpw/visidata/issues/1805

Issue #1910, opened 2023-06-03
[loaders]  Support builtin data formats


frosencrantz:
This is meant to be an aspirational wishlist request.  For developers who would like to contribute, this can be used as a map to what loaders could be added.

In https://github.com/saulpw/visidata/issues/1587#issuecomment-1369263933 it was suggested that:

> When Python includes the batteries, VisiData should as well.

This was within the context of talking about the new toml module data format.  If Python supports a data format it would be reasonable to expect that VisiData also supports that data format.

So I asked a generative chatbot what data formats the Python Standard Library module supports.  Below is the answer, some of these may work better as column formats than file formats.  And it likely missed some formats.  

There are some formats that are deprecated as described in [PEP 594] that do not make sense to include since they will eventually be removed from Python.

Other formats are useful as building blocks for other formats, like the email parsers can be used for email, but are also useful for things like Python METADATA files  (e.g. [Source distribution format])

[Source distribution format]: https://packaging.python.org/en/latest/specifications/source-distribution-format/)

[PEP 594]: https://peps.python.org/pep-0594

| Format | Module | Description | Documentation |
|--------|--------|-------------|---------------|
| CSV | csv | Comma Separated Values | [CSV Module Documentation](https://docs.python.org/3/library/csv.html) |
| JSON | json | JavaScript Object Notation | [JSON Module Documentation](https://docs.python.org/3/library/json.html) |
| XML | xml.etree.ElementTree | eXtensible Markup Language | [XML Etree Module Documentation](https://docs.python.org/3/library/xml.etree.elementtree.html) |
| HTML | html | Hypertext Markup Language | [HTML Module Documentation](https://docs.python.org/3/library/html.html) |
| INI | configparser | Configuration file format | [ConfigParser Module Documentation](https://docs.python.org/3/library/configparser.html) |
| Pickle | pickle | Python object serialization | [Pickle Module Documentation](https://docs.python.org/3/library/pickle.html) |
| ZIP | zipfile | Archive file format | [Zipfile Module Documentation](https://docs.python.org/3/library/zipfile.html) |
| TAR | tarfile | Archive file format | [Tarfile Module Documentation](https://docs.python.org/3/library/tarfile.html) |
| GZIP | gzip | Compression format | [Gzip Module Documentation](https://docs.python.org/3/library/gzip.html) |
| BZ2 | bz2 | Compression format | [BZ2 Module Documentation](https://docs.python.org/3/library/bz2.html) |
[...]

PR #1946, opened 2023-07-02
[types] add ipaddr and ipnet types

### LINKED ISSUE #1782: Type as: IPv4
It would be helpful to be able to cast values as an IPv4 datatype, and then allow filtering of that data in CIDR notation. Alternatively if someone has a suggestion on how to achieve this currently, I would be very grateful!

Thank you kindly for developing such a fantastic tool.

ajkerrigan:
- Handle IPv4 and IPv6 addresses
- Include a `select-supernets` command to select rows containing an input network

This is _mostly_ a copy/paste of the example [here](https://github.com/saulpw/visidata/issues/1782#issuecomment-1613397296), plus a suggestion from Saul and a couple docstrings / safety checks / style tweaks that seem to make it more suitable for inclusion in core. Suggestions for alternatives/additions/further tweaks welcome of course. 

Closes #1782, and also relates to #1910


visidata/features/type_ipaddr.py
```diff
@@ -0,0 +1,64 @@
...
+TableSheet.addCommand(
+    None,
+    "type-ipnet",
+    "cursorCol.type=ipaddress.ip_network",
+    "set type of current column to IP network",
+)
+TableSheet.addCommand(
+    None,
+    "select-supernets",
+    'cursorCol.selectSupernets(inputExpr("ip or cidr block: "))',
```
saulpw: We should use `input` rather than `inputExpr`, since it's not an expression (and won't benefit from e.g. tab-completion of column names).  Also let's add a `type="ip"` so that the history can be limited to ip addresses.

ajkerrigan: Ah right, good call. I also need to dig into the test failures which at a glance seem to stem from assuming that ipaddress will be globally available.

saulpw: `addGlobals(ipaddress=ipaddress)` should cover that.

saulpw (approved):
Looks good, thanks for putting this together, @ajkerrigan !

PR #1949, opened 2023-07-05
[freqtbl-] Default disp_histogram to U+25A0 BLACK SQUARE (■)

daviewales:
See discussion here:
https://github.com/saulpw/visidata/discussions/1807#discussioncomment-6362568

Copying my comment from the discussion above below:

So there I was, innocently reading through the [Notcurses](https://nick-black.com/htp-notcurses.pdf) book, when I encountered the following paragraph in section 7.6 _Stupid Unicode Tricks_:

> Those readers who remember Code Page 437 might recall 0xFD, a “middle half block”. Given a constant
> background, this can be used with the other two half blocks to increase “vertical resolution” by a factor of 3
> (imagine a tank drawn with upper, middle, and lower half blocks moving on a monochromatic background).
> Well, that character was actually “Solid Square/Histogram/Square Bullet” in the CP437 definitions[31], and
> lives on as U+25A0 BLACK SQUARE (■). Despite its name, it will take on the foreground color.

I was already thinking of the visidata histogram, and when I read the words "middle half block", I had to try it.
Here is the result:

**Before**:
Screen Shot 2023-07-05 at 10 56 58 pm

**After**:
Screen Shot 2023-07-05 at 10 29 29 pm

It's not _perfect_, because the block is not _quite_ in the vertical middle, and there's a small gap between each character.
But I think it's pretty close to the best we can get for default visidata, because there's vertical space both above and below which avoids problems with the reverse video line highlight.
And it's a solid improvement over `*`.

@saulpw, would you consider making this the default character for `options.disp_histogram`?
(Bonus points if we can squeeze it into 2.11.1!)

### DIFF
--- a/visidata/freqtbl.py
+++ b/visidata/freqtbl.py
@@ -5,7 +5,7 @@ from visidata import vd, asyncthread, vlen, VisiData, Column, AttrColumn, Sheet,
 from visidata.pivot import PivotSheet, PivotGroupRow
 
 
-vd.option('disp_histogram', '*', 'histogram element character')
+vd.option('disp_histogram', '■', 'histogram element character')
 vd.option('histogram_bins', 0, 'number of bins for histogram of numeric columns')
 vd.option('numeric_binning', False, 'bin numeric columns into ranges', replay=True)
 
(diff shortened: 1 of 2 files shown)


saulpw:
I like it!!   Nice find.  Please add the old `*` to the `ascii8` theme and then I think it's reasonable to take it for 2.11.1. 
 https://github.com/saulpw/visidata/blob/develop/visidata/themes/ascii8.py

PR #1955, opened 2023-07-09
[path] fix open() using both binary and text mode

midichef:
The switch from `open_bytes()` to use the scaffolding in what was `open_text()` led to PNG files being opened in `mode='rbt'`:

`vd sample_data/vdlogo-160x24.png` gives this error
```
  File "/home/midichef/.local/lib/python3.8/site-packages/visidata/path.py", line 288, in read_bytes
    with self.open(mode='rb') as fp:
  File "/home/midichef/.local/lib/python3.8/site-packages/visidata/path.py", line 224, in open
    return self._open(mode=mode, encoding=encoding or vd.options.encoding, errors=vd.options.encoding_errors, newline=newline)
  File "/home/midichef/.local/lib/python3.8/site-packages/visidata/path.py", line 264, in _open
    return FileProgress(path, fp=self._path.open(*args, **kwargs), **kwargs)
  File "/usr/lib/python3.8/pathlib.py", line 1222, in open
    return io.open(self, mode, buffering, encoding, errors, newline,
ValueError: can't have text and binary mode at once
```


midichef:
Also I forgot, the description of open() should be changed to mention binary mode, too. Right now it only mentions text.
https://github.com/saulpw/visidata/blob/e70b4ed40681f6422d9d9a9810eedc9174c0931f/visidata/path.py#L196

anjakefala:
Hi @midichef!

I made an additional change that it looks like binary-mode needed. I also added a very basic load-png test (so it can catch binary loading stuff earlier), and added the docstring you mentioned! I think this should be good to merge now.

Thank you so much!

PR #1958, opened 2023-07-12
[graph] fix legend display of full-width characters

midichef:
When a graph legend has a full-width character, its length is calculated incorrectly, and it gets truncated with "…". You can see the difference if you graph the following TSV data with `g.`
```
x	y_normalwidth	Ｙ_fullwidth
0	0	0
```
The label for the string "Ｙ_fullwidth", which starts with a full-width character, will be shown as "Ｙ_fullwid…". This commit replaces `len()` with `dispwidth()` to handle these characters properly.

### DIFF
--- a/visidata/canvas.py
+++ b/visidata/canvas.py
@@ -305,7 +305,7 @@ class Plotter(BaseSheet):
                 for o, fldraw in line:
                     if fldraw:
                         char_x, char_y, txt, attr, row = o
-                        clipdraw(scr, char_y, char_x, txt, attr, len(txt))
+                        clipdraw(scr, char_y, char_x, txt, attr, dispwidth(txt))
 
 
 # - has a cursor, of arbitrary position and width/height (not restricted to current zoom)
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Ah, yup!  We probably need an audit of all `len()` to determine which ones should be `dispwidth()`.

midichef:
Good idea!There are around 340 instances of `len()`, and they reduce down to about 100 unique names that aren't for collections of rows, sheets, columns, etc. In about 10 minutes I turned up several that need fixing. Including one I wrote several weeks ago that was a few lines above this commit...

I'll tackle this more systematically in the next month or two.

midichef:
I wrote:
> I'll tackle this more systematically in the next month or two.

It's been seven months, but I have not forgotten about auditing `len()` vs `dispwidth()`. Other fixes have been more important, and that's still the case now...

saulpw:
I wonder if we couldn't shorten the list of `len()` calls, by only looking at the ones which are near `clipdraw()` or `addstr`.

PR #1960, opened 2023-07-15
Fix zsh completion

### LINKED ISSUE #1959: Current HEAD zsh-completion.py needs option_aliases update
**Small description**

`option_aliases` was removed in ce497f444db6d2f3fc0b8309f5ca839196c33c8b but is still referred to in the zsh completion code.

https://github.com/saulpw/visidata/blob/34808745232e798b0f25e893bb444fc9f3c034eb/dev/zsh-completion.py#L11C41-L11C41

I think the script needs a slight rejig to use the (present) `vd` import instead.

I wonder whether this can be included in future CI?

**Expected result**

The command succeeds.

**Actual result**

```
       > /build/visidata-src
       > Traceback (most recent call last):
       >   File "/build/visidata-src/dev/zsh-completion.py", line 11, in <module>
       >     from visidata.main import option_aliases
       > ImportError: cannot import name 'option_aliases' from 'visidata.main' (/build/visidata-src/visidata/main.py)
```

**Steps to reproduce**

```
python dev/zsh-completion.py
```

**Additional context**
~~Please include the version of VisiData and Python.~~

https://github.com/saulpw/visidata/tree/34808745232e798b0f25e893bb444fc9f3c034eb but I listed the commit above that causes the breakage — I suspect this is a two minute fix for somebody familiar with the codebase, though not me. I can help with extending CI, though it might just be a case of adding

```yaml
    - name: Ensure VisiData can create completions
      run: python dev/zsh-completion.py
```

(I guess you might want to run a linter, instead.)

dbaynard:
Fixes #1959

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -66,3 +66,5 @@ jobs:
         sudo apt install aha
         dev/mkman.sh
 
+    - name: Ensure VisiData can create completions
+      run: python -v dev/zsh-completion.py
--- a/dev/zsh-completion.py
+++ b/dev/zsh-completion.py
@@ -8,7 +8,6 @@ import re
 
 sys.path.insert(0, dirn(dirn((os.path.abspath(__file__)))))
 from visidata import vd
-from visidata.main import option_aliases
 
 ZSH_COMPLETION_FILE = "_visidata"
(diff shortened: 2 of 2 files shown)


anjakefala:
Wonderful, thank you!

PR #1962, opened 2023-07-16
[misc-] remove trailing commas from addCommand

midichef:
Just a little cleanup.


saulpw (approved):
Heh, those must have been there for quite awhile!  Thanks for the cleanup.

Issue #1953, opened 2023-07-08
Can't open the clickhouse example 


maxim-uvarov:
This thread started here: https://github.com/saulpw/vdsql/issues/33

@anjakefala, thank you for the quick answer.

> * vdsql has been moved to the [core VisiData repo](https://github.com/saulpw/visidata/tree/develop/visidata/apps/vdsql). Please reinstall it from there. There are lots of updates there to fix the problems you are seeing with the clickhouse loader.

> * you will need at least Python 3.9. [ibis-framework](https://github.com/ibis-project/ibis), the most important dependency of the project, dropped support for https://github.com/ibis-project/ibis/pull/6491.
> * The test url is now clickhouse://play:clickhouse@play.clickhouse.com/?secure=1

I tried Python 3.9 and 3.10. I tried to use a Linux machine with an Intel processor.

I installed vd: `pip3 install git+https://github.com/saulpw/visidata.git@develop`

I tried to connect to another clickhouse instance (which I was able to connect succesfuly using DBeaver).

Nothing worked. What else can I try?

Suggested by you address gives me this error even before showing me a list of tables:

`vd -f ibis clickhouse://play:clickhouse@play.clickhouse.com/?secure=1` 

<img width="1322" alt="image" src="https://github.com/saulpw/visidata/assets/4896754/cf7821a3-14c9-4750-930e-4f8443a0b474">

and 

`vd -f ibis clickhouse://explorer@play.clickhouse.com:9440/?secure=0`

gives me the error when I open the `recipes` table:

```
Traceback (most recent call last):
  File "/Users/user/miniconda3/envs/p310/lib/python3.10/site-packages/visidata/threads.py", line 200, in _toplevelTryFunc
    t.status = func(*args, **kwargs)
  File "/Users/user/miniconda3/envs/p310/lib/python3.10/site-packages/visidata/sheets.py", line 235, in reload
    for r in self.iterload():
  File "/Users/user/miniconda3/envs/p310/lib/python3.10/site-packages/vdsql/clickhouse.py", line 39, in iterload
    for row in self.query_result:
  File "/Users/user/miniconda3/envs/p310/lib/python3.10/site-packages/clickhouse_driver/util/helpers.py", line 16, in chunks
    item = list(islice(it, n))
  File "/Users/user/miniconda3/envs/p310/lib/python3.10/site-packages/clickhouse_driver/client.py", line 213, in iter_receive_result
    for rows in result:
[...]

PR #1966, opened 2023-07-20
[vdsql] replace execute_iter for clickhouse/ibis 6 compatibility

ajkerrigan:
ibis 6 moved away from the clickhouse-driver project in favor of clickhouse-connect. The latter client does not provide an `execute_iter` method, though it _does_ provide a [row-based stream method](https://github.com/ClickHouse/clickhouse-connect/blob/8621aeafa124edd0008949da96edc101b7b8c54f/clickhouse_connect/driver/client.py#L233-L250) which seems to address the same general use case.

I wouldn't say I'm _confident_ this is the right solution to #1953 - it still may need some fine-tuning but it sounds like a reasonable start. More productive than a shrug at least :laughing: . Thanks to @cpcloud for providing helpful context around the underlying ibis changes.

### DIFF
--- a/visidata/apps/vdsql/clickhouse.py
+++ b/visidata/apps/vdsql/clickhouse.py
@@ -34,13 +34,13 @@ class ClickhouseSheet(IbisTableSheet):
 
                 with Progress(gerund='clickhousing', sheet=self) as prog:
                     settings = {'max_block_size': 10000}
-                    self.query_result = con.con.execute_iter(sqlstr, settings, chunk_size=1000, query_id=qid)
-                    qid = str(time.time())
-                    for row in self.query_result:
-                        prog.total = con.con.last_query.progress.total_rows
-                        prog.made = con.con.last_query.progress.rows
+                    with con.con.query_rows_stream(sqlstr, settings) as stream:
+                        prog.total = int(stream.source.summary['total_rows_to_read'])
+                        prog.made = 0
+                        for row in stream:
+                            prog.made += 1
+                            yield row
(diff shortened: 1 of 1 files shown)


visidata/apps/vdsql/clickhouse.py
```diff
@@ -34,13 +34,13 @@ def iterload(self):
 
                 with Progress(gerund='clickhousing', sheet=self) as prog:
                     settings = {'max_block_size': 10000}
-                    self.query_result = con.con.execute_iter(sqlstr, settings, chunk_size=1000, query_id=qid)
-                    qid = str(time.time())
-                    for row in self.query_result:
-                        prog.total = con.con.last_query.progress.total_rows
-                        prog.made = con.con.last_query.progress.rows
+                    with con.con.query_rows_stream(sqlstr, settings) as stream:
+                        prog.total = int(stream.source.summary['total_rows_to_read'])
```
ajkerrigan, line 38: For context, if I look at `stream.source.summary` when we enter this block after opening the `benchmark_results` table it looks like this:

```json
{
    "read_rows": "16384",
    "read_bytes": "360448",
    "written_rows": "0",
    "written_bytes": "0",
    "total_rows_to_read": "500",
    "result_rows": "0",
    "result_bytes": "0",
    "query_id": "c413bc6a-5bd7-40a5-81bc-a64ef53dc9c8"
}
```

anjakefala:
Thanks so much for looking into this @ajkerrigan! And to @cpcloud for explaining the problem. =)

PR #1967, opened 2023-07-21
[docs] add line break to remove unintended heading

asfaltboy:
This currently looks wrong

image

Btw, I noticed that [the rendered style](https://www.visidata.org/docs/columns/) looks off in some cases:
1. No number for numbered lists
2. Headings are not visible, at least for H6 (`######`)

### DIFF
--- a/docs/columns.md
+++ b/docs/columns.md
@@ -140,6 +140,7 @@ or set in the **Options Sheet**.
 1. Press `Shift+O` to open the **Options Sheet**.
 2. Move the cursor down to the relevant *disp_date_fmt* option.
 3. Type `e` followed by *%m/%d/%Y*.
+
 ---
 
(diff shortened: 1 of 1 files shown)


saulpw:
Ah, thanks for being willing to submit a PR, @asfaltboy!  You're also right about the other two points.  @lkhrs may have a simple fix, otherwise feel free to submit another PR if you are motivated.

lkhrs:
Hi all, I opened visidata/visidata.org#88 to track these.

1. The ordered list issue is valid
2. The heading issue is also valid, however:
- I think it's best to use headings semantically, and the H6s used here are skipping 3 tiers—they should be H3s. Totally understand if it's a style choice, and I'll let @saulpw make the call on formatting style :)
- I will be adding styles for H6s just in case

PR #1968, opened 2023-07-21
migrate pgk_resources to importlib_resources

### LINKED ISSUE #1911: (Unnecessary) hidden dependency on setuptools; debian package broken
**Small description**

When installing visidata from the package-manager without having python3-setuptools installed, the following error appears when starting visidata:

```
  File "/usr/bin/vd", line 3, in <module>
    import visidata.main
  File "/usr/lib/python3/dist-packages/visidata/__init__.py", line 102, in <module>
    import visidata.main
  File "/usr/lib/python3/dist-packages/visidata/main.py", line 19, in <module>
    from pkg_resources import resource_filename
ModuleNotFoundError: No module named 'pkg_resources'
```

**Expected result**

visidata starts normally

**Additional context**
Python 3.11.2
Visidata 2.11
Debian Sid ARM64 on Crostini (Chromebook)

The dependency is only invoked in main.py and help.py, mostly for displaying the provided help file. I don't think this justifies importing the whole setuptools universe, as the function is only used to find files in the installation directory.

**Possible fixes**
This has previously been fixed for Arch by adding setuptools to the dependencies.
Long term fix is probably not to rely on resource_filename instead, a possible backup for when setuptools is not installed is using something like https://stackoverflow.com/questions/5137497/find-the-current-directory-and-files-directory to figure out the directory the files are in to invoke the right help files.

zormit:
## Motivation
While playing around with the codebase, I saw a warning pop up (weirdly I'm having trouble to reproduce it right now) that `pgk_resources` is deprecated. It's not an important thing, but I took it as a small first issue to get a bit used to the codebase and the contribution mechanism again :)

## Approach

When following the link in the warning, you'll be able to find a [migration guide](https://importlib-resources.readthedocs.io/en/latest/migration.html#pkg-resources-resource-filename), which I partly followed.

## Side-Effects

This _probably_ closes #1911, because the dependency on `pkg_resources` is gone.

## Open Questions
- [ ] I'm not sure whether I should have called `importlib.resources.as_file` in some of the places.
- [ ] It's a bit tricky because the state of things differs in the different python versions. I still have to find the right fix for that.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,6 +25,7 @@ setup(name='visidata',
           'python-dateutil',
           'windows-curses<2.3.1; platform_system == "Windows"',  #1841
           'importlib-metadata >= 3.6',
+          'importlib_resources; python_version<"3.9"'
       ],
       packages=['visidata', 'visidata.loaders', 'visidata.vendor', 'visidata.tests', 'visidata.ddw', 'visidata.man', 'visidata.themes', 'visidata.features', 'visidata.experimental', 'visidata.apps', 'visidata.apps.vgit', 'visidata.apps.vdsql', 'visidata.desktop'],
       data_files=[('share/man/man1', ['visidata/man/vd.1', 'visidata/man/visidata.1']), ('share/applications/', ['visidata/desktop/visidata.desktop'])],
--- a/visidata/__init__.py
+++ b/visidata/__init__.py
@@ -14,10 +14,13 @@ class EscapeException(BaseException):
 
 def addGlobals(*args, **kwargs):
     '''Update the VisiData globals dict with items from *args* and *kwargs*, which are mappings of names to functions.
-    Importers can call ``addGlobals(globals())`` to have their globals accessible to execstrings.'''
+    Importers can call ``addGlobals(globals())`` to have their globals accessible to execstrings.
+
+    Dunder methods are ignored, to prevent accidentally overwriting housekeeping methods.'''
+    drop_dunder = lambda d: {k: v for k, v in d.items() if not k.startswith("__")}
     for g in args:
-        globals().update(g)
-    globals().update(kwargs)
+        globals().update(drop_dunder(g))
+    globals().update(drop_dunder(kwargs))
 
 
 def getGlobals():
--- a/visidata/help.py
+++ b/visidata/help.py
@@ -119,10 +119,9 @@ class HelpPane:
 @VisiData.api
 @functools.lru_cache(maxsize=None)
 def getHelpPane(vd, name, module='vdplus'):
-    import importlib.resources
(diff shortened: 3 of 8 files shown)


zormit:
@saulpw @anjakefala before I try to fix the version problems... what do you think of this in general? 

regarding the version problems...
do you know why `import_resources` would not exist in 3.10, while it seems to work again in 3.11 (not sure though, the test was cancelled... 🤔 )? It is probably somehow an implicit dependency that got dropped from the providing package(s)?

zormit:
>weirdly I'm having trouble to reproduce it right now

Ok I realised my mistake. The warning only pops up when launching via the installed `bin/vd`:
```
% vd
/Users/mn/.venv/visidata/bin/vd:4: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  __import__('pkg_resources').require('visidata==2.12.dev0')
saul.pw/VisiData v2.12dev
```

I'm not even sure how this file is generated, but I'd be curious to find out & maybe also remove the `pkg_resources` dependency from there. Probably through updating the underlying deployment dependencies, so I opened some kind of rabbit hole here...

anjakefala:
Ahhh, thank you for tackling this @zormit! And researching what needed to change. I think `bin/vd` is generated by some installers.

I learned that `importlib_resources` is actually referred to as `importlib.resources` wrt the stdlib: https://docs.python.org/3/library/importlib.resources.html. I made the change, and tested locally. Let's see how the build looks!

anjakefala:
Failing now because `importlib.resources.files` is new in Python 3.9! Hmmmm.

anjakefala:
Seems like we need a try/except to use importlib.resources.files > Python 3.9, and `importlib_resources.files` in 3.8, 3.7: https://stackoverflow.com/questions/6028000/how-to-read-a-static-file-from-inside-a-python-package. Which is a bit gross. We would also need to then install [importlib_resources](https://pypi.org/project/importlib-resources/) in 3.7, 3.8.

anjakefala:
Upon further research, it seems like pkgutil can possibly give us what we want without the version challenges with `importlib.resources`: https://docs.python.org/3/library/pkgutil.html

anjakefala:
I decided to go with `importlib.resources` and `importlib_resources`. It seemed like the change that was closest to pkg_resources, and when we drop support for 3.7 and 3.8, we will be in a good position.

saulpw (approved):
Thanks, @zormit and @anjakefala !  If everything seems to work, this is good.

zormit:
Thanks for the review @anjakefala and pulling it through! :)

PR #1979, opened 2023-07-29
[ttf] implement _closePath() to draw missing lines

midichef:
The ttf viewer is missing some lines when it draws shapes, in glyphs like `xrhombus` and `boxcheckbld`.

It seems [fonttools](https://fonttools.readthedocs.io/en/latest/pens/basePen.html) calls `_closePath()` to draw the final lines, so it needs to be implemented. This patch implements it. I looked at all the shapes in `sample_data/wingdings.ttf`, and the change does not seem to break any of them.

A note to myself for future reference, opentype.js has [a useful web viewer for TTF files](https://opentype.js.org/glyph-inspector.html).


saulpw (approved):
What are you even doing using the ttf viewer?! :)

PR #1980, opened 2023-07-29
[canvas] fix graph legend drawn too early

midichef:
If you graph the data in a large file, like in `seq 1222333 |vd`, you can see the graph legend is drawn too early, while the data are loading. It's also in the wrong part of the screen if your window is not 80 columns wide. This commit stops the early drawing by taking `plotlegends()` out of `plotColor()`.

I checked that the change does not cause problems for other viewers that use `plotColor`, which are:  `geojson`, `mbtiles`, `png`, `shp`, and `ttf`. I couldn't figure out how to test `mbtiles` but the rest seem fine.

In fact, `geojson` and `shp` have legends that might not be necessary, since they seem to be blank for the data in `sample_data`. It would be simple to get rid of them. Just change `reload()`, to insert `self.legends.clear()` between the last call to `self.plotColor()` and `self.refresh()`.

### DIFF
--- a/visidata/canvas.py
+++ b/visidata/canvas.py
@@ -362,7 +362,6 @@ class Canvas(Plotter):
             self.legendwidth = max(self.legendwidth, dispwidth(legend))
             self.legends[legend] = attr
             self.plotAttrs[k] = attr
-            self.plotlegends()
         return attr
 
     def resetCanvasDimensions(self, windowHeight, windowWidth):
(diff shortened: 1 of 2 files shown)


anjakefala:
> In fact, geojson and shp have legends that might not be necessary, since they seem to be blank for the data in sample_data. It would be simple to get rid of them. Just change reload(), to insert self.legends.clear() between the last call to self.plotColor() and self.refresh().

Thanks for looking into this thoroughly! =) I tested it as well, and your change is definitely an improvement.

A while ago, I have used a shpfile that did have a legend, so I am leaning towards keeping it for now.

Thank you so much!

PR #1982, opened 2023-08-03
[sheets] fix slowness from adding rows during getMaxWidth()

### LINKED ISSUE #1964: loading large sheets is occasionally slow and nonresponsive
**Small description**
About 1 out of 10-20 times I run visidata, the file is slow to load and the screen is totally blank and nonresponsive during loading.

**Expected result**
I expect the usual loading behavior during loading:  the screen appears, "loading" is shown, and the percentage of the file updates as the file loads.

But 1 out of every 10-20 times, the loading seems to hit a rough patch. The file takes 3-5x as long as usual to load. The screen starts out completely blank, and stays blank during that entire loading period. It is also nonresponsive, neither Ctrl-C or Ctrl-Q have any effect. Then after a long delay, the sheet finally appears, fully loaded with rows of data, and Visidata can be used as usual.

I can trigger it with this:
```
python -c 'for _ in range(1_000_000): print("1,2,one,two,three,four,five")' > bigfile.csv
while true; do; vd -N bigfile.csv; sleep 0.1; done
```
To quickly check for the problem, I run the while loop and hit `Ctrl-Q` in the first few seconds if I see Visidata draw any text to the screen. I.e. if it says "loading" and updates the progress percentage, I hit Ctrl-Q to quit and try again. What I'm looking for is a long pause with a blank screen, lasting more than a few seconds.

The effect seems easier for me to reproduce with this larger, more complicated file:
```
wget https://github.com/datablist/sample-csv-files/raw/main/files/customers/customers-2000000.zip
unzip customers-2000000.zip
while true; do; vd -N customers-2000000.csv; sleep 0.1; done
```

One observation: the nonresponsive time does seem to scale with the size of the CSV.
For this file with 2M rows, [datablist's customers-2000000.zip](https://github.com/datablist/sample-csv-files/raw/main/files/customers/customers-2000000.zip) normal loading takes 14 seconds. Buggy loading takes 56 seconds.
For similar test data with half the number of rows, [datablist's customers-1000000.zip](https://github.com/datablist/sample-csv-files/raw/main/files/customers/customers-1000000.zip), it's about half as long:  normal loading takes 7 seconds, buggy loading takes 24 seconds.

And I suspect there are other ways to trigger the problem than CSV data. I have the same blank nonresponsive screen if I try to load a large JSON file that is not actually JSON:
```
seq 1000000 > not-json.json
vd -N -f json not-json.json
```
That's a bit harder to trigger though. It can take up to 100 runs to trigger the problem with that data.

**Additional context**
Ubuntu 22.04.2
visidata 2.11.1 or  latest develop version
Python 3.10.12

midichef:
Part of #1964.

### DIFF
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -572,7 +572,7 @@ class TableSheet(BaseSheet):
         for vcolidx in range(0, self.nVisibleCols):
             col = self.visibleCols[vcolidx]
             if col.width is None and len(self.visibleRows) > 0:
-                vrows = self.visibleRows if self.nRows > 1000 else self.rows
+                vrows = self.visibleRows if self.nRows > 1000 else self.rows[:1000]  #1964
                 # handle delayed column width-finding
                 col.width = max(col.getMaxWidth(vrows), minColWidth)
                 if vcolidx != self.nVisibleCols-1:  # let last column fill up the max width


anjakefala:
I changed the message to not close https://github.com/saulpw/visidata/issues/1964 in case we still want to add the timeout to `draw`!

anjakefala:
Thank you so much for this beautiful work @midichef!

Issue #1553, opened 2022-10-06
[mouse] Error: getmouse() returned ERR


geekscrapy:
**Small description**
I get the following "bomb out" with no stacktrace (even with --debug) when "slinging" the trackpad mouse up or down.

**Expected result**
To not crash, or give a stacktrace

**Actual result with screenshot**
This just gets printed once returned to the cli: `Error: getmouse() returned ERR`

**Steps to reproduce with sample data and a .vd**
Open a file with more rows than can be fit on the screen
"sling" scroll the view and the crash comes. You need to give it momentum so it scrolls on its own. This occurs in macos

**Additional context**
This seems to occur for me on any commit, so it maybe my setup... If someone could test doing a "sling" scroll that would be good. If it doesn't occur dont waste time on this, it might be my unique setup and I'll try feedback when/if I find a solution... TIA!

anjakefala:
Hey @geekscrapy!

We are not quite sure what is going on, but we just made sure all of the `vd.parseMouse()` calls (the only function which calls `getmouse()`) were in `try`/`except` blocks. https://github.com/saulpw/visidata/commit/c4e2a43bf46278708f14b071e5175ed3ba372485

Could you pull `develop` and check it out?

geekscrapy:
Yea it's very odd. It happens in both my terminals. Same bash -l session though 

I'll test with a generic bash session and see if it's any env variables messing it up 

~~Sorry forgot to say, yea this is in the current commit too~~

I'll try with the new commit

geekscrapy:
OK, so, when using the latest commit (c4e2a43bf46278708f14b071e5175ed3ba372485) it works fine (no error, no stacktrace, its like the issue does not exist), but bc39516261516343aefe1bde9e5728fe5b76afd7 crashes for me.

Anything you want me to try to get the stacktrace?

anjakefala:
We are okay with just having the silent failure! Thanks for the offer and for checking the fix. =)

frosencrantz:
Thanks for fixing this @anjakefala.   Thanks for reporting this @geekscrapy! 
I was seeing this also, but wasn't quite sure what the cause was.  I'll report back if I see it again.

PR #1984, opened 2023-08-05
[canvas] fix clicks on labels and unplotted canvas

midichef:
When a canvas is taking a long time to plot for the first time, like in `seq 1222333 |vd`, if you click the mouse buttons or use the scroll-wheel, there are errors from data structures that aren't initialized yet.  Like `AttributeError: 'NoneType' object has no attribute 'xmin'`. This PR checks whether they exist instead of accessing them.

And there's another issue during slow plots. Because the handling of cursor events is delayed, the `BUTTON1_PRESSED`/`BUTTON1_RELEASED` events turn into `BUTTON1_CLICKED`, which shows a message of `no command for "BUTTON1_CLICKED"`. So this adds added bindings for the click event, as well as equivalent bindings for double and triple clicks.

Also, there is a sporadic error that is hard to reproduce. If I click many times on the graph during the long plotting time, sometimes I see a `KEY_MOUSE` event, for some reason. I'm not sure why. In handling it, `curses.getmouse()` raises an error, and the error handling had a problem:
```    Traceback (most recent call last):
    File "/home/midichef/.local/lib/python3.8/site-packages/visidata/mainloop.py", line 202, in mainloop
        keystroke = vd.handleMouse(sheet)  # if it was handled, don't handle again as a regular keystroke
    File "/home/midichef/.local/lib/python3.8/site-packages/visidata/mouse.py", line 105, in handleMouse
        return r.keystroke
    UnboundLocalError: local variable 'r' referenced before assignment
```
So this PR fixes the error handling as well too for this rare case. 

This PR does not address why the error is raised by `getmouse()`. The most I could find out is this:
```
    File "/home/midichef/.local/lib/python3.10/site-packages/visidata/mouse.py", line 49, in parseMouse
        devid, x, y, z, bstate = curses.getmouse()
    _curses.error: getmouse() returned ERR
```
But https://github.com/saulpw/visidata/commit/94f2c2b5d26557bc2ba52d21f19e204b216900dc is a useful starting point to explore that error if someone wants to investigate it. It takes around 30 clicks to make it happen on my system (Ubuntu 20.04.6, Python 3.8.10, and Ubuntu 22.04.2, Python 3.10.12). (Perhaps it is related to https://github.com/saulpw/visidata/issues/1553?)

As a separate, unfixed issue: when a graph is still plotting for the first time, pressing arrow keys will also cause errors, similar to the mouse clicks. I'm not sure what is the best way to address them. Putting `if sheet.cursorBox` checks inside all of the major `go-*` commands makes the command strings unwieldy. One option is to move the `addCommand()` command strings into Python code inside a full-fledged function. So for `go-left` I could do `def go_left()` and then use `@Canvas.command(...'go-left'...)` connect the longname to the function. But I see `addCommand()` in hundreds of places, and `.command()` only twice. So it seems the Visidata philosophy is to prefer command strings over functions. What do you think?

### DIFF
--- a/visidata/canvas.py
+++ b/visidata/canvas.py
@@ -399,11 +399,13 @@ class Canvas(Plotter):
     def canvasMouse(self):
         x = self.plotterMouse.x
         y = self.plotterMouse.y
+        if not self.canvasBox: return None
         p = Point(self.unscaleX(x), self.unscaleY(y))
         return p
 
     def setCursorSize(self, p):
         'sets width based on diagonal corner p'
+        if not p: return
         self.cursorBox = BoundingBox(self.cursorBox.xmin, self.cursorBox.ymin, p.x, p.y)
         self.cursorBox.w = max(self.cursorBox.w, self.canvasCharWidth)
         self.cursorBox.h = max(self.cursorBox.h, self.canvasCharHeight)
@@ -453,7 +455,12 @@ class Canvas(Plotter):
                            self.scaleY(self.cursorBox.ymax))
 
     def startCursor(self):
-        self.cursorBox = Box(*self.canvasMouse.xy)
+        cm = self.canvasMouse
+        if cm:
+            self.cursorBox = Box(*cm.xy)
+            return True
+        else:
+            return None
 
     def point(self, x, y, attr=0, row=None):
         self.polylines.append(([(x, y)], attr, row))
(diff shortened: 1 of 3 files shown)


midichef:
Oh and this PR fixes another issue, where the cursor becomes invisible if you click on the label or legend text of a graph.

saulpw:
> But I see addCommand() in hundreds of places, and .command() only twice. So it seems the Visidata philosophy is to prefer command strings over functions. What do you think?

Yeah, I added `.command()` for that use and don't entirely like it.  The menu system that's been added uses the command string to determine whether sheets are pushed or input is required (though that is cosmetic), and the command log uses it to determine whether the column or row movement should be recorded.  So maybe it should be removed, and if we want to pull code out into functions, we just have to call them explicitly in the command string (but still keep `push()` and `input()` and `cursorRow` etc in the command string).

saulpw:
What if, instead of adding `if cursorBox` checks, we make sure cursorBox is always at least somewhat valid?

visidata/canvas.py
```diff
@@ -748,13 +763,20 @@ def deleteSourceRows(self, rows):
...
-
-Canvas.addCommand('BUTTON3_PRESSED', 'start-move', 'sheet.anchorPoint = canvasMouse', 'mark grid point to move')
-Canvas.addCommand('BUTTON3_RELEASED', 'end-move', 'fixPoint(plotterMouse, anchorPoint)', 'mark canvas anchor point')
-
-Canvas.addCommand('ScrollwheelUp', 'zoomin-mouse', 'tmp=canvasMouse; incrZoom(1.0/options.zoom_incr); fixPoint(plotterMouse, tmp)', 'zoom in with scroll wheel')
-Canvas.addCommand('ScrollwheelDown', 'zoomout-mouse', 'tmp=canvasMouse; incrZoom(options.zoom_incr); fixPoint(plotterMouse, tmp)', 'zoom out with scroll wheel')
+Canvas.addCommand('BUTTON1_RELEASED', 'end-cursor', 'cm=canvasMouse; setCursorSize(cm) if cm else None', 'end cursor box with left mouse button release')
+Canvas.addCommand('BUTTON1_CLICKED',        'remake-cursor', 'startCursor(); cm=canvasMouse; setCursorSize(cm) if cm else None', 'end cursor box with left mouse button release')
+Canvas.addCommand('BUTTON1_DOUBLE_CLICKED', 'remake-cursor', 'startCursor(); cm=canvasMouse; setCursorSize(cm) if cm else None', 'end cursor box with left mouse button release')
```
saulpw: use `bindkey` to create an alias, when the command and exec string are identical.

visidata/mouse.py
```diff
@@ -73,12 +73,12 @@ def parseMouse(vd, **kwargs):
 
 @VisiData.api
 def handleMouse(vd, sheet):
+    vd.keystrokes = ''
```
saulpw: Is it necessary to move all this code outside the `try` block?  I think it's probably harmless, but does increase the diff size.

midichef:
> What if, instead of adding `if cursorBox` checks, we make sure cursorBox is always at least somewhat valid?

Hmm, that could work. But I went with the simpler option for now, and just added the checks, since doing it for the four arrow keys is enough.

I had a separate advantage in mind when considering moving the `go-*` commands to functions. I'd been thinking it would make it easy to add a hook after each move, to show the cursor coordinates. But I've since realized I can do that in my .visidatarc by wrapping `InvertedCanvas.draw()`. So it's not necessary.

PR #1990, opened 2023-08-06
[pandas] handle read methods that produce a list of dataframes

### LINKED ISSUE #1986: [pandas/html] Stack trace when loading html data with pandas
**Small description**
Stack trace when loading html data with pandas

**Expected result**
Return an IndexSheet of PandaSheets of the HTML tables found in an html file.

https://asciinema.org/a/650fulFvz3V9FXwJVkCkuDpeP

```
  File "/usr/lib/python3.9/site-packages/visidata/threads.py", line 200, in _toplevelTryFunc                         
  File "/usr/lib/python3.9/site-packages/visidata/loaders/_pandas.py", line 171, in reload
    df = df.reset_index(drop=True)
AttributeError: 'list' object has no attribute 'reset_index'
```

```
vd -f pandas sample_data/sunshinelist.html
```

Note: Visidata code expects the read_* functions return a DataFrame.  The read_html returns a list of DataFrames:
https://pandas.pydata.org/docs/reference/api/pandas.read_html.html
 

**Additional context**
Please include the version of VisiData and Python.
Latest develop version.  3.9.2

ajkerrigan:
Pandas read methods (like `read_html`) can return a list of dataframes rather than a single one. In that case, use the first element of the list for the current sheet, and push additional dataframes to new sheets.

There might be a more sensible way to handle this case. :thinking: 

Closes #1986

### DIFF
--- a/visidata/loaders/_pandas.py
+++ b/visidata/loaders/_pandas.py
@@ -158,6 +158,12 @@ class PandasSheet(Sheet):
                 readfunc = getattr(pd, 'read_'+filetype) or vd.error('no pandas.read_'+filetype)
             # readfunc() handles binary and text open()
             df = readfunc(self.source, **options.getall('pandas_'+filetype+'_'))
+            # some read methods (html, for example) return a list of dataframes
+            if isinstance(df, list):
+                for idx, inner_df in enumerate(df[1:], start=1):
+                    vd.push(PandasSheet(f'{self.name}[{idx}]', source=inner_df))
+                df = df[0]
+                self.name += '[0]'
             if (filetype == 'pickle') and not isinstance(df, pd.DataFrame):
                 vd.fail('pandas loader can only unpickle dataframes')
         else:


saulpw:
Thanks, @ajkerrigan!  This is fine for now (better than an error).  Typically the `vd.open_foo` function knows that the format has multiple sheets, and returns the IndexSheet, which does the work to open the source and yield the subsheets.  Is it the case that some pandas `read_` methods (like `read_html`) always return a list, or can read_html return a single DataFrame as well?  If the former, maybe `open_pandas` should have a list of those, so it can return a new PandasIndexSheet.

ajkerrigan:
Hmm good point having a `PandasIndexSheet` for that case :thinking: . It looks like `read_html()` _always_ returns a list of DataFrames, so an index sheet may be smartest there. There are other read methods where things are more variable though. Out of curiosity I looked at the return type annotations on all read methods:

```
read_clipboard: <class 'inspect._empty'> 🤷‍
read_csv: DataFrame | TextFileReader
read_excel: DataFrame | dict[IntStrT, DataFrame]
read_feather: <class 'inspect._empty'> 🤷‍
read_fwf: DataFrame | TextFileReader
read_gbq: DataFrame
read_hdf: <class 'inspect._empty'> 🤷‍
read_html: list[DataFrame]
read_json: DataFrame | Series | JsonReader
read_orc: DataFrame
read_parquet: DataFrame
read_pickle: <class 'inspect._empty'> 🤷‍
read_sas: DataFrame | ReaderBase
read_spss: DataFrame
read_sql: DataFrame | Iterator[DataFrame]
read_sql_query: DataFrame | Iterator[DataFrame]
read_sql_table: DataFrame | Iterator[DataFrame]
read_stata: DataFrame | StataReader
read_table: DataFrame | TextFileReader
read_xml: DataFrame
```

So maybe it's worth landing this PR to dodge the error, and looking to introduce a `PandasIndexSheet` in a separate PR? (I'm OK having that assigned to me for the record, since this PR feels like step 1 rather than a true fix :sweat_smile: )

saulpw (approved):
Thanks, @ajkerrigan !  As a first pass this seems good to at least get past the error.

frosencrantz:
Thanks, @ajkerrigan!  I am happy you caught the bigger issue of the pandas read methods.

Issue #1308, opened 2022-02-27
[html rowspan] Rowspan is not handled.


frosencrantz:
**Small description**
HTML table loading doesn't handle rowspan properly.

**Expected result**
The data in the rowspan column is duplicated on the spanned rows.

**Actual result with screenshot**

https://asciinema.org/a/qotdqplkmpKJxPUQUC5kaGHEd

Animation shows html files, how w3m renders the data, and how Visidata shows data.
The columns with the errors have this exception:

```
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/visidata/wrappers.py", line 108, in wrapply
    return func(*args, **kwargs)
  File "/usr/lib/python3.9/site-packages/visidata/column.py", line 273, in getValue
    return self.calcValue(row)
  File "/usr/lib/python3.9/site-packages/visidata/column.py", line 235, in calcValue
    return (self.getter)(self, row)
  File "/usr/lib/python3.9/site-packages/visidata/loaders/html.py", line 103, in <lambda>
    self.addColumn(Column(name, getter=lambda c,r,i=colnum: r[i][0]))
IndexError: list index out of range
```

**Steps to reproduce with sample data and a .vd**

I would expect the data from columns 2 & 3 for both of these tables would be the same.

Regular 3x3.html (works)

```html
<table >
        <tbody>
                <tr>
                        <td>1.1 </td>
                        <td>1.2 </td>
                        <td>1.3 </td>
                </tr>
                <tr>
                        <td> 2.1 </td>
                        <td> 2.2 </td>
                        <td> 2.3 </td>
                </tr>
                <tr>
                        <td> 3.1 </td>
                        <td> 3.2 </td>
                        <td> 3.3 </td>
                </tr>
        </tbody>
</table>
```

With row span 3x3-rowspan.html (breaks):
```html
<table >
        <tbody>
                <tr>
                        <td rowspan=3>1.1 </td>
                        <td>1.2 </td>
                        <td>1.3 </td>
                </tr>
                <tr>
                        <td> 2.2 </td>
                        <td> 2.3 </td>
                </tr>
                <tr>
                        <td> 3.2 </td>
                        <td> 3.3 </td>
                </tr>
        </tbody>
</table>
```

**Additional context**
Please include the version of VisiData.  Using latest version from develop branch

I have data sources I try to use with Visidata that make use of rowspan to format html tables.

There is code in the html loader to handle rowspan for column headers.

frosencrantz:
[...]

PR #2002, opened 2023-08-19
[html] fix failure from colspan with only td tags

midichef:
When an HTML table has colspan, with only td tags, the visidata rows contain invalid data for the resulting empty cells, and the table can't be viewed.

More specifically, cells in a row created by the html loader should be tuples like `(cellval, links)`:
https://github.com/saulpw/visidata/blob/d0e8a354b475b03a3a4202ba046796a1d3976a53/visidata/loaders/html.py#L146
But for empty cells, what `iterload()` adds to the row is a `None` object, not a tuple. That causes the following exception that makes the table unviewable:
```
  File "/home/j/.local/lib/python3.10/site-packages/visidata/threads.py", line 200, in _toplevelTryFunc
  File "/home/j/.local/lib/python3.10/site-packages/visidata/sheets.py", line 235, in reload
    for r in self.iterload():
  File "/home/j/.local/lib/python3.10/site-packages/visidata/loaders/html.py", line 160, in iterload
    it = list(list(x) for x in self.rows.pop(0))
  File "/home/j/.local/lib/python3.10/site-packages/visidata/loaders/html.py", line 160, in <genexpr>
    it = list(list(x) for x in self.rows.pop(0))
TypeError: 'NoneType' object is not iterable
```
Here is an example file to reproduce the error:  [td_colspan.html.txt](https://github.com/saulpw/visidata/files/12385238/td_colspan.html.txt)
It can be viewed with `vd -f html td_colspan.html.txt`

This PR fixes that issue and allows the table to be viewed.

The resulting table will still show errors for the empty cells, which I did not attempt to fix. They are the same errors already covered by https://github.com/saulpw/visidata/issues/1308 , but caused by colspan instead of the rowspan that the open issue discusses.

### DIFF
--- a/visidata/loaders/html.py
+++ b/visidata/loaders/html.py
@@ -142,7 +142,7 @@ class HtmlTableSheet(Sheet):
                         cellval = ''   # use empty non-None value for subsequent rows in the rowspan
                 else:
                     while colnum >= len(row):
-                        row.append(None)
+                        row.append((None, []))
                     row[colnum] = (cellval, links)
 
                 colnum += colspan


saulpw:
Thanks @midichef for catching this oversight!

PR #2006, opened 2023-08-24
[clipboard-] improve error when deleting row on empty sheet

midichef:
This improves the error message currently seen by the user when doing `d` or `x` on an empty sheet:
```
  File "/home/midichef/.local/lib/python3.10/site-packages/visidata/clipboard.py", line 117, in delete_row
    oldrow = sheet.rows.pop(rowidx)
IndexError: pop from empty list
```
And it changes the message to be displayed with `vd.fail()`, which will abort a replay if `delete-row` or `cut-row` are called on a sheet with no rows.

### DIFF
--- a/visidata/clipboard.py
+++ b/visidata/clipboard.py
@@ -113,6 +113,8 @@ def pasteFromClipboard(vd, cols, rows):
 
 @Sheet.api
 def delete_row(sheet, rowidx):
+    if len(sheet.rows) == 0:
+        vd.fail("no row to delete")
     if not sheet.defer:
         oldrow = sheet.rows.pop(rowidx)
         vd.addUndo(sheet.rows.insert, rowidx, oldrow)
--- a/visidata/clipboard.py
+++ b/visidata/clipboard.py
@@ -113,7 +113,7 @@ def pasteFromClipboard(vd, cols, rows):
 
 @Sheet.api
 def delete_row(sheet, rowidx):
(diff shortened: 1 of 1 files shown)


saulpw (approved):

visidata/clipboard.py
```diff
@@ -113,6 +113,8 @@ def pasteFromClipboard(vd, cols, rows):
 
 @Sheet.api
 def delete_row(sheet, rowidx):
+    if len(sheet.rows) == 0:
```
ajkerrigan: This looks like a friendly move 👍 . Makes me wonder though if we need to account for cases where `sheet.rows` is `None` or `UNLOADED` 🤔. If so, maybe... 

```suggestion
    if not sheet.rows:
```

In any case yay for friendlier errors 🎉

saulpw: Ah good idea, @ajkerrigan!  It's unlikely but as long as we're here..

saulpw (approved):

visidata/clipboard.py
midichef: Good catch. So, should it handle UNLOADED, should it say `if not sheet.rows or sheet.rows is UNLOADED`?

saulpw: `UNLOADED` is False-y, so `if not sheet.rows` should suffice for all cases.

PR #2009, opened 2023-08-26
[describe-] collect aggr funcs that operate on list of values in Dict

anjakefala:
In draft until https://github.com/saulpw/visidata/pull/2008 is merged. We need to make sure the tests are all green. This is needed to fix the vgit build.

aggregator() converts a func, which operates on a list, into a _func, which operates on a srccol and a list of rows.

The original functions were then added into Globals, but this caused problems for a function like `sum()` which appears naturally in Python code.

I created an OrderedDict, and named it aggregators_vals as a place to store them.

Other possible options:

* We could include the optional funcvals along with func(srccol) for Aggregator. Describe Sheet could then grab the funcvals if it exists.

* Describe Sheet could pass the srccol and list of rows, instead of vals. This is not ideal because it means for each aggregator, we call getValues once-more. This would cause a performance degradation.

* Add them to vd.aggregators, possibly with the suffix "_vals", and create an Aggregator out of them as well. Have Describe Sheet pull aggrname_vals.

* Similarly use the vd.aggregator_vals, but have a less terrible name.

Optional: Do we want to do the work of porting quantiles and percentiles to be useable by Describe Sheet? Currently, an aggregator that does not go through aggregator() is not useable by Describe Sheet.


anjakefala:
@geekscrapy noticed weird behaviour here: https://github.com/saulpw/visidata/issues/2013 I am marking this PR as draft until this issue is resolved.

visidata/features/describe.py
```diff
@@ -94,17 +94,16 @@ def reloadColumn(self, srccol):
...
-            d['mode'] = self.calcStatistic(d, mode, vals)
+            d['mode'] = self.calcStatistic(mode, vals)
             if vd.isNumeric(srccol):
                 for func in [min, max, sum, median]:  # use type
-                    d[func.__name__] = self.calcStatistic(d, func, vals)
+                    d[func.__name__] = self.calcStatistic(func, vals)
                 for aggrname in vd.options.describe_aggrs.split():
-                    func = vd.getGlobals()[aggrname]
-                    d[func.__name__] = self.calcStatistic(d, func, vals)
+                    aggr = vd.aggregators_vals[aggrname]
```
saulpw: Can't we just use `vd.aggregators[aggrname].func`, instead of maintaining a separate dict of funcs?

anjakefala: They have different function signatures! One is (srcCol, rows), the other (values). I added an optional funcValues argument!

visidata/aggregators.py
```diff
@@ -30,7 +30,7 @@ def getValues(self, rows):
         yield v
 
 
-vd.aggregators = collections.OrderedDict()  # [aggname] -> annotated func, or list of same
+vd.aggregators = collections.OrderedDict()  # [aggname] -> annotated func, or list of same(srccol, list of rows)
```
saulpw: what's this comment addition about?

anjakefala: I guess it is not necessary anymore!

saulpw (approved):

PR #2025, opened 2023-09-12
[jsonl] Allow slash comments

geekscrapy:
- Exit earlier for blank lines
- Added '`//`' to the acceptable comment type for jsonl
- Added to JSON test file

Although not part of the spec, the loader already uses "`#`" as a comment. Ironically, vscode does not recognise this as a valid comment delimiter and prefers "`//`". Therefore I don't think it is a push to include this comment type.


saulpw (approved):
Hey @geekscrapy, thanks for putting this together!  If vscode accepts this as a comment, we'll allow it too.  BTW for `startswith`, "prefix can also be a tuple of prefixes to look for", which I'm guessing is faster.

geekscrapy:
> Hey @geekscrapy, thanks for putting this together! If vscode accepts this as a comment, we'll allow it too. BTW for `startswith`, "prefix can also be a tuple of prefixes to look for", which I'm guessing is faster.

TIL added! Thanks

PR #2028, opened 2023-09-15
Add postgresql schemata options

### LINKED ISSUE #2027: Add option to show tables from all PgSQL schemas
I usually use multiple schemas in PostgreSQL databases. Unfortunately, `vd/vdsql` only shows `public.*` tables.

An option to configure table visibility would be nice :-)

- by default: show public tables;
- show all tables, except system tables (`pg_catalog`, `information_schema`);
- show system tables (`pg_catalog`, `information_schema`).

(I don't know Visidata internals... yet. If you can give me some pointers on how to implement this, I may have a look ;-)

jrjsmrtn:
Add PgSchemataTable class
Add support for options.postgres_schema "*" and "**":
- open a schemata table
- "*" show user schemata
- "**" show user and system schemata
- entering a schemata row open the table 

Sheet names follow the "database[.schema[.table]]" format

Fixes #2027

### DIFF
--- a/visidata/loaders/postgres.py
+++ b/visidata/loaders/postgres.py
@@ -54,7 +54,11 @@ def openurl_postgres(vd, url, filetype=None):
                 port=url.port,
                 password=url.password)
 
-    return PgTablesSheet(dbname+"_tables", sql=SQL(conn))
+    schema = options.postgres_schema
+    if schema in ("*", "**"):
+        return PgSchemataSheet(dbname, sql=SQL(conn))
+    else:
+        return PgTablesSheet(dbname+"."+schema, sql=SQL(conn))
 
 
 VisiData.openurl_postgresql=VisiData.openurl_postgres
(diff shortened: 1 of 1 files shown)


jrjsmrtn:
- `vd postgres:///dbname` still show `public.*` tables;
- `vd --postgres-schema=myschema postgres:///dbname` still show `myschema.*` tables;
- `vd --postgres-schema=* postgres:///dbname` will show the database schemas;
- `vd --postgres-schema=** postgres:///dbname` will show both database and system schemas;

visidata/loaders/postgres.py
```diff
@@ -84,18 +88,61 @@ def postgresGetColumns(vd, cur):
         yield ColumnItem(coldesc.name, i, type=codeToType(coldesc.type_code, coldesc.name))
 
 
+class PgSchemataSheet(Sheet):
+    rowtype = 'schemas'
+
+    def reload(self):
+        if options.postgres_schema == "*":
```
saulpw, line 95: Use `self.options` (here at least, also other places it's used) so that the option can be overridden on a per-sheet basis.

visidata/loaders/postgres.py
```diff
@@ -84,18 +88,61 @@ def postgresGetColumns(vd, cur):
...
+                FROM information_schema.schemata
+                WHERE
+                    (schema_name <> 'information_schema')
+                    AND (schema_name NOT LIKE 'pg_%');
+                """
+        elif options.postgres_schema == "**":
+            qstr = """SELECT schema_name FROM information_schema.schemata;"""
+
+        with self.sql.cur(qstr) as cur:
+            self.nrowsPerTable = {}
```
saulpw, line 107: Is this used anywhere?

visidata/loaders/postgres.py
saulpw: And let's make it so that an empty string means all-but-internal schema, instead of `*`.  Maybe `*` can mean "including internal schemas" like you have for `**` below.  Do we think it's important to support internal schemas though?

visidata/loaders/postgres.py
```diff
@@ -115,17 +162,19 @@ def reload(self):
                 self.addRow(r)
 
     def openRow(self, row):
-        return PgTable(self.name+"."+row[0], source=row[0], sql=self.sql)
+        return PgTable(
+            self.name+"."+row[0],
+            source=(self.source or options.postgres_schema)+"."+row[0],
```
saulpw, line 167: Seems like the given name and the source should be the same here.  Let's use `source`, since I think it's reasonable to require `source` to be the schema for this class, and only use `options.postgres_schema` in the outer `openurl_postgres` function.

visidata/loaders/postgres.py
```diff
@@ -84,18 +88,61 @@ def postgresGetColumns(vd, cur):
         yield ColumnItem(coldesc.name, i, type=codeToType(coldesc.type_code, coldesc.name))
 
 
+class PgSchemataSheet(Sheet):
+    rowtype = 'schemas'
+
+    def reload(self):
```
saulpw, line 94: Generally we're standardizing on defining `iterload()`, which is almost entirely like `reload`, but you don't have to clear the rows, and instead of calling `addRow()` directly, just `yield` the row.

saulpw (changes requested):
Hi @jrjsmrtn, thanks for putting this together!  This will be a useful addition.  There are a few things that need to be adjusted, if you can make these changes, I think we're well on our way to getting it merged.

anjakefala:
Ping @jrjsmrtn!

jrjsmrtn:
Pong ! :-)Sorry for being MIA. I have not forgotten your comments. I hope to apply them next week-end.Le 15 nov. 2023 à 07:26, Anja Kefala ***@***.***> a écrit :﻿
Ping @jrjsmrtn!

visidata/loaders/postgres.py
```diff
@@ -115,17 +162,19 @@ def reload(self):
...
 
 # rowdef: tuple of values as returned by fetchone()
 class PgTable(Sheet):
     @asyncthread
     def reload(self):
-        if self.options.postgres_schema:
-            source = f"{self.options.postgres_schema}.{self.source}"
-        else:
-            source = self.source
+        source = self.source
```
isosphere, line 176: This will conflict with #2129, where the table name and schema were quoted to support both being something other than lowercase. 

i.e.:

`"public"."TableName"` rather than `public.TableName` or `"public.TableName"`

anjakefala: Thanks for checking! As part of this PR, should we then move the #2129 fix into `openurl_postgres`?

anjakefala: We will have to check that the loading of lowercase table names is preserved before merging.

isosphere: > should we then move the #2129 fix into `openurl_postgres`?

I don't see how that could be done. I think my tweak belongs in these places, where `PgTable.source` is set:

certainly here:
https://github.com/jrjsmrtn/visidata/blob/621dd88a75c6a133ce1a79cbd795c51fccd1d743/visidata/loaders/postgres.py#L167

maybe here?:
https://github.com/jrjsmrtn/visidata/blob/621dd88a75c6a133ce1a79cbd795c51fccd1d743/visidata/loaders/postgres.py#L123

I don't understand what some of the variables in this code refers to, such as `row[0]`, `columns[0:1]`, etc, so I'm on shaky ground here.

saulpw: `row[0]` is the `table_name` as per the `rowdef` comment at the top of the class.  `columns[0:1]` creates a list containing the first column (and an empty list if there are no columns).

I don't think the quoting should be done in the `schema.table` string given to the sheet; the SQL should do the quoting.  So this means that the schema and the table name should be passed separately to the postgres sheet.

anjakefala:
Hi @jrjsmrtn!

We're going to close this for now due to lack of activity. Please re-open when you make the changes!

PR #2049, opened 2023-10-08
 [batch-] sync should not wait for outputProgress Thread   

anjakefala:


saulpw (approved):
Great, thanks for fixing the build once again @anjakefala !

PR #2051, opened 2023-10-09
[cmdlog-] ensure record of global options in all cmdlogs

anjakefala:
Previously, replayable options would only exist on the global cmdlog.

This ensures that they are recorded in the cmdlogs of all sheets that they apply to.

### DIFF
--- a/visidata/cmdlog.py
+++ b/visidata/cmdlog.py
@@ -442,7 +442,12 @@ def cmdlog(sheet):
 
 @BaseSheet.lazy_property
 def cmdlog_sheet(sheet):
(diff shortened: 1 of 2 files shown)


saulpw (commented):
What if someone changes an option after viewing the command log for a sheet?  I think that'll be missed because the cmdlog_sheet is persistent (lazy).

saulpw (approved):
I think it's strictly better, anyway, let's take it and we can fix it later.

anjakefala:
> What if someone changes an option after viewing the command log for a sheet? I think that'll be missed because the cmdlog_sheet is persistent (lazy).

FWIW, this behaviour existed before!

PR #2056, opened 2023-10-16
fix(docs): clean formats markdown table

### LINKED ISSUE #2057: DOC: Formats table has misaligned columns
**Small description**
See https://www.visidata.org/docs/formats/

E.g.:

image

image

**Expected result**
Table is correctly aligned

Linked PR: #2056

corneliusroemer:
I love visidata! When I tried to install various extra dependencies I noticed that the formats table is a bit broken: https://www.visidata.org/docs/formats

I ran the table through prettier's markdown table formatter which makes it much easier to spot misaligned columns.

Resolves #2057

### DIFF
--- a/docs/formats.md
+++ b/docs/formats.md
@@ -4,58 +4,57 @@ eleventyNavigation:
     order: 99
 ---
 
-|filetype            |format              |VisiData\_loader |VisiData saver  |version\_added |created  |creator             |PyPI dependencies   |
-|--------------------|--------------------|-----------------|----------------|---------------|---------|--------------------|--------------------|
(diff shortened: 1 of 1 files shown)


anjakefala:
Thanks so much for contributing!

I will be happy to take this, can you just walk me through how you did:

> I ran the table through prettier's markdown table formatter which makes it much easier to spot misaligned columns.

Are there any options you set, or do you use default? 

The formats table is partially generated by the VisiData markdown table saver. I need to add `prettier` to that process, I think.

anjakefala:
@corneliusroemer This will be updated in the next release of the website! =)

Issue #247, opened 2019-01-31
Add command palette


paulklemm:
Often I find myself browsing the (great) `visidata` documentation for the shortcut of the command I want to use because I forgot it again.

One thing that would really help is a command palette like in Sublime Text or Visual Studio Code. One shortcut brings up a search box where one can query basically all commands available in the current context. Showing the keyboard shortcut next to the command also helps a lot memorizing them again.

Here is an example from VSCode:

![image](https://user-images.githubusercontent.com/2169489/52074578-7c249e80-258a-11e9-8a5b-92fb3762e1b1.png)

saulpw:
This is a good idea, @paulklemm.  Do you know about `z Ctrl+H`, which takes you to a sheet of all commands available on the current sheet?  And `g/` which you can use to search that sheet.  It's not quite the same workflow, but I'm wondering how close that gets you to what you want.  What if we improve that workflow (suppose `g Space` to do something similar without going to a separate sheet), and maybe add a 'keywords' column to commands.tsv so that common synonyms can be given without cluttering the helpstring.

paulklemm:
Thanks, the `z Ctrl+H` together with `g/` already is super useful, I wasn't aware of that. Still having something without going to a separate sheet would be nice I think.

jtrakk:
I like how emacs-helm does this by splitting into two panes, where the bottom pane shows the command palette. For visidata, showing the `z Ctrl+H` table in the bottom half of the screen would work nicely, I think. Typing into the palette does a live fuzzy-filter over the options.

![image](https://user-images.githubusercontent.com/43392409/58979229-661a2b00-8782-11e9-9987-ec235f551988.png)

saulpw:
Added `help-search` command that opens a commands sheet filtered by the input regex.  Unbound by default, you could try binding it to `KEY_BACKSPACE` if that works on your system, or `^@` (Ctrl+Space).

I agree, split-pane is the way to do this.  It's already on the roadmap, so I'm closing this issue, though we may not get to it until 2.1.

reagle:
I'm new to visidata and this is the first thing I wanted. I can't find it, and disagree with the suggestion that a static help filter/search is equivalent to the dynamic fuzzy match of a genuine command palette.

reagle:
I’m wondering if this might be in 3.0? Should I create a new issue if this one won’t be reopened?

saulpw:
[...]

PR #2059, opened 2023-10-18
Commandpalette Proof of Concept

zormit:
With the help of @anjakefala and @saulpw I've created this little prototype for #247.

Here's a demo that shows how it works, https://asciinema.org/a/616455. Notes on the features demonstrated:
- space opens the palette (as previously the plain "command name")
- if you type a digit, it'll execute the corresponding command
- the debug mode shows you the current weighting of the matches (oops, forgot to demonstrate with this one. See [older recording](https://asciinema.org/a/615453) for that)
- it's possible to adapt the number of matches (also check older recording for a demo. or use the option yourself)

I've decided to port over the [fzf algorithm](https://github.com/junegunn/fzf/blob/master/src/algo/algo.go) for this, such that we can do this without dependencies.

2024-vd3 0


zormit:
@saulpw i think this is ready for review from my point of view. There's one TODO left that I won't manage to finish today, but feel free to move the keystrokes around if want to have it in this PR :) My head is already fuzzy from all the fuzziness.

PS: just seeing that the tests are failing. Not sure what's going on there. But I'll be logging off for today, sorry! 🙈

saulpw (approved):
This is great, @zormit!  Thanks so much for putting the time into this.

reagle:
Thank you!

PR #2069, opened 2023-10-21
Fix inconsistent indentation in gitsheet.py

cool-RR:


cool-RR:
The CI failure isn't really related to my change, right?

saulpw:
(Correct, the CI failure isn't your fault.)

Thanks for taking the effort to put this forward.  This inconsistency is a result of minimizing the diffs in previous commit, in which I wanted to wrap the block in try/except, but wanted to emphasize in the diff/blame that the block of code itself was unchanged.  Perhaps short-sighted, but then I do have an extensive history of myopia :)

Hopefully this makes the code at least a little easier to understand and extend in the future!

cool-RR:
You're welcome. 

At this point, I think that if I contributed to any random open-source project and the CI didn't fail on a problem unrelated to my code with an unclear error message, I would be surprised.

PR #2074, opened 2023-10-21
Bind gz^R to reload-modified

cool-RR:
What do you think about this binding? Feel free to reject.


saulpw (commented):
Hm, I don't think this binding makes much sense to me.  Thanks for the suggestion though!

PR #2091, opened 2023-10-28
Standardize __repr__ for sheets and columns

cool-RR:

I find the current `__repr__` for sheets confusing:

```
In [1]: vd.sheet
Out[1]: 2023-10-21-14-21-50-423073
```

It's not clear enough it's a sheet. I don't want to have to do `type(x)` to figure out what something is. 

Here's how it looks like after this change:

```
In [1]: vd.sheet
Out[1]: <DirSheet: 2023-10-21-14-21-50-423073>

In [2]: vd.sheet.columns
Out[2]:
[<Column: directory>,
 <Column: filename>,
 <Column: abspath>,
 <Column: ext>,
 <Column: size>,
 <Column: modtime>,
 <Column: owner>,
 <Column: group>,
 <Column: mode>,
 <Column: filetype>]
```

### DIFF
--- a/visidata/column.py
+++ b/visidata/column.py
@@ -108,7 +108,7 @@ class Column(Extensible):
         return ret
 
     def __repr__(self):
-        return f'{self.__class__.__name__}:{self.name}'
+        return f'<{type(self).__name__}: {self.name}>'
 
     def __deepcopy__(self, memo):
         return self.__copy__()  # no separate deepcopy
--- a/visidata/column.py
+++ b/visidata/column.py
@@ -107,6 +107,9 @@ class Column(Extensible):
             ret._cachedValues = collections.OrderedDict()  # an unrelated cache for copied columns
         return ret
 
+    def __str__(self):
+        return f'{type(self).__name__}:{self.name}'
+
     def __repr__(self):
         return f'<{type(self).__name__}: {self.name}>'
 
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -346,7 +346,7 @@ class TableSheet(BaseSheet):
         return ret
 
     def __repr__(self):
-        return self.name
+        return f'<{type(self).__name__}: {self.name}>'
 
     def evalExpr(self, expr, row=None, col=None):
         if row is not None:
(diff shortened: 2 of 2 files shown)


cool-RR:
I assume the CI failure is not related to my change.

anjakefala:
Hey @cool-RR, could you give me examples of times when you would have found this change valuable? Especially the tagging that columns are a `Column`? It's an interesting proposal, but I'm trying to assess if we would also like it. 

Before we could merge this in, we would definitely need to have a `__str__`, with the previous `__repr__` text! A grep shows a bundle of places with something like `basepath = str(self.source)`

cool-RR:
> Hey @cool-RR, could you give me examples of times when you would have found this change valuable? 

When I'm developing my scripts, I snoop around a live VisiData session to look at the different objects and how they relate to each other. When I see an object is just `2023-10-21-14-21-50-423073` I have to do an additional `type()` call to confirm that it's   a sheet.

> Especially the tagging that columns are a `Column`?

Columns are already tagged as columns. Currently their `__repr__` shows up as `Column:directory`. I'm just making it look more standard, which is `<Column: directory>`. It could be argued that the colon should be dropped, or that there should be a colon but no space, or that there should be a comma instead. The pointy brackets however are important in my opinion, to make the boundaries from other objects clear. 

Here's the similar representation for some built-in Python objects:

```
<sqlite3.Connection object at 0x7f3a20ca6c70>
<re.Match object; span=(0, 1), match='f'>
<_MainThread(MainThread, started 139858874212352)>
```


> It's an interesting proposal, but I'm trying to assess if we would also like it.
> 
> Before we could merge this in, we would definitely need to have a `__str__`, with the previous `__repr__` text! A grep shows a bundle of places with something like `basepath = str(self.source)`

Added, though I thought `self.source` points at something different than a sheet or a column.

saulpw:
> I thought self.source points at something different than a sheet or a column.

It can be either a Path (for a source sheet), or another Sheet (for a derived sheet or meta-sheet).

saulpw (approved):
Okay, thanks @cool-RR!

PR #2092, opened 2023-10-28
[rename-col] skip gratuitous rewrites

pacien:
Going to and from the AST changes the formatting and pollutes the
undo stack even if the expression is logically the same.

Now comparing against the canonicalized version to detect worthy
changes.


saulpw (approved):
Thanks, @pacien!

Issue #2089, opened 2023-10-27
[vds] hidden columns used in expressions not saved in .vds export


pacien:
**Small description**

Hidden (zero-width) columns are removed from .vds exports.

**Expected result**

Those columns and their expressions should be preserved,
especially when used as intermediary columns for other expressions.

I think it would make more sense in terms of intent between `-` (hide, setting
width to zero) vs. removing the column from the columns sheet.

**Actual result ~~with screenshot~~**

The hidden columns are missing.

After opening the file again,
this results in expression computation errors (red `!`) on columns depending
on those.

**Steps to reproduce with sample data and a .vd**

```vdj
#!vd -p
{"sheet": "global", "row": "numeric_binning", "longname": "set-option", "input": "True", "keystrokes": ""}
{"longname": "open-file", "input": ".", "keystrokes": "o"}
{"col": "", "row": "", "longname": "open-new", "input": "", "keystrokes": "Shift+A", "comment": "Open new empty sheet"}
{"sheet": "unnamed", "col": "", "row": "", "longname": "add-row", "input": "", "keystrokes": "a", "comment": "append a blank row"}
{"sheet": "unnamed", "col": "A", "row": "0", "longname": "edit-cell", "input": "a", "keystrokes": "e", "comment": "edit contents of current cell"}
{"sheet": "unnamed", "col": "A", "row": "", "longname": "addcol-expr", "input": "A*2", "keystrokes": "=", "comment": "create new column from Python expression, with column names as variables"}
{"sheet": "unnamed", "col": "A*2", "row": "", "longname": "rename-col", "input": "B", "keystrokes": "^", "comment": "rename current column"}
{"sheet": "unnamed", "col": "B", "row": "", "longname": "addcol-expr", "input": "B*2", "keystrokes": "=", "comment": "create new column from Python expression, with column names as variables"}
{"sheet": "unnamed", "col": "", "row": "", "longname": "save-sheet", "input": "all.vds", "keystrokes": "Ctrl+S", "comment": "save current sheet to filename in format determined by extension (default .tsv)"}
{"sheet": "unnamed", "col": "B", "row": "", "longname": "hide-col", "input": "", "keystrokes": "-", "comment": "Hide current column"}
{"sheet": "unnamed", "col": "", "row": "", "longname": "save-sheet", "input": "with_hidden.vds", "keystrokes": "Ctrl+S", "comment": "save current sheet to filename in format determined by extension (default .tsv)"}
[...]

PR #2093, opened 2023-10-28
[vds-] save hidden columns content also #2089

pacien:
Follow-up of 727fd33.

Saving just the expressions of hidden columns isn't enough:
an expression may depend on some data (value) from some hidden column,
resulting in computation errors when loading back the .vds.

Only saving the column metadata is also weird because they'd show in
the columns sheet but result in empty columns values after setting
back a width >0.


saulpw (approved):
Okay, thanks @pacien, let's try it!

PR #2097, opened 2023-10-31
[docs] enable viewcode

jgarte:
Hi,

This pull request enables the [viewcode](https://www.sphinx-doc.org/en/master/usage/extensions/viewcode.html) sphinx extension so that the jump to source link is available when browsing the API reference documentation.


anjakefala:
Works great, thanks @jgarte! I had no idea it was that simple.

anjakefala:
This will be on https://visidata.org/develop/docs the next time we update the website, and on https://visidata.org/docs when we ship VisiData 3.0. =)

PR #2100, opened 2023-11-03
[build] pin pandas version based on python version

anjakefala:
1.5.3 is the earliest version with a built wheel for Python 3.11.

1.3.5 is the latest supported version for Python 3.7.

Opened a PR so I could make sure CI works.


anjakefala:
Dependency installation is now 10 minutes faster for the Python 3.11 CI!!

Issue #1576, opened 2022-10-27
[json] saving a file discards columns that contain only null values


midichef:
**Small description**
When I save a JSON file that has a column containing only null values, the column is lost.

**Expected result**
When I load this JSON file and save it, I expect the save process to preserve nulls_column.
```
[
{"index": 1, "nulls_column": null}
]
```

**Actual result**
The saved file is instead:
```
[
{"index": 1}
]
```

**Steps to reproduce with sample data and a .vd**
Create a file "input.json" containing:
```
[
{"index": 1, "nulls_column": null}
]
```

In the same directory, create a file "cmdlog.vdj" containing:
```
#!vd -p
{"longname": "open-file", "input": "input.json", "keystrokes": "o"}
{"sheet": "input", "col": "", "row": "", "longname": "save-sheet", "input": "output.json", "keystrokes": "Ctrl+S", "comment": "save current sheet to filename in format determined by extension (default .tsv)"}
{"col": "", "row": "", "longname": "open-file", "input": "output.json", "keystrokes": "o", "comment": "Open file or URL"}
```

Run ```vd -p cmdlog.vdj```

**Additional context**
Version:
VisiData v2.10.2, python 3.8.10, Ubuntu 20.04.5

The loss of the columns is caused by loaders/json.py, line 81:
https://github.com/saulpw/visidata/blob/924563a0b7eeede80834dc02d8c1f237fe1949c1/visidata/loaders/json.py#L81

saulpw:
Hi @midichef, thanks for filing this issue.  It's an open question in my mind, as to whether None values are deliberately saved in JSON as `null` or if the field should be omitted in that case.  We use JSONL as a format for several things, and often commit the files to a git repo.   Sometimes we'll want to add a field only to one or a few rows, and our current approach works well to minimize the diffs in this case.

If the opposite functionality is important to you, we may have to add some kind of option that you can set; maybe `json_save_nulls` or something.  Before we do that, though, I'd like to ask: if you have a column that is entirely null values, it has no actual information in it, so why would you need it to be saved?  Could your use case be solved in a different way, for example by being able to save schema information within VisiData, so that when loading a specific file, it would add this empty column for you if it it didn't already exist?

saulpw:
This should be fixed since 91838bb4c88fc47481c005ad1c223e8b021e571d, which saves all columns in the first row, even if the values are null.  Thus the first row effectively sets the schema for the whole file, which maintains empty columns but doesn't interfere with the rest of the rows.

PR #2109, opened 2023-11-12
[json] include null columns in first row, in more cases

midichef:
This is a followup to https://github.com/saulpw/visidata/issues/1576
It generalizes the change for JSONL files in https://github.com/saulpw/visidata/commit/91838bb4c88fc47481c005ad1c223e8b021e571d, and applies it to JSON files too.


saulpw:
Thanks, @midichef, makes sense to me!

Issue #2110, opened 2023-11-14
[unzip-http] RecursionError when opening a file within a remote zip file


saulpw:
**Small description**

RecursionError when opening a file within a remote zip file.  e.g. `vd https://github.com/jart/cosmopolitan/releases/download/3.0.1/cosmos-3.0.1.zip` and press Enter.

**Additional context**

The visidata unzip_http is no longer a simple copy of unzip_http.py from the upstream unzip-http project.  It has two `def open()` functions, and the second one calls `self.open` which is just calling itself.

This was broken in a1affe7d8, which renamed `open_text` to `open`.

frosencrantz:
Referenced in #2105.

anjakefala:
The completion of the fix was in https://github.com/saulpw/visidata/issues/2127

PR #2116, opened 2023-11-17
[zip] fix recursion error fetching remote zipfile

midichef:
Partial fix for #2110. This is a narrow fix for the specific recursion error.
It does not fix the broader issue that files within a remote zipfile still cannot be opened after this fix, as mentioned in #2105. I've been sitting on some experimental code that fixes issues in opening files. It may be related to this particular bug, I'm not sure. I should be able to put it up as a draft in a day or two.


saulpw (approved):
This makes sense as a short-term fix.  Looking forward to seeing your changes in this area, some things definitely need to be straightened out.

PR #2117, opened 2023-11-17
[hint] improve warning for typing source key columns before joining

anjakefala:
- [x] instead of making this into a hint, convert it into a more useful warning that uses column names

### DIFF
--- a/visidata/features/join.py
+++ b/visidata/features/join.py
@@ -137,7 +137,8 @@ class JoinKeyColumn(Column):
             if row[c.sheet] is not None:
                 vals.add(c.getTypedValue(row[c.sheet]))
         if len(vals) != 1:
-            vd.warning(f"The source key columns of {', '.join([col.sheet.name + ':' + col.name for col in self.keycols])}, have different types. Type the source columns before joining.")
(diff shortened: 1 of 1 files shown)


visidata/features/join.py
```diff
@@ -137,7 +137,8 @@ def calcValue(self, row):
             if row[c.sheet] is not None:
                 vals.add(c.getTypedValue(row[c.sheet]))
         if len(vals) != 1:
-            vd.warning(f'inconsistent keys: ' + str(vals))
+            keycolnames = ', '.join([f'{col.sheet.name}:{col.name}' for col in self.keycols])
+            vd.warning(f"The source key columns ({keycolnames}) have different types. Type the source columns before joining.")
```
saulpw: Let's tighten this up.  In general internal messages don't have "the"s or repeated words (like "source columns" and "type").  Plus the user should already know they're joining. This is just a warning and the join will still happen (though it might be incorrect), so the imperative is unclear (is it necessary or not?).

anjakefala: This makes a lot of sense to me. I shortened the warning, and it does feel better, and sufficiently informative.

saulpw (approved):
this is great, much better, thank you!

PR #2124, opened 2023-11-19
[sheets] fix errors on sheets with no columns

midichef:
This PR fixes an error when `hide-col` hides the only column:
```
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/sheets.py", line 789, in calc_height
    return max(len(lines) for _, _, lines in displines.values())
ValueError: max() arg is an empty sequence
```
and a minor error when `resize-col-max` is run on a sheet with no columns:
```
AttributeError: 'NoneType' object has no attribute 'toggleWidth'
```

### DIFF
--- a/visidata/features/layout.py
+++ b/visidata/features/layout.py
@@ -35,7 +35,7 @@ def hide_col(vd, col):
     if not col: vd.fail("no columns to hide")
     col.hide()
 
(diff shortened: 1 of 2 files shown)


anjakefala (approved):
I tested it, and both work great! Thanks. =)

PR #2125, opened 2023-11-19
[search] handle no rows and invalid regex

midichef:
This PR fixes a few corner cases in search:
`/` on empty sheet:
```
```

And the case of `/` with invalid regex, which gives a bit scary message:
```
re.error: missing ), unterminated subpattern at position 4
```
and then subsequent calls to `n` give errors because they try to reuse the regex but it hasn't been saved:
```
AttributeError: 'str' object has no attribute 'search'
```

This PR changes the invalid regex error to be more informative:
```
invalid regex: missing ), unterminated subpattern
```
And makes future calls to `n` fail.

I'd prefer to print the invalid regex as part of the error message, but it could contain formatting characters for the statusbar.
Is there a definitive way to escape all strings like '__' that have special formatting rules in the statusbar?

### DIFF
--- a/visidata/movement.py
+++ b/visidata/movement.py
@@ -6,6 +6,7 @@ from visidata import vd, VisiData, BaseSheet, Sheet, Column, Progress, ALT, asyn
 
 def rotateRange(n, idx, reverse=False):
     'Wraps an iter starting from idx. Yields indices from idx to n and then 0 to idx.'
+    if n == 0: return []
     if reverse:
         rng = range(idx-1, -1, -1)
         rng2 = range(n-1, idx-1, -1)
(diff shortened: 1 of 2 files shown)


saulpw (approved):
I think these are fine.  We could put the assignment (line 43) inside the try block to save a line of code.  And we could preserve both the re.error message and the regex itself in the vd error message to make it a bit clearer.  But I don't think it matters that much, we can also take this as it is.

anjakefala:
Hi @midichef!

>  Is there a definitive way to escape all strings like '__' that have special formatting rules in the statusbar?

This doesn't have a good answer at the moment! Would you be comfortable with merging as-is?

midichef:
@anjakefala
> This doesn't have a good answer at the moment! Would you be comfortable with merging as-is?

Yup, I made the change to line 43 that Saul suggested.

saulpw (approved):

Issue #2128, opened 2023-11-19
PostgreSQL table names not quoted: `UndefinedTable` relation "public.whatever" does not exist for table "Whatever"


isosphere:
**Small description**

Connected to a local PostgreSQL database. `PgTablesSheet` looks good. Press "Enter" on any table that is not exclusively lowercase, get the error in the title.

Try with another table, and always get `InFailedSqlTransaction: current transaction is aborted`

**Expected result**
I should see the table I selected

**Actual result with screenshot**
![image](https://github.com/saulpw/visidata/assets/163370/15aed0db-5e0e-4319-b9e8-f98f1b62b164)

If you get an unexpected error, please include the full stack trace that you get with `Ctrl-E`.
![image](https://github.com/saulpw/visidata/assets/163370/c53386ba-8eea-4c89-b44f-b68a2e855206)

As you can see from the stack trace, the table identifier is not quoted. In PostgreSQL, `public.HelloWorld` is interpreted as `public.helloworld`. If the actual table name is `public.HelloWorld`, you have to use quotes. This applies for columns, too. You can solve this by always quoting identifiers. 

**Additional context**
Please include the version of VisiData and Python.

VisiData 2.11,1, Python 3.10.4

isosphere:
I think we just have to change:

https://github.com/saulpw/visidata/blob/9c31e8dabe74774f4bebf546d8e15f5b7030bd70/visidata/loaders/postgres.py#L129

To read:

`with self.sql.cur(f'SELECT * FROM "{source}"') as cur:`

saulpw:
@isosphere would you be willing to test that change and submit a PR?

isosphere:
> `with self.sql.cur(f'SELECT * FROM "{source}"') as cur:`

PR submitted; my proposed solution in this quote was not enough, as "public.TableName" refers to the literal object named "public.TableName", not schema public, table TableName. I've quoted them separately, which works.

saulpw:
Looks like this was fixed and we just forgot to close it.  Thanks @isosphere!

PR #2129, opened 2023-11-20
[postgresql] quote schema and table name

isosphere:
Without this, non-lowercase objects cannot be loaded. This builds on #852.

#2128

Tested on my own local PostgreSQL database which is rife with CamelCase table names.


saulpw (approved):
Thanks, @isosphere!

PR #2137, opened 2023-11-26
[join] clarify ensureLoaded status msg

anjakefala:
Was a confusing message to see while the joined sheet was loading.

### DIFF
--- a/visidata/features/join.py
+++ b/visidata/features/join.py
@@ -12,7 +12,7 @@ def ensureLoaded(vd, sheets):
     threads = [vs.ensureLoaded() for vs in sheets]
     threads = [t for t in threads if t]
     if threads:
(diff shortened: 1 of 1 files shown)


saulpw (commented):
agree with adding the `if threads`, but the wording is awkward.  is the message only awkward for joins?  is it because they're not source sheets?

anjakefala:
@saulpw The awkward thing is that the message pops up while you are loading the joined sheet. I would get confused because the joined sheet was loading, and I would see a message `'loading 0 sheets'`. I wasn't sure what those sheets were until I read the source code. I think this only affects joined sheets where the `ensureLoaded` is not for the sheet itself, but its source sheets.

saulpw (approved):

PR #2145, opened 2023-11-29
[docs] update plugins API docs to go with 7db5741781

anjakefala:


docs/api/plugins.rst
```diff
@@ -1,39 +1,22 @@
...
-- *pydeps*: space-separated list of PyPI dependencies (like in ``requirements.txt``).
-- *vdplugindeps*: space-separated list of vd plugin dependencies.
-- *sha256*: SHA256 hash of plugin .py of most recent release. A script for obtaining this has can be found `in dev/vdhash.py <https://raw.githubusercontent.com/saulpw/visidata/develop/dev/vdhash.py>`__ .
-
 .. note::
 
-    VisiData installs plugins in the ``$HOME/.visidata/plugins/`` directory.
-
-    ``import`` statements for all enabled plugins go into ``$HOME/.visidata/plugins/__init__.py``.
+    At startup, all .py files placed in the ``$HOME/.visidata/`` directory or Python code added to a ``~/.visidatarc`` will be auto-imported into VisiData.
```
saulpw: It's not true that all .py files in .visidata will be imported.  The `.visidata/plugins` module will be imported only (and thus imported modules in its `__init__.py`).

docs/api/plugins.rst
```diff
@@ -54,31 +37,15 @@ hello.py
...
-     "url": "https://raw.githubusercontent.com/saulpw/visidata/387de72b369039ae864f1b84d8191d1085c2105b/plugins/hello.py",
-     "latest_ver": "1.0",
-     "visidata_ver": "2.0",
-     "pydeps": "",
-     "vdplugindeps": "",
-     "sha256": "6b3baf4e1c4550947e611b9f1bdc96b7193f85048d0fc83880c4be0e7a5537d4"
-    }
-
-- Use ``sha256 hello.py`` to compute the hash.
+* a ``__version__`` of the current release
```
saulpw: We're not versioning plugins in the mono-repo any more.  They can add a `__author__` and `__description__` however.

docs/api/plugins.rst
anjakefala: I got muddled there; good catch!

saulpw (approved):

PR #2148, opened 2023-12-01
[xml] fix file loading

midichef:
This fixes a typo to match the other calls to `open_text_source()` in a361d4ceca3382c0c3c5cb447c774e4d0fc8f0de. Right now loading a local xml file with `vd -f xml file.xml` gives this error:
```
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/loaders/xml.py", line 54, in iterload
    self.root = etree.parse(self.source.open_text_source(), parser=p)
  File "/home/midichef/.local/lib/python3.10/site-packages/visidata/path.py", line 209, in __getattr__
    r = getattr(self._path, k)
AttributeError: 'PosixPath' object has no attribute 'open_text_source'
```

### DIFF
--- a/sample_data/benchmark.xml
+++ b/sample_data/benchmark.xml
@@ -1,0 +1,513 @@
+<?xml version='1.0' encoding='utf-8'?>
+<data>
+  <row>
+    <index>0</index>
+    <Date>7/3/2018 1:47p</Date>
+    <Customer>Robert Armstrong</Customer>
+    <SKU>FOOD213</SKU>
+    <Item>BFF Oh My Gravy! Beef &amp; Salmon 2.8oz</Item>
+    <Quantity>4</Quantity>
+    <Unit>$12.95</Unit>
+    <Paid>$51.8</Paid>
+  </row>
+  <row>
+    <index>1</index>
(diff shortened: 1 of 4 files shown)


anjakefala:
(I updated the branch with develop, so we could run the CI on the fixed tests.)

saulpw (approved):

midichef:
I just updated the commit message to be `[xml-]` instead of `[xml]`. Because I just read this in `dev/style.md` today.
> The trailing `-` is for cases when the entire commit is safe and/or desirable to take in a patch release--tightly constrained and the code afterwards is unilaterally improved.
When we put together a patch release, we scan commit messages for this marker to see which ones should be included.
Ideally only and all `-` marked commits since the last release, would be in a patch release.

Up till now I just assumed it meant "minor change"!

saulpw:
> Up till now I just assumed it meant "minor change"!

It started off meaning "bugfix" but it's since acquired a more rigorous definition :)

anjakefala:
I added a test for a basic loading of xml files. We should be able to catch problems like this sooner now!

PR #2166, opened 2023-12-09
[guide] add guide for Errors Sheet

anjakefala:
helpstr needs to be pullable from parent Sheet types.

### DIFF
--- a/visidata/features/errors_guide.py
+++ b/visidata/features/errors_guide.py
@@ -1,0 +1,26 @@
+from visidata import GuideSheet, vd
+
+class ErrorsGuide(GuideSheet):
+    guide_text='''# What was that error?
+
+Status messages include [:warning]warnings[/] and [:error]errors[/].
+
+A command may issue a [:warning]warning[/] status and continue running.
+A command that [:warning]fails[/] or [:error]errors[/] is aborted.
+
+## Investigating errors further
+
+If a Python Exception like [:error]RuntimeError[/] appears in the sidebar:
+
+- {help.commands.error_recent}
+- {help.commands.errors_all}
+
+If [:note_type]{vd.options.note_format_exc}[/] or [:error]{vd.options.note_getter_exc}[/] appear inside a cell, it indicates an error happened during calculation, type-conversion, or formatting.  When the cursor is on an error cell:
+
+- {help.commands.error_cell}
+'''
+
+
+
+vd.addGuide('ErrorsSheet', ErrorsGuide)
+


visidata/features/errors_guide.py
```diff
@@ -0,0 +1,29 @@
+from visidata import GuideSheet, vd
+
+class ErrorsGuide(GuideSheet):
+    guide_text='''# A guide to errors in VisiData
+
+There are two severity levels VisiData will use when communicating something has gone wrong.
+If in performing an operation it reaches a [:166]warning[/], the operation will continue.
+With [:202]errors[/], operations will stop.
+
+[:202]Errors[/] and/or [:166]warnings[/] can manifest in different ways. How they appear determines how they are folllowed up on.
```
saulpw: `folllowed` is a typo.  How could we run these through a spell-checker?

visidata/features/errors_guide.py
```diff
@@ -0,0 +1,29 @@
...
+
+1. [:202]a phrase all in lowercase[/] [:166]appears in the sidebar[/]
```
saulpw: Let's use semantic instead of numeric colors, and be more specific than "a phrase": `[:warning]an error message[/][:sidebar]appears in the sidebar[/]`

visidata/features/errors_guide.py
```diff
@@ -0,0 +1,29 @@
...
+This is a hand-written message written for a known potential scenario. An example is *"invalid regex"* when the user passes an invalid regex string to a VisiData command.
+For more information on what condition triggered the message `git grep` for it within the VisiData repo.
+
+2. The first line of a stacktrace for an [:202]Exception[/] appears in the sidebar.
+Most `Exceptions` in VisiData are caught, and VisiData will continue running even if an operation has failed.
```
saulpw: Let's be more sparing with `code` within VisiData itself.  Keystrokes, longnames, CLI options--anything a user would actually type.  `Exception` is more of a dev word.

visidata/features/errors_guide.py
```diff
@@ -0,0 +1,29 @@
...
+    guide_text='''# A guide to errors in VisiData
+
+There are two severity levels VisiData will use when communicating something has gone wrong.
+If in performing an operation it reaches a [:166]warning[/], the operation will continue.
+With [:202]errors[/], operations will stop.
+
+[:202]Errors[/] and/or [:166]warnings[/] can manifest in different ways. How they appear determines how they are folllowed up on.
+
+1. [:202]a phrase all in lowercase[/] [:166]appears in the sidebar[/]
+This is a hand-written message written for a known potential scenario. An example is *"invalid regex"* when the user passes an invalid regex string to a VisiData command.
```
saulpw: Two "written"s in the same sentence.

saulpw: We shouldn't use "VisiData" in these online docs.  The reader will be using VisiData, so the VisiData context can be assumed, and we want to keep these docs as short as possible.

visidata/features/errors_guide.py
```diff
@@ -0,0 +1,29 @@
+from visidata import GuideSheet, vd
+
+class ErrorsGuide(GuideSheet):
+    guide_text='''# A guide to errors in VisiData
+
+There are two severity levels VisiData will use when communicating something has gone wrong.
+If in performing an operation it reaches a [:166]warning[/], the operation will continue.
```
saulpw: VisiData should use "command" instead of "operation".

visidata/features/errors_guide.py
```diff
@@ -0,0 +1,29 @@
+from visidata import GuideSheet, vd
+
+class ErrorsGuide(GuideSheet):
+    guide_text='''# A guide to errors in VisiData
```
saulpw: Drop "A guide to"

visidata/features/errors_guide.py
anjakefala: Seems I should run `pylint` locally (https://stackoverflow.com/questions/27162315/automated-docstring-and-comments-spell-check).

I don't think we want to add `pylint` to the CI. It became annoying for every project I was part of.

visidata/features/errors_guide.py
```diff
@@ -0,0 +1,29 @@
...
+
+There are two severity levels VisiData will use when communicating something has gone wrong.
+If in performing an operation it reaches a [:166]warning[/], the operation will continue.
+With [:202]errors[/], operations will stop.
+
+[:202]Errors[/] and/or [:166]warnings[/] can manifest in different ways. How they appear determines how they are folllowed up on.
+
+1. [:202]a phrase all in lowercase[/] [:166]appears in the sidebar[/]
+This is a hand-written message written for a known potential scenario. An example is *"invalid regex"* when the user passes an invalid regex string to a VisiData command.
+For more information on what condition triggered the message `git grep` for it within the VisiData repo.
```
saulpw: A user really can't be expected to do this.  I'd rather say "come to github and help us make a better error message". Or let's just cut the sentence entirely.

visidata/features/errors_guide.py
```diff
@@ -0,0 +1,29 @@
...
+
+2. The first line of a stacktrace for an [:202]Exception[/] appears in the sidebar.
```
saulpw: "A Python Exception like RuntimeError appears in the sidebar"?  `stacktrace` is not well-defined here.

visidata/features/errors_guide.py
```diff
@@ -0,0 +1,29 @@
...
+Most `Exceptions` in VisiData are caught, and VisiData will continue running even if an operation has failed.
+- {help.commands.error_recent}
+- {help.commands.errors_all}
+
+3. An [:166]?[/] or [:202]![/] appear in a cell.
+These are caught warnings or **Exceptions** that occurred while calculating the value of a cell.
```
saulpw: "This is an error during calculation, type-conversion, or formatting."

visidata/features/errors_guide.py
```diff
@@ -0,0 +1,29 @@
+from visidata import GuideSheet, vd
+
+class ErrorsGuide(GuideSheet):
+    guide_text='''# A guide to errors in VisiData
+
+There are two severity levels VisiData will use when communicating something has gone wrong.
```
saulpw: There are actually 5: "debug" "aside" "status" "warning" "error".  And a warning may abort the command (fail) or it may not.

saulpw (changes requested):
Let's be sure and codify each of these into our style guide!

visidata/features/errors_guide.py
saulpw: totally, I don't like it either.  Maybe run it manually?  It might be a good step to include for doc authors.

anjakefala: `pylint` did not catch 'folllowed' =/

visidata/features/errors_guide.py
anjakefala: Which word should I use instead of `Exception`? Just describe that they are looking at a traceback or stacktrace? I'm feeling conflicted, because they will bump into `Exception`s if they are interacting with the **Errors Sheet**.

visidata/features/errors_guide.py
anjakefala: In reference to https://github.com/saulpw/visidata/pull/2166#discussion_r1432198312, does this mean for this doc I can mention "Python Exception" or "RuntimeError"?

visidata/features/errors_guide.py
anjakefala: Ooooo, so cool! I had no idea I could use these semantics. That's a lot better.

visidata/features/errors_guide.py
anjakefala: Thank you, that's much better!

visidata/features/errors_guide.py
```diff
@@ -0,0 +1,35 @@
+from visidata import GuideSheet, vd
+
+class ErrorsGuide(GuideSheet):
+    guide_text='''# Errors
+
+There are five levels for sidebar status updates: [:status]"debug"[/], [:status]"aside"[/], [:status]"status"[/], [:warning]"warning"[/], and [:error]"error"[/].
+
+[:status]Status[/] is meant to display neutral information.
```
anjakefala: I added these, because I mentioned them, but maybe I should just say, "this document focuses on errors + warnings". There is going to be a guide on the Sidebar.

PR #2168, opened 2023-12-10
[guide] Add guide to menu

anjakefala:
#2313

Screenshot from 2024-02-22 22-08-23

### DIFF
--- a/visidata/guides/MenuGuide.md
+++ b/visidata/guides/MenuGuide.md
@@ -1,21 +1,23 @@
 # The Menu System
 
-Located at the very top of the screen, the menu organizes many commands within submenu trees.
+The menu (on the first row of the window) contains most commands, organized within a tree of submenus.
 
-Navigate the menu like a GUI by using the mouse to click on menu items.
+Click on menu items to navigate the menu like a GUI.
 
-To navigate with the keyboard, `Alt+H` will open the **Help** menu, and from there use the arrow keys to move between submenus.
+With the keyboard:
 
-Additionally, each top-level menu name has a single letter underlined. `Alt+<underlined letter>` will open that menu. For example, `Alt+F` will open the **File** menu.
+- {help.commands.menu_help}
+- [:keys]Arrow[/] keys to move between submenus
 
-To open a submenu (indicated by `{vd.options.disp_menu_more}`) press the `→` key or click on it with the mouse. Each non-submenu item is a command.
+Additionally, each top-level menu name has a single letter underlined. [:keys]Alt+<underlined letter>[/] to open that menu. For example, [:keys]Alt+F[/] to open the **File** menu.
 
-The menu shows the description of the current command, along with the command name and its keyboard shortcut (if available). Execute the current command by clicking on it or pressing `Enter`.
+Each non-submenu item is a command. It shows the description of the current command, along with the command name and its keyboard shortcut (if available). Click on the item or [:keys]Enter[/] to execute it.
 
-Some other indicators:
+Indicators:
 
-- `{vd.options.disp_menu_push}`: the command opens another sheet.
-- `{vd.options.disp_menu_input}`: the command requires input.
+- {help.options.disp_menu_more}
+- {help.options.disp_menu_push}
+- {help.options.disp_menu_input}
 
 Some commands are only available on a specific sheet type, and appear [:color_menu_spec]highlighted[/] in the menu when the current sheet is that type. If the command is not available, it will not appear at all.
 
--- a/visidata/guides/MenuGuide.md
+++ b/visidata/guides/MenuGuide.md
@@ -1,0 +1,24 @@
+# The Menu System
+
+Located at the very top of the screen, the menu organizes many commands within submenu trees.
+
+Navigate the menu like a GUI by using the mouse to click on menu items.
+
+To navigate with the keyboard, `Alt+H` will open the **Help** menu, and from there use the arrow keys to move between submenus.
+
+Additionally, each top-level menu name has a single letter underlined. `Alt+<underlined letter>` will open that menu. For example, `Alt+F` will open the **File** menu.
+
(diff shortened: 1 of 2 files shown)


visidata/features/menu_guide.py
```diff
@@ -0,0 +1,29 @@
+from visidata import GuideSheet, vd
+
+class MenuGuide(GuideSheet):
+    guide_text = '''# The Menu System
+
+Located at the [:black on 68]very top of the screen[/], it contains a collection of commands organised within submenu trees.
```
saulpw: We should use the semantic colors (would be `[:menu]` in this case).  But I don't know about using colors inline like this anyway.  Let's drop this color.

Also, "the menu organizes many commands within submenu trees."

- Use American spellings for consistency (sorry).
- "contains" is passive voice; "contains" "collection" "organized" are all saying the same basic thing.

visidata/features/menu_guide.py
```diff
@@ -0,0 +1,29 @@
...
+
+The menu can be navigated both as a standard, clickable GUI menu, and via keystrokes.
```
saulpw: Don't use "can be X": "Navigate the menu like a GUI by using the mouse to click on menu items."

visidata/features/menu_guide.py
```diff
@@ -0,0 +1,29 @@
...
+
+To navigate via keystrokes, `Alt+H` will open the **Help** menu, and from there `↑ ↓ → ←` keys can be used to move between submenus.
```
saulpw: - "with" is preferable to "via" (simpler English):  "To navigate with the keyboard". 
- we should probably just say "arrow keys" instead of using unicode chars in this case.
- Let's combine this and the next one, so there's only one example.

visidata/features/menu_guide.py
```diff
@@ -0,0 +1,29 @@
...
+
+Additionally, each top-level menu name has a single letter underlined. `Alt+<underlined letter>` will open that menu. For example, `Alt+F` will open the **File** menu.
+
+`{vd.options.disp_menu_more}` indicates a submenu, which can be traversed with the `→` or a mouse-click. The leaf node of every menu tree is a command.
```
saulpw: "To open a submenu (indicated by X), `→` or click on it with the mouse.  Each non-submenu item is a command."
- no "can be" passive voice
- leaf/tree/node is CS technical jargon

visidata/features/menu_guide.py
```diff
@@ -0,0 +1,29 @@
...
+
+When the cursor is on a command, the menu throws up its [:color_menu_help]helpbox[/]. This contains the description of the command, the keyboard shortcut if available, and its command longname. Clicking on the command in the menu, or pressing `Enter`, will execute the actual command.
```
saulpw: "throws up"?  How about "The menu shows the description of the current command, along with the command name and its keyboard shortcut (if available)."
- avoid future/conditional tenses; use imperative/infinitive instead: "Execute the current command by clicking on it or pressing `Enter`".

visidata/features/menu_guide.py
```diff
@@ -0,0 +1,29 @@
...
+
+`{vd.options.disp_menu_push}` indicates that the command will push a sheet onto sheet stack. `{vd.options.disp_menu_input}` indicates that input will be required for the command.
```
saulpw: "Some other indicators:
- `{vd.options.disp_menu_push}`: the command opens another sheet.
- `{vd.options.disp_menu_input}`: the command requires input.

visidata/features/menu_guide.py
```diff
@@ -0,0 +1,29 @@
...
+
+[:color_menu_spec]Sheet-specific commands[/] will only appear in the menu when that specific sheet type is loaded.
```
saulpw: "Some commands are only available on a specific sheet type, and appear [:color_menu_spec]highlighted[/] in the menu when the current sheet is that type.  If the command is not available, it will not appear at all."

visidata/features/menu_guide.py
```diff
@@ -0,0 +1,29 @@
...
+
+## Options
+
+- {help.options.disp_menu}
+- {help.options.disp_menu_fmt}
```
saulpw: `disp_menu` makes sense, but do we want to call out the other ones?  They are more thematic and aren't important for using or understanding the menu.

saulpw (changes requested):

visidata/guides/MenuGuide.md
```diff
@@ -0,0 +1,24 @@
+# The Menu System
+
+Located at the very top of the screen, the menu organizes many commands within submenu trees.
+
+Navigate the menu like a GUI by using the mouse to click on menu items.
```
saulpw: Our phrasing is, "Click on menu items to navigate the menu like a GUI".  Always "Input X to do Y".

visidata/guides/MenuGuide.md
```diff
@@ -0,0 +1,24 @@
...
+
+To navigate with the keyboard, `Alt+H` will open the **Help** menu, and from there use the arrow keys to move between submenus.
```
saulpw: Similarly, "Alt+H to open the Help menu" and "Arrow keys to move between submenus", probably on a separate line.

visidata/guides/MenuGuide.md
```diff
@@ -0,0 +1,24 @@
...
+
+Additionally, each top-level menu name has a single letter underlined. `Alt+<underlined letter>` will open that menu. For example, `Alt+F` will open the **File** menu.
```
saulpw: and apply throughout.

visidata/guides/MenuGuide.md
```diff
@@ -0,0 +1,24 @@
+# The Menu System
+
+Located at the very top of the screen, the menu organizes many commands within submenu trees.
```
saulpw: "The menu (on the first row of the window) contains most commands, organized within a tree of submenus."

saulpw (changes requested):

saulpw (approved):
This is good, thank you @anjakefala!

PR #2185, opened 2023-12-22
[setup] install regex help file

### LINKED ISSUE #2181: inputMultiple() displays a malformed empty status box
**Small description**
`search-keys` and `search-col` displays an empty box while waiting for the user to input a regex at the prompt. The box is shown where the usual status messages are shown.

**Expected result**
No box should be displayed.

sidebar-inputMultiple

For comparison, here is what the normal status messages look like:
sidebar-normal

Open any sheet and hit `r` or `/`.

**Additional context**
saul.pw/VisiData v3.0dev
Python 3.10.12
Ubuntu 22.04.3

midichef:
Fixes #2181.

In the course of exploring the installation process, I ran into a surprise. If I reverted this fix by removing `regex.ddw` from `setup.py` and `MANIFESET.in`, it did not restore the broken behavior. It turns out that `regex.ddw` kept being installed, because my local `visidata.egg-info/SOURCES.txt` retained `visidata/ddw/regex.ddw`. I'll have to remember that in the future.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,5 +4,6 @@ include visidata/man/vd.1
 include visidata/man/vd.txt
 include visidata/man/visidata.1
 include visidata/ddw/input.ddw
+include visidata/ddw/regex.ddw
 include visidata/tests/sample.tsv
 include visidata/desktop/visidata.desktop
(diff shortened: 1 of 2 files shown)


saulpw (approved):

anjakefala:
Wild!

Thanks for fixing it in time for v3.0 =)

PR #2186, opened 2023-12-22
[statusbar-] move sheet.longname into vd object

midichef:
Immediately after moving to a new sheet, the `sheet.longname` in the statusbar is incorrect. On the latest 3.0dev as well as 2.11.1.

For example, right after `sheets-stack`, the longname displayed is blank. That's because the new sheet's `sheet.longname` is `''`. Then if we quit out of that sheet with `q`, the longname shown in the original sheet is incorrect. I expect it to show `"quit-sheet"`, but instead it shows `"sheets-stack"`, the name of last command executed in the original sheet.

This PR moves `longname` into the vd object, instead of storing it in each sheet. Will this work?

And is it safe to remove `BaseSheet.longname`?
```
BaseSheet.init('longname', lambda: '')
```
I think it's now obsolete but it's possible I missed a use somewhere.

### DIFF
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -1032,6 +1032,8 @@ def push(vd, vs, pane=0, load=True):
 
     if load:
         vs.ensureLoaded()
+    if vd.activeCommand:
+        vs.longname = vd.activeCommand.longname
 
 
 @VisiData.api
@@ -1042,6 +1044,8 @@ def quit(vd, *sheets):
         vs.confirmQuit('quit')
         vs.pane = 0
         vd.remove(vs)
+    if vd.activeCommand:
+        vd.activeSheet.longname = vd.activeCommand.longname
 
 
 @BaseSheet.api


saulpw:
Thanks for noticing and addressing this.  I believe sheet.longname isn't used anywhere else, and could be removed with this change.  However, I believe that this change means both windows in splitpane will show the same last command; the intent is that if you don't do anything in one of the panes, it will still show the last command executed on that sheet.  It's a minor issue, and I don't think we need to preserve that behavior, so perhaps we shouldn't show the previous command on the inactive sheet at all.  But we should make sure splitpane acts reasonably here too.

midichef:
Oh, now I see the reason why each sheet stores longname.

What if we copy `sheet.longname` from the active sheet to the next sheet, when doing `vd.push()` and `vd.quit()`?
Something like https://github.com/saulpw/visidata/commit/2a7b8b0507fffdbaea2448fb6c11e2af5db1fbfe?

saulpw:
> What if we copy sheet.longname from the active sheet to the next sheet, when doing vd.push() and vd.quit()?

Does this work for both cases?  If so I'm willing to try it out.

midichef:
It works with a single pane and for split panes.

But it is a bit hacky, and `push()` is called in a few dozen lines where it's not in a command string in `addCommand`. I can imagine situations where `push()` gets called, and `longname` gets copied when it should not be. I looked the existing uses over and don't see any such situations, but I could have missed one. But even if it's not happening now, it seems likely to eventually happen.

The situation for `quit()` is safer. It is only called in 3 commands and I checked them all.

So maybe only `quit()` should copy `longname`, and `push()` should not.

midichef:
I've replaced the original commit with the discussed (https://github.com/saulpw/visidata/commit/2a7b8b0507fffdbaea2448fb6c11e2af5db1fbfe) that keeps `sheet.longname` and carries it across `push()` and `quit()`.

For future reference, here are all the uncertain calls to `vd.push()` that I thought about for this patch. And some notes.
[push.json](https://github.com/saulpw/visidata/files/14455577/push.json)
The rest of the calls to `vd.push()` are all done directly in the `addCommand()` command string, or done in a function only called in the command string. So they should work fine.

It's a bit inelegant, but I think it's worth it, to help orient users. Sometimes I really want to see what command I just activated to create a new sheet.

midichef:
The drawback of my previous proposed fix is that `push()` and `quit()` can be called from contexts that are not a command string. Like plugins, or .visidatarc user code, or future changes to Visidata source code.

There is a better solution (https://github.com/saulpw/visidata/commit/e7da9855f5ffecc677755d674418680da81af998). Instead of copying longname from the last sheet on `push()`/`quit()`, read it from `activeCommand`. Then status will only be set when `push()` and `quit()` are called from commands.

`activeCommand` has a threading problem at the moment, so it's safest to wait for a resolution on #2345. That's why I've placed https://github.com/saulpw/visidata/commit/e7da9855f5ffecc677755d674418680da81af998 on the tip of that branch.

midichef:
I've rebased this patch onto current develop, so it's ready for review.

saulpw (approved):

PR #2189, opened 2023-12-24
[macros] add fail if no visidata_dir to save in

anjakefala:


visidata/macros.py
```diff
@@ -89,7 +89,10 @@ def saveMacro(self, rows, ks):
         vs = copy(self)
         vs.rows = rows
         macropath = Path(vd.fnSuffix(str(Path(vd.options.visidata_dir)/ks)))
-        vd.save_vdj(macropath, vs)
```
saulpw: it would be nicer to fail before they've gone through the whole macro :)

saulpw (approved):

PR #2193, opened 2023-12-27
Allow custom accept-then keybindings from edit mode

ajkerrigan:
Allow a mapping in the form "keystroke -> iterable-of-vd-functions" to define custom keybindings that accept input and then queue one or more vd functions. With these changes, an option like the following can define custom "accept/move/re-edit" flows:

```python
vd.options.edit_accept_keybindings = {
    "Ctrl+N": ("go-down", "edit-cell"),
    "Ctrl+P": ("go-up", "edit-cell"),
}
```

### DIFF
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -13,6 +13,7 @@ vd.theme_option('color_edit_cell', '233 on 110', 'cell color to use when editing
 vd.theme_option('disp_edit_fill', '_', 'edit field fill character')
 vd.theme_option('disp_unprintable', '·', 'substitute character for unprintables')
 vd.theme_option('mouse_interval', 1, 'max time between press/release for click (ms)', sheettype=None)
+vd.option('edit_accept_keybindings', {}, 'mapping of keystrokes to vd functions to run after accepting edit input')
 
 vd.disp_help = 1  # current level of help shown (up to vd.options.disp_help as maximum)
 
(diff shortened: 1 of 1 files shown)


visidata/_input.py
```diff
@@ -623,6 +625,7 @@ def editCell(self, vcolidx=None, rowidx=None, value=None, **kwargs):
 
     # update local bindings with kwargs.bindings instead of the inverse, to preserve kwargs.bindings for caller
     bindings.update(kwargs.get('bindings', {}))
+    bindings.update({k: acceptThenFunc(*v) for k, v in options.edit_accept_keybindings.items()})
```
saulpw, line 628: I like the idea of allowing people to customize their keybindings for edit/input, but the 'always accept' gives me pause every time I come back to it.  I think it'd be better if we had user-set bindings apart from options, and required them to manually specify something like `acceptThenFunc`.  Like:

```
vd.edit_bindings = {
    'Ctrl+X': acceptThenFunc('syscopy-cell', 'delete-cell')
}
```

but that API is pretty terrible too.

saulpw: Thanks again for starting a discussion on this, @ajkerrigan.  I think this isn't quite the way we should go with it, and the input code has been refactored recently anyway.   So we're going to close this for now and find another way (eventually) for people to add custom keybindings in their .visidatarc to the input widget.

saulpw:
Thanks again for starting a discussion on this, @ajkerrigan. I think this isn't quite the way we should go with it, and the input code has been refactored recently anyway. So we're going to close this for now and find another way (eventually) for people to add custom keybindings in their .visidatarc to the input widget.

PR #2196, opened 2023-12-29
[save-] save by options.save_filetype if save_ext does not exist

anjakefala:


saulpw (approved):
I like it, we'll refactor if necessary later.

PR #2199, opened 2023-12-30
[json] warn if some cols/sheets were not saved

midichef:
The JSON dict format can't hold multiple sheets with the same name, or multiple columns with the same name.

This drawback of JSON may be obvious to some users, but it surprised me when I first discovered it. My thoughts only went as far as "I like how JSON remembers what columns are numbers. That makes JSON better than TSV, so I'm going to save in JSON from now on."  Also, name collisions are easy to forget about. I forgot it would happen for 2 sheets with the same name (but different filetypes), while testing this very PR. So it's probably going to catch other users off guard too.

So this PR warns the user when some columns or sheets will be lost in the resulting saved data file.

The warnings are pretty long, and I welcome any better phrasings.

An issue not addressed by this PR:  JSONL has a problem in saving multiple sheets in `save_jsonl()`. No rows will be lost, but the file will lose information on which sheet each row belongs to.

### DIFF
--- a/visidata/loaders/json.py
+++ b/visidata/loaders/json.py
@@ -1,4 +1,5 @@
 import json
+from collections import Counter
 
 from visidata import vd, date, anytype, VisiData, PyobjSheet, AttrDict, stacktrace, TypedExceptionWrapper, AlwaysDict, ItemColumn, wrapply, TypedWrapper, Progress, Sheet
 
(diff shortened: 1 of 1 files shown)


saulpw:
I apologize for sitting on this PR for awhile. The fundamental premise is good--let's warn the user for something like this that seems highly suspect and could result in lost data.  But yeah, the warnings are too long, and I'm worried about the performance implications of the unrolled list comprehension, and it also becomes 10x as much code.  So I'm not excited to take this in its current form.

Do you think this is important enough to spend more time on right now?

anjakefala:
Thoughts @midichef?

midichef:
@saulpw Your feedback was right on. I just force-pushed a much cleaner version (ec6256a) than the original messy commit (https://github.com/saulpw/visidata/commit/f3ffadfdd55e045e7b4718a9bf51fd22a5d657fa). How's this one look?

I also added a warning for .jsonl saving multiple sheets, since it will lose information about which rows belong to which sheets.

PR #2240, opened 2024-01-07
[csv] handle more dialect parameters from Sniffer

midichef:
When visidata guesses a csv filetype, there are warnings due to unhandled options from `csv.Sniffer`. (Python 3.10.12, Ubuntu 22.04.3, visidata v3.1/dev)

`echo "a,b" |vd` produces:

```
setting unknown option csv_doublequote
setting unknown option csv_quoting
guessed "csv" filetype based on contents
```

This PR silences these warnings by adding these options in the csv loader.

This requires some different importing of `csv`, because to set `csv_quoting` I need `csv.QUOTE_MINIMAL`. (Its value on my system is `0`, and [in cpython it is defined in an enum](https://github.com/python/cpython/blob/f19b93fce04fb0bc9b59071915a6aa6b01860d8a/Modules/_csv.c#L91)).

The existing code waits to do `import csv` till loading/guessing/saving a CSV file. If we want to still put off that import, then we could move `vd.option('csv_quoting', csv.QUOTE_MINIMAL, ...` into `guess_csv()`.

I also explored the other option: `from csv import QUOTE_MINIMAL`, but surprisingly, `timeit` says that's several times slower than `import csv`, at least on my setup:
```
python -m timeit "import csv"
5000000 loops, best of 5: 71.7 nsec per loop
python -m timeit "from csv import QUOTE_MINIMAL"
500000 loops, best of 5: 524 nsec per loop
```

### DIFF
--- a/visidata/loaders/csv.py
+++ b/visidata/loaders/csv.py
@@ -3,7 +3,9 @@ from visidata import TypedExceptionWrapper, Progress
 
 vd.option('csv_dialect', 'excel', 'dialect passed to csv.reader', replay=True)
 vd.option('csv_delimiter', ',', 'delimiter passed to csv.reader', replay=True)
+vd.option('csv_doublequote', True, 'quote-doubling setting passed to csv.reader', replay=True)
 vd.option('csv_quotechar', '"', 'quotechar passed to csv.reader', replay=True)
+vd.option('csv_quoting', 0, 'quoting style passed to csv.reader and csv.writer', replay=True)
 vd.option('csv_skipinitialspace', True, 'skipinitialspace passed to csv.reader', replay=True)
 vd.option('csv_escapechar', None, 'escapechar passed to csv.reader', replay=True)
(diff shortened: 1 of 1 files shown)


anjakefala:
I can take a look at the build failure after 3.0.1 gets shipped! I just wanted to comment that the build failure is related to the PR. =) We're green on develop.

midichef:
I used the wrong value for `csv_doublequote`, the default is `True`, not `False`. I'm so glad the tests caught it, because this would have caused data to be corrupted!

saulpw:
It's basically okay to `import csv` at the top-level for this one, since it's a stdlib module.  But maybe can we use `0` instead of `csv.QUOTE_MINIMAL` instead?  I know it's not the best practice, but the value is unlikely to change, and the user is going to have to see and set a number anyway if they set this option themselves.

anjakefala:
Hey @midichef!

I went ahead and implement Saul's suggestion. =) We'll take this PR as-is.

PR #2244, opened 2024-01-10
[undo-] prevent undo of command that created sheet

midichef:
The undo command currently incorrectly attempts to undo the command that created the sheet.

To reproduce:
`vd sample_data/sample.tsv` then `g"` and `undo-last`. This will empty the sheet out.

The cause is an assumption in the undo code, that the first line of the command log is for the command that created the sheet:
https://github.com/saulpw/visidata/blob/5772c0f5e817fba2a7a1f9f5949cb959067cebca/visidata/undo.py#L40

Because `set-option` commands are carried into new sheets, that assumption is no longer true. (since 1925808cfcc2d9c2025c45c45927c8c5c40560a4)

This PR skips over the first sequence of command log rows for the `global` sheet, where `cmdlogrow.sheet == 'global'`. I think these commands are all for `set-option`, so we could instead match longname: `cmdlogrow.longname == "set-option"`. But I think excluding the first commands from the global sheet, is more robust to future changes.

### DIFF
--- a/visidata/undo.py
+++ b/visidata/undo.py
@@ -40,7 +40,10 @@ def undo(vd, sheet):
     cmdlogrows = itertools.dropwhile(lambda r: r.longname == 'set-option', sheet.cmdlog_sheet.rows)
     # skip the first remaining command, to exclude it from undo,
     # because it is always the one that created the sheet
-    next(cmdlogrows)
+    try:
+        next(cmdlogrows)
+    except StopIteration:
+        pass
     for i, cmdlogrow in enumerate(reversed(list(cmdlogrows))):
         if cmdlogrow.undofuncs:
             for undofunc, args, kwargs, in cmdlogrow.undofuncs[::-1]:
--- a/visidata/undo.py
+++ b/visidata/undo.py
@@ -40,11 +40,7 @@ def undo(vd, sheet):
     cmdlogrows = itertools.dropwhile(lambda r: r.longname == 'set-option', sheet.cmdlog_sheet.rows)
     # skip the first remaining command, to exclude it from undo,
     # because it is always the one that created the sheet
-    try:
-        next(cmdlogrows)
-    except StopIteration:
-        pass
-    for i, cmdlogrow in enumerate(reversed(list(cmdlogrows))):
+    for i, cmdlogrow in enumerate(reversed(list(cmdlogrows)[1:])):
         if cmdlogrow.undofuncs:
             for undofunc, args, kwargs, in cmdlogrow.undofuncs[::-1]:
                 undofunc(*args, **kwargs)
--- a/visidata/undo.py
+++ b/visidata/undo.py
@@ -37,10 +37,10 @@ def undo(vd, sheet):
     if not vd.options.undo:
         vd.fail("options.undo not enabled")
 
-    cmdlogrows = itertools.filterfalse(lambda r: r.longname == 'set-option', sheet.cmdlog_sheet.rows)
-    # exclude the first remaining command from undo,
+    cmdlogrows = itertools.dropwhile(lambda r: r.longname == 'set-option', sheet.cmdlog_sheet.rows)
+    # skip the first remaining command, to exclude it from undo,
     # because it is always the one that created the sheet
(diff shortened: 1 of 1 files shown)


visidata/undo.py
```diff
@@ -37,14 +37,23 @@ def undo(vd, sheet):
     if not vd.options.undo:
         vd.fail("options.undo not enabled")
 
-    # don't allow undo of first command on a sheet, which is always the command that created the sheet.
-    for i, cmdlogrow in enumerate(sheet.cmdlog_sheet.rows[:0:-1]):
+    cmdlogrows = sheet.cmdlog_sheet.rows
```
saulpw: How about:

    cmdlogrows = itertools.filterfalse(lambda r: r.longname == 'set-option')
    for i, cmdlogrow in enumerate(reversed(cmdlogrows[1:])):

?

anjakefala:
@midichef If you get the modification in by tomorrow, this can be part of v3.0.2 =)

visidata/undo.py
midichef: Okay I've incorporated this change.
Right now `cmdlog.py` only adds `'set-option'` commands before the sheet creation command, but never after it. So `filterfalse()` is fine as long as that is true.
Is there any situation where user-created code might cause the insertion of `'set-option'` after sheet creation? If that ever happens, `row_idx` will not be correct and the undo log would be corrupted.
https://github.com/saulpw/visidata/blob/e4f603b7b73cb20710d49328dd89fa3eb066f844/visidata/undo.py#L49

saulpw: If an option gets set by the user, it should be added as a `set-option` command on the source sheet cmdlog.  (I think this doesn't happen currently, which would mean that in-session option changes aren't recorded properly for replay.)  So your original code might have been better, though I can't see it after the force push.  Sorry to make an incorrect suggestion if that's the case!

Also what's the difference between this `islice` and `[1:]`?

midichef: Oops, I shouldn't have force pushed to overwrite the original code, then! That code skipped just the first sequential `set-option` commands. But it did it by calculating indices. It is shorter and cleaner to translate it to iterators as you suggested. So I've done that now with https://github.com/saulpw/visidata/pull/2244/commits/43ad4ae37d4652516d2d33739326308fa0b46ec0. 

As for the `islice`, it was needed because the generator from `filterfalse` didn't allow subscripts. But `next()` is all we actually need, so I got rid of the `islice`.

midichef:
@anjakefala okay this is ready to go, if it gets approved. Can you squash it into 1 commit before doing the final merge? Or I can do this myself. I don't quite know the usual protocol, but I do like to keep the commit history clean.

visidata/undo.py
```diff
@@ -37,8 +37,14 @@ def undo(vd, sheet):
     if not vd.options.undo:
         vd.fail("options.undo not enabled")
 
-    # don't allow undo of first command on a sheet, which is always the command that created the sheet.
-    for i, cmdlogrow in enumerate(sheet.cmdlog_sheet.rows[:0:-1]):
+    cmdlogrows = itertools.dropwhile(lambda r: r.longname == 'set-option', sheet.cmdlog_sheet.rows)
+    # skip the first remaining command, to exclude it from undo,
+    # because it is always the one that created the sheet
+    try:
```
saulpw: Is this just `list(cmdlogrows)[1:]`? If so I'd rather use that.  (If at some point on any PR you'd like me to just go ahead and make changes myself, let me know, I know I have some quirky taste sometimes.)

Thank you so much for diving into this whole issue, by the way.

anjakefala:
> okay this is ready to go, if it gets approved. Can you squash it into 1 commit before doing the final merge? Or I can do this myself. I don't quite know the usual protocol, but I do like to keep the commit history clean.

For sure! I can make a habit of cleaning your commits before merging.

visidata/undo.py
midichef: Sure, feel free to make any changes yourself. I don't mind and it's faster turnaround time.

saulpw (approved):

PR #2245, opened 2024-01-10
[docs-] add list of pip packages needed for tests

midichef:
I added a list of packages needed to get the tests to pass. Otherwise it takes several failing runs of `dev/test.sh` to discover all the missing packages.


anjakefala:
@midichef Oh!

The `requirements.txt` is what we recommend installing to run the tests. It does have a lot, and requires using a virtual environment to keep things contained, and easily eraseable. 

I don't mind maintaining an extras for tests. I would rather maintain an extras in setup.py for tests because I forgot this document existed. I would forget to update it.

midichef:
Right! I'd appreciate that. I like to install a minimal set of packages when running tests. If you do decide to maintain a list in setup.py, could you put a note in `dev/test.sh`, saying to look there?

anjakefala:
Done @midichef!

PR #2253, opened 2024-01-14
[cmdpalette-] check for used triggers in a resettable list

anjakefala:
Bindings is not reset upon a new loop.


saulpw (approved):
should really be a set or dict, but this is fine for N<10.

PR #2257, opened 2024-01-15
[loaders] improve fixed-width saver

### LINKED ISSUE #2255: Fixed width saver and loader don't round trip. (Columns expand with increasing number of spaces)
**Small description**

Open `test.csv`:

``` csv
colours,counts
red,3
green,5
blue,8
```

Then save it as `test.fixed`:

```
colours   counts   
red       3        
green     5        
blue      8        
```

Each column is separated by three spaces.

Now, open the newly saved `test.fixed`, and save it as `test.fixed.csv`.
Upon inspection, you will see that the three spaces have been included as column data, rather than being discarded as a separator:

```
colours,counts
red       ,3     
green     ,5     
blue      ,8     
```

If you repeat this process, three spaces get appended to each column every time you repeat a round trip.

**Expected result**

I expect to be able to round trip from CSV to Fixed and back without extra spaces being added to the data.

[test-fixed-saver.zip](https://github.com/saulpw/visidata/files/13938788/test-fixed-saver.zip)

**Additional context**

- saul.pw/VisiData v3.0.1
- Python 3.10.12

daviewales:
Don't add so many spaces between columns when saving.
Use the max width of the data, rather than the max width of the window when saving.
Don't include the separating space as data when loading. (Prevents extra space being added each round trip, as the saver also adds a single space between columns.)

Fixes #2255.

### DIFF
--- a/visidata/loaders/fixed_width.py
+++ b/visidata/loaders/fixed_width.py
@@ -1,5 +1,5 @@
 
-from visidata import VisiData, vd, Sheet, Column, Progress, SequenceSheet
+from visidata import VisiData, vd, Sheet, Column, Progress, SequenceSheet, Column, dispwidth
 
 
 vd.option('fixed_rows', 1000, 'number of rows to check for fixed width columns')
@@ -8,6 +8,22 @@ vd.option('fixed_maxcols', 0, 'max number of fixed-width columns to create (0 is
 @VisiData.api
 def open_fixed(vd, p):
     return FixedWidthColumnsSheet(p.base_stem, source=p, headerlines=[])
+
+@Column.api
+def getMaxDataWidth(col, rows):  #2255 need real max width for fixed width saver
+    '''Return the maximum length of any cell in column or its header,
+    even if wider than window. (Slow for large cells!)'''
+
+    w = 0
+    nlen = dispwidth(col.name)
+    if len(rows) > 0:
+        w_max = 0
+        for r in rows:
+            row_w = dispwidth(col.getDisplayValue(r))
+            if w_max < row_w:
+                w_max = row_w
+        w = w_max
+    return max(w, nlen)
 
 class FixedWidthColumn(Column):
     def __init__(self, name, i, j, **kwargs):
@@ -84,7 +100,7 @@ def save_fixed(vd, p, *vsheets):
             widths = {}  # Column -> width:int
             # headers
             for col in Progress(sheet.visibleCols, gerund='sizing'):
-                widths[col] = col.getMaxWidth(sheet.rows)  #1849
+                widths[col] = col.getMaxDataWidth(sheet.rows)  #1849 #2255
                 fp.write(('{0:%s} ' % widths[col]).format(col.name))
             fp.write('\n')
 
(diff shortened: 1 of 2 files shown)


visidata/loaders/fixed_width.py
```diff
@@ -38,7 +38,7 @@ def columnize(rows):
     # collapse fields
     for i in allNonspaces:
         if i > prev+1:
-            yield colstart, i
+            yield colstart, prev+1 #2255
```
daviewales, line 57: This could also be:

``` python
            yield colstart, i-1 #2255
```

`prev+1` collapses multiple spaces between columns.
`i-1` assumes only a single space is the column separator, and keeps the remaining spaces as data.

For example, given the following fixed width data with three spaces between the columns:

```
column_1    column_2
one         two
```

`prev+1` would be saved as the following CSV. (Separating spaces are removed, but padding spaces are retained):

```
column_1,column_2
one     ,two
```

`i-1` would be saved as the following CSV. (One separating space is removed. All others are retained.):

```
column_1  ,column_2
one       ,two
```

The current version of the code with `prev+1` is my personal preference, but open to guidance if the `i-1` version better matches the intent of 'fixed width'.

daviewales: Note, it must be `i-1` rather than `i` to avoid adding an extra space between the columns each round trip.

daviewales:
I think this is ready for initial review. Let me know if you're happy with the suggested changes, and I'll update the tests accordingly. (They are currently broken due to these changes.)

daviewales:
Note: Something which should be considered is a file like the following:

```
colours shades               counts 
red     light                3      
red     dark                 4      
green   very light           5      
green   very very very light 5      
blue    dark                 8      
```

VisiData reads this as four columns rather than three, because it splits columns based on unbroken vertical space, rather than header position.

Screen Shot 2024-01-23 at 4 23 01 pm

This behaviour remains unchanged by this PR.

visidata/loaders/fixed_width.py
saulpw: Okay, I prefer `prev+1` too.  Thanks for the explanation.

visidata/column.py
```diff
@@ -473,6 +473,20 @@ def getMaxWidth(self, rows):
         w = min(w, self.sheet.windowWidth)
         return w
 
+    def getMaxDataWidth(self, rows):  #2255  need real max width for fixed width saver
```
saulpw: Let's move this function into the fixed_width.py for now (with `@Column.api`), since it's the only consumer so far.

saulpw (changes requested):
Thanks for your attention on this, @daviewales!  If you wouldn't mind updating the tests too, we'd much appreciate it.

visidata/column.py
anjakefala: Ping @daviewales =)

anjakefala:
Hey @daviewales!

I went ahead and made the requested changes, and updated the tests. =) Thanks for taking care of this.

daviewales:
Thanks @anjakefala, much appreciated!

PR #2268, opened 2024-01-23
[guides] add JSON guide

ajkerrigan:
Add a guide for JSON, mostly focused on handling/navigating nested data. I also added a "Visual cues" section that _might_ be useful or just noisy. Because I couldn't decide one way or the other, I put it at the bottom :sweat_smile: .

### DIFF
--- a/visidata/loaders/json.py
+++ b/visidata/loaders/json.py
@@ -192,6 +192,11 @@ class JsonGuide(GuideSheet):
 
 ## Working with nested data
 
+VisiData uses these conventions to display nested data:
+
+- {{2}} indicates a nested object with 2 keys
+- [3] indicates a nested array with 3 elements
+
 [:note_type]Expanding[/] unpacks nested data column-wise, often useful for nested objects:
 
 - {help.commands.expand-col}
@@ -210,7 +215,7 @@ To revert earlier `expand-` operations:
 
 - {help.commands.unfurl-col}
 
-Note that `unfurl-col` creates a new sheet with `_unfurled` appended to the name. There is no command to revert an unfurl; instead, close the unfurled sheet.
+Note that `unfurl-col` creates a new sheet with `_unfurled` appended to the name. There is no command to revert an unfurl; instead, quit the unfurled sheet.
 
 For particularly deep or complex nested data, it can be helpful to open an individual cell as a new sheet:
 
@@ -221,37 +226,6 @@ For particularly deep or complex nested data, it can be helpful to open an indiv
 - {help.options.json_indent}
 - {help.options.json_sort_keys}
 - {help.options.json_ensure_ascii}
-
-## Visual cues
-
-VisiData provides visual cues to help identify and preview nested JSON data. Given the following data:
-
-`[`
-`  {{`
-`    "geolocation": {{`
-`        "type":"Point",`
-`        "coordinates": [6.08333,50.775]`
-`    }},`
-`  }},`
-`  {{`
-`    "geolocation": {{`
-`        "type":"Point",`
-`        "coordinates": [2.08,30.77]`
-`    }}`
-`  }}`
-`]`
-
-VisiData displays:
-
-`[:underline]geolocation[/]`
-`{{2}} type=Point coordinates=[2] 6.08333; 50.775`
-`{{2}} type=Point coordinates=[2] 2.08; 30.77`
-
-Where:
-
-- `{{2}}` indicates that the cell contains a [:note_type]nested object[/] with 2 keys
-- `type=Point coordinates=[2] 6.08333; 50.775` is a preview of that nested object
-- `coordinates=[2]` indicates a [:note_type]nested array[/] with 2 elements
 '''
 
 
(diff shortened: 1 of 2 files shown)


visidata/loaders/json.py
```diff
@@ -187,6 +187,74 @@ def JSON(vd, s:str):
...
+`        "coordinates": [6.08333,50.775]`
+`    }},`
+`  }},`
+`  {{`
+`    "geolocation": {{`
+`        "type":"Point",`
+`        "coordinates": [2.08,30.77]`
+`    }}`
+`  }}`
+`]`
```
ajkerrigan: This is admittedly weird. Triple backtick and `[:code]...[/]` didn't work for a multi-line code segment though. Maybe there's a good answer for formatting multi-line code blocks already, or maybe the better answer is to not try to do this in the first place (and save this sort of thing for posts/tutorials).

visidata/loaders/json.py
```diff
@@ -187,6 +187,74 @@ def JSON(vd, s:str):
...
 
 
+class JsonGuide(GuideSheet):
+    guide_text = '''# Some special features for JSON
+
+## Working with nested data
+
+[:note_type]Expanding[/] unpacks nested data column-wise, often useful for nested objects:
+
+- {help.commands.expand-col}
```
saulpw, line 202: holy cow, does this work?!  I thought it would have to be `expand_col` to be a valid identifier.

visidata/loaders/json.py
```diff
@@ -187,6 +187,74 @@ def JSON(vd, s:str):
...
+- {help.commands.contract-col}
+- {help.commands.contract-col-depth}
+- {help.commands.contract-cols}
+- {help.commands.contract-cols-depth}
+
+[:note_type]Unfurling[/] unpacks nested data row-wise, often useful for nested arrays:
+
+- {help.commands.unfurl-col}
+
+Note that `unfurl-col` creates a new sheet with `_unfurled` appended to the name. There is no command to revert an unfurl; instead, close the unfurled sheet.
```
saulpw: Let's use "quit" instead of "close".  (We need to make a glossary of VisiData terms that includes some translation rules like "use X instead of Y").

visidata/loaders/json.py
```diff
@@ -187,6 +187,74 @@ def JSON(vd, s:str):
...
+
+VisiData displays:
+
+`[:underline]geolocation[/]`
+`{{2}} type=Point coordinates=[2] 6.08333; 50.775`
+`{{2}} type=Point coordinates=[2] 2.08; 30.77`
+
+Where:
+
+- `{{2}}` indicates that the cell contains a [:note_type]nested object[/] with 2 keys
```
saulpw: Yeah, I think this explanation of the visual interface is useful, but maybe not the multiline JSON and the reiteration of what's on the display.  Ideally we would show it directly with a JSON example, and this could then also be the sidebar help for JsonSheet itself (which might clear up the confusion I had above with JsonSheet/JsonGuide).

In fact, let's move these next two lines up to the top as the primary explanation of when you'd want to use these features.  Then I think we can cut the rest of this section and put it into a more specific how-to if we want.

saulpw: Let's move these 2 lines up to the top, as a preface for when you'd need to use these commands.  Then the rest of the section can be cut and possibly included in a separate how-to.

visidata/loaders/json.py
anjakefala: Yup, it works!

visidata/loaders/json.py
anjakefala: JSON and some other formats allow cell values to be nested objects.                                                                                            
 
- `{2}` indicates the cell contains a mapping object with 2 keys                                                                                             

- `[3]` indicates that the cell contains a list of 3 values

visidata/loaders/json.py
saulpw: Also I'm thinking we should stick to Markdown for the guides and not use the special color tags, as these will be transcluded into the manpage and the webpages.  So bold or underline or italic?

anjakefala: @saulpw So I should remove mention of the special colour tags from the documentation guide? And remove them from the errors guide? 

I kinda like them! They add some flair. Maybe we can just ignore them in the manpage concatenator? Or interpret them as bold?

saulpw: You're right, they should be pretty easy to strip later.  Okay, let's keep them for now.

visidata/loaders/json.py
ajkerrigan: Trimmed & moved to the top based on suggestions.

visidata/loaders/json.py
ajkerrigan: Thanks for all the feedback. I've tried to incorporate it all, but it still might need some tweaking.

saulpw:
Perfect, I think we can ship this as it is!

saulpw (approved):
Thanks, @ajkerrigan !

PR #2275, opened 2024-01-29
[tsv-] let options set NUL delimiters

### LINKED ISSUE #2272: [tsv] add CLI option to use NUL as delimiter
It's useful to parse output from GNU grep's `-Z` option. That produces lines that in Python are `f'{filename}\0{line}\n'`, instead of the usual `f'{filename}:{line}\n'`.

Right now the command line can't be used to specify a NUL delimiter, as in `vd --delimiter="\0"`, because `sys.argv` strings are NUL-terminated and can't ever contain NUL.

My workarounds for now are to use .visidatarc, either add a temporary line:
`vd.option('delimiter', '\x00', 'field delimiter to use for tsv/usv filetype', replay=True)`.
or add a new filetype to allow `vd -f nsv`:
```
@VisiData.api
def open_nsv(vd, p):
    tsv = TsvSheet(p.base_stem, source=p)
    tsv.delimiter = '\x00'
    tsv.reload()
    return tsv
```
Can `open_nsv()` be written without `reload()` right now? I couldn't think of another way to set `delimiter` for TsvSheet.

midichef:
Closes #2272.

I also added `--row-delimiter=` to use `'\x00'` as the row separator. That'll be useful for the output of `find -print0` and for input for `xargs -0`.

### DIFF
--- a/visidata/features/xsv_guide.py
+++ b/visidata/features/xsv_guide.py
@@ -13,6 +13,8 @@ class XsvGuide(GuideSheet):
 
     Use `-f usv` for Unicode separators U+241F and U+241E.
     Use `-f lsv` for awk-like records.
+    Use `--delimiter=` (an empty string) to make '\0' the value separator.
+    Use `--row-delimiter=` to make '\0' the row separator.
 
     ## `csv` (Comma Separated Values) for maximum computibility
 
--- a/visidata/loaders/tsv.py
+++ b/visidata/loaders/tsv.py
@@ -75,6 +75,10 @@ class TsvSheet(SequenceSheet):
     def iterload(self):
         delim = self.delimiter or self.options.delimiter
         rowdelim = self.row_delimiter or self.options.row_delimiter
+        if delim == '': delim = '\x00'  #2272
+        if rowdelim == '': rowdelim = '\x00'
+        if delim == rowdelim:
+            vd.fail('field delimiter and row delimiter cannot be the same')
 
         with self.open_text_source() as fp:
                 for line in splitter(adaptive_bufferer(fp), rowdelim):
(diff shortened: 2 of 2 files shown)


saulpw (approved):

anjakefala:
Hey @midichef!

I'm trying to confirm this worked with `find visidata -print0 | vd --row-delimiter=''`, but it does not to be seperating the output into rows:

Screenshot from 2024-02-01 21-42-15

Is there something I am testing incorrectly?

midichef:
Right now the option will only apply when the filetype is `tsv`. So the command would need to be `find visidata -print0 | vd -f tsv --row-delimiter=''`

I added some warnings when the field/row delimiter is NUL, because it may surprise people. One situation I'm thinking of is: someone has run `vd file.tsv` and is exploring the options sheet after hitting `O`. They hit `e` to edit the blank cell for the global option for `delimiter`. In that cell appears the `·` character (`disp_unprintable`, for `'\t'`). They erase it and hit `Enter`. The blank cell looks the same as before they started editing it, so they don't think much of it.  But they have unwittingly overwritten the default value of `vd.options.delimiter` with the empty string `''`, so the actual delimiter will be `'\0'`. Then they save their TSV file, and it's full of `'\0'`. And life is less great. A warning can help them realize what happened.

midichef:
The issue is more complicated than I realized though. How should we handle comments when the delimiter is NUL?

i.e. `TsvSheet.options.regex_skip = '^#.*'` will currently skip over lines that look like comments. But it should definitely not do that when handling the output of `find -print0`.

My intuition is, `regex_skip` should not be used when the row delimiter is NUL, as we're not in classic TSV format any more.

anjakefala:
> My intuition is, regex_skip should not be used when the row delimiter is NUL, as we're not in classic TSV format any more.

I would say, let's add that to an issue as a follow-up task, but get this piece merged in for now. No need to block the PR.

PR #2277, opened 2024-01-30
go-col-name using palette

midichef:
This PR adds a command `go-col-name`, which lets the user search for a column by its name. It uses the great new input palette so it gives better visual feedback than `go-col-regex`. To make it easy to test, for now it's bound to `gc`.

One problem with it is the matching characters are only shown when they are all lowercase.
To see this problem, `vd ~vdsrc/sample_data/sample.tsv`
then `gc` and see the matching characters are shown when you type `nits` but not when you type `Units`.

As an experiment, I also made a version that shows the contents of the column next to its name, for the cursor's row.
To get that working, I had to change `fuzzymatch()`.  Right now `fuzzymatch()` searches all keys in `haystack`. I changed it to ignore any keys that start with `'_'`. The function description mentions "each non-_key", so maybe skipping such keys is how it's meant to work, and just hasn't been implemented yet?
https://github.com/saulpw/visidata/blob/c209e6ec105d72c030a768ebe41e0d425038af33/visidata/fuzzymatch.py#L370

### DIFF
--- a/visidata/movement.py
+++ b/visidata/movement.py
@@ -1,11 +1,10 @@
 import itertools
-import re
 
-from visidata import vd, VisiData, BaseSheet, Sheet, Column, Progress, ALT, asyncthread
+from visidata import vd, BaseSheet, Sheet, Column, Progress, ALT, asyncthread
 
 
 def rotateRange(n, idx, reverse=False):
-    'Wraps an iter starting from idx. Yields indices from idx to n and then 0 to idx.'
+    'Wraps an iter starting from idx. Yields indices from idx+1 to n and then 0 to idx, or from idx-1 to 0 and n-1 to idx.'
     if n == 0: return []
     if reverse:
         rng = range(idx-1, -1, -1)
@@ -81,18 +80,6 @@ def moveToNextRow(vs, func, reverse=False, msg='no different value up this colum
         vd.status(msg)
 
 
-@Sheet.api
-def nextColRegex(sheet, colregex):
-    'Go to first visible column after the cursor matching `colregex`.'
-    pivot = sheet.cursorVisibleColIndex
-    for i in itertools.chain(range(pivot+1, len(sheet.visibleCols)), range(0, pivot+1)):
-        c = sheet.visibleCols[i]
-        if re.search(colregex, c.name, sheet.regex_flags()):
-            return i
-
-    vd.fail('no column name matches /%s/' % colregex)
-
-
 @Column.property
 def visibleWidth(self):
     'Width of column as is displayed in terminal'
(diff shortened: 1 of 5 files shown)


saulpw:
@zormit Is this how this was supposed to work anyway?

> The function description mentions "each non-_key", so maybe skipping such keys is how it's meant to work, and just hasn't been implemented yet?

zormit:
> @zormit Is this how this was supposed to work anyway?
> 
> > The function description mentions "each non-_key", so maybe skipping such keys is how it's meant to work, and just hasn't been implemented yet?

My honest answer after looking into this just a little bit: I'm not sure, but it sounds plausible? :D I think I haven't looked into this for too long and also this whole wrapper function including its comment seems to be from your feather 😅

saulpw:
Okay, let's take the `_` fix at least, I think that makes sense.

It seems like this command is useful but could use some more work; let's move it to `visidata/features/go_col.py` (perhaps along with `go-col-regex`), and then we can be more lenient with the level of polish.  Also let's structure it like `go-col-regex`, such that it assigns to `cursorVisibleColIndex` in the execstr.

visidata/movement.py
```diff
@@ -102,6 +102,52 @@ def visibleWidth(self):
...
+            # pad the right side with spaces
+            # name has invisible ANSI formatting sequences changing its length,
+            # so use row.name
+            n_spaces = max(20 - dispwidth(row.name), 0)
+            r += f'{name + n_spaces*" "}'
+            r += '   '
+            #todo:  does not show disp_note_none for None
+            r += row._cursor_cell
+        else:
+            r += f'{name}'
```
saulpw: This can just be `r += name`.  Same for another one above.

saulpw (changes requested):
With a few changes we can take this.

midichef:
Okay I've made the requested changes. This PR is ready to go.

A couple more changes since the last review: I make cmdpalette catch the case of 0 items to choose from. And I took away the `gc` keystroke binding for `go-col-name`. I don't think the `g` prefix makes sense for it. It doesn't fit into the Visidata philosophy for keystrokes of `_` vs`z_` vs `g_`.

visidata/movement.py
```diff
@@ -114,7 +101,6 @@ def visibleWidth(self):
 Sheet.addCommand(None, 'go-bottom', 'sheet.cursorRowIndex = len(rows); sheet.topRowIndex = cursorRowIndex-nScreenRows', 'go all the way to the bottom of sheet')
 Sheet.addCommand(None, 'go-rightmost', 'sheet.leftVisibleColIndex = len(visibleCols)-1; pageLeft(); sheet.cursorVisibleColIndex = len(visibleCols)-1', 'go all the way to the right of sheet')
 
-Sheet.addCommand('c', 'go-col-regex', 'sheet.cursorVisibleColIndex=nextColRegex(inputRegex("column name regex: ", type="regex-col", defaultLast=True))', 'go to next column with name matching regex')
 Sheet.addCommand('zc', 'go-col-number', 'sheet.cursorVisibleColIndex = int(input("move to column number: "))', 'go to given column number (0-based)')
```
anjakefala: I noticed in your commit message that you mention this one was being moved, too. Did you intend to move it?

midichef: At first I meant to move it. Later I decided against it and forgot to change the message. I figured it belonged next to `go-row-number`, in `movement.py`. Now that I see your menu changes, the move seems better after all. Putting it in `go_col.py` keeps it with the group of commands for `Menu> Column > Goto`.

Could you squash these last few commits for the move? Thank you.

anjakefala:
@saulpw Ready for review

saulpw (approved):

PR #2287, opened 2024-02-02
[memory] shorten cells too wide to fit in prompt

midichef:
If I press `Alt+m` (`memo-cells`) on a cell that has contents too wide to fit on the screen, the command fails due to an exception:
```
File "memo-cell", line 1, in <module>
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/_input.py", line 543, in input
ret = vd.editText(y, promptlen, w=w,
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/_input.py", line 358, in editText
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/_input.py", line 285, in editline
prew = clipdraw(scr, y, x, dispval[:dispi], attr, w, clear=clear, literal=True)
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/cliptext.py", line 211, in clipdraw
return clipdraw_chunks(scr, y, x, chunks, attr, w=w, clear=clear, **kwargs)
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/cliptext.py", line 234, in clipdraw_chunks
scr.addstr(y, x, disp_column_fill*actualw, cattr.attr)  # clear whole area before displaying
```
This PR shortens the cell contents, so the prompt will fit on screen and prevent this error.

In the worst case, when the default column width is about as wide as the screen, then the right side of the `memo-cells` prompt will be overwritten by `rightStatus()`. The prompt won't be very readable. But the user can widen their terminal to make the prompt readable, and continue the memo action from there.

In this PR, I used a default max width of `options.default_width`. We could instead use `cursorCol.width`. That would match the prompt's quoted cell width to the width in the sheet. The advantage there is if the cell is truncated with `'…'`, it's truncated the same in both places, so the cell's last few characters match in both places. The drawback is that the column may be one the user made very wide, which would make the prompt unreadable from being overwritten by the right status bar.

What makes more sense?

### DIFF
--- a/visidata/memory.py
+++ b/visidata/memory.py
@@ -9,6 +9,19 @@ def memo(vd, name, col, row):
     vd.memory[name] = col.getTypedValue(row)
     vd.status('memo %s=%s' % (name, col.getDisplayValue(row)))
 
+@Sheet.api
+def memo_cell(sheet):
+    if not sheet.cursorCol or not sheet.cursorRow: return
+    value_params = dict(prompt='assign value: ', value=sheet.cursorDisplay)
+    name_params = dict(prompt='to memo name: ')
+    # edits to memo_value are blocked, to preserve its type  #2287
+    inputs = vd.inputMultiple(memo_name=name_params,
+                              memo_value=value_params,
+                              readonly_keys=('memo_value',))
+    if not inputs['memo_name']:
+        vd.fail('memo name cannot be blank')
+    vd.memory[inputs['memo_name']] = sheet.cursorTypedValue
+    vd.status(f'memo {inputs["memo_name"]}={sheet.cursorDisplay}')
 
 class MemorySheet(Sheet):
     rowtype = 'memos' # rowdef: keys into vd.memory
@@ -32,4 +45,4 @@ def memosSheet(vd):
 
 
 Sheet.addCommand('Alt+Shift+M', 'open-memos', 'vd.push(vd.memosSheet)', 'open the Memory Sheet')
-Sheet.addCommand('Alt+m', 'memo-cell', r'vd.memory[input("assign \""+cursorCol.getDisplayValue(cursorRow)+"\" to: ")] = cursorCol.getTypedValue(cursorRow)', 'store value in current cell in Memory Sheet')
+Sheet.addCommand('Alt+m', 'memo-cell', 'memo_cell()', 'store value in current cell in Memory Sheet')
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -404,8 +404,12 @@ def inputsingle(vd, prompt, record=True):
     return v
 
 @VisiData.api
-def inputMultiple(vd, updater=lambda val: None, record=True, **kwargs):
-    'A simple form, where each input is an entry in `kwargs`, with the key being the key in the returned dict, and the value being a dictionary of kwargs to the singular input().'
+def inputMultiple(vd, updater=lambda val: None, record=True, readonly_keys = (), **kwargs):
+    '''A simple form, where each input is an entry in `kwargs`, with
+    the key being the key in the returned dict, and the value being a
+    dictionary of kwargs to the singular input(). *readonly_keys* is
+    a tuple of keys in `kwargs`; for each key, its corresponding value
+    will be displayed as a field that cannot be tabbed into or edited.'''
     sheet = vd.activeSheet
     scr = sheet._scr
 
@@ -429,8 +433,8 @@ def inputMultiple(vd, updater=lambda val: None, record=True, **kwargs):
     maxw = sheet.windowWidth//2
     attr = colors.color_edit_unfocused
 
-    keys = list(kwargs.keys())
-    cur_input_key = keys[0]
+    editable_keys = [k for k in kwargs.keys() if k not in readonly_keys]
+    cur_input_key = editable_keys[0]
 
     if scr:
         scr.erase()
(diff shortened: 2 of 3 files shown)


visidata/memory.py
```diff
@@ -32,4 +32,4 @@ def memosSheet(vd):
 
 
 Sheet.addCommand('Alt+Shift+M', 'open-memos', 'vd.push(vd.memosSheet)', 'open the Memory Sheet')
-Sheet.addCommand('Alt+m', 'memo-cell', r'vd.memory[input("assign \""+cursorCol.getDisplayValue(cursorRow)+"\" to: ")] = cursorCol.getTypedValue(cursorRow)', 'store value in current cell in Memory Sheet')
+Sheet.addCommand('Alt+m', 'memo-cell', r'w = min(vd.options.default_width, windowWidth-len("assign \"\" to: ")-1)-1; dv = cursorCol.getDisplayValue(cursorRow); dv = dv[:w-dispwidth(options.disp_truncator)] + options.disp_truncator if dispwidth(dv) > w else dv[:w]; vd.memory[input("assign \""+dv+"\" to: ")] = cursorCol.getTypedValue(cursorRow)', 'store value in current cell in Memory Sheet')
```
saulpw: This seems pretty complicated..what about using `inputMultiple` and putting the value into the secondary field?  Like `assign: x+2` and `to: `.

saulpw:
Does `inputMultiple` effectively workaround this problem?

visidata/memory.py
midichef: That is much cleaner:
```
test_input_cmd = 'vd.memory[inputMultiple(name=dict(prompt="to memo name: "), cell=dict(prompt="assign cell value: "+cursorCol.getDisplayValue(cursorRow),display=False))["name"]] = cursorCol.getTypedValue(cursorRow);'
Sheet.addCommand('V', 'test-input', test_input_cmd, 'tester')
```
But it triggers the same error if the user hits Tab to switch to the long prompt showing the cell value. Maybe if we add an argument `inputMultiple()` to leave off the bindings passed to `input()`?
https://github.com/saulpw/visidata/blob/7ad1de0b98fb6b77c786e0198cae49167c391c6e/visidata/_input.py#L468

saulpw: Oh, I mean use the value as the initial value of the "assign cell value" field (`value=`).  Then tabbing to it will allow you to edit the value, too, and `inputMultiple` should already be able to handle that (or we should fix it).

Also I believe you can use `cursorTypedValue` instead of `cursorCol.getTypedValue(cursorRow)`.

And when the execstr gets to be this size, let's move some of it into function(s), to make it more readable (I know I am guilty of this too more than anyone :)

midichef: I tried making the value editable, but it seemed like the edited value of `inputMultiple()` was a string.  (because I was editing a string from `getDisplayValue()`). That's why I displayed the value in the prompt, to avoid giving the user a chance to change the type.

Now I tried applying the column type to the, copying `editCell()`: https://github.com/saulpw/visidata/blob/5e611e39cbb6a35c25bd34c323956f9a7bd44c30/visidata/_input.py#L639

```
def memo_cell(sheet):
    value_params = dict(prompt='assign value: ',
                        value=sheet.cursorCol.getDisplayValue(sheet.cursorRow))
    name_params = dict(prompt='to memo name: ')
    inputs = vd.inputMultiple(memo_name=name_params, memo_value=value_params)
    try:
        v_edited = sheet.cursorCol.type(inputs['memo_value'])  # convert input to column type
    except ValueError:
        vd.warning('the new memo value\'s type does not match the column type')
        v_edited = inputs['memo_value']
    if not inputs['memo_name']:
        vd.fail('memo name cannot be blank')
    vd.memory[inputs['memo_name']] = v_edited
Sheet.addCommand('V', 'test-input', 'memo_cell(vd.sheet)', 'test variation of memo-cell')
```

But a cell with a list inside, like `[1,2,3]` in a JSON sheet, turns into a string `'[3] 1; 2; 3'`, even if the user never tabs into it. Is there a better way to proceed?

Also, now that I've moved the code into a function, is `cursorDisplay` still preferred over `getDisplayValue()`? Or is it only preferred in `addCommand` strings, for readability?

saulpw: `cursorDisplay` is preferred, because it can be cached.

saulpw: > But a cell with a list inside, like [1,2,3] in a JSON sheet, turns into a string '[3] 1; 2; 3', even if the user never tabs into it. Is there a better way to proceed?

Oh, now I understand.  That's not great.  One idea is, we could add a `readonly` or something kwarg to `input()`, and make it so that readonly prompts can't be edited, and thus won't be accepted.

Another possibility would be to add a special case for lists and dicts to parse them correctly after input in general.  I've run into cases myself where I want to edit the value of a cell with `e` and have to use `z=` instead, which is non-intuitive.  But this would be a bigger change, because this display format isn't compatible with any existing parser, and I don't think we should be committing to this display format which is a little janky anyway.  So maybe we just go with the easier first option for now?

midichef: Okay, I went with the first option, making the field readonly, inside `inputMultiple()`. It can't be tabbed into to be edited.

saulpw:
Thanks for your efforts on this, @midichef!  I pushed a separate commit with you as the co-author, that doesn't introduce more parameters to inputMultiple, but still should fix the issue.  (We use the typed value from the sheet, and not the returned input value, so we don't need to make it read-only).

PR #2288, opened 2024-02-02
[save] allow saving files with .cols extension

midichef:
if I save any sheet as `test.cols` I get an exception.
```
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/threads.py", line 220, in _toplevelTryFunc
TypeError: save_cols() takes 2 positional arguments but 3 were given
```
The problem is that `save_cols()` exists and gets mistaken for a function to save files ending in `.cols`.

### DIFF
--- a/visidata/save.py
+++ b/visidata/save.py
@@ -89,7 +89,7 @@ def getDefaultSaveName(sheet):
 
 
 @VisiData.api
-def save_cols(vd, cols):
+def saveCols(vd, cols):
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Ah yes, thank you for spotting/fixing this.

midichef:
Oops, updating to change the commit message from "[save]" to "[save-]"

PR #2294, opened 2024-02-04
[cmdlog-] clarify error for malformed replay file

midichef:
If I mistakenly pass a data file in for a replay file, like `vd -p data.tsv`, I get a cryptic error:
```
if r.sheet and longname not in ['set-option', 'unset-option']:
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/utils.py", line 163, in __getattr__
raise AttributeError from e
AttributeError
```
This PR changes it to a clearer `'failed to find command to replay'`.

### DIFF
--- a/visidata/cmdlog.py
+++ b/visidata/cmdlog.py
@@ -262,6 +262,9 @@ def replayOne(vd, r):
         'Replay the command in one given row.'
         vd.currentReplayRow = r
         longname = getattr(r, 'longname', None)
+        if longname is None and (hasattr(r, 'keystrokes') and not r.keystrokes):
+            vd.fail('failed to find command to replay')
+            return
 
         if r.sheet and longname not in ['set-option', 'unset-option']:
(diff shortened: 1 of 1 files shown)


anjakefala:
@saulpw will need to approve the fix itself, but I just wanted to say that I AM SO GLAD you decided to address this. This does happen to me!

midichef:
Yes, I confused myself yesterday for quite some time making that mistake. Good to know I'm not the only one!

visidata/cmdlog.py
```diff
@@ -262,6 +262,9 @@ def replayOne(vd, r):
         'Replay the command in one given row.'
         vd.currentReplayRow = r
         longname = getattr(r, 'longname', None)
+        if longname is None and (hasattr(r, 'keystrokes') and not r.keystrokes):
+            vd.fail('failed to find command to replay')
+            return
```
saulpw: We don't need `return` after `fail` (it raises).  I'll fix it.

saulpw (approved):
Thanks, @midichef !

Issue #2266, opened 2024-01-23
[undo] can't undo some threaded commands


midichef:
**Small description**
After a pasting into cells or rows using `syspaste-*`, undo does not restore the original value of the cell.

**Steps to reproduce with sample data and a .vd**
Starting with this tsv sheet:
```
cells
first
second
```
hit `zY` on the cell that contains`"first"`, then `j` and then `zP`
```
cells
first
first
```
then hit `U`, and no undo happens. Instead there is a message: `nothing to undo on current sheet`.

To trigger the bug, `syspaste-cells` must be done manually. If it's done inside `-p script.vdj`, then `undo-last` works as expected.

**Additional context**
saul.pw/VisiData v3.1dev

The behavior started with a9e4e050 and its parent commit.

I've looked into the undo code a bit. The problem is that when `syspaste-*` runs `addUndo()`, `vd.activeCommand` is `None`, so the function returns without adding to `r.undofuncs`. I'm not sure what the fix should be.

midichef:
The problem is a race condition in `basesheet.py`:
https://github.com/saulpw/visidata/blob/74c6eb221b9e5ddac6b27295f98928878c1d0e32/visidata/basesheet.py#L211
https://github.com/saulpw/visidata/blob/74c6eb221b9e5ddac6b27295f98928878c1d0e32/visidata/basesheet.py#L219

The events happen in this order:
1) `execCommand()` runs `execCommand2()`, which starts `pasteFromClipboard()` in its own thread.

2) `pasteFromClipboard()` starts. It will eventually look at `vd.activeCommand`, but not yet. It runs a command to check the clipboard, but that takes a while to finish, so it is paused.

3) `execCommand()` resumes. It runs `afterExecSheet()` which changes `vd.activeCommand` to `None`.

4) `pasteFromClipboard()` resumes. When it tries to make its undo functions, in `addUndo()`, it sees the wrong new value of `vd.activeCommand`, so it does not add its undo functions to the command log sheet.

5) Because there are no undo functions, when I do `undo-last`, it fails.

As a confirmation that this is what's happening, undo is fixed when I remove the `@asyncthread` decorator from `pasteFromClipboard()`.

A fix would be to make sure `afterExecSheet()` doesn't run until the threaded command finishes. But at the moment I'm not sure how to do that.

midichef:
> A fix would be to make sure `afterExecSheet()` doesn't run until the threaded command finishes. But at the moment I'm not sure how to do that.

[...]

PR #2295, opened 2024-02-04
[selection] add commands to sort selected rows

midichef:
This is a proof-of-concept command that sorts only the rows in the selection.  @daviewales [requested such a feature](https://github.com/saulpw/visidata/discussions/1851) and I wanted it too.

I recommend using the https://github.com/saulpw/visidata/commit/c79ce4f6cea505505014f863553d128f7be35c09 commit for testing. It uses a temporary workaround for #2266 to allow undo. Otherwise you can't undo the sort. And it uses assertions to detect if rows are lost.

If you think this is worth including, I would appreciate a close inspection of this, @saulpw and @anjakefala , as my first few versions of this caused data corruption.

The practical use of it is a bit limited. Right now you can use `sort-selected-asc` and `-desc` to sort by the current column. But if you want to sort by multiple columns you'll have to do it by hand, via `exec-python` and then
`sortSelected(ordering)`. `ordering` is a list of (column, boolean) tuples.  The boolean is a direction; `False` means ascending, `True` means descending. For example, `sortSelected([(row1, False), ('title2', True), ('title3'), False)])`.

I can imagine a way to reorder rows more flexibly. The user could pull the selection into a new sheet with `"`, reorder the sheets using keys like `[` and `z[`, or even by sliding rows up and down. And then use a new command to replace the source row selection with the reordered rows. What do you think of that?

### DIFF
--- a/visidata/experimental/sort_selected.py
+++ b/visidata/experimental/sort_selected.py
@@ -1,0 +1,54 @@
+from copy import copy
+from visidata import vd, Sheet, Progress, asyncthread, options, UNLOADED
+
+@Sheet.api
+@asyncthread
+def sortSelected(self, ordering):
+    """sort the rows that are selected, in place so that each originally-selected row takes the place of another.
+    *ordering* is a list of tuples: (col_name_str, reverse_sort_bool) (string and bool)"""
+    if self.rows is UNLOADED or not self.rows:
+        return
+    if self.nSelectedRows == 0:
+        vd.fail('no rows selected')
+        return
+
+    if options.undo:
+        vd.status('undo added')
+        vd.addUndo(setattr, self, 'rows', copy(self.rows))
+
+# temporary:  save data to test integrity after the sort
+    rows_initial = { self.rowid(r) for r in self.rows }
+    selected_initial = set(self._selectedRows.keys())
+
+    with Progress(gerund='sorting', total=self.nSelectedRows) as prog:
+        selected = list(self._selectedRows.values())
+        selected_rowids = set(self._selectedRows.keys())
+
+        selected_indices = []
+        for i, r in enumerate(self.rows):
+            rowid = self.rowid(r)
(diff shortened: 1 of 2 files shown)


saulpw (commented):
I think these commands should go in their own file, either `features/sort_selected.py` or `experimental/sort_selected.py`.  I'm leaning towards the latter, because while the code looks right to me, the idea in general makes me nervous--what if the user aborts this thread with `Ctrl+C` after the sort but before the replacement?  It feels like these should be opt-in, use-at-your-own-risk commands, for the time being at least.

I made the other changes to the sort API to support this, and because they feel better in general.

midichef:
Good points about cancelling being dangerous! I hadn't considered that.

We could bypass this if we could edit a copy of the rows and then assign it to `self.rows`. But `Sheet.sort()` says we must not assign to self.rows. What are the specific reasons for that?
https://github.com/saulpw/visidata/blob/aab3cda093274fbf23c9ff0310c81ddd86de5b6e/visidata/sort.py#L72

saulpw:
> We could bypass this if we could edit a copy of the rows and then assign it to `self.rows`. But `Sheet.sort()` says we must not assign to self.rows. What are the specific reasons for that?

With `git log -S reassign` I got to 3930ad044bf29115fe9c9ceff18c7275353de547:

> by not sorting in-place, sorting on columns sheet would not sort columns on source sheet

It's an edge case for sure, but one that I ran across and felt important to make work for internal design consistency.

midichef:
Oh, I see. Assigning to `self.rows` on a `ColumnsSheet` would break the link to the source sheet's columns. It would break sorting in that command. And afterward commands that edit rows, like `slide-up`, would have no effect on the source sheet. Even `cursorRowIndex` would give misleading answers that don't correspond to the source sheet columns.

I'll search the git logs more in the future, that note is still helping, 5 years later.

PR #2296, opened 2024-02-04
[docs] migrate guides to visidata/guides/*.md

anjakefala:
- [x] update the API guide to writing guides


visidata/guide.py
```diff
@@ -100,7 +102,9 @@ def iterload(self):
         for line in guides_list.splitlines():
             m = re.search(r'(\w+?) \("(.*)"\)', line)
             if m:
-                yield [i] + list(m.groups())
+                guide = list(m.groups())  # guidename, description
```
saulpw: Let's make this:

```
    guidename, description = m.groups()
    vd.addGuide(guidename)
    yield [i, guidename, description]
```

visidata/guide.py
```diff
@@ -156,15 +160,27 @@ class GuideSheet(Sheet):
...
+
+        # parsing front matter
+        sections = guidetext.split('---\n', maxsplit=2)
+        d = {}
+        for config in sections[:-1]:
+            config = config.strip()
+            if config:
+                key, val = config.split(': ', maxsplit=1)
+                d[key] = val
+        self.__dict__.update(d)
```
saulpw: What might this be used for, besides `sheettype`?

visidata/guide.py
```diff
@@ -156,15 +160,27 @@ class GuideSheet(Sheet):
...
+        helper = AttrDict(commands=CommandHelpGetter(globals()[self.sheettype]),
                           options=OptionHelpGetter())
-        guidetext = MissingAttrFormatter().format(self.guide_text, help=helper, vd=vd)
-        for startingLine, text in enumerate(guidetext.splitlines()):
+        guidetext = MissingAttrFormatter().format(text, help=helper, vd=vd)
+
+        # parsing front matter
+        sections = guidetext.split('---\n', maxsplit=2)
+        d = {}
+        for config in sections[:-1]:
```
saulpw: This isn't right, there needs to be a `splitlines()` in here somewhere.

saulpw: `for section in sections[:-1]: for config in section.splitlines():`

saulpw (changes requested):
This is great, thank you!  A few things to change, and please update the docs about how to make a guide.

visidata/guide.py
anjakefala: I had a vision for `VisiData version`, `updated on`. I also thought it might be good to add this mechanism in case we do add to these parameters. Maybe we designate categories of guides, where only some categories get concatenated into a manpage later.

saulpw: Well I'm wondering if we shouldn't store it as an AttrDict `metadata` or something, and then it would be clearer at the place where it's used, what's going on and where it comes from.

anjakefala:
Doc updates are ready for review!

visidata/guide.py
anjakefala: Great idea! I implemented that.

saulpw (approved):
Thanks for doing this, @anjakefala !

PR #2297, opened 2024-02-05
[open-] allow opening files with .dir extension

midichef:
A file that ends in `'.dir'` will be opened with the `dir` filetype, showing a blank DirSheet. To demo: `vd a.dir`. This PR loads it instead as `txt`.


saulpw (approved):
Fair enough!

PR #2299, opened 2024-02-05
[xlsx-] prevent errors when column_letter is absent

midichef:
When I insert a column into an `xlsx` sheet with `=`, I get errors.
```
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/loaders/xlsx.py", line 206, in colorize_xlsx_cell
  fg = getattrdeep(row, col.column_letter+'.font.color', None)
AttributeError: 'SettableColumn' object has no attribute 'column_letter'
```
This PR checks that the attribute `column_letter` is present.

### DIFF
--- a/visidata/loaders/xlsx.py
+++ b/visidata/loaders/xlsx.py
@@ -203,6 +203,8 @@ HLSMAX = 240
 
 @XlsxSheet.api
 def colorize_xlsx_cell(sheet, col, row):
+    if not hasattr(col, 'column_letter'):
+        return ''
     fg = getattrdeep(row, col.column_letter+'.font.color', None)
     bg = getattrdeep(row, col.column_letter+'.fill.start_color', None)
     fg = sheet.xlsx_color_to_xterm256(fg)


visidata/loaders/xlsx.py
```diff
@@ -203,8 +203,11 @@ def save_xls(vd, p, *sheets):
 
 @XlsxSheet.api
 def colorize_xlsx_cell(sheet, col, row):
-    fg = getattrdeep(row, col.column_letter+'.font.color', None)
-    bg = getattrdeep(row, col.column_letter+'.fill.start_color', None)
+    if hasattr(col, 'column_letter'):
+        fg = getattrdeep(row, col.column_letter+'.font.color', None)
+        bg = getattrdeep(row, col.column_letter+'.fill.start_color', None)
+    else:
+        fg = bg = None
```
saulpw: Let's early return instead, to not call `xlsx_color_to_xterm256` unnecessarily, and to reduce the size of the diff.

midichef: Oh right. I didn't even look at the rest of the function. I wasn't sure whether to go with `None` or `''`. It seems both are fine for now.

saulpw (approved):

PR #2302, opened 2024-02-08
[tsv] turn off regex skipping sometimes

### LINKED ISSUE #2272: [tsv] add CLI option to use NUL as delimiter

midichef:
The first patch (a1607d1f2814963977a06669ce3b1931ceba7e69) is a bugfix. Right now users can't turn off regex skipping for tsvs, because they'll trigger an error:
`vd --regex-skip= ~vdsrc/sample_data/sample.tsv`:
```
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/loaders/tsv.py", line 89, in iterload
if not line or fp._regex_skip.match(line):
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/path.py", line 136, in __getattr__
return getattr(self.fp, k)
AttributeError: '_io.TextIOWrapper' object has no attribute '_regex_skip'
```

The second patch (fb10773a62661e5d732a52f09cbb964658b5ec02) turns off regex skipping when NUL is the row or field delimiter. The rationale is that a file with NUL separators is not designed for humans to read, so there will not be comments in it.

Note that NUL delimiters will quietly disable regex skipping, even when the user tries to set a different regex on the command line. For example, they will not be able to skip blank lines with: `vd -f tsv --row-delimiter= --regex_skip='^\s+$' `. I think that's the right choice. It's better to give them unwanted lines that they can choose to strip manually, than to silently strip lines that they didn't expect to lose. As an example where skipping is unexpected, despite `regex_skip` being set to something other than the default: I set the option `regex_skip` in .visidatarc for some CSV data files, then time goes by and I forget I did that, then I use `vd -f tsv --delimiter=` to look at the output of `grep -Z`. I won't be expecting lines to be skipped.

### DIFF
--- a/tests/golden/pr2302.tsv
+++ b/tests/golden/pr2302.tsv
@@ -1,0 +1,45 @@
+# comment						
+OrderDate	Region	Rep	Item	Units	Unit_Cost	Total
+2016-01-06	East	Jones	Pencil	95	1.99	189.05
+2016-01-23	Central	Kivell	Binder	50	19.99	999.50
+2016-02-09	Central	Jardine	Pencil	36	4.99	179.64
+2016-02-26	Central	Gill	Pen	27	19.99	539.73
+2016-03-15	West	Sorvino	Pencil	56	2.99	167.44
+2016-04-01	East	Jones	Binder	60	4.99	299.40
+2016-04-18	Central	Andrews	Pencil	75	1.99	149.25
+2016-05-05	Central	Jardine	Pencil	90	4.99	449.10
+2016-05-22	West	Thompson	Pencil	32	1.99	63.68
+2016-06-08	East	Jones	Binder	60	8.99	539.40
+2016-06-25	Central	Morgan	Pencil	90	4.99	449.10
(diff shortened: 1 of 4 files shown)


saulpw (approved):
I think this makes sense to me.

PR #2303, opened 2024-02-08
[main-] to use stdin as data source, require a pipe

midichef:
This PR is for the case where stdin is a data source on the command line, and input is a terminal, not a pipe. For example, `vd -f tsv -`. The interface gets confused. Visidata shows "processing..." and ignores some keypresses, making it difficult to quit.

With this PR, visidata will exit with an error message.


saulpw (approved):
Thanks, @midichef!

Issue #2293, opened 2024-02-04
[cmdlog] sheets created with no keypresses show errors 


midichef:
**Small description**
clicking on `dir_hidden` in the DirSheet guide raises an exception

**Actual result with screenshot**
```
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/statusbar.py", line 56, in sheetlist
if len(vs.shortcut) == 1:
TypeError: object of type 'NoneType' has no len()
```

**Steps to reproduce with sample data and a .vd**
`vd .`, navigate to `filename` column to bring the DirSheet guide up, click on `dir_hidden`

**Additional context**
visidata 3.1dev

It looks like vs.shortcut is `None` because some code for `shortcut()` is obsolete, where it checks `cmdlog.rows[0].keystrokes`:
https://github.com/saulpw/visidata/blob/aa9d2615f3b2773001cf75a1b24219903a91c1bb/visidata/cmdlog.py#L415

midichef:
Okay I've got a clearer idea of exactly when this happens. It's when you create a sheet without using keystrokes. Like a command started by `exec-command`, or a help menubar entry, or clicking a statusbar link. The sheet has to have `precious == False`, like a HelpSheet or OptionsSheet. So. running `exec-longname` and then choosing `help-commands`.
https://github.com/saulpw/visidata/blob/c702c3512ba81584ec8f2dad30e7dc75eba1e943/visidata/cmdlog.py#L412
Because the sheet isn't precious, it can't be found in `vd.allSheets`, so the command log is searched for keystrokes. But there are no keystrokes because the action that triggered the command wasn't a keypress.

midichef:
As a secondary issue, I wonder if
https://github.com/saulpw/visidata/blob/c702c3512ba81584ec8f2dad30e7dc75eba1e943/visidata/cmdlog.py#L417
has a remaining problem, when it uses `rows[0]`. It assumes the first row of the command log is the one that created the sheet. But they're usually `set-options` commands. (This same assumption caused a problem with undo in https://github.com/saulpw/visidata/pull/2244.) It seems like that may cause an error.

I haven't yet found a case it does. But before I dig deeper, what is this shortcut code for? What's an example where a keystroke from the command log can be thought of as a shortcut?

saulpw:
> What is this shortcut code for? What's an example where a keystroke from the command log can be thought of as a shortcut?

[...]

PR #2304, opened 2024-02-09
[cmdlog-] log a keystroke if cmd had keyless cause

midichef:
Addresses #2293. It shuts off the status errors in all the situations I'm aware of.

It doesn't close the issue because there might be a few more ways for the trouble to show up. But I'm pushing this out now to let people run v3.1dev without the errors.

### DIFF
--- a/visidata/basesheet.py
+++ b/visidata/basesheet.py
@@ -204,6 +204,9 @@ class BaseSheet(DrawablePane):
             vdglobals = vd.getGlobals()
 
         vd.cmdlog  # make sure cmdlog has been created for first command
+        if keystrokes is None and longname:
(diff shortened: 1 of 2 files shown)


saulpw:
Can we fix this with a simpler solution, by returning an empty string for shortcut instead of None?  I'm hesitant to add another mapping like this, since it's not generally accurate, and I think it would be confusing to run a command from the guide and see a keystroke on the status bar that you didn't press.

Alternatively we could make a function that gets the keystroke for a command, like the HelpSheet `keystrokes` column does, and use that: `sheet.revbinds.get(longname, [None])[0])`

midichef:
Okay, I'll close this and submit the simpler solution.

PR #2308, opened 2024-02-12
[aggr-] fix handling of various special cases

midichef:
e4d8b793c63c767d20b385a97058c865d44b3b20
Right now, the `list` aggregator produces a list, but keeps the type of the column it was applied to. This leads to an `!` error marking for the type mismatch `Exception: invalid type for date list`. This commit changes the `list` aggregator to `anytype` to get rid of the error.
repro:  `vd sample_data/sample.tsv`  `@` `+` `list` `zF`
You will see error markings on the right side of `OrderDate list`.

10bd6185f85905126d9fcc074767ad2c4ac66f53
When choosing an aggregator, if the name is invalid or a typo, it's just silently dropped. This commit adds a warning.
repro:  On any sheet:  `+` `min asdf max`
Only `min` and `max` will show up as aggregators.

9d3cd652ba15eb117e018743e7eb434127f6a888
`memo-aggregator` fails with `KeyError` when any of the aggregators are invalid or a typo. This commit changes the error to a more informative warning, and then accepts all remaining valid aggregators.
repro:  on any sheet, `z+` `asdf` will give `KeyError`

4887806149d993af5ed1b8806bfed5c2d0d43eeb
The quantile aggregators, like `q4`, produce a list of `Aggregator` objects. Some code paths can be given that list, even though they only accept a single `Aggregator`. One place this happens is when editing the `aggregators` column in `ColumnsSheet`. This commit fixes all such code paths to accept lists.
repro: `vd sample_data/sample.tsv`, `C`; go to the `Units` column, edit its cell under `aggregators`, input `q4`, then quit to the original sheet with `q`. On the original sheet hit `zF`. You'll see `AttributeError: 'list' object has no attribute 'name'`.

30fe7ff2a94990c5804dadc0271ca18c9c975702
When no rows are selected, and `zF` is run, the `keymax` aggregator gives an error, because `max()` is running on an empty list of rows. There should be no error, because it is common for no rows to be selected.

And when there are no key columns, `keymax` gives the empty tuple as the key: `()`.
This PR changes both results in the `zF` keymax column to `None`, and gets rid of the error.
`vd sample_data/sample.tsv` `+` `keymax`, `zF`, see that under the `OrderDate_keymax` column, the `Total` cell is `[0]` for the empty tuple, and the `Selected` cell is blank, with an error.

3795c493fb6b71b77bf86aa5771afdf08389e53a
This commit adds a `keymin` aggregator that is the equivalent of `keymax`, mentioned in #1331 by @frosencrantz.

44befa0e7cbc5713ff20feafe089ce6b89e6f43c
This commit rewrites `keymax` and `keymin` because they need more elaborate code to handle rows that are dictionaries. On JSON and other sheets whose rows are dicts, `keymax` runs into trouble when it has to compare rows, because they are dictionaries and can't be compared. (In my test case, it happens only because the two cells in the `A` column have the same value, so they compare as equal. Then the `max` function has to compare the full rows to resolve the tie.)
repro: 
Using [repro.json](https://github.com/saulpw/visidata/files/14236729/repro.json), do `vd repro.json`, hit `!` on `id` column, then on `A` column: hit `+` and enter `keymax`, then `zF`
You'll see this error in the `A_keymax` cell:
```
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/aggregators.py", line 166, in <lambda>
vd.aggregators['keymax'] = _defaggr('keymax', anytype, lambda col, rows: col.sheet.rowkey(max(col.getValueRows(rows))[1]), helpstr='key of the maximum value')
TypeError: '>' not supported between instances of 'dict' and 'dict'
```
This commit needs some help on design. I handle dicts as a special case here. Maybe a more general approach is called for. Maybe each Sheet should define a row sorting function that defaults to None?

Let me know if these commits make sense. Before they go in, they should be squashed into to one or two commits.


saulpw:
These are really good, @midichef, thank you.  I only have a couple of minor points:

- chooseAggregators docstring looks incorrect, since it does check for validity (to warn if invalid).
- Nice work using the `key` kwarg on min/max.  Can we use the `default` kwarg too, instead of checking for an empty rowset ourselves?
- Let's structure `keyfunc` as a function that returns the aggregator function, so we can pass `keyfunc(max)` to `_defaggr` without the lambda.

midichef:
>  Can we use the `default` kwarg too, instead of checking for an empty rowset ourselves?

Okay, here's a patch that does that. The benefit is it makes the `aggr_func` more consistent with `max()`. But it doesn't get rid of the `None` check, it just moves it from rowset to the return line. I think we still need to check if `row` is `None`, as I don't know how safe it is to pass that to `rowid()`.
https://github.com/midichef/visidata/blob/4d4afd1e9974793c2074257e5ebbc0ef16ed3d01/visidata/aggregators.py#L184

midichef:
Here's a little fix to improve error handling.
repro: generate an error, since `mean` does not handle date types: `vd ~vdsrc/sample_data/sample.tsv` `@` `+` `mean` `zF` 
there are two issues here
1) The `mean` column says `invalid type for date NotImplementedType`.
2) If you do `error-cell` on that value, you see the word 'unknown' in one letter per line.

The error message in the cell comes because `_funcRows` returns an Exception instead of raising it. Then `wrapply()` and `getCell()` don't know what to do with that Exception later.

The first commit fixes the display of 'unknown' that's currently one character per line.
The second changes `_funcRows` to raise the Exception. That changes the cell shown by `zF` to be empty, and the error changes from `unknown` to the usual multi-line stack trace. If we want to keep the current behavior where the `zF` cell shows a line describing the exception, then there's still the problem that `error-cell` only shows the stack trace as the single line `unknown`. In that case an alternate fix would be to change `wrapply()` or `getCell()` to save the stack trace.

saulpw (approved):
Thanks so much for making these QoL improvements, @midichef !

midichef:
Alright. I've just force-pushed, to squash most of the little changes into one. I didn't change anything else.

PR #2315, opened 2024-02-15
[clipboard-] prevent empty paste and add specific warnings

midichef:
When visidata has just started, and clipboards are empty, `None` is pasted in, when a user pastes into a cell with `zp`, or into a column of selected rows with `gzp`. This PR stops the paste, and issues a warning. That brings the commands in line with the existing behavior of `p`, discussed in #1793 (where it was decided to not fail, but issue a warning).

The second commit here makes the "nothing to paste" errors more specific:  "nothing to paste from [clipcells/cliprows/clipvals/system clipboard]". This should orient the user a bit when they press the wrong paste key. `p`/`zp`/`gzp`/`zP` each refer to different clipboards, and they're extremely difficult to keep straight in my mind.

To make it a bit easier to test, here's a table of the affected commands:
```
paste-cell              zp
setcol-clipboard        gzp  
paste-after             p/P
pasteFromClipboard()    zP/gzP/BUTTON2
```
After this PR, each command shows a warning when pasting from its own empty clipboard, and each warning is distinct from the others.

### DIFF
--- a/visidata/clipboard.py
+++ b/visidata/clipboard.py
@@ -6,6 +6,7 @@ import sys
 import tempfile
 import functools
 import os
+import itertools
 
 from visidata import VisiData, vd, asyncthread, SettableColumn
 from visidata import Sheet, Path, Column
@@ -58,6 +59,13 @@ def syscopyValue(sheet, val):
 
     vd.status('copied value to system clipboard')
 
+@Sheet.api
+def setColClipboard(sheet):
+    if not vd.memory.clipcells:
+        vd.warning("nothing to paste")
+        return
+    for r, v in zip(sheet.onlySelectedRows, itertools.cycle(vd.memory.clipcells or [None])):
+        sheet.cursorCol.setValuesTyped([r], v)
 
 @Sheet.api
 def syscopyCells(sheet, cols, rows, filetype=None):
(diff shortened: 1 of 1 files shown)


visidata/clipboard.py
```diff
@@ -58,6 +59,13 @@ def syscopyValue(sheet, val):
 
     vd.status('copied value to system clipboard')
 
+@Sheet.api
+def setColClipboard(sheet):
+    if not vd.memory.clipcells:
+        vd.warning("nothing to paste from clipcells")
+        return
```
saulpw, line 66: Whenever I see a `warning` followed by `return`, I wonder if it could be a `fail()` instead.  The difference being that any command which uses a function with `warning`/`return` will succeed, whereas with `fail` it will raise an exception and won't be added to the cmdlog, and will abort during a replay.  By default I think we should prefer failing, but if this case is either inconsequential, or potentially even useful in rare cases, then we would keep it as a `warning`/`return`.

visidata/clipboard.py
```diff
@@ -183,7 +191,7 @@ def paste_after(sheet, rowidx):
 Sheet.addCommand('gzP', 'syspaste-cells-selected', 'pasteFromClipboard(visibleCols[cursorVisibleColIndex:], someSelectedRows)', 'paste from system clipboard to selected cells')
 
 Sheet.addCommand('gzy', 'copy-cells', 'copyCells(cursorCol, onlySelectedRows)', 'yank (copy) contents of current column for selected rows to clipboard')
-Sheet.addCommand('gzp', 'setcol-clipboard', 'for r, v in zip(onlySelectedRows, itertools.cycle(vd.memory.clipcells or [None])): cursorCol.setValuesTyped([r], v)', 'set cells of current column for selected rows to last clipboard value')
+Sheet.addCommand('gzp', 'setcol-clipboard', 'setColClipboard()', 'set cells of current column for selected rows to last clipboard value')
```
saulpw, line 194: Thanks for factoring this out into a function.

saulpw (approved):
Good changes, thank you!

saulpw:
> p/zp/gzp/zP each refer to different clipboards, and they're extremely difficult to keep straight in my mind.

They each have a different shape and it felt reasonable at the time.  I guess I only really use y/p and zy/zp and zY and gzY myself.  
I would love to consolidate these if possible.

PR #2318, opened 2024-02-17
[json-] load objects as AttrDict consistently

midichef:
When the JSON loader creates rows in the typical case, by `json.load()`, they are `dict`. But any new rows made by `addRow()` are `AttrDict`.

To demonstrate:
`echo '[\n{"n":2}]' |vd -f json` then add a row with`a`.
`^Y` on the first row shows it is a `dict`, but on the second row it shows `AttrDict`.

Similarly, if the file can be loaded as rows of standalone JSON objects, each row is `AttrDict`:
`echo '{"n":2}' |vd -f json`
`^Y` shows the row is an `AttrDict`.

This PR makes the rows all `AttrDict`, by making the `json.load()` match the existing call to `json.loads()`.

### DIFF
--- a/visidata/loaders/json.py
+++ b/visidata/loaders/json.py
@@ -58,7 +58,7 @@ class JsonSheet(Sheet):
                         yield TypedExceptionWrapper(json.loads, L, exception=e)  # an error on one line
                     else:
                         with self.open_text_source() as fp:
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Yes, good catch!

PR #2319, opened 2024-02-17
[csv-] display loading/saving progress

### LINKED ISSUE #2213: [csv] load % not working for csv on disk
When loading a larger .csv file from disk, the loading status bar says `0%` the entire time.

midichef:
Closes #2213, for loading % progress. Progress was also not being shown when saving large files.

As a separate bug, FileProgress counts (for all text filetypes) are inaccurate when the data is not 1 byte per character. For example, for UTF-16 data, loading progress is shown as half of what it actually is. I'll fix that separately.

### DIFF
--- a/visidata/loaders/csv.py
+++ b/visidata/loaders/csv.py
@@ -71,9 +71,10 @@ def save_csv(vd, p, sheet):
         if ''.join(colnames):
             cw.writerow(colnames)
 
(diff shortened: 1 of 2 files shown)


midichef:
I force-pushed a version with the same diffs but a better commit message.

saulpw (approved):
Looks great, thank you!

Issue #2316, opened 2024-02-15
Log file inside zip file erroneously guessed as "tar" instead of text


krackout:
Log file inside zip file erroneously guessed as "tar" instead of text. Visidata version 3.0.2.

On 2.11 it's correctly opened as text.

anjakefala:
Hi @krackout!

Thanks for filing this issue. Please provide some sample data, so that we can reproduce this. We might not be able to fix it otherwise.

krackout:
Inside somelogs.zip file, there's a freshclam.log which produces the error.
[somelogs.zip](https://github.com/saulpw/visidata/files/14301136/somelogs.zip)

I also attach a screenshot of the error.
![visidata-logfile-error](https://github.com/saulpw/visidata/assets/36501063/d85147d8-8ff2-481f-8ca3-264740beac17)

ajkerrigan:
Seeing the same thing here, repro:

```console
echo "hiya" > test.log
zip test.zip test.log
vd -N test.zip
```

Then hitting `Enter` gives me the same thing @krackout showed. Full traceback:

```console
Traceback (most recent call last):
File "/home/aj/.local/pipx/venvs/visidata/lib/python3.12/site-packages/visidata/threads.py", line 220, in _toplevelTryFunc
t.status = func(*args, **kwargs)
^^^^^^^^^^^^^^^^^^^^^
File "/home/aj/.local/pipx/venvs/visidata/lib/python3.12/site-packages/visidata/sheets.py", line 260, in reload
self.loader()
File "/home/aj/.local/pipx/venvs/visidata/lib/python3.12/site-packages/visidata/sheets.py", line 285, in loader
for r in self.iterload():
File "/home/aj/.local/pipx/venvs/visidata/lib/python3.12/site-packages/visidata/loaders/archive.py", line 140, in iterload
with tarfile.open(name=str(self.source)) as tf:
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.12/tarfile.py", line 1808, in open
raise ReadError(f"file could not be opened successfully:\n{error_msgs_summary}")
tarfile.ReadError: file could not be opened successfully:
- method gz: ReadError('not a gzip file')
- method bz2: ReadError('not a bzip2 file')
- method xz: ReadError('not an lzma file')
- method tar: ReadError('truncated header')
```

[...]

PR #2320, opened 2024-02-18
[tar-] use open_bytes() for guessing filetype

ajkerrigan:
Change the way we peek at a file to see if it's of type `tar`, to resolve some error conditions raised in #2316. Specifically, with this change:

- Opening a file with an unknown extension inside an archive will no longer get detected as `tar` and trigger read errors
- Opening a tar file with a non-tar extension will correctly get detected as filetype `tar`

There's more detail about edge cases [here](https://github.com/saulpw/visidata/issues/2316#issuecomment-1949337402) and in follow-up comments, and potentially more changes coming. But this feels like a good incremental step.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -71,3 +71,10 @@ jobs:
 
     - name: Ensure VisiData can create completions
       run: python -v dev/zsh-completion.py
+
+    - name: Ensure VisiData does not drop lines from RepeatFiles
+      run: |
+        NUMLINES=$(seq 10000 | vd -b | wc -l)
+        if ! [[ $NUMLINES -eq 10000 ]]; then
+            exit 1
+        fi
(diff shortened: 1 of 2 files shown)


saulpw (approved):
Thanks for the analysis, @ajkerrigan, and the fix!

anjakefala:
I added a test to the CI to make sure we do not regress any RepeatFile behaviour in the future! With that we can make these changes in the future, without needing to manually check =) Thanks AJ!

ajkerrigan:
> I added a test to the CI to make sure we do not regress any RepeatFile behaviour in the future! With that we can make these changes in the future, without needing to manually check =) Thanks AJ!

Big thanks for that! I wasn't sure how many edge cases to add, but I dig the way you handled it 👍

PR #2323, opened 2024-02-19
[path-] fix undercounted progress for multibyte chars

midichef:
For text files encoded with more than one byte per character, `FileProgress` undercounts loading progress.

To demonstrate, you can use a UTF-32 file, where every character takes 4 bytes:
```
seq 1000001 | iconv -t UTF-32 >! progress.utf32.tsv
vd --encoding=utf-32 progress.utf32.tsv
```
The progress only goes up to 25%, not 100%.

That's because `read()` progress is counting the characters, but the goal is measured in bytes. The file is around 7 million characters long, but when encoded in UTF-32, it is 28 million bytes, so even at the end, 7 million/28 million becomes 25%.

This PR changes `FileProgress` to track progress as bytes.

### DIFF
--- a/visidata/path.py
+++ b/visidata/path.py
@@ -325,11 +325,9 @@ class Path(os.PathLike):
         return zopen(FileProgress(path, fp=open(path, mode='rb'), **kwargs), **kwargs)
 
     def __iter__(self):
-        with Progress(total=filesize(self)) as prog:
-            with self.open(encoding=vd.options.encoding) as fd:
-                for i, line in enumerate(fd):
-                    prog.addProgress(len(line))
-                    yield line.rstrip('\n')
+        with self.open(encoding=vd.options.encoding) as fd:
+            for line in fd:
+                yield line.rstrip('\n')
 
     def open_bytes(self, mode='rb'):
         'Open the file pointed by this path and return a file object in binary mode.'
--- a/visidata/path.py
+++ b/visidata/path.py
@@ -118,13 +118,9 @@ class FileProgress:
         self.fp.read = self.read
         self.fp.close = self.close
 
-        self.est_charbytes = 0
-        self.sampled = ''
-        self.est_ctr = 0
-        self.est_interval = 1000
+        self.est_charbytes = 1
+        self.est_sample = ''
         self.est_total = 0
-
-        self.readline_chunks = []
 
         if self.prog:
             self.prog.__enter__()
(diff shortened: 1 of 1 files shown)


visidata/path.py
```diff
@@ -130,13 +130,13 @@ def read(self, size=-1):
         r = self.fp_orig_read(size)
         if self.prog:
             if r:
-                self.prog.addProgress(len(r))
+                self.prog.addProgress(len(r.encode(self.encoding)) if self.encoding else len(r))
```
saulpw: Unfortunately this re-encodes every decoded line, which I think will have a significant impact on performance.  Is there another way to do this?

midichef: Offhand, all I can think of is to estimate the progress. We can make an educated guess about the average bytes per character, by measuring it around every 100 calls to `read()`. That would only add 1/100th of the encoding delay. Then it would just be `addProgress(len(r)*bytes_per_character)`.

Are there other options?

saulpw: I'm okay with estimation, doing it every 100 or 1000 calls to `read()`.

midichef:
[ecc8628](https://github.com/saulpw/visidata/pull/2323/commits/ecc8628d88b457fd55717d40c2fc9441fa04d160) estimates character length and uses that to estimate progress. It fixes the progress for UTF-16/UTF-32 as discussed. It also fixes it for UTF-8. For example, with a sample UTF-8 dataset of mostly Thai characters, loading progress was too slow by a factor of 2.5. Progress would max out at 40% instead of 100%.

The progress estimator samples characters every so often, to estimate the average bytes per character. Right now it samples more characters early in the file. That's because it needs more samples to come up with a decent estimate of byte length early.

It doesn't slow down code much. For a UTF-8 tsv file with 10 million short lines (`seq 10000000`), it's within 1% of the running time of the develop branch.  For a 2.7GB UTF-32 tsv file with 6 million rows, this change increases mean loading time by approximately 4%. That's higher than I expect. I would like to bring it down to below 1%. I will look into it more, but not soon. So this is ready for review.

midichef:
A related bug in v3.0.2. Progress on compressed textfiles was overcounted.  It would pass 100% and go to 200-600%.
[338b250](https://github.com/saulpw/visidata/pull/2323/commits/338b2506ac78cbb118b1c6b875072fdbe0729240) fixes it.
To reproduce: `seq 10000000 |gzip -c >! repro_gz_progress.gz; vd repro_gz_progress.gz`
The progress meter will pass 100% around the 5,000,000th row.

saulpw:
Thanks for doing the work on this, @midichef!  I put together #2407 which might be a simpler way to do a few things.  I'm not super-excited about some things in my PR, but between the two let's find the common ground.

and let's take 338b250 regardless.

anjakefala:
Hi @midichef! Could you review the most [recent commit](https://github.com/saulpw/visidata/pull/2323/commits/667672183a1d2b7f6be0bc2d9a73f2e561b5eb19)? We removed the batching to simplify the code a bit.

saulpw:
As this is almost 2 years old, I'm going to close it.  Please open a new PR if you want to get these changes in.

midichef:
I'm fine with this one staying closed. It's always had more important work keeping it from being top of my list, and I didn't like the idea of slowing down loading by 4%. And profiling the loading time was difficult to do precisely and reproducibly.

PR #2324, opened 2024-02-19
[path-] minimize string concat in RepeatFile read

midichef:
Right now `read()` in `RepeatFile` is inefficient. It builds up long strings by repeatedly appending a small string to a huge one. When piping to a tsv reader via `cat test.tsv |vd -f tsv -` that ends up being about 300-1000 appends to assemble a buffer's worth of characters, 64k.

In practice, this change doesn't make much difference in speed/memory. Longer reads would presumably be worse, though I couldn't construct an example that showed that. Maybe Python is already really good about reusing string memory in such cases?

It's also good to have [a new extra test of Repeat File](https://github.com/saulpw/visidata/issues/2320#issuecomment-1951705858_), to make sure this PR doesn't lose data, thanks @anjakefala.

### DIFF
--- a/visidata/path.py
+++ b/visidata/path.py
@@ -414,21 +414,23 @@ class RepeatFile:
 
     def read(self, n=None):
         '''Returns a string or bytes object. Unlike the standard read() function, when *n* is given, more than *n* characters/bytes can be returned, and often will.'''
-        r = None
         if n is None:
             n = 10**12  # some too huge number
-        while r is None or len(r) < n:
+        r = []
+        size = 0
+        output_type = str; eol = '\n'; joiner = ''
+        while not r or size < n:
             try:
                 s = next(self.iter)
-                if r is None:
-                    r = '' if isinstance(s, str) else b''
-                else:
-                    assert isinstance(r, type(s)), (r, type(s))
-
-                r += s + '\n' if isinstance(s, str) else b'\n'
+                if not r and isinstance(s, bytes):
+                    output_type = bytes; eol = b'\n'; joiner = b''
+                assert isinstance(s, output_type), (s, output_type)
+                r.append(s)
+                r.append(eol)
+                size += len(s) + len(eol)
             except StopIteration:
                 break  # end of file
-        return r or ''
+        return joiner.join(r)
 
     def write(self, s):
         return self.iter_lines.write(s)
--- a/visidata/path.py
+++ b/visidata/path.py
@@ -413,6 +413,7 @@ class RepeatFile:
         return RepeatFile(self.iter_lines, lines=self.lines)
 
     def read(self, n=None):
+        '''Returns a string or bytes object. Unlike the standard read() function, when *n* is given, more than *n* characters/bytes can be returned, and often will.'''
         r = None
         if n is None:
             n = 10**12  # some too huge number


ajkerrigan:
I was curious about the speed/memory impact of this change, because it _feels_ like such an intuitively good/"pythonic" move. But yeah from a few tests opening a 200k line, ~8.5 MB TSV file the speed/peak memory delta seemed negligible. [Memray](https://bloomberg.github.io/memray) did reveal some pretty dramatic changes from the local allocation perspective though.

With this change, a [table report](https://bloomberg.github.io/memray/table.html) shows:

image

<details>
<summary>Text version</summary>

```
Thread ID	Size	Allocator	Allocations	Location
0x1	208.9 MiB	malloc	400000	iterload at /home/aj/code/visidata/visidata/loaders/tsv.py:92
0x1	208.9 MiB	malloc	400000	iterload at /home/aj/code/visidata/visidata/loaders/tsv.py:92
0x1	28.6 MiB	malloc	4115	splitter at /home/aj/code/visidata/visidata/loaders/tsv.py:62
0x1	28.6 MiB	malloc	4115	splitter at /home/aj/code/visidata/visidata/loaders/tsv.py:62
0x1	13.1 MiB	malloc	3143	read at /home/aj/code/visidata/visidata/path.py:431
0x1	6.7 MiB	malloc	2125	read at /home/aj/code/visidata/visidata/path.py:431
0x1	8.3 MiB	malloc	1063	__next__ at /home/aj/code/visidata/visidata/path.py:486
```

</details>

Where before the change...

image

<details>
<summary>Text version</summary>

```
Thread ID	Size	Allocator	Allocations	Location
0x1	208.9 MiB	malloc	400000	iterload at /home/aj/code/visidata/visidata/loaders/tsv.py:92
0x1	208.9 MiB	malloc	400000	iterload at /home/aj/code/visidata/visidata/loaders/tsv.py:92
0x1	6.1 GiB	malloc	198410	read at /home/aj/code/visidata/visidata/path.py:427
0x1	6.0 GiB	malloc	197356	read at /home/aj/code/visidata/visidata/path.py:427
0x1	28.6 MiB	malloc	4115	splitter at /home/aj/code/visidata/visidata/loaders/tsv.py:62
0x1	28.6 MiB	malloc	4115	splitter at /home/aj/code/visidata/visidata/loaders/tsv.py:62
0x1	8.4 MiB	malloc	1065	__next__ at /home/aj/code/visidata/visidata/path.py:481
```

</details>

There are a few different knobs for memory profiling and I don't at all claim to be an expert, so consider this more "oh, neat" than anything else. Still... neat :grin:

midichef:
Oh cool, thanks for demonstrating memray!

I was curious what was going on in that top line: `tsv.py:92`, which is line 95 in the latest develop branch:
https://github.com/saulpw/visidata/blob/be295d8fabf468596861015980d6c27f59c21636/visidata/loaders/tsv.py#L95
We can get rid of that `list()` call. I missed it in my recent patches to tsv.py. Would you like to do the honor of submitting a PR?

Again, not much memory is saved, since Python reuses the object. But on my system it saves about 150ns per row, or 150ms on 1 million rows.

saulpw:
The row actually has to be a `list`, rather than the `tuple` returned by `split()`, for multiple reasons:

1. Small tuples can be interned and resolve to the same object, so you may find that selecting one row (which happens by object id) selects all rows with that value.  Lists are guaranteed to have unique object ids.
2. Tuples aren't modifiable, so you wouldn't be able to `e`dit a value on a TSV without the `list`.

midichef:
Oh, I didn't know about that behavior for tuples! It looks like `split()` returns a `list` here, though?

anjakefala:
As a side note, memray is incredible. It has a `--native` flag which will show the full stacktrace for any C-code underpinning a Python program. So it works for Cython as well!

saulpw:
> It looks like split() returns a list here, though?

Huh, I guess it does!  I don't know why, I was certain that str.split returned a tuple.  Well if it doesn't break anything then I guess it truly is superfluous.

saulpw (approved):
This looks reasonable to me at a glance, but I haven't tested it.  Thanks for putting it together, @midichef !

midichef:
I added a note that `RepeatFile.read(size)` can return more than `size` characters/bytes. That's longstanding behavior, nothing new due to my patch.

I don't think we can fix it, as `RepeatFile` iterators don't allow seeking backward. Still, it doesn't cause any problems, so it's fine for now.

PR #2331, opened 2024-02-23
[pyobj-] improve display of sets and simple objects

midichef:
This set of commits is aimed at making `pyobj-expr` more user-friendly. As it stands now, the results from looking at certain values are quite strange.
`None`: gives a sheet with no rows
`set([1,2,3])`: a sheet with no rows (because what is shown are its attributes, not its elements)
`str()`: a sheet with no rows, with the header `text`, which is different from the previous two cases
`1` or `True`: a sheet of attributes, hard to make sense of:
```
attribute   value docstring
denominator 1     <type int>
imag        0     <type int>
numerator   1     <type int>
real        1     <type int>
```

I've added a new sheet type, `PythonAtomSheet` to hold `None` or a `bool`, a number, or an empty string. The goal is to improve the display. The change makes it easy to distinguish an expression value of `None` from `""`, due to the `⌀` shown. And it shows bools/numbers in a simple 1x1 cell, instead of a 4x3 sheet of display attributes.

The new sheet also doesn't allow diving deeper into such simple objects. By contrast, the way visidata already works is, after `pyobj-expr` for a number `1`, you can keep hitting `Enter` to dive in on the `denominator` field infinitely. It's quite confusing for a new user.

And I fixed a bug where if a sheet had no rows, `open-row` takes the user to a new `PyobjExpr` sheet.
(to reproduce, `vd nonexistent_file.tsv` then `ENTER`)

Some quirks of this PR:
1. `None` is shown as it is in a standard sheet, an empty cell with `⌀` next to it. Maybe it should show the text "None".
2. If you view bools, the call to `deduceType` will put `?` next to the column header.
3. The empty string goes into `PythonAtomSheet` unlike regular strings that go into a TextSheet. I did that to prevent further dives into the empty string. But that creates an inconsistency, the header for empty strings is "value", but for all other strings it's "text". To get rid of the special handling for empty strings, just change the code to:
```
        if pyobj is None or isinstance(pyobj, numbers.Number):
...
        elif isinstance(pyobj, str):
            return TextSheet(*names, source=pyobj.splitlines() if pyobj else [''])
        elif isinstance(pyobj, bytes):
            s = pyobj.decode(cls.options.encoding).splitlines()
            return TextSheet(*names, source=s if s else [''])
```

### DIFF
--- a/visidata/pyobj.py
+++ b/visidata/pyobj.py
@@ -1,6 +1,7 @@
 from typing import Mapping
 import inspect
 import math
+import numbers
 
 from visidata import vd, asyncthread, ENTER, deduceType
 from visidata import Sheet, Column, VisiData, ColumnItem, TableSheet, BaseSheet, Progress, ColumnAttr, SuspendCurses, TextSheet
@@ -16,6 +17,28 @@ class PythonSheet(Sheet):
     def openRow(self, row):
         return PyobjSheet("%s[%s]" % (self.name, self.keystr(row)), source=row)
 
+class PythonAtomSheet(PythonSheet):
+    '''a sheet to display one Python object that does not offer deeper inspection,
+        like None, a bool, or an int/float'''
+    rowtype = 'object'  #singular, because it should only ever hold one
+    columns = [
+        Column('value', getter=lambda col,row: row,
+                        setter=lambda c,r,v: None)
+    ]
+    def loader(self):
+        self.rows = [self.source]
+        self.column('value').type = deduceType(self.source)
+
+    def openRow(self, row):
+        vd.fail('cannot dive deeper on this object')
+    def openCell(self, col, row, rowidx=None):
+        vd.fail('cannot dive deeper on this object')
+    def openRowPyobj(self, rowidx):
+        vd.fail('cannot dive deeper on this object')
+    def openCellPyobj(self, col, rowidx):
+        vd.fail('cannot dive deeper on this object')
+    def newRow(self):
+        vd.fail('adding rows to this sheet is not supported')
 
 #### generic list/dict/object browsing
 @VisiData.global_api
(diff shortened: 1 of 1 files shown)


visidata/indexsheet.py
```diff
@@ -102,7 +102,7 @@ def nextRow(sheet, n=1):
 SheetsSheet.addCommand('gC', 'columns-selected', 'vd.push(ColumnsSheet("all_columns", source=selectedRows))', 'open Columns Sheet with all visible columns from selected sheets')
 SheetsSheet.addCommand('z^C', 'cancel-row', 'cancelThread(*cursorRow.currentThreads)', 'abort async thread for current sheet')
 SheetsSheet.addCommand('gz^C', 'cancel-rows', 'for vs in selectedRows: cancelThread(*vs.currentThreads)', 'abort async threads for selected sheets')
-SheetsSheet.addCommand('Enter', 'open-row', 'dest=cursorRow; vd.sheets.remove(sheet) if not sheet.precious else None; vd.push(openRow(dest))', 'open sheet referenced in current row')
+SheetsSheet.addCommand(ENTER, 'open-row', 'dest=cursorRow; vd.sheets.remove(sheet) if not sheet.precious else None; vd.push(openRow(dest))', 'open sheet referenced in current row')
```
saulpw: We're heading more towards using prettykeys like `Enter` for keybindings, and deprecating the `ENTER` variable (though we haven't officially done so yet).  So let's not change these instances.

saulpw (changes requested):
Let's revert the changes to use `ENTER` for the keybinding, and then this is good to merge.  Thanks @midichef!

midichef:
Okay, I reverted the ENTER change, then consolidated the others. This PR is done.

saulpw (approved):

PR #2333, opened 2024-03-01
[sheets-] draw type symbol for col wider than screen

midichef:
If a column is wider than the screen the type symbol is drawn offscreen. Here's a little fix for that.

repro:  `vd sample_data/countries` `#` `z_` `1000` and the type symbol '#' is not visible.

It also changes `len()` to `dispwidth()` to handle the hypothetical case where the type symbol is wider than a normal character.

2nd bug, only related because it came up in testing this.
In two cases empty columns are created with a width of `None`. `z_` on such columns will fail with an error:
```
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/_input.py", line 376, in editText
    v = type(starting_value)(v)
TypeError: NoneType takes no arguments
```
repro:  `vd -f txt` then `z_` `100`
repro:  `touch zero.txt; vd zero.txt` then `z_` `100`

### DIFF
--- a/visidata/_open.py
+++ b/visidata/_open.py
@@ -177,7 +177,7 @@ def open_txt(vd, p):
                 if delimiter and delimiter in next(fp):    # peek at the first line
                     return vd.open_tsv(p)  # TSV often have .txt extension
             except StopIteration:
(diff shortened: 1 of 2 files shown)


saulpw (approved):
Thanks for the attention to detail!

PR #2338, opened 2024-03-04
[cmdpalette-] fix incomplete command list from race condition

### LINKED ISSUE #2336: Python error after pressing SPACE to pull up command-name input
**Small description**
The first time I press SPACE on a sheet, I often get the following Python error. It usually succeeds on the second press.

**Expected result**
Open command prompt.

Screenshot from 2024-03-03 15-07-48
[errors_recent.txt](https://github.com/saulpw/visidata/files/14471969/errors_recent.txt)

Open any sheet and press SPACE.
It may take a few attempts. It seems to only happen the first time it is entered on a sheet, and only some of the time.

**Additional context**
Python 3.8.0
VisiData 3.0.2

midichef:
Probably fixes #2336.

Not all rows on the HelpSheet were loaded by the time the cmdpalette read its list of commands.
repro:  on any sheet, reload the command palette by pressing `Space` and then `Esc` repeatedly. Eventually the list in the palette will change. Or the palette will come up blank, or the exceptions mentioned in #2336 will be raised.

### DIFF
--- a/visidata/features/cmdpalette.py
+++ b/visidata/features/cmdpalette.py
@@ -167,7 +167,7 @@ def inputLongname(sheet):
     prompt = 'command name: '
     # get set of commands possible in the sheet
     this_sheets_help = HelpSheet('', source=sheet)
-    this_sheets_help.ensureLoaded()
+    vd.sync(this_sheets_help.ensureLoaded())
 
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Thanks @midichef!  This seems right to me.

PR #2341, opened 2024-03-05
[save-] before saving, load unloaded sheets

midichef:
repro:
```
vd sample_data/sample.tsv sample_data/errors.csv sample_data/countries
save-all
unloaded.vds
```
then
`vd unloaded.vds` will show that two of the sheets have no rows.

Also, saving a sheet with 0 rows results in a .vds file that's not readable. I'll file that as a separate issue.

### DIFF
--- a/visidata/save.py
+++ b/visidata/save.py
@@ -3,7 +3,7 @@ import os
 from copy import copy
 
 from visidata import vd
-from visidata import Sheet, BaseSheet, VisiData, IndexSheet, Path, Progress, TypedExceptionWrapper
+from visidata import Sheet, BaseSheet, VisiData, IndexSheet, Path, Progress, TypedExceptionWrapper, UNLOADED
 
 vd.option('safe_error', '#ERR', 'error string to use while saving', replay=True)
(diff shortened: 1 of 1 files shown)


saulpw:
Should we fail outright, or call `ensureLoaded()` on them all and sync for them to finish?

midichef:
Good point, loading is definitely better. I've changed it to call `ensureLoaded()` instead of fail.

visidata/save.py
```diff
@@ -109,6 +109,8 @@ def saveSheets(vd, givenpath, *vsheets, confirm_overwrite=True):
     if not vsheets: # blank tuple
         vd.warning('no sheets to save')
         return
+    unloaded = [ vs for vs in vsheets if vs.rows is UNLOADED ]
+    vd.sync(*vd.ensureLoaded(unloaded))
```
saulpw, line 113: Closer!  But ensureLoaded is on BaseSheet, not the vd object.  And it also checks for UNLOADED itself.  So maybe we should make `vd.sync` ignore None values, and just call `vd.sync(*[vs.ensureLoaded() for vs in vsheets])`.

saulpw (approved):

anjakefala:
This is so cool!

PR #2345, opened 2024-03-11
[vdobj] give each thread its own activeCommand

### LINKED ISSUE #2266: [undo] can't undo some threaded commands

midichef:
Closes #2266.

a simple repro for #2266:  open a large sheet and delete a large number of rows:
`seq 1222333 |vd` then `gs` `gd` `U`. The rows don't come back.

### DIFF
--- a/visidata/vdobj.py
+++ b/visidata/vdobj.py
@@ -172,12 +172,8 @@ class VisiData(visidata.Extensible):
     @property
     def activeCommand(self):
         '''Returns the active command for the current thread.
-            Returns UNLOADED, None, or a CommandLogRow.
+            Returns a CommandLogRow, or None.
             Returns None when no command has ever been stored for the current thread.'''
-        if not self.activeCommands:
-            # this tuple is the same as UNLOADED, but we can't use the name
-            # here, because importing it fails
-            return tuple()
         tid = threading.get_ident()
         if tid not in self.activeCommands:
             return None
--- a/visidata/vdobj.py
+++ b/visidata/vdobj.py
@@ -9,6 +9,7 @@ __all__ = ['ENTER', 'ALT', 'ESC', 'asyncthread', 'VisiData']
 ENTER='Enter'
 ALT=ESC='^['
 
+import threading
 
 # define @asyncthread for potentially long-running functions
 #   when function is called, instead launches a thread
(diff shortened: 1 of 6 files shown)


visidata/vdobj.py
```diff
@@ -166,3 +168,24 @@ def screenHeight(self):
...
+        tid = threading.get_ident()
+        if tid not in self.activeCommands:
+            return None
+        return self.activeCommands[tid]
+
+    @activeCommand.setter
+    def activeCommand(self, value):
+        self.setThreadActiveCommand(threading.current_thread(), value)
+
+    def setThreadActiveCommand(self, thread, value):
```
saulpw: If no one else uses this function, let's merge it into the activeCommand setter.

saulpw (approved):
I think this all makes sense to me.  If it fixes the issue and seems to work otherwise, we'll take it.

visidata/vdobj.py
midichef: Yes, that's the only place it was used. I left the thread-specifying setter there in case we wanted to call it explicitly in `_with_active_cmd` for clarity. Because as I have it now, it is a bit strange to see `cmd = vd.activeCommand` and then, two lines later `if cmd: vd.activeCommand = cmd`.

I've merged it.

saulpw: I see what you mean.  We'll add a comment there referring to this issue/PR.

PR #2346, opened 2024-03-12
[selection-] fix threading and progress for toggle

### LINKED ISSUE #1671: [freqtbl] toggling rows selects/unselects an inconsistent number of rows
**Small description**
In a FreqTableSheet where its source sheet has many rows, when rows are toggled, the number of rows that is selected/unselected on the source sheet is inconsistent.

**Expected result**
Every time the toggle is performed, the number of rows selected/unselected on the source sheet should be the same.

The number of rows selected on the source sheet is different every time. Sometimes, the number of unselected rows is negative.
https://asciinema.org/a/ksxBTGy0E1tGh5unX2Gsu0eeR

`python -c "n = 200_000; s = 'line\n'; print(f'{s*n}',end='')" |vd`
hit `F` and then repeatedly hit `t` (or `gt` or `gzt` or, in a sheet with more than one kind of cell, `zt`)

**Additional context**
VisiData v2.10.2

Workaround:
Use `s` and `u` to select/unselect rows instead of toggling.

There is a simple fix for the inconsistent behavior, by adding vd.sync() to selectRow() and unselectRow().
https://github.com/midichef/visidata/commit/52d8596ed01fed17bca8b5215130a77801f1e3b3
However, I'm unsure if this is the correct solution for 2 reasons.
First, I don't have a good understanding of how threading works in the visidata codebase.
Second, it's not clear to me how FreqTableSheet toggling should behave. Suppose that some rows on the source sheet are already selected, but not all of them. The way it works now is that after a toggle, all matching rows in the source sheet will end up in the same state: either all are selected, or all are unselected. But arguably the effect should be to toggle the selected status of all the matching rows in the source sheet, leaving some selected and some unselected.

midichef:
Fixes #1671.

To recap the bug there:  on a FreqTableSheet, `unselectRows()` and `selectRows()` each start a thread to process the rows on the source table. Both functions are run in order when `toggle-row()` is run on an unselected row. The two threads each process the same rows, which are left in an inconsistent mix of selected and not-selected.

This PR prevents that by making sure the toggle functions in `Sheet` only run `unselectRows()` or `selectRows()`, not both. The downside is that `toggle()` will run a bit slower than it does now. Because `unselectRows()` already runs the equivalent of `isSelected(row)`, and this PR adds an explicit `isSelected()` in front of that.

As a separate bug, progress on large toggles was undercounted.
repro: `seq 5000111 | vd` then toggle half the rows via `zr` `2500111` `zt`.   The progress will max out at 50%.

### DIFF
--- a/visidata/selection.py
+++ b/visidata/selection.py
@@ -23,7 +23,7 @@ def isSelected(self, row):
 def toggle(self, rows):
     'Toggle selection of given *rows*.  Async.'
     self.addUndoSelection()
-    for r in Progress(rows, 'toggling', total=len(self.rows)):
+    for r in Progress(rows, 'toggling', total=len(rows)):
         if not self.unselectRow(r):
             self.selectRow(r)
 
--- a/visidata/selection.py
+++ b/visidata/selection.py
@@ -24,7 +24,9 @@ def toggle(self, rows):
     'Toggle selection of given *rows*.  Async.'
     self.addUndoSelection()
     for r in Progress(rows, 'toggling', total=len(rows)):
-        if not self.unselectRow(r):
+        if self.isSelected(r):  #1671
+            self.unselectRow(r)
+        else:
             self.selectRow(r)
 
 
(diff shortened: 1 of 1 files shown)


saulpw:
I see, this makes sense.

`isSelected` is pretty fast by design, as it's just a hash lookup of `id(row)`.  We can take this fix.  Thanks for sorting it out.

saulpw (approved):

PR #2349, opened 2024-03-15
[ping-] fix open-ping sheet opening

midichef:
Currently, `open-ping` does nothing after the user inputs an IP/hostname. That's because the user input is passed to `openSource()`, and then treated like a file that does not exist:
https://github.com/saulpw/visidata/blob/e9e8ef88849e75420017d42cbe250c778d822c10/visidata/_open.py#L93

This PR fixes that by removing the call to `openSource()`.

A shorter fix would have been to change the command `openSource()` to `openSource(..., create=True)`. The reason to reject that fix, is that it gives the wrong behavior when the user inputs a hostname that is also a file in the current directory. `openSource()` would then open the local file instead.

### DIFF
--- a/visidata/features/ping.py
+++ b/visidata/features/ping.py
@@ -10,9 +10,14 @@ from visidata import vd, VisiData, BaseSheet, Sheet, Column, AttrColumn, Progres
 @VisiData.api
 def new_ping(vd, p):
     'Open a sheet with the round-trip time for each hop along the path to the given host.'
-    pingsheet = PingSheet(p.given, source=p.given)
-    return PingStatsSheet("traceroute_"+pingsheet.name, source=pingsheet)
+    return makePingStats(p.given)
 
+# open_ping() is called when a local file exists matching the path p, otherwise new_ping() is called
+vd.open_ping = vd.new_ping
+
+def makePingStats(ip):
+    pingsheet = PingSheet(ip, source=ip)
+    return PingStatsSheet("traceroute_"+pingsheet.name, source=pingsheet)
 
 class PingStatsSheet(Sheet):
     help='''# ping/traceroute
@@ -71,7 +76,7 @@ class PingSheet(Sheet):
     def ping_error(self, ip, data):
         if ip in self.sources:
             self.sources.remove(ip)
-            vd.warning("%s removed: %s" % (ip, data))
+            vd.warning("%s removed: %s" % (ip, data.rstrip()))
 
     def update_traces(self, row, ip):
         rtes = self.routes.get(ip)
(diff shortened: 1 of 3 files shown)


midichef:
I disabled the test for `open-ping` because the test runner was failing to run the traceroute command in the shell:
```
FAILED: open-ping Traceback (most recent call last):
...
  File "/home/runner/work/visidata/visidata/visidata/features/ping.py", line 103, in send_trace
    sh.traceroute('--sim-queries=1',
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/sh.py", line 3657, in __getattr__
    return self.__env[name]
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/sh.py", line 3438, in __getitem__
    raise CommandNotFound(k)
sh.CommandNotFound: traceroute
1/440 commands had errors
FAILED
```

visidata/features/ping.py
```diff
@@ -145,7 +151,7 @@ def iterload(self):
 vd.option('ping_interval', 0.1, 'wait between ping rounds, in seconds', sheettype=PingSheet)
 
 PingSheet.options.null_value = False
-BaseSheet.addCommand('', 'open-ping', 'vd.push(openSource(input("ping: ", type="hostip"), filetype="ping"))', 'open sheet to ping input IP Address')
+BaseSheet.addCommand('', 'open-ping', 'open_ping()', 'open sheet to ping input IP Address')
```
saulpw: Please include at least `push` and `input` strings in the execstr; this is how the e.g. menu system knows that it takes input and pushes a new sheet.  You can make an `inputIP` if that is helpful; and move the `open_ping` to the VisiData object and return a Sheet (like other loaders) so that someone can do `vd -f ping 4.4.4.4`.

midichef: Oh, I didn't realize the strings needed to be in the execstr. I've put them back.

I kept `new_ping` that handled `vd -f ping 4.4.4.4`. And I made `open_ping` part of the VisiData object as you suggested, since it's called if the ping target happens to exist as a local file.

I modeled `makePingStats` on the `freqtbl` call to `makeFreqTable`. Please feel free to make any changes if that's not a good fit for current vd style.

saulpw (approved):
This is good, thank you!

anjakefala:
In theory, I added support for the `open-ping` test :crossed_fingers:

midichef:
Thanks for the test fix. Now that I know how, next time a library is needed, I'll give that a try myself.

anjakefala:
@midichef Yeah, in most CIs you can't `apt install`, so it is not obvious github actions will let you!

Issue #354, opened 2019-09-09
Failure reading stdin (piping to vd)


ajkerrigan:
**Small description**
In `v2.-1` I was able to pipe command output to `vd` and get a sheet. With the current develop branch (37acde0bb9a0b758d664341c835b4a13de0a2a08), I get no data.

**Expected result**
Piping something like a directory list to VisiData, I expect to see a sheet with the data:

```ls -l | vd -f fixed```

**Actual result with screenshot**
I get no error, but vd stops after showing:

```- does not exist, creating new sheet```

**Steps to reproduce with sample data and a .vd**
This small test script should exit with 0 if things are working, and 1 otherwise. Using it with `git bisect` suggests the issue traces back to 4f7f82cf4f1d4d278c51a9bfc00ad3514b69b8ef:

```bash
#!/bin/bash
VALUE=$(echo "hi" | ./bin/vd -bo -)
[[ "$VALUE" == "hi" ]] && exit 0
exit 1
```

saulpw:
A model bug report, @ajkerrigan!  concise and informative, which made it easy to fix.  Should be fixed in 2.-2.

PR #2361, opened 2024-03-25
[main-] remove unneeded reload in batch mode

midichef:
Batch mode (when not replaying commands with `-p`) is substantially slower to load a file than interactive mode.
```
seq 2000111 > 2m.tsv
# interactive mode:  quit.vdj is a timing helper script that opens 2m.tsv and immediately quits
=time vd -p quit.vdj
6.41user
# batch mode
=time vd -b 2m.tsv
9.41user
```

That's because batch mode runs `reload()` twice in a row in these two lines:
https://github.com/saulpw/visidata/blob/581cdf12b28278aea87ce09f7c32bf8ae1be2f0f/visidata/main.py#L336-L337

`vd.push()` calls `BaseSheet.ensureLoaded()` which calls `reload()`:
https://github.com/saulpw/visidata/blob/581cdf12b28278aea87ce09f7c32bf8ae1be2f0f/visidata/sheets.py#L1034
https://github.com/saulpw/visidata/blob/581cdf12b28278aea87ce09f7c32bf8ae1be2f0f/visidata/basesheet.py#L272

This PR removes the unneeded explicit call to `reload()`, running 2x faster for the special case of no commands.
```
=time vd -b 2m.tsv
4.73user
```

### DIFF
--- a/visidata/main.py
+++ b/visidata/main.py
@@ -334,7 +334,6 @@ def main_vd():
         if args.batch:
             if sources:
                 vd.push(sources[0])
-                sources[0].reload()
 
         for (f, *parms) in after_config:
             f(sources, *parms)


saulpw (commented):
This line was added in 330117fa328ecd8498fb76f08954bce2b328851c.  I agree that it shouldn't be necessary, and a 2x slowdown is bad of course, but let's make sure that we can still read from stdin in batch mode as per #354.

midichef:
It looks like we can still read from stdin in batch mode:
```
# cat sample_data/UTF-8-demo.txt |vd -b -o test.batch.txt
saul.pw/VisiData v3.1dev
unknown "" filetype
Support VisiData: https://github.com/sponsors/saulpw
opening - as txt
saving 1 sheets to test.batch.txt as txt
# diff -s sample_data/UTF-8-demo.txt test.batch.txt
Files sample_data/UTF-8-demo.txt and test.batch.txt are identical
```
In addition, the specific sample command from #&#x2060;354 also works for me: `ls -l | vd -f fixed`. So this PR is ready for final testing.

saulpw (approved):

PR #2362, opened 2024-03-25
[selection-] hide selection markers when cols are hidden

midichef:
When rows are selected and all columns are hidden, the selection markers are still drawn, next to the blank sheet. This PR checks for visible columns before drawing the symbol.

### DIFF
--- a/visidata/selection.py
+++ b/visidata/selection.py
@@ -7,7 +7,7 @@ vd.option('some_selected_rows', False, 'if no rows selected, if True, someSelect
 Sheet.init('_selectedRows', dict)  # rowid(row) -> row
 
 vd.rowNoters.append(
-        lambda sheet, row: sheet.isSelected(row) and sheet.options.disp_selected_note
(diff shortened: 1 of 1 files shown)


saulpw (commented):
Does it matter whether the selection markers are shown in this degenerate case?  It seems like the user might still want to know which rows were selected even if they can't see any data from the rows.

midichef:
The selection markers become misleading, once the last column disappears. Their vertical offset, which is usually 1, but could be any integer, becomes 0 when the header disappears with the last column, when `self.nHeaderRows` becomes 0.
https://github.com/saulpw/visidata/blob/43347b6148a317523b11053dd7b04fdb23a98455/visidata/sheets.py#L391
https://github.com/saulpw/visidata/blob/43347b6148a317523b11053dd7b04fdb23a98455/visidata/sheets.py#L768
So I think it's better not to show the markers at all.

saulpw:
Okay, I made the minimum number of header rows 1, and when there are no unhidden columns, it shows a basic Column that just stringifies the row.  I think this is more useful behavior than constructing a perfectly null interface for this case.

PR #2365, opened 2024-03-27
[sas-] make XptSheet rows be list instead of tuple

midichef:
Currently `XptSheet` in `loaders/sas.py` produces rows that are tuples.

That's undesirable because as @saulpw pointed out recently, rows can't be edited if they are tuples. And also because selecting one row may cause other rows to be selected if they are interned tuples, that have the same values and ID. This PR converts the tuples to lists.

to reproduce:
`pip install xport==2.0.2`
download [brakes.xpt](https://github.com/prasertcbs/basic-dataset/blob/8c80eb454b2defd3db217d8f95883eb6a1e3f263/brakes.xpt)
`vd brakes.xpt` then inspect any row with `^Y` for `pyobj-row`, the status bar will say `tuple`.


saulpw (approved):
Yes, these should be lists.

PR #2369, opened 2024-04-03
[mainloop-] during replay, redraw less often

### LINKED ISSUE #2274: [input] screen flickers in edit-cell
**Small description**
If I do `edit-cell` with a guide on screen in the sidebar, and then hit `Tab`, the cursor moves to the correct cell. But the cursor and screen flicker like they are being frequently redrawn. The flicker is mainly in the sidebar and the current cell.

 I can upload a terminal recording, if this isn't easy to reproduce on different machines.

`vd -N sample_data/sample.tsv`, then `edit-cell` and hit `Tab`.

**Additional context**
The behavior is different in gnome-terminal vs alacritty. For gnome-terminal, the flicker shows up after tabbing to a different cell.
For alacritty, I don't have to hit Tab, the flicker shows up as soon as I run `edit-cell`.

For both gnome-terminal and alacritty, bisect shows the flicker started with b859610102dc570573f2db8d7e24a13f9a12397b.

saul.pw/VisiData v3.1dev
Python 3.10.12
Ubuntu 22.04.3
gnome-terminal 3.44.0 using VTE 0.68.0 +BIDI +GNUTLS +ICU +SYSTEMD
alacritty 0.13.1

midichef:
When replaying commands, visidata 3.0 versions are noticeably slower to load some large sheets than v2.11 was. It's caused by redrawing the screen too often. The symptom to look for is that the progress (`% loading…`) is updated dozens of times per second.

Sometimes the quick updates are steady. For other files they come and go in short bursts, then doing proper slower updating for longer stretches. Only the short bursts of quick updates will cause slower loading.

Here is a case that, on my system, consistently exhibits the problem.
```
seq 2000111 > 2m.tsv
# interactive mode:  quit.vdj is a timing helper script that opens 2m.tsv and immediately quits
=time vd -p quit.vdj
6.63user
```
After applying 02432ff86ec91c36ed53c507319402d7d7c054b0:
```
=time vd.fixed -p quit.vdj
5.12user
```

In this case the bug causes 30% slower loading. It causes the screen to be drawn 37 times per second, vs the fixed version that draws 7 times per second.

While I was patching mainloop, I also fixed an off-by-one error in counting idle timeouts (51421d9a3afb211a1cc820fffc2bd1b6dd09643a). The effect will be ~9% fewer recalculations of columns when idling. To demonstrate the change:  `echo a |vd` then make an ExprColumn: `=vd.status('test')`. For each row in the column, the ExprColumn will be calculated once, and then recalculated once after every timeout. Before the fix, in v3.0.2 it will output 12 lines to the status bar, and after the fix, only 11.

--

Edited to add:  it turns out this PR also closes #2274. In that bug, `editCell` queues commands in `_nextCommands`, triggering the same high rate of redrawing.

### DIFF
--- a/dev/checklists/manual-tests.md
+++ b/dev/checklists/manual-tests.md
@@ -94,3 +94,4 @@
 22. Test `open-row` on an html link: https://hls.gsfc.nasa.gov/data/
 23. That DirSheet requires a commit-sheet before changes on filesystem
 24. Test adding multiple aggregators via palette (+)
+25. time vd -p tests/quit-nosave.vdj  - note down the time. compare to PR #2369
--- a/tests/quit-nosave.vdj
+++ b/tests/quit-nosave.vdj
@@ -1,0 +1,3 @@
+#!vd -p
+{"sheet": "", "col": "", "row": "", "longname": "open-file", "input": "sample_data/2m.tsv", "keystrokes": null, "comment": null}
+{"sheet": "", "col": "", "row": "", "longname": "quit-all", "input": ""}
--- a/visidata/mainloop.py
+++ b/visidata/mainloop.py
@@ -258,7 +258,7 @@ def mainloop(vd, scr):
             vd.curses_timeout = nonidle_timeout
         else:
             numTimeouts += 1
-            if vd.timeouts_before_idle >= 0 and numTimeouts > vd.timeouts_before_idle:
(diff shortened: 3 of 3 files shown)


midichef:
It turns out this PR also closes #2274.

saulpw (approved):
This makes sense, thanks for fixing it!

PR #2372, opened 2024-04-05
[clipboard-] speed up setcol-clipboard and make it async

midichef:
`setcol-clipboard` is very slow for large numbers of rows. It also freezes the display while it runs.

To see the slowness: do `seq 1000111 |vd -` then paste a column of a million rows via `gs` `gzy` `za` `newcol` `gzp`. On my system it processes a few thousand rows per second.

This PR makes `setcol-clipboard` about 100x faster. It also adds the `@asyncthread` decorator to `setValuesTyped()` (like  `setValues()`) so the screen can continue to update while `setValuesTyped()` runs.


saulpw (approved):
I'm surprised that this is a 100x speedup, but it does simplify the code, and there's already the cycle() in setValuesTyped.  Thanks for finding/fixing it.

PR #2373, opened 2024-04-06
[color-] make status colors more readable

midichef:
In `develop`, but not in v3.0.2, the display of the last command is hard to read. It's cyan on light blue:

longname

That's because `color_longname` is currently `6`, which is cyan. It's hard to pick a text color that is readable against the light blue status bar and also readable against the green guide, so this PR splits `color_longname` into two settings, `color_longname_guide` and `color_longname_status`.

A couple of other color changes:
For the `ascii8` theme: the progress indicator (`loading...`) is hard to read, as `color_working` is currently `green`, on a cyan background. So I've changed it to `bold white`. Of the 8 foreground color choices, magenta and yellow were a bit too faint, red looked like an error, and the rest were too hard to read.

Also for `ascii8`, longname is currently invisible, cyan on cyan. The new default color for `color_longname_status` fixes this too.

And the lower right corner of the guide box has a stray bracket character. It currently reads: `toggle sidebar |[┘`, which this PR changes to `toggle sidebar |─┘`.

The reason for choosing `white` for `color_longname_status`, is that it's readable against the light blue status bar for the default theme, as well as against cyan for the `ascii8` theme. If there are better colors, please feel free to go with those.


saulpw (approved):
Thanks for this, @midichef!  I've been annoyed by the lack of readability myself, and these are definite improvements.

PR #2380, opened 2024-04-15
handle dates and timedeltas in aggregators and DescribeSheet

midichef:
This PR is makes the current aggregators (mean, median, sum) work with dates and timedeltas. It's a continuation of the date-handling in https://github.com/saulpw/visidata/discussions/1996.

One notable change is that the `stdev` of a `date` is a `timedelta`, not a `date`. The benefits of this are, it formats more readably in the describe sheet (such as '200 days'). And it simplifies doing arithmetic with the result of `stdev`, like expressing another timedelta as a multiple of the stdev. The downside is that it breaks the assumption that an aggregator's result is the same type as the column being aggregated. That leads to some special case handling:
https://github.com/saulpw/visidata/blob/518341122ec84ba150e6b728d994f78195bbaeda/visidata/aggregators.py#L249
And note that the `stdev` of a `timedelta` is still a `timedelta`.

In order to make the aggregators and the DescribeSheet handle `date` and `timedeltas`, I've had to change the type of several aggregators to `anytype`. Is this acceptable? Are there drawbacks?

I've submitted this as a series of small commits for easier review. If this all looks okay, let me know, I'll squash it to a smaller number of commits before deploying.


visidata/aggregators.py
```diff
@@ -107,13 +108,48 @@ def _funcRows(col, rows):  # wrap builtins so they can have a .type
 def mean(vals):
     vals = list(vals)
     if vals:
-        return float(sum(vals))/len(vals)
+        if type(vals[0]) is date:
+            vals = [d.timestamp() for d in vals]
```
saulpw, line 110: I don't think these special cases belong in the aggregators.  The idea is that people can define their own aggregators with 'obvious' implementations and they should work reasonably.  (and regardless, where these checks are, they should use `isinstance` with the Python stdlib datetime as you're doing below).

To that end, the visidata `date` and `datedelta` classes both already have `__float__` methods which convert them to float as you're doing here.  So maybe the aggregator callsite can call `aggregator.intype` on each value, and for these "more numeric" aggregators would set their intype to `float`.  Then they would need to convert the float result back into some result type, which as you've noted below can be tricky for date formats.

visidata/aggregators.py
```diff
@@ -107,13 +108,48 @@ def _funcRows(col, rows):  # wrap builtins so they can have a .type
...
+        elif isinstance(vals[0], datetime.timedelta):
+            return datetime.timedelta(seconds=vsum(vals)/datetime.timedelta(seconds=len(vals)))
+        else:
+            return float(sum(vals))/len(vals)
 
 def _vsum(vals):
-    return sum(vals, start=type(vals[0] if len(vals) else 0)())  #1996
+    if vals:
+        if type(vals[0]) is date:
+            vd.error('dates cannot be summed')
```
saulpw, line 121: This would be a `vd.fail` as it's more of a user error.  But it should also be outside the aggregation function.

visidata/aggregators.py
```diff
@@ -107,13 +108,48 @@ def _funcRows(col, rows):  # wrap builtins so they can have a .type
...
+        # two midpoints to average them, so convert to timestamps
+        vals = [d.timestamp() for d in vals]
+        return datetime.date.fromtimestamp(statistics.median(vals))
+    return statistics.median(vals)
+
+def stdev(vals):
+    if vals and len(vals) >= 2:
+        if type(vals[0]) is date:
+            vals = [d.timestamp() for d in vals]
+            return datetime.timedelta(seconds=statistics.stdev(vals))
```
saulpw, line 141: This is very interesting, I wonder how we can codify that the output type of stdev for both date and datedelta are datedelta.  It's not quite as simple as making the result type `datedelta` and passing it the float from the generic aggregations, since Python's timedelta (which datedelta is a subclass of) has a constructor with a first parameter of 'days' instead of 'seconds'.

visidata/aggregators.py
```diff
@@ -146,17 +182,17 @@ def percentile(pct, helpstr=''):
 def quantiles(q, helpstr):
     return [percentile(round(100*i/q), helpstr) for i in range(1, q)]
 
-vd.aggregator('min', min, 'minimum value')
-vd.aggregator('max', max, 'maximum value')
-vd.aggregator('avg', mean, 'arithmetic mean of values', type=float)
-vd.aggregator('mean', mean, 'arithmetic mean of values', type=float)
-vd.aggregator('median', statistics.median, 'median of values')
+vd.aggregator('min', min, 'minimum value', type=anytype)
```
saulpw, line 186: Why are min/max forcibly set to `anytype` instead of letting them take the type of their source column?

I understand why avg/mean/median want to be more generic; maybe this is where we have both an 'input type' which remains `float` for these, and a 'result type' which is more variable.

visidata/aggregators.py
```diff
@@ -218,14 +254,20 @@ def aggname(col, agg):
...
+        vd.fail('no rows to aggregate')
     for agg_choice in agg_choices:
         agg = vd.aggregators.get(agg_choice)
         if not agg: continue
         aggs = agg if isinstance(agg, list) else [agg]
         for agg in aggs:
             aggval = agg(col, rows)
             typedval = wrapply(agg.type or col.type, aggval)
-            dispval = col.format(typedval)
+            if agg.name == 'stdev' and (col.type is date):
```
saulpw, line 285: We need to figure out how to not have a very special type check here.

visidata/features/describe.py
```diff
@@ -44,10 +45,6 @@ class DescribeSheet(ColumnsSheet):
             DescribeColumn('nulls',  type=vlen),
             DescribeColumn('distinct',type=vlen),
             DescribeColumn('mode',   type=str),
-            DescribeColumn('min',    type=str),
```
saulpw, line 47: Is there any functional reason to moving these into describe_aggrs?  I think this must be why min/max types were changed to `anytype` (and maybe they should not be `str` here in the original code), so maybe this is not necessary.

saulpw (changes requested):
Thanks for putting this together, @midichef. I've been looking at it the past few days and I have a few comments and questions.  I think you have a keen understanding of what we would want the behavior to be, and I'd like to take this opportunity to think deeply and clean up some things around aggregators, so that we don't make them a lot messier to work with and reason about.

midichef:
I just force-pushed, but these commits are just the original set of commits, moved to the latest develop branch, fixing some merge conflicts. I didn't change anything substantial.

I'll respond to saulpw's review comments soon.

visidata/features/describe.py
midichef: I think the reason for doing this was  that`stdev` needed to be date-aware, so I couldn't use the standard `statistics.stdev` function. Instead I was using the `stdev` aggregator. While I was replacing standard library 
functions with aggregators, I moved `min`/`max`/etc too.

midichef:
It's been several months since I made these changes. I've forgotten the details of why I made some choices here.

Aggregators for dates is a pretty uncommon use. Since I still have a backlog of more important features and bugs I'd like to tackle, I'm closing this PR as not important enough to work through.

PR #2381, opened 2024-04-15
[mouse-] click on error note to run error-cell

midichef:
This is a UI change, to let users click on the `!` symbol to run `error-cell`. The goal is to make the `error-cell` command easier to discover, and easier to get to without remembering the shortcut `zCtrl+E` or the command name.

A drawback is that users may accidentally click on the `!`, thinking they're clicking on a cell. After `error-cell` runs they may not quite understand what has happened. To make it clearer, I've created a new kind of sheet, `ErrorCellSheet`, with its own guide to explain where they've come to, and how to get out. And in that sheet, the column name will be `cell_error`, which is more specific than the current `text`.

What do you think?

### DIFF
--- a/visidata/column.py
+++ b/visidata/column.py
@@ -368,7 +368,7 @@ class Column(Extensible):
                     dispval = options.disp_error_val
                 return DisplayWrapper(cellval.val, error=exc.stacktrace,
                                         text=dispval,
-                                        note=options.disp_note_getexc,
+                                        note=f'[:onclick error-cell]{options.disp_note_getexc}[:]',
                                         notecolor='color_error')
             elif typedval.val is None:  # early out for strict None
                 return DisplayWrapper(None, text='',  # force empty display for None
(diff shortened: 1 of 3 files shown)


saulpw:
I agree with making the `!` symbol clickable.  What about wrapping it with `[:onclick error-cell]`, instead of the custom code?  We'd have to update sheets.py:881 to use clipdraw instead of addstr, but then it would be configurable via an option (currently `note_getter_exc` but I think that should be renamed to `disp_note_getexc`).  Also currently clipdraw auto-underlines clickable cells, which doesn't look good, but we could change it so that it doesn't.

saulpw:
Okay, I renamed `options.note_*` to `options.disp_note_*` and made it use clipdraw, as mentioned above.  So this PR can be changed to be the changes to the error sheet help text, and updating the options.  I do like the idea of making the notes clickable.

midichef:
Okay, I moved to the use of `[:onclick]`. I left underlining in, for the time being.

saulpw (approved):
Thanks @midichef!  We'll deprecate `view-cell` and remove the underline from (some) clickables separately.

anjakefala:
Thanks @midichef!

I also included clickable type exceptions in this PR. =)

PR #2394, opened 2024-04-28
Add bindings for hide/show current cursor column in columns sheet.

blais:
Minor change adding a way to toggle to show a single column in the Columns metasheet.
I'm finding myself a bit annoyed having to (gv) show all and then rehide some of the columns.

(Feel free to change the binding overrides to be more in line with the holistic choices across the tool.
Thank you for this wonderful tool! :-) )

### DIFF
--- a/visidata/metasheets.py
+++ b/visidata/metasheets.py
@@ -24,6 +24,8 @@ For example, edit the _{sheet.cursorCol.name}_ column to **{sheet.cursorCol.form
 Some new commands on this sheet operate on all selected columns on the source sheet:
 
 - {help.commands.hide_selected}
+- {help.commands.hide_current}
+- {help.commands.show_current}
 - {help.commands.key_selected}
 - {help.commands.key_off_selected}
 - {help.commands.type_int_selected}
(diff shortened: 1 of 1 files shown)


saulpw:
Thanks for putting together this contribution, @blais!  I'd like to know more about when you have this issue, and to think about other possible solutions.

First off, did you know that you can `e`dit the column width on the Columns Sheet to do the same?  It's a few extra keystrokes, which is annoying, but consistent with how VisiData works already.  And also, did you know that if you hide a column accidentally, you can use `Shift+U` to undo the hiding without going to the Columns Sheet?

I'm open to some of what you propose here.  Particularly since using aggregators makes little sense on the Columns Sheet, so rebinding `+` shouldn't be a problem.  But I'm worried that it might be confusing, since that binding doesn't work like that in any other area of the program.  Also, if we change the `-` keybinding on the Columns Sheet to hide/show on the source sheet, should other column-oriented commands, like `_` for column resizing and `#` to type as integer, work the same way?

saulpw:
Well, @blais, your PR caused me to think about the problem of unhiding hidden columns, and I wound up implementing something (just pushed) that I hope will make it easier in general.  Thank you so much for your PR; I'm going to close it for now, and if you have other ideas I would love if you opened another PR or issue.

blais:
Thank you Saul, much appreciated, I'll definitely update and check it out.
(Definitely more ideas for more PRs.)

blais:
It's a nice touch, the hidden columns on the right, and g- from the columns sheet.
However, to complete the picture, one should be able to also unhide the columns as well.
It would be convenient if
- from the main / data view, another binding ("+" on the hidden column?) would unhide
- from the columns sheet, one should have an equivalent to "g-" to unhide (g+?) them
Thank you,

saulpw:
> from the main / data view, another binding ("+" on the hidden column?) would unhide

`_` will resize the hidden column to be visible again.

As for the Columns Sheet, let me consider what to use as the inverse of `g-`.  `g+` already adds an aggregator to the selected "rows"/source columns.  Maybe `gz-`?  That's available but also kind of gross :)

blais:
`_` is an awesome choice for that (now a new binding), thanks!
`g_` for the columns sheet might be a great analogue, but means something else in the context of that sheet. Hmmm.

PR #2397, opened 2024-04-30
[issues] show latest vd version to bug reporters

midichef:
Recently there were a couple of bug reports that were for older versions of visidata, that had already been fixed. Maybe a note in the bug report template will help people find the newer release?


saulpw:
I'd love to have people file better bug reports, but I'm not sure the people who might benefit from this would read it--many people just leave the template text in their issues verbatim, even though it makes no sense.  And I believe we'd have to update the version number manually in the issue template for every release.

Anyway, I've updated the template in a couple of ways; if we're going to ask, I'd like people to check against the develop branch, even more than the latest release.  Thanks for spurring us on here.

PR #2402, opened 2024-05-04
[build-] add sqlalchemy to vdsql requirements.txt

midichef:
This seems to fix the tests breaking for vdsql right now:

```
Run cd visidata/apps/vdsql && ./test.sh
--- tests/dup-limit.vdj
saul.pw/VisiData v3.1dev
opening tests/dup-limit.vdj as vdj
opening tests/business.sqlite as sqlite
ModuleNotFoundError: No module named 'sqlalchemy'
no "キcustomers" row on business
replay aborted during open-row
Error: Process completed with exit code 1.
```


saulpw:
sqlalchemy was removed in ibis 9.0, so we don't need to import sqlalchemy to handle missing backends anymore.

PR #2407, opened 2024-05-16
[path-] fix undercounted progress for multibyte chars

saulpw:
I tried moving all the functionality from #2323 into a TextProgress class to keep it out of the path.py, and I simplified it some as I didn't want to join all the lines together.  This probably regresses some performance issues that @midichef had already addressed; let's call out the perf cases we should test with.  Ideally they could become part of a build load/save perf report.

### DIFF
--- a/visidata/path.py
+++ b/visidata/path.py
@@ -8,7 +8,7 @@ from urllib.parse import urlparse, urlunparse
 from functools import wraps, lru_cache
 
 from visidata import vd
-from visidata import VisiData, Progress
+from visidata import VisiData, Progress, TextProgress
 
 vd.help_encoding = '''Common Encodings:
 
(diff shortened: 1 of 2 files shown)


midichef:
Okay, I'll do some testing for loading speed.

Profiling has become easier now that #2369 has made replay behavior faster and more consistent. Before that, the loading time for a large file (using replay) varied quite a bit, with a standard deviation of 2-5% of the mean loading time. Testing a change used to require a few dozen runs to detect a few percentage points of improvement in loading time, which took about a half hour. After #2369, the standard deviation is 0.5-1%, and the individual runs are 30-80% faster on my system.

midichef:
Update, when I wrote:
> the standard deviation [of the loading time for a large file has decreased, from 2-5% to]
> 0.5-1%, and the individual runs are 30-80% faster on my system.

I was wrong. I spoke after too little testing. The standard deviation is still in the range of a few percent.

PR #2411, opened 2024-05-22
[viewtsv] fix missing import and argument to open_tsv

tavisrudd:
Fixes:
`NameError: name 'VisiData' is not defined`
and 

```
    run(*(open_tsv(Path(fn)) for fn in sys.argv[1:]))
          ^^^^^^^^^^^^^^^^^^
TypeError: open_tsv() missing 1 required positional argument: 'p'
```

Happy to sign a CAA or for you to just apply the patch under your name. Thanks for a great tool.

### DIFF
--- a/bin/viewtsv.py
+++ b/bin/viewtsv.py
@@ -3,7 +3,7 @@
 # as plugin: `from viewtsv import open_tsv` in .visidatarc
 # standalone: `viewtsv.py <tsv-files>'
 
-from visidata import Sheet, ColumnItem, asyncthread, options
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Ah, thank you @tavisrudd !  `viewtsv` is an older example of how to make a separate application that uses the visidata library, and we've been trending away from that use case for awhile now, but it's still good to keep this example working.  I'm curious what you're doing with viewtsv yourself, or if you just tried to run it out of curiosity and that's how you discovered the missing imports.

tavisrudd:
Just running it out of curiosity while reading the api docs

PR #2412, opened 2024-05-28
[loaders-] handle exceptions in filetype guess

midichef:
Exceptions during filetype guessing are usually invisible, and only show up when running `vd --debug`. This PR handles a couple of these usually invisible exceptions, to clean up the debug status log.

Empty files cause empty messages for a few filetypes: `echo -n '' |vd --debug`
```
guess_jsonla: 
guess_json: 
guess_csv: 
```

And a JSON file causes an error in the JSONLA guesser: `echo '[\n{"id": 1}\n]' |vd --debug`
```
guess_jsonla: Expecting value: line 2 column 1 (char 2)```

### DIFF
--- a/visidata/loaders/csv.py
+++ b/visidata/loaders/csv.py
@@ -16,7 +16,10 @@ vd.option('safety_first', False, 'sanitize input/output to handle edge cases, wi
 def guess_csv(vd, p):
     import csv
     csv.field_size_limit(2**31-1)  #288 Windows has max 32-bit
-    line = next(p.open())
+    try:
+        line = next(p.open())
+    except StopIteration:
+        return
     if ',' in line:
         dialect = csv.Sniffer().sniff(line)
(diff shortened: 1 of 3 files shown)


saulpw:
We don't necessarily have to log any errors, even in debug output; maybe should we just remove that instead of trying to handle all exceptions in each guesser?  I only put it there because it seemed like it would be handy for when someone was trying to debug why the file type wasn't being guessed correctly for some file, or when writing or updating a guesser.  But maybe it's just noisy for the vast majority of times we use `--debug`.

midichef:
The error log is useful for me at the moment. I'm using `--debug` to check for errors in file guessing, as I try to make a binary version of `RepeatFile`. That's how I noticed the errors in these corner cases. So it is handy for now.

saulpw (approved):

PR #2417, opened 2024-05-31
[aggr-] allow ranking rows by key column

midichef:
This PR adds a `rank` aggregator that returns a list, and a command `addcol-rank`, which adds a new column with the rank of each row. Ranks are calculated by comparing key columns.

It also fixes a bug in `memo-aggregate` where long output takes an extremely long time to show up in the statusbar.
For example:  `seq 1222333 |vd -`, then `z+` `list`. After the list is calculated, visidata will get stuck for many seconds showing `processing…`, because it's very slow to run `format()` on a long sequence.

I think it's worth having an aggregator for rank, and [the need for a simpler solution than the current method has come up before](https://github.com/saulpw/visidata/discussions/2280). On the other hand, I know part of Visidata philosophy is that it's not a spreadsheet. How do people feel about having a rank aggregator?

Also, in its current form, the rank aggregator will give errors when comparing key columns with different types across 2 rows:
```
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/aggregators.py", line 169, in rank
    keys_sorted = sorted(((rowkey, i) for i, rowkey in enumerate(keys)), key=_key_progress(prog))
TypeError: '<' not supported between instances of 'float' and 'list'
```
What's the standard way to handle sorting mixed types for Visidata?


saulpw:
> What's the standard way to handle sorting mixed types for Visidata?

The standard way is to convert the column into a known type, and then anything that can't be converted (errors and nulls) become `TypedWrapper`s which are sortable with any type.  Does that work acceptably here too?

midichef:
Yes, that seems like it should work. Should the rank aggregator pick the known type, and if so, which one? Or is it the user who should convert the column?

saulpw:
Since it's not obvious which type to pick, the user can convert the column.

visidata/aggregators.py
```diff
@@ -142,11 +143,39 @@ def __init__(self, pct, helpstr=''):
     def aggregate(self, col, rows):
         return _percentile(sorted(col.getValues(rows)), self.pct/100, key=float)
 
+class RankAggregator(Aggregator):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
saulpw: A pass-through `__init__` is unnecessary and will happen by default.

visidata/aggregators.py
```diff
@@ -142,11 +143,39 @@ def __init__(self, pct, helpstr=''):
...
+        def identity(val):
+            prog.addProgress(1)
+            return val
+        return identity
+
+    keys = map(sheet.rowkey, rows)
+    with Progress(gerund='grouping', total=sheet.nRows) as prog:
+        keys_sorted = sorted(((rowkey, i) for i, rowkey in enumerate(keys)), key=_key_progress(prog))
+    # group elements by rowkey
+    with Progress(gerund='ranking', total=sheet.nRows) as prog:
```
saulpw: Using these Progress objects separately in serial will reset the progress meter.  You only need one of them, with a `total=3*sheet.nRows` (since there are 3 steps).

If you want to keep the gerunds to indicate the various steps, you can have the first one be 'outermost', and then then other Progress() within that scope with different gerunds (and no total), and only addProgress on the outer one.

visidata/aggregators.py
```diff
@@ -142,11 +143,39 @@ def __init__(self, pct, helpstr=''):
     def aggregate(self, col, rows):
         return _percentile(sorted(col.getValues(rows)), self.pct/100, key=float)
 
+class RankAggregator(Aggregator):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def aggregate(self, col, rows):
+        if not col.sheet.keyCols:
+            vd.error('ranking requires one or more key columns')
```
saulpw: Is this actually true?  I could see row number being used if there are no key columns.  If we remove this check, does that just work?

visidata/aggregators.py
```diff
@@ -276,10 +307,21 @@ def _fmt_aggr_summary(match, row, trigger_key):
...
+        vd.fail('ranking requires one or more key columns')
+    c = SettableColumn(name=f'{sheet.name}_{list_agg}', type=int)
+    rows = vd.aggregators[list_agg].aggregate(col, col.sheet.rows)
+    sheet.addColumnAtCursor(c)
+    c.setValues(sheet.rows, *rows)
+
 Sheet.addCommand('+', 'aggregate-col', 'addAggregators([cursorCol], chooseAggregators())', 'Add aggregator to current column')
 Sheet.addCommand('z+', 'memo-aggregate', 'cursorCol.memo_aggregate(chooseAggregators(), selectedRows or rows)', 'memo result of aggregator over values in selected rows for current column')
 ColumnsSheet.addCommand('g+', 'aggregate-cols', 'addAggregators(selectedRows or source[0].nonKeyVisibleCols, chooseAggregators())', 'add aggregators to selected source columns')
+Sheet.addCommand('', 'addcol-rank', 'addcol_list_aggr(cursorCol, "rank")', 'create new column ranking rows by their key columns')
```
saulpw: Is this specific to "rank"?  What happens if we apply a different list aggregator?  Should this take an input?  (the default value could be "rank" to make it the easiest option).

Also what happens with non-list aggregators?  This could be an instant `SUM(curcol) GROUP BY keycols` which I think would be a major hit!

saulpw (changes requested):
I love what this is adding, and I think with a few tweaks it would be even more powerful!

visidata/aggregators.py
midichef: Using row number for ranking does the same thing as `addcol-incr`. Would it work to make `rank()`, when there are no key columns, act as if all columns are key columns?

If so, should it use only visible columns, or all columns including hidden ones?

visidata/aggregators.py
midichef: > Is this specific to "rank"? What happens if we apply a different list aggregator? Should this take an input? (the default value could be "rank" to make it the easiest option).

I agree, it is more useful to make the command take an input. Right now the only other list aggregator is `list`, and using that would just be equivalent to `=curcol`. Taking more aggregators would be excellent for the future, and for user-defined aggregators. So I've pushed a commit for that.

> Also what happens with non-list aggregators? This could be an instant `SUM(curcol) GROUP BY keycols` which I think would be a major hit!

This is a great idea! I've pushed a commit for that.

There are two kinds of aggregators, list and non-list. To distinguish them and allow the new column, I've added a field to each aggregator: `listttype`. It specifies what kind of Visidata type the column should have.

Also, by allowing both kinds of aggregators in the same command, what should the behavior be when there are no key columns? For `rank`, we discussed above that perhaps we should use row number, or treat the entire row as a key. But for non-list aggregators, that do `SUM/MIN/MAX/COUNT(curcol) GROUP BY keycols`, neither of those behaviors makes sense. So for now I've stuck with giving an error. I think that helps tutor new users, if they blindly try the command `addcol-aggregator` without knowing the sheet should have key columns.

This PR now creates some conflicts with #2380. I'll focus on getting this PR mergable first, then after it is accepted, I will rework #2380, which I agree is a flawed PR.

Once this PR is in shape to be added, I'll squash it to a smaller set of commits, and fix a tiny conflict with the current develop branch, and rebase it.

visidata/aggregators.py
midichef: My questions above are obsolete, see https://github.com/saulpw/visidata/pull/2417#discussion_r1681972604 for newer context.

midichef:
There are two kinds of ranking operations people may want.
1. keycol-based rank within sheet:  what is the rank of this row, vs. all rows in the sheet, ranking by the value of its key columns? In this example, the key column is `keycol`, and the current column `col` is ignored:
```
keycol	col	keycol_sheetrank
1	10	1
1	20	1

2	60	2
2	50	2
2	30	2
```
2. column-based rank within group: when grouping the rows by key columns, what is the rank of this row, within its group? The current column determines the rank. In this example, the current column is `col`:
```
keycol	col	col_grouprank
1	10	1
1	20	2

2	60	3
2	50	2
2	30	1
```
What is a good name for these two aggregators? `sheetrank` and `grouprank`? Or maybe `rank_key` and `rank_col`?
Any suggestions?

midichef:
Okay, I implemented a command that adds a column and applies an aggregator to rows after grouping them by key columns. It's `addcol-aggregate`.

To get this to work with list aggregators, I made a new class `ListAggregator` for aggregators that return lists. Their most common use would be with `addcol-aggregate`. Right now the only two ListAggregators are `list` and `rank`.

I also tried making a `sheetrank` aggregator, but it's too different from normal aggregators. Normal aggregators apply to a column, but sheetrank is more for the sheet. So I broke it out into a separate command, `addcol-sheetrank`.

I'm a bit unsure about the new behavior of the `list` aggregator when used with `addcol-aggregate`. Right now, if the input column has cells with Exceptions, they show up in the new column. But the error text shows up on the display, it's not hidden behind an error note. `!`. So I could use guidance on a couple of issues here:
1) Should these Exceptions be passed through by the list aggregator, or should they be translated to null?
2) If they should be passed through the aggregator, how do I make them look/behave like the original cell with an exception?
The relevant code is here:
https://github.com/saulpw/visidata/blob/c6c608ee64216d097107540fab8fb34c25e0f8ba/visidata/aggregators.py#L120
To see it in action, `vd sample_data/test.jsonl`, then `addcol-aggregate` `list`. The `key1` and `key1_list` columns ought to look the same, but the fourth cell in `key1_list` reads `Expecting ':' delimiter: line 1 column 34 (char 33)` instead of empty.

midichef:
There is one detail about the grouping in `addcol-aggregate`. If the key column holds multiple cells with null, all nulls are grouped together as one group of rows. But if the key column holds multiple error cells, each error cell forms its own unique group of 1 row, even if all the errors have the same traceback text. I didn't design this, it's just how it behaved on sorting. Does that error cell treatment sound reasonable?

saulpw (changes requested):
Can we move most of this into `features/addcol_sheetrank.py`?  It's a fair amount of code and I'd like to make it self-contained as much as possible.

midichef:
Okay, I moved the `addcol-sheetrank` code into a `features/` file.

I had to move the `RankAggregator` class there too, because it relies on functions that moved with `addcol-sheetrank`. Because the new file is not just for the `addcol-sheetrank` functionality, I named it `features/rank.py` instead of `features/addcol_sheetrank.py`.

One notable consequence of this move is that the new `features/rank.py` file modifies `vd.aggregators` outside of `aggregators.py`:
https://github.com/saulpw/visidata/blob/ec284462ad251ad1a43e5b8f4fa1daf0fdff2184/visidata/features/rank.py#L41

visidata/aggregators.py
```diff
@@ -288,10 +371,34 @@ def _fmt_aggr_summary(match, row, trigger_key):
...
+@asyncthread
+def addcol_aggregate(sheet, col, aggrnames):
+    for aggrname in aggrnames:
+        aggrs = vd.aggregators.get(aggrname)
+        aggrs = aggrs if isinstance(aggrs, list) else [aggrs]
+        if not aggrs: continue
+        for aggr in aggrs:
+            rows = aggregate_groups(sheet, col, sheet.rows, aggr)
+            if isinstance(aggr, ListAggregator):
+                t = aggr.listtype or col.type
```
saulpw, line 386: Do we need a separate `listtype`?  Seems like we could just use the same `aggr.type` in both cases, and remove this `isinstance` (which is usually a code smell for me).

saulpw:
Okay, this one is quite old and quite big at this point!  I gave it another pass through, and my remaining questions have to do with `aggr.listtype` (in review comment), and the `vd.aggregate_list` function, which seems like it might be an unnecessary and confusing API function.  I've asked @anjakefala to look over the behavior too.   It'll be nice to finally get this one merged; thanks for your patience on this!

midichef:
Let me look into why I added a custom `stdev` function with b41afba and then removed it in ec28446. That's probably an oversight by me.

anjakefala:
@midichef I think your PR successfully meets your goals. This is such thoughtfully considered work.

Once you have your fnal changes in, I am going to add an update to the guide in this PR, with documentation.

I have one question for my own clarity: what is rank's intended behaviour when there are multiple key columns?

Say

keycol	keycol_2	Item
1       	10	          Pen
1	        20        	  Pencil
2       	60	          Book
2      	50	           Pen
2	        30	           Book

midichef:
@anjakefala 

> I have one question for my own clarity: what is rank's intended behaviour when there are multiple key columns?
> 
> Say
> 
>```
>keycol keycol_2 Item
>1 10 Pen
>1 20 Pencil
>2 60 Book
>2 50 Pen
>2 30 Book
>```

When there are multiple key columns, `rank` will group each row based on all the key columns, put into a tuple. So in this case, there are 5 unique tuples and therefore 5 groups: `(1,10)`, `(1,20)`, `(2,60)`, `(2,50)`, `(2,30)`. Each row is ranked `1` in its group.

But if the key columns were not unique, the results would be more interesting. Here the groups are `(1,10)` and `(2,30)`.
```
keycol keycol_2 Item Item_rank
1 10 Pen 1
1 10 Pencil 2
2 30 Book 1
2 30 Pen 2
2 30 Book 1
```

visidata/aggregators.py
midichef: The way list aggregators work now, there is a need for two distinct types, `type` and `listtype`. `type` is for the result of the aggregator. For example, this is used by `memo-aggregate`. That's why `type` is `anytype` for `ListAggregators`. This type would be used whenever we want to hold the entire result (a list) in a cell.

But we also need a separate type for the elements of the list. This is for when the aggregator result goes in a column, like for `addcol-aggregate`, where each cell holds not the result itself, but an element of the list result.

If I try to get rid of one or the other types, I run into problems. For `RankAggregator`, if I get rid of the `listtype=int` switch to `type=int`, I get an error in the statusbar for `z+` `rank`:
'''
text_rank=int() argument must be a string, a bytes-like object or a real number, not 'list'
'''
But if instead I make `RankAggregator` use `type=anytype`, the column added by `addcol-aggregate` `rank` does not get the type `int`.

The need for two types is awkward. And I see your point about `isinstance` being a code smell. (That is a helpful heuristic, and I'll use it in the future.) It's accurately pointing out strain in the design: most aggregators produce a single value, list aggregators produce a list.

Maybe `rank` should not be an aggregator. It's unlikely people want a list object holding the ranks. Most people want a column holding the ranks. What if we replace `addcol-aggregate`+`rank` with an equivalent command `addcol-grouprank` (in addition to the existing `addcol-sheetrank`)? And we would reserve `addcol-aggregate` for finding group values like `sum`, `mean`, `median`, [as you suggested earlier](https://github.com/saulpw/visidata/pull/2417#discussion_r1660530166). What do you think?
(I would also consider changing the name `addcol-aggregate`. Maybe to `addcol-group-aggregate`.)

midichef:
On review of this feature, I found that after several months, I found it difficult to make sense of the command names I introduced. The three operations added in this PR as it stands now are:
1. `addcol-sheetrank`
2. `addcol-aggregate` then `rank`
3. `addcol-aggregate` then other aggregators like `sum` or `mean`.

Operations 1 and 2 do not have suggestive enough names. If I couldn't keep them straight in my mind, nobody else will either. Any suggestions on better names?

midichef:
Okay I submitted two minor changes, https://github.com/saulpw/visidata/pull/2417/commits/20ba354e957f856c92a028cce5a6de75c06f78ba (which corrects an accidental reversion of a line from a previous commit) and https://github.com/saulpw/visidata/pull/2417/commits/9ed9f4b76eac9465ef89c69e4904879b4280061a (which adds a few errors/warnings as guardrails, since I expect these operations to be difficult for users to understand at first).
Once these are approved, can you squash them into https://github.com/saulpw/visidata/pull/2417/commits/d35738b2ad1b1269f34e530e6c2d365632e07060 ?
I would do it myself now, but I want to make it easy to see that I have only made minor changes today.

anjakefala:
@midichef Yes, this is why I was going to update the docs as part of the merging of this PR. I think it's really hard to understand what these commands are solely from their name and one-line description (which isn't your fault, these are inherently complicated concepts). 

But, I will mull over if I have better names at hand....Maybe a good question for the VisiData discord.

midichef:
I fixed some merge conflicts with the current develop branch, and rebased all previous fixes onto the latest develop. I also did the squashing that I had requested above. These new commits don't change the substance of any code from the previous commits.

midichef:
Now I've added https://github.com/saulpw/visidata/pull/2417/commits/f5210e5e7080d50243951d44d88d499513d2f74d. 
When doing within-group rankings via `addcol-aggregate` `rank`, the rank comes after ordering the column elements in ascending order. This commit instead uses the sort order on the column. If it's descending, the elements will be ranked in descending order.

I believe it's the behavior users will expect when the column is already sorted in descending order. That's how I realized I wanted this feature. I had a column sorted descending, and ranked it, and was surprised when it was ranked ascending instead.

midichef:
I still have a little more to add to this PR. Now that the `rank` operator uses the sort order of the current column, for consistency I'll make `addcol-sheetrank` also use the sort order of the key columns. And I need to add `Progress` updates while ranking is underway.

midichef:
I've added the pending features. This PR is done in terms of implementing features.

What remains to do is to come up with intuitive names and explanations for `addcol-aggregate` `rank` vs `addcol-sheetrank`. I don't have any ideas on how to better do that.

Also, I'm not sure why the ci-build tests are failing, as the failing test `tests/load-http.vd` passes for me on my system, but is failing here on Github.

anjakefala:
@midichef It might be a one-off CI blunder. Though, the test output is supposed to show when the error code is 1 :sob: I'll have to investigate that.

saulpw (approved):

midichef:
I've got one more small change queued up to add to this PR: better error handling if the sort fails during ranking. It's a minor change and shouldn't require much review.

anjakefala:
I changed `addcol-sheetrank` to `addcol-rank-sheet` to better fit the naming scheme. =) 

I do have a guide I'm working on, but my brain has been a bit cloudy, and I don't want to block this PR on me finishing it. So I'll merge for now, and push the guide when it's ready.

Thanks for all your work @midichef!

PR #2421, opened 2024-06-10
[save-] handle empty delimiter when safety_first

midichef:
With `--safety-first`, saving a sheet fails when using` '\x00'` as a delimiter.

`echo "col\none\0two" |vd --safety-first --delimiter= -f tsv -` then execute `save-sheet`:
```
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/save.py", line 18, in safe_trdict
    ord(delim): vs.options.tsv_safe_tab, # \t
TypeError: ord() expected a character, but string of length 0 found
```
This PR changes the behavior to fail, with an explanatory message, when the delimiter is `'\x00'`, `'\r'`, or `'\n'`.

### DIFF
--- a/visidata/save.py
+++ b/visidata/save.py
@@ -13,12 +13,15 @@ def safe_trdict(vs):
     'returns string.translate dictionary for replacing tabs and newlines'
     if vs.options.safety_first:
         delim = vs.options.delimiter
-        return {
+        trdict = {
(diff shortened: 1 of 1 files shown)


saulpw (approved):
At some point I'd like to rework `safety_first`, since it's used in only a few places.  But this is fine for what it is now.

Issue #2357, opened 2024-03-23
[main] command-line argument +a:b is parsed as row:col, contradicting man page


midichef:
**Small description**
`vd sample_data/benchmark.csv +:0:2` opens the sheet with the cursor at the third column.

**Expected result**
The documentation suggests it should open at the third row, not column:
https://github.com/saulpw/visidata/blob/581cdf12b28278aea87ce09f7c32bf8ae1be2f0f/visidata/man/vd.txt#L10

What's the correct order here: +col:row or +row:col? I've got a PR to fix some bugs in handling the initial row/columns passed in the `+a:b` argument. Should I match the man page and docstrings which say col:row? Or the existing behavior and the tests in `features/slide.py`, which do row:col?

**Additional context**
saul.pw/VisiData v3.1dev

anjakefala:
@aborruso, this feature is your request! Can you please tell midichef your preferences for its syntax? =)

anjakefala:
(Nice catch, midichef)

aborruso:
Tomorrow I try to remember what the request was and then I answer you. Thank you very much

aborruso:
Tomorrow I try to remember what the request was and then I answer you. Thank you very much

anjakefala:
When I went to look up the original request, it seems like you cared about selecting the top-level sheet from the CLI! https://github.com/saulpw/visidata/issues/214#issuecomment-552174553 Not selecting a specific row/column. It wasn't you who originally proposed the orderl. =)

Saul designed it originally `row:col`. I think he chose that because VisiData is [row-oriented](https://www.visidata.org/blog/2020/row-major/), and that makes sense to me. I possibly accidentally wrote the documentation as `col:row` because I often think in a column-oriented fashion. Since VisiData is row-oriented, I vote we keep the existing behaviour, and fix the documentation. I can go ahead and fix it.

anjakefala:
I thought about it some more.

i think probably keeping it row:col is the default approach, because we might silently break folks CLI usage with an abrupt change. 

But since you did start this conversation midichef, I would love to hear yours, aborruso's and anyone else's preferences. Personally `col:row` felt more intuitive to me. If this is true for lots of other folks, it possibly makes sense to change it.

midichef:
For a single argument, `+row` makes sense. It matches the behavior of other programs, like `vim +3` and `tail +3` (for tail in GNU coreutils). Which is probably where it came from.

[...]

PR #2358, opened 2024-03-23
[docs] fix syntax for CLI argument +a:b

anjakefala:
Closes #2357 

@midichef and @aborruso please review!

PR #2425, opened 2024-06-18
[main-] handle position args as +sheet:subsheet:col:row

midichef:
Two bug fixes:
3ea655baacc1ebe7ac2c4488d938aadb26dd8b9e - fixes setting column position in sheet
Currently the starting column cannot be set, only the row.
repro command: `vd sample_data/sample.tsv sample_data/StatusPR.csv +2:2`
This occurs because the column position changes too early, before the sheet is loaded.

1aa2d9808d79d89810bf1837d39b5f89b71d83f5 - fixes setting column position in subsheet
A similar problem, changing position before the subsheet is loaded.
repro: `vd ~vdsrc/sample_data/employees.sqlite +:emp:5:2`

And then some functionality changes:
d16743d96114949aafb6f63b507ed01125f31406 - change arg parsing, from row:col to col:row
Strings of the form a:b are parsed as row:column. This commit changes it to column:row.
[Previously, I discussed some pros and cons](https://github.com/saulpw/visidata/issues/2357#issuecomment-2016659308) of this change.
This is technically a breaking change, but as I noted in that comment, the form `+a:b` has not worked since 2.9 so it's not likely that anyone is currently using that syntax. However, if they are using `+:a:b` with the extra colon, this will indeed be a breaking change for them.
repro:  `vd ~vdsrc/sample_data/sample.tsv +:1:4`
(note the `:` before the `1`. It's needed (on `develop`) to demonstrate the column-positioning bug, until the 2 commits above are applied.)

0a8f6ee3ca4b1b62fb514fbec7eb29e4c3b9b20b - handling args of form `+a:b` vs `+:a:b`.
Comments in the code suggest that '+col:row' should apply to only the last sheet on the list, vs. adding a colon, '+:col:row', which should apply to all sheets:
https://github.com/saulpw/visidata/blob/c01c2b5fe506632590c8d7f5bcf23b1afe9c1985/visidata/main.py#L90
(The suggestion comes from the phrase `Empty sheetstr in startsheets`. An empty `sheetstr` happens only when `arg.split(':')` has more than 2 items and `pos[:-2]` is `''`, which happens when `arg` follows the pattern `+:a:b`:
https://github.com/saulpw/visidata/blob/c01c2b5fe506632590c8d7f5bcf23b1afe9c1985/visidata/main.py#L105
Currently the behavior is the reverse: +col:row applies to all sheets.
repro: `vd sample_data/sample.tsv sample_data/StatusPR.csv +:2:2`
and look at the cursor position on each sheet.
This commit changes `+:a:b` to apply to all sheets, and `+a:b` to apply only to the last sheet.

da52ac3b7b72243e25bfd4cb8e85bb95dd4eb9be - improve error for invalid sheets
Attempting to index a sheet that does not exist, as with:
repro:  `vd sample_data/sample.tsv +10:0:0`
gives a traceback with `IndexError: list index out of range`.
This commit handles the error and shows a status message of `no sheet "10"`.

a0d4913a94d7147975c28eb3ec3ceb8014addc66 - handling subsheet names that are integers
For consistency with sheets/rows/cols, this commit allows indexing subsheets with integers, like:
[numeric-subsheet-names.vds.txt](https://github.com/user-attachments/files/15883000/numeric-subsheet-names.vds.txt)
repro:  `vd -f vds numeric-subsheet-names.vds.txt +0:3:0:0`
This is also a change that breaks current behavior, for a subsheet that happens to have a name that is an integer. Such a subsheet would become unable to opened by using its name, as with sheets `5`, `6`, `7`, and `8` in `numeric-subsheet-names.vds.txt`; `+0:5:0:0` currently works, but after this commit only `+0:0:0:0` would work.

Thoughts? I'm not so worried about breaking changes, because I don't think the `vd +` feature is heavily used.

Just to write out all the cases, here's what the args would mean after all these commits:
```
    +2         = row 2
    +1:        = last sheet; col 1
    +:1:       = all sheets; col 1
    +:2        = last sheet; row 2
    +1:2       = last sheet; col 1, row 2
    +:1:2      = all sheets; col 1, row 2
    +0:1:2     = sheet 0; col 1, row 2
    +:a:1:2    = last sheet, subsheet a; col 1, row 2
    +0:a:1:2   = sheet 0, subsheet a; col 1, row 2
    +0:a:b:1:2 = sheet 0, subsheet a from sheet 0, subsheet b from sheet a; col 1, row 2
    +0:0:1:2   = sheet 0, first subsheet of sheet 0, col 1, row 2
    +1::       = 2nd sheet, col 0, row 0
    +-2::      = 2nd sheet from last, col 0, row 0
```

### DIFF
--- a/visidata/main.py
+++ b/visidata/main.py
@@ -93,6 +93,9 @@ vd.optalias('r', 'dir_depth', 100000)
 @visidata.VisiData.api
 def parsePos(vd, arg:str, inputs:'list[tuple[str, dict]]'=None):
     '''Return (startsheets:list, startcol:str, startrow:str) from *arg* like "+sheet:subsheet:col:row".
+    The elements of *startsheets* are identifiers that pick out a sheet, either
+    a) a string that is a sheet name, or
+    b) a list of strings that are sheet names or numbers (for row indices). For example ['1', 'sales', '3'].
     Returns an empty list for *startsheets* when the starting pos applies to all sheets.
     Returns None for *startsheets* when the position expression did not specify a sheet.
     *inputs* is a list of (path, options) tuples.
@@ -154,11 +157,18 @@ def moveToPos(vd, sources, startsheets, startcol, startrow):
         vd.clearCaches()
         # descend the tree of subsheets
         for subsheet in startsheets[1:]:
-            rowidx = vs.getRowIndexFromStr(vd.options.rowkey_prefix + subsheet)
-            if rowidx is None:
-                vd.warning(f'{vs.name} has no subsheet "{subsheet}"')
+            if subsheet and subsheet.isdigit():
+                rowidx = int(subsheet)
+            else:
+                rowidx = vs.getRowIndexFromStr(vd.options.rowkey_prefix + subsheet)
+                if rowidx is None:
+                    vd.warning(f'{vs.name} has no subsheet "{subsheet}"')
+                    return
+            try:
+                vs = vs.rows[rowidx]
+            except IndexError:
+                vd.warning(f'{vs.name} has no subsheet "{rowidx}"')
                 return
-            vs = vs.rows[rowidx]
             if not isinstance(vs, BaseSheet):
                 vd.warning(f'row {subsheet} for subsheet is not a sheet')
                 return
@@ -171,16 +181,17 @@ def moveToPos(vd, sources, startsheets, startcol, startrow):
     if startrow:
         for vs in sheets:
             if vs:
+                if startrow.isdigit():  # treat strings that look like integers as indices, never row keys
+                    startrow = int(startrow)
                 vs.moveToRow(startrow) or vd.warning(f'{vs} has no row "{startrow}"')
 
     if startcol:
         for vs in sheets:
             if vs:
+                if startcol.isdigit():  # treat strings that look like integers as indices, never column names
+                    startcol = int(startcol)
                 if not vs.moveToCol(startcol):
-                    if startcol.isdigit():
-                        vs.moveToCol(int(startcol)) # handle indexing by column number
-                    else:
-                        vd.warning(f'{vs} has no column "{startcol}"')
+                    vd.warning(f'{vs} has no column "{startcol}"')
 
 def main_vd():
     'Open the given sources using the VisiData interface.'
--- a/visidata/main.py
+++ b/visidata/main.py
@@ -91,25 +91,27 @@ vd.optalias('r', 'dir_depth', 100000)
 
 
 @visidata.VisiData.api
-def parsePos(vd, arg:str, inputs=None):
-    'Return (startsheets:list, startrow:str, startcol:str) from *arg* like "+sheet:subsheet:col:row".  Empty sheetstr in startsheets means the starting pos applies to all sheets.'
-    startsheets, startrow, startcol = [], None, None
-
-    if ':' not in arg:
-        return (None, arg, None)
+def parsePos(vd, arg:str, inputs:'list[tuple[str, dict]]'=None):
+    '''Return (startsheets:list, startcol:str, startrow:str) from *arg* like "+sheet:subsheet:col:row".
+    Returns an empty list for *startsheets* when the starting pos applies to all sheets.
+    Returns None for *startsheets* when the position expression did not specify a sheet.
+    *inputs* is a list of (path, options) tuples.
+    '''
+    startsheets, startcol, startrow = None, None, None
 
     pos = arg.split(':')
     if len(pos) == 1:
-        startsheet = [Path(inputs[-1]).base_stem] if inputs else None
-        start_pos = (startsheet, pos[0], None)
+        startrow = arg
     elif len(pos) == 2:
-        startsheet = [Path(inputs[-1]).base_stem] if inputs else None
-        startrow, startcol = pos
-        start_pos = (None, startrow, startcol)
-    else:  # if len(pos) >= 3:
+        startcol, startrow = pos
+    else:
+        # the first element of pos is the startsheet,
+        # the later elements (if present) describe the branch to a subsheet
         startsheets = pos[:-2]
-        startrow, startcol = pos[-2:]
-        start_pos = (startsheets, startrow, startcol)
+        startcol, startrow = pos[-2:]
+    if startcol == '':  startcol = None
+    if startrow == '':  startrow = None
+    start_pos = (startsheets, startcol, startrow)
 
     # index subsheets need to be loaded *after* the cursor indexing
     vd.options.set('load_lazy', True, obj=start_pos[0])
@@ -134,8 +136,8 @@ def outputProgressEvery(vd, sheet, seconds:float=0.5):
         time.sleep(seconds)
 
 @visidata.VisiData.api
-def moveToPos(vd, sources, startsheets, startrow, startcol):
-    sheets = []  # sheets to apply startrow:startcol to
+def moveToPos(vd, sources, startsheets, startcol, startrow):
+    sheets = []  # sheets to apply startcol:startrow to
     if not startsheets:
         sheets = sources  # apply row/col to all sheets
     else:
@@ -147,10 +149,11 @@ def moveToPos(vd, sources, startsheets, startrow, startcol):
 
         vd.sync(vs.ensureLoaded())
         vd.clearCaches()
-        for startsheet in startsheets[1:]:
-            rowidx = vs.getRowIndexFromStr(vd.options.rowkey_prefix + startsheet)
+        # descend the tree of subsheets
+        for subsheet in startsheets[1:]:
+            rowidx = vs.getRowIndexFromStr(vd.options.rowkey_prefix + subsheet)
             if rowidx is None:
-                vd.warning(f'{vs.name} has no subsheet "{startsheet}"')
+                vd.warning(f'{vs.name} has no subsheet "{subsheet}"')
                 vs = None
                 break
             vs = vs.rows[rowidx]
--- a/visidata/main.py
+++ b/visidata/main.py
@@ -157,9 +157,11 @@ def moveToPos(vd, sources, startsheets, startcol, startrow):
             rowidx = vs.getRowIndexFromStr(vd.options.rowkey_prefix + subsheet)
             if rowidx is None:
                 vd.warning(f'{vs.name} has no subsheet "{subsheet}"')
-                vs = None
-                break
+                return
             vs = vs.rows[rowidx]
+            if not isinstance(vs, BaseSheet):
+                vd.warning(f'row {subsheet} for subsheet is not a sheet')
+                return
             vd.sync(vs.ensureLoaded())
             vd.clearCaches()
         if vs:
--- a/visidata/main.py
+++ b/visidata/main.py
@@ -101,13 +101,16 @@ def parsePos(vd, arg:str, inputs:'list[tuple[str, dict]]'=None):
 
     pos = arg.split(':')
     if len(pos) == 1:
+        startsheets = [Path(inputs[-1][0]).base_stem] if inputs else None
         startrow = arg
     elif len(pos) == 2:
+        startsheets = [Path(inputs[-1][0]).base_stem] if inputs else None
         startcol, startrow = pos
     else:
         # the first element of pos is the startsheet,
         # the later elements (if present) describe the branch to a subsheet
         startsheets = pos[:-2]
+        if startsheets == ['']: startsheets = []
         startcol, startrow = pos[-2:]
     if startcol == '':  startcol = None
     if startrow == '':  startrow = None
--- a/visidata/main.py
+++ b/visidata/main.py
@@ -16,7 +16,7 @@ import signal
 import warnings
 import builtins  # to override print
 
-from visidata import vd, options, run, BaseSheet, AttrDict, stacktrace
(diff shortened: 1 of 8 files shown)


midichef:
I fixed the type hint that was breaking in Python 3.8, by adding `'` quotes to make it a string:
https://github.com/midichef/visidata/blob/bcf812f0052d54f25a6c1c5d7734f9cec2e86a06/visidata/main.py#L89
I don't understand type hints enough to know what the string does, I'm just doing it because it worked in `sort.py`. 
https://github.com/saulpw/visidata/blob/c01c2b5fe506632590c8d7f5bcf23b1afe9c1985/visidata/sort.py#L50

visidata/main.py
```diff
@@ -129,26 +131,43 @@ def outputProgressEvery(vd, sheet, seconds:float=0.5):
...
 @visidata.VisiData.api
-def moveToPos(vd, sources, startsheets, startrow, startcol):
-    sheets = []  # sheets to apply startrow:startcol to
+def moveToPos(vd, sources, startsheets, startcol, startrow):
+    sheets = []  # sheets to apply startcol:startrow to
     if not startsheets:
-        sheets = sources  # apply row/col to all sheets
+        sheets = sources  # apply col/row to all sheets
+        for vs in sheets:
+            vd.sync(vs.ensureLoaded())
```
saulpw: So, the problem with this sync, is that opening a large file with a start position specified, will wait until the entire file has loaded, before it shows any interface at all.  (Prior to this change, it loads the file immediately and reports that it can't find the column).

Neither behavior is what the user wants, but the current behavior is at least immediately responsive, whereas adding a sync() here will appear to delay startup indefinitely, with no feedback.

My proposed behavior, would be to wait a second or two after loading on the sheet has started (or until loading has finished, if that's sooner), before trying to move the cursor.  This will fix the issue in most cases, since columns are usually discovered very early in the loading process.  If the column hasn't been discovered yet, it will fail as it does currently.  Alternatively, we could start a thread that keeps trying to move the cursor to the given starting position until it either succeeds, or loading has finished.

saulpw:
(This is a large combined PR, so I'm posting my comments piecemeal.)

I agree with your comments in #2358, that it should be `+col:row`.  It's unfortunately a breaking change, but as you said, no one has complained, so probably (hopefully) no one is using it in scripts.  I'd rather suffer the break now and get it right for the future, than to let it stay weirdly broken out of fear.

I'm less enthusiastic about `+row`, but I accept the argument that it's like how vim and tail work, and since we're borrowing this CLI syntax from them, that we should be consistent.

saulpw:
> Comments in the code suggest that '+col:row' should apply to only the last sheet on the list, vs. adding a colon, '+:col:row', which should apply to all sheets:

It should apply to the most recently specified sheet on the CLI, and not the last sheet.  So you could specify a different position for each sheet:

    vd foo.csv +name: bar.csv +address:

vs

    vd foo.csv +2 bar.csv +3 +:name:

such that the row movement would be different for the two sheets, but they should both end up on the `name` column.

I would prefer to specify "all sheets" more explicitly, like `+*:name:`, but of course `*` is problematic (as the shell may/will try to expand it).  Suggestions for alternatives welcome, though in the absence of anything compelling I guess the empty string still works.

saulpw:
BTW, @midichef, thank you so much for calling out all the different cases.  We should lift that and put it directly into the docs, both the manpage and the online .md.

midichef:
> I'm less enthusiastic about `+row`, but I accept the argument that it's like how vim and tail work, and since we're borrowing this CLI syntax from them, that we should be consistent.

I thought about this syntax some more. Here are a few arguments against `+row`:
1. It's unneeded. We can already specify a row number by inserting `0:` in front of the row: `+0:2`.
2. It will be unintuitive for users unfamiliar with vim and tail. That's probably quite a few users. In fact, it would be worse thatn unintuitive, but actively confusing, if they think in terms of "the number after the plus", which changes, from being a row in `+row`, to being a column in `+col:row`.
3. The visidata syntax of +a:b:c:d is so different from vim that they're not analogous. For example, vim only allows `+` for line number, it does not support `+col:row`. And vim uses `+` in ways that visidata does not, like `+/pattern` (to search for a line with `pattern`) and `+command` (to execute a command after reading the first file).

And changing the behavior would be easy. It is a one-line change. But if we get rid of `+row`, what should `+1` mean now?  `+col`? `+col` would be surprising for vim/tail users, and it's a breaking change from current behavior in 3.0.2. Should we require column and row to be specified: `+col:row`?

saulpw:
Here's the design and explanation that I think is coherent:

The core design for this option is `+sheet:col:row`.  (This ordering is consistent with VisiData internals; the colorizer function signature, for instance, is `func(sheet, col, row, value)`). This moves the starting cursor on the given `sheet` to the given `col` and `row`, without changing the active sheet.  Either names or numbers can be given.  Numbers are 0-based; negative numbers mean "from the end" as in Python.

Values are interpreted right-to-left, with an empty sheet field meaning "all" and a missing field meaning "current".  So `vd bar.csv foo.csv +col:row` would mean move to col/row in foo.csv (the "current" or most recently given sheet), whereas `vd bar.csv foo.csv +:col:row` would mean move to col/row on both sheets.

This also means that `+10` would move to the 11th row on the last sheet, and `+:10` would behave identically.  `+::10` would mean to move to the 11th row on all sheets.  We can keep this `+row` as a lightly-documented usage, or as a nicety, a pleasant surprise compatibility with vim, but almost entirely promote the more straightforward `+s:c:r` syntax.  

I believe the above is entirely consistent with your given examples, so it feels like we're converging on the design.

Open question: can we use `+s::` (which would be a no-op) to mean "switch the active sheet to s"?  It might be useful to switch to be able to switch to a newly opened subsheet from the CLI.

Does this leave anything unspecified or unclear?

visidata/main.py
midichef: Good point, responsiveness is critical. I'll think about how to implement it. I do like the idea of a thread that keeps trying. Because then the delay before the cursor move could be shorter than a second, and unobtrusive.

midichef:
> Does this leave anything unspecified or unclear?

It's clear, thanks!

saulpw (changes requested):
I think we're still waiting on changes for the last comment about vd.sync().

midichef:
I've almost got a final patch ready to remove unnecessary `vd.sync()` and simply retry loading till it succeeds.

> BTW, @midichef, thank you so much for calling out all the different cases. We should lift that and put it directly into the docs, both the manpage and the online .md.

@anjakefala How do I update the manual pages in `man/`: `vd.1`, `vd.inc`, `vd.txt`, `visidata.1`? I'm not sure how to generate the `.1` files.

anjakefala:
@midichef Just update `vd.inc`! The others are updated when I run `dev/mkman.sh` when I'm shipping a release.

midichef:
Before this PR, for the purposes of the `+col:row` arg, row number was 0-based, but column number was 1-based.  `+1:1` was the 1st column but the 2nd row.

As this PR currently stands, changing to `+row:col`, both rows and columns would be 0-based. `+1:1` would be the 2nd column and the 2nd row.

What's the right base to use for numbering rows and columns?

midichef:
> Open question: can we use +s:: (which would be a no-op) to mean "switch the active sheet to s"?

Sure, I'll put that in too.

saulpw:
> What's the right base to use for numbering rows and columns?

Let's use 0-based numbers for both, consistent with Python.

midichef:
Okay, I've incorporated all the feedback.

This set of commits will retry the move changes every 0.1 seconds. It's hardcoded. Should I make this a configurable option? Also, it will retry failed moves forever. Should there be a time limit? Or a number of failures limit? The repeated errors are annoying and currently make it impossible to leave the sheet that is giving errors:
(to see the error messages repeat:  `vd sample_data/sample.tsv sample_data/test.usv +50:10`)

I could also use some guidance on the troff/groff formatting. Existing `Sy` tags seem to be ended by doing `No`, as in `Sy 1 No`. But they left a space after the `No`. To close the tags without adding a space, I used `Ns`. It was used near the start of `vd.inc` and seemed to do what I want. I copied `Ns` without understanding, as I could not find any documentation online that explained how to use any of `Sy` `No` or `Ns`.

saulpw:
> This set of commits will retry the move changes every 0.1 seconds. It's hardcoded. Should I make this a configurable option? Also, it will retry failed moves forever. Should there be a time limit? Or a number of failures limit?

Can we use `vd.sync()` in the retry_moves thread?  It should wait until all threads except the current one (and the main interface thread) have finished, by which point we shouldn't have any reason to retry anymore.

anjakefala:
> I could also use some guidance on the troff/groff formatting. Existing Sy tags seem to be ended by doing No, as in Sy 1 No. But they left a space after the No. To close the tags without adding a space, I used Ns. It was used near the start of vd.inc and seemed to do what I want. I copied Ns without understanding, as I could not find any documentation online that explained how to use any of Sy No or Ns.

Yes, this is correct!

`Sy` bolds. `No` == No formatting. `Ns` == No formatting and no space.

midichef:
To get the branch for this PR in shape for me to finish it soon, I've slightly modified the previous outdated commits, changing a couple of lines of `import`s to let them work with the current `develop` branch.

midichef:
Okay, I rebased again to fix a merge conflict. The only new 2 commits are the last 2:  https://github.com/saulpw/visidata/pull/2425/commits/7a8135a7ee1b5c7274f33d8765c417f9a9705855 and https://github.com/saulpw/visidata/pull/2425/commits/db8a84ed7233195f062bf6964da9019e9dc82414.

> Can we use `vd.sync()` in the retry_moves thread? It should wait until all threads except the current one (and the main interface thread) have finished, by which point we shouldn't have any reason to retry anymore.

Okay, I've implemented `vd.sync()`. I also use an exponential backoff. Please feel free to make any changes, like to the warning messages, or to `retry_move_to_pos()` with its arbitrary `retry_interval` and `timed_attempts`. My main motivation here is to close this old PR out.

midichef:
Currently `seq 3 | vd +5` will retry the position move 8 times, for 25.6 seconds, even though the file is loaded nearly instantly. I'd like to get rid of those hardcoded attempts, if the file is already done loading.

Is there a reliable way to test whether the data is done loading? Maybe `len(vd.unfinishedThreads) == 1`?

saulpw:
We have `sheet.afterLoad()` which currently sorts the sheet, and we could use `@Sheet.after`.  It's not 100% reliable as some loaders, like fec, still overload `reload` instead of `loader` or `iterload`, without calling `afterLoad` appropriately.  But it should work for a majority of loaders and be strictly better than doing a sync/backoff dance.

midichef:
I'm not sure I understand. Is the idea to add a sheet attribute to track the loading status, and set it in `Sheet.afterload`?

saulpw:
I was thinking that we would install a function that would be called after afterLoad, that moves the cursor to the given position.

midichef:
Using `afterLoad` has a complication that I don't see any way around.

The problem comes with nested sheet references like in the argument `+sheet1:subsheet2:5`. This consists of a sheet, and a subsheet. Currently I store this data as a "sheet description", which is resolved into a specific sheet later in startup, when it's time to attempt a move; we keep attempting on a timer, not knowing when sheets will load.

If instead we try to queue this move up via `afterLoad` early, then I need to resolve the sheet description into a `Sheet` object early, to modify the `afterLoad` of that specific `Sheet`. But to resolve the subsheet part, we've first got to fully load `sheet1` by `sync(vs.ensureLoaded())`. That means `sheet1.afterLoad()` gets called. From that point on there is a problem for queueing up other moves; another `+` argument may also need to queue up a move for `sheet1`, like `+sheet1:4:5`, but we can no longer do it by way of `sheet1.afterLoad()` since it's already been called.

midichef:
On second thought, there is a way to get `afterLoad` working, by properly ordering the moves. We have to queue up moves for ancestor sheets first, and descendants afterward. I'll try it out.

midichef:
Okay, I've pushed a version that uses `afterLoad` to attempt column moves after the relevant sheet loads. Just one attempt. No retrying on a timed schedule. No `sync` calls.

saulbert:
I pushed 3 commits to your branch:

**683309ff** `[main-] use @Sheet.after hook instead of monkey-patching afterLoad`
Replaced the decorator/wrapper closure in `queue_move_to_pos` with a `@Sheet.after` hook on `afterLoad`, which is the idiomatic VisiData pattern (see `modify.py:64`, `known_cols.py:15`, `scroll_context.py:24` for examples). Moves are stored on the sheet instance as `_startpos_moves` and processed by the class-level hook. Everything else is unchanged.

**d9ca1183** `[tests-] add parsePos unit tests`
18 test cases covering the full syntax matrix from your PR description: row-only, col:row, sheet:col:row, subsheets, all-sheets prefix, negative indices, named values, empty string.

**1fa8501e** `[dev-] add test style rule`
Minor dev docs update (unrelated to this PR's functionality).

All 168 existing tests pass, plus the new parsePos tests.

saulpw:
Okay, I made a few small changes, mostly changing out the monkey-patching of afterLoad for a more idiomatic usage, and adding the tests that we came upon here.  Please do a final review and then I'll merge it!

midichef:
This looks good to me. I just re-tested by hand all the example syntax cases documented in the man page (the same as the unit tests added for parsePos). Let's close out this oldest of the pending PRs!

PR #2428, opened 2024-06-22
[misc-] fix minor name definition errors caught by pyflakes

midichef:
I ran pyflakes, and it uncovered some errors related to symbols that weren't defined, or were defined more than once. These are only minor errors in code that is rarely run, or never run.

### DIFF
--- a/visidata/path.py
+++ b/visidata/path.py
@@ -179,10 +179,6 @@ class Path(os.PathLike):
         'Filename without any extensions.  Not the same as pathlib.Path.'
         return self.base_stem
 
-    @lru_cache()
-    def stat(self, force=False):
-        return self._path.stat()
-
     @property
     def given(self):
         'The path as given to the constructor.'
--- a/visidata/path.py
+++ b/visidata/path.py
@@ -229,7 +229,7 @@ class Path(os.PathLike):
         return self._path.__fspath__()
 
     def __lt__(self, a):
-        if isinstance(a, visidata.Path):
+        if isinstance(a, Path):
             return self._path.__lt__(a._path)
         return self._path.__lt__(a)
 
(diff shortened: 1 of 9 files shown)


saulpw (commented):
Do you mean that pyflakes only found these 10 issues and they were all minor?  Or that, of the numerous issues pyflakes found, these are the 10 minor ones that could be easily fixed?

visidata/path.py
```diff
@@ -179,10 +179,6 @@ def name(self):
         'Filename without any extensions.  Not the same as pathlib.Path.'
         return self.base_stem
 
-    @lru_cache()
-    def stat(self, force=False):
```
saulpw, line 183: Removing this makes me nervous.  It should still work, as members that aren't defined explicitly fall back to self._path, but the @lru_cache will make this much faster for DirSheets which would otherwise stat the same file repeatedly.  How can pyflakes know this isn't used?

midichef: Oh, my commit message was unclear on why it's redundant. I believe this `stat()` is redundant because `stat()` is redefined a couple of hundred lines later in the same class.
https://github.com/saulpw/visidata/blob/c9790737398636c15946180bb3836ebe27a777e4/visidata/path.py#L377

midichef:
> Do you mean that pyflakes only found these 10 issues and they were all minor? Or that, of the numerous issues pyflakes found, these are the 10 minor ones that could be easily fixed?

Pyflakes found more issues, but all were minor. Of all the minor issues, these are the most significant ones. These are the ones that seemed to relate to actual functioning of code.

My goal in fixing the trivial ones, like symbols imported twice, is to thin out the list of warnings for the future. So if a real function redefinition problem happens, it stands out in the Pyflakes output very clearly.

visidata/path.py
saulpw: Ah, you're totally right!

saulpw (approved):

PR #2432, opened 2024-06-27
[misc-] fix minor typos, and add-rows prompt

midichef:
Three minor commits:

1. The first commit fixes a few self-explanatory typos.

2. The `add-rows` prompt (`add rows: 1`) is a bit vague. The new wording makes it just a bit clearer that what the user is inputing is a count of rows. Otherwise, they may think they're being prompted for a row index. Or they just may not understand what is being asked.

### DIFF
--- a/visidata/modify.py
+++ b/visidata/modify.py
@@ -65,7 +65,7 @@ def rowAdded(self, row):
     self._deferredAdds[self.rowid(row)] = row
     def _undoRowAdded(sheet, row):
         if sheet.rowid(row) not in sheet._deferredAdds:
-            vd.warning('cannot undo to before commit')
+            vd.warning('cannot undo row addition after a commit')
             return
         del sheet._deferredAdds[sheet.rowid(row)]
     vd.addUndo(_undoRowAdded, self, row)
@@ -87,7 +87,7 @@ def cellChanged(col, row, val):
             if oldval == col.getSourceValue(row):
                 # if we have reached the original value, remove from defermods entirely
                 if col.sheet.rowid(row) not in col.sheet._deferredMods:
-                    vd.warning('cannot undo to before commit')
+                    vd.warning('cannot undo cell change after a commit')
                     return
                 del col.sheet._deferredMods[col.sheet.rowid(row)]
             else:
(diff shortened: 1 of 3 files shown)


visidata/modify.py
```diff
@@ -105,7 +105,7 @@ def rowDeleted(self, row):
     self.unselectRow(row)
     def _undoRowDeleted(sheet, row):
         if sheet.rowid(row) not in sheet._deferredDels:
-            vd.warning('cannot undo to before commit')
+            vd.warning('cannot undo before commit')
```
saulpw: This is/was inferring that the user was trying to undo a row deletion after a commit operation, which is not possible.  So I agree the message should be changed, but as stated now is incorrect.

visidata/modify.py
```diff
@@ -330,7 +330,7 @@ def new_rows(sheet, n):
     return [sheet.newRow() for i in range(n)]
 
 Sheet.addCommand('a', 'add-row', 'addRows([newRow()], index=cursorRowIndex); cursorDown(1)', 'append a blank row')
-Sheet.addCommand('ga', 'add-rows', 'n=int(input("add rows: ", value=1)); addRows(new_rows(n), index=cursorRowIndex); cursorDown(1)', 'append N blank rows')
+Sheet.addCommand('ga', 'add-rows', 'n=int(input("add how many rows: ", value=1)); addRows(new_rows(n), index=cursorRowIndex); cursorDown(1)', 'append N blank rows')
```
saulpw: `add # rows: ` ?

saulpw (approved):
Thanks for this attention to detail, @midichef !

visidata/modify.py
midichef: Oops, I did not read the context. How about, as you said, `cannot undo row deletion after a commit`?

There are two more places in this file to change the warning as well, `rowAdded()`, and `cellChanged()`.

visidata/modify.py
midichef: That's better: clearer and shorter.

visidata/modify.py
saulpw: Yes, that wording sounds fine to me.  Thanks, @midichef.

PR #2434, opened 2024-06-28
[fec-] fix diving into row sheets

### LINKED ISSUE #2433: [fec] opening rows in FEC files does not seem to work
**Small description**
FEC files can be opened, but rows cannot be browsed.

**Data to reproduce**
Using [a sample file 1326070.fec](https://docquery.fec.gov/dcdev/posted/1326070.fec).

**Steps to reproduce**

**Expected result**
When the user presses `Enter`, rows should open as in [jsvine's demo video at 13 seconds].(https://asciinema.org/a/Xyh2BFsUaOF0AlHTmMUbqQZPC)

**Actual result**
When the user presses `Enter`, `open-rows` runs, for example, opening the itemization row, as:
```
column    | value
------------------------
component | itemization
size      | 56958
```

**Configuration**
Yes, it reproduces with `-N`.

Yes, the latest develop.

**Additional context**
`fecfile` 0.8.0 installed via pip
Python 3.10.12
Ubuntu 22.04.4

midichef:
Closes #2433.

Also takes `open_fec()` out of the global namespace. It's a trivial change. I found `open_fec()` in the global namespace when I was looking for stray functions of the form `open_*()` and `save_*()`.

### DIFF
--- a/visidata/loaders/fec.py
+++ b/visidata/loaders/fec.py
@@ -41,10 +41,8 @@ from visidata import (
     Path,
     Sheet,
     TextSheet,
-    Column,
     ColumnAttr,
     ColumnItem,
-    ENTER,
     asyncthread,
     Progress,
     addGlobals,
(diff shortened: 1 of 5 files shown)


visidata/loaders/fec.py
```diff
@@ -230,7 +210,7 @@ class FECFiling(Sheet):
     @asyncthread
     def reload(self):
         from fecfile import fecparser
-        self.rows = []
+        self.rows = []  # rowdef:  Sheet, of a type from COMPONENT_SHEET_CLASSES.values()
```
saulpw, line 213: In general if the rowdef is a Sheet then the parent class is IndexSheet, which provides default columns and other functionality.  I'm not sure if it applies here but it might be worth exploring.

saulpw (approved):

visidata/loaders/fec.py
midichef: Good to know! I'll make note to look into that for future work with rows that are Sheets. But since this FEC loader has been broken for a while without any user complaints, I suspect it's not often used, so I don't plan to spend more time on it.

PR #2436, opened 2024-06-30
[sheet-] in execCommand(), fail if cmd does not exist

midichef:
A command that does not exist, like `qqquit-sheet`, will be skipped during replay of a macro or command log, but the subsequent commands will continue to be replayed.

Some ways that could happen:  a user makes a typo editing a command name in a macro or command log. Or in the past, the user defined a command to be used in a macro, for example, `loggable-quit-sheet`, and now they delete the command definition, breaking the macro that uses it.

It seems better to `fail()` instead of `warning()` in this case. Is there a case where `warning()` is better?


midichef:
For this PR, I needed to make changes to get tests to pass.

One test was trying to run commands that couldn't be found. It was running commands on a test `TsvSheet` that the sheet didn't have, like commands for `CanvasSheet` or `ColumnsSheet`. I've made it skip those, there are around 150 of those commands.

The other test that started to fail was the macro test, `test_macros.vd`. It had been quietly not testing anything, because it had been trying to run a command `macro`. I cleaned it up and changed it to run a macro `1.vdj` via a command `exec-1`.

But then I ran into another problem, which is that macros couldn't be run when `--batch` was used. Macros were only loaded when `run()` executes, which doesn't happen in batch mode. So now they get loaded, by `vd.reloadMacros()`.

Note that there is still one case where macros aren't loaded, which is bare `--batch` with no `--play`. That should be fine, since there are no commands, so there's no way to run a macro.

visidata/basesheet.py
```diff
@@ -194,7 +194,7 @@ def execCommand(self, longname, vdglobals=None, keystrokes=None):
 
         cmd = self.getCommand(longname or keystrokes)
         if not cmd:
-            vd.warning('no command for %s' % (longname or keystrokes))
+            vd.fail('no command for %s' % (longname or keystrokes))
```
saulpw, line 197: I have a vague recollection that `fail` was not desired in replay at one time.  Maybe it was that some commands would `fail`, but you wanted them to continue anyway?  Like `select-regex`, if nothing matches, you can still `delete-selected` as a no-op, but you don't want to abort the replay entirely.

That's almost entirely tangential to this, but it makes me wonder if something like this should be `error` instead of `fail`?  There's almost no difference between them--error is red while fail is yellow.  But we could use them more semantically, in case we want to catch `fail`s and continue the replay, vs let `error`s abort the replay.

Either way, I agree that an unknown command should abort the replay.

saulpw (approved):

anjakefala:
Thank you so much for your careful improvement of the tests!

PR #2442, opened 2024-07-09
handle corner cases when reading from stdin

midichef:
The first commit makes it possible to use stdin as a replay file:  `vd -p -`.
Right now that causes a traceback:
```
...
File "/home/midichef/.local/lib/python3.10/site-packages/visidata/main.py", line 347, in main_vd
    vdfile.name = 'stdin.vd'
AttributeError: can't set attribute 'name'
```

The second commit is for interactive use of visidata. When a user runs `open-file` and inputs `-` as the filename, visidata hangs. The patch catches that and gives an error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # VisiData version history
 
+# vX.X.X (XXXX-XX-XX)
+
+- support replay from stdin for .vdj
+
 # v3.0.2 (2024-01-15)
 
 ### Fixes and minor improvements
--- a/visidata/_open.py
+++ b/visidata/_open.py
@@ -1,5 +1,6 @@
 import os
 import os.path
+import sys
 
 from visidata import VisiData, vd, Path, BaseSheet, TableSheet, TextSheet, SettableColumn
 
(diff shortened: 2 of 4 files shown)


saulpw (approved):
Thanks for this, @midichef !

midichef:
I fixed one more case, the case of stdin for replay commands, `-p -`, but with another file argument present, `a.json`:
```
cat quit.vdj |vd -p - a.json
    ...
    File "/home/midichef/.local/lib/python3.10/site-packages/visidata/main.py", line 46, in eval_vd
        log = log.format(*args, **kwargs)
    KeyError: '"sheet"'
```
by adding this line:
https://github.com/saulpw/visidata/blob/32a59948f63b75011cb1f55bfd07ba9268924326/visidata/main.py#L40
And a previous commit message had a typo, so I fixed it and force-pushed a replacement.

anjakefala:
When testing, I found an additional cornercase bug!

```
cat tests/append.vd | vd -p -  
```

Currently, VisiData expects `.vdj` from stdin

```
message	source
opening - as vdj	visidata/_open.py:138:openPath
JSONDecodeError: Expecting value: line 1 column 1 (char 0)	visidata/errors.py:26:exceptionCaught

```

anjakefala:
@midichef I am going to address that new cornercase in this PR!

anjakefala:
Hi @midichef! 

It ended up being 15 lines of code to support both `.vd` and `.vdj` from stdin, so we're just going to merge as-is!

Issue #431, opened 2020-01-03
[howto] hide invariant columns


frosencrantz:
This is a request to the visidata community.  I would like a plugin that hides invariant columns.

I look at data with many columns.  Sometimes the columns have only a single value. Sometimes this happens after drilling down into the data.

It would be useful if these invariant columns could be hidden with a single command. Now, I just visit the frequency table to see if what values are in a column, and manually hide columns.

A secondary feature, would be to create a second table that has a copy of each of the invariant columns but with a single row.  This second table should be the alternate table so, that Ctrl-^ would make it possible to quickly view.

anjakefala:
@frosencrantz, can you define for me "invariant columns" in this context?

frosencrantz:
In this context  an "invariant column" is a column that has same value in all rows.  There needs to be more than one row.

saulpw:
To hide columns which have the same value in all rows, open the Describe Sheet (`Shift+I`), select columns using `z|` where `distinct <= 1`, and then hide them with `g-`.

frosencrantz:
Thanks, @saulpw!  That works.
Is there a way to make that into a single shortcut?

PR #2453, opened 2024-07-13
[features-] add command to hide repetitive columns

midichef:
This is a command to hide any column that contains one value repeated in every row. Is this worth having? It's often the first command I run when opening large tables with such uninformative columns.

I'm not sure what to call the command.  Right now it is `hide-unvarying-cols`, but perhaps `hide-uniform-cols` is better? `hide-repetitive-cols` is too hard to spell.

### DIFF
--- a/visidata/features/dedupe.py
+++ b/visidata/features/dedupe.py
@@ -24,7 +24,7 @@ __author__ = "Jeremy Singer-Vine <jsvine@gmail.com>"
 
 from copy import copy
 
-from visidata import Sheet, BaseSheet, asyncthread, Progress, vd
+from visidata import Sheet, TableSheet, asyncthread, Progress, vd
 
 
 def gen_identify_duplicates(sheet):
@@ -103,8 +103,8 @@ def dedupe_rows(sheet):
 
 
 # Add longname-commands to VisiData to execute these methods
-BaseSheet.addCommand(None, "select-duplicate-rows", "sheet.select_duplicate_rows()", "select each row that is a duplicate of a prior row")
-BaseSheet.addCommand(None, "dedupe-rows", "vd.push(sheet.dedupe_rows())", "open new sheet in which only non-duplicate rows in the active sheet are included")
+TableSheet.addCommand(None, "select-duplicate-rows", "sheet.select_duplicate_rows()", "select each row that is a duplicate of a prior row")
+TableSheet.addCommand(None, "dedupe-rows", "vd.push(sheet.dedupe_rows())", "open new sheet in which only non-duplicate rows in the active sheet are included")
 
 vd.addMenuItems('''
     Row > Select > duplicate rows > select-duplicate-rows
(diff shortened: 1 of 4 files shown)


frosencrantz:
This would be a useful command.  Thank you for proposing it @midichef!

I like it and would use it.  Saul provided a useful suggestion when I asked in Issue 431 (2022 issues ago):

[[howto] hide invariant columns](https://github.com/saulpw/visidata/issues/431#top)

visidata/features/dedupe.py
```diff
@@ -102,9 +102,21 @@ def _reload(self=vs):
...
+    for col in sheet.visibleCols:
+        vals = (col.getTypedValue(r) for r in sheet.rows)
+        first = next(vals)
+        if all(v == first for v in vals):
+            col.hide()
+
 # Add longname-commands to VisiData to execute these methods
 BaseSheet.addCommand(None, "select-duplicate-rows", "sheet.select_duplicate_rows()", "select each row that is a duplicate of a prior row")
 BaseSheet.addCommand(None, "dedupe-rows", "vd.push(sheet.dedupe_rows())", "open new sheet in which only non-duplicate rows in the active sheet are included")
+BaseSheet.addCommand(None, 'hide-unvarying-cols', 'sheet.hide_unvarying_cols()', 'hide any column that has multiple rows but only one distinct value')
```
saulpw: These commands should all be on TableSheet instead of BaseSheet.

visidata/features/dedupe.py
```diff
@@ -102,9 +102,21 @@ def _reload(self=vs):
...
 
 vd.addMenuItems('''
```
saulpw, line 109: Let's add this command to an appropriate place in the menu.

midichef: I think it'd be good to put it under `Column`, with `hide` and `unhide-cols`. I think the only way to do that is to do `addMenuItems` with the other hide commands in `features/layout.py`. Is that right? So then the function definition and `addCommand()` should move there.

visidata/features/dedupe.py
midichef: Okay. If I move the new command to `features/layout.py` (see below), `Sheet` is used for `Sheet.addCommand()` there, so I'll use that to match. What's the intended distinction between `Sheet` and `TableSheet`?

midichef:
I changed the command prefix to `hide-uniform-` instead f `hide-unvarying-`. One reason is, "unvarying" is hardly ever used as a technical term, so it would be a new word for users to to recall. Another is, "hide" and "un-varying" describe the result based on two negatives (the positive equivalents would be "show" and "varying"). "uniform" seems better as it describes the targets of "hide" in terms of what they are (uniform), instead of what they aren't (varying).

If there's a common word that means "having one distinct value", I'd like to know.

visidata/features/dedupe.py
saulpw: There's no functional difference between Sheet and TableSheet.  Sheet is the original name, in use in many places and not worth it to fix.  TableSheet is the more descriptive name that is preferred for new usage.  But Sheet is fine to match, as you say.

saulpw:
> I changed the command prefix to `hide-uniform-` instead f `hide-unvarying-`.

I was considering suggesting 'invariant' but I think I like `uniform` better (less formal).  Thanks for giving it some thought.

saulpw (approved):

visidata/features/dedupe.py
saulpw: Menu items can be added anywhere, so if this is the best place for the code, then the addCommand and its associated Menu item should be here.

PR #2455, opened 2024-07-15
[xlsx] add options.xlsx_color_cells (default: True) to allow disabling of xlsx colorizers

### LINKED ISSUE #2317: Allow disabling of colorizers on xlsx files.
It seems that colour support for xlsx files was added with this:
https://github.com/saulpw/visidata/issues/1718

I recently updated from Visidata 2.11 to 3.0.2. I prefer to not have text and fill colours on xlsx files (as in v2.11), yet I cannot disable colorizers.
I've tried  starting `visidata --xlsx-meta-columns=False` parametre, but unfortunately colours are still on. Also tried to switch theme to asciimono, yet xlsx files still load with colours.

If `xlsx-meta-columns=False` is supposed to disable colours on xlsx, I'm filling a bug report. Else sorry for misunderstanding and please guide me to the proper way of disabling them.

anjakefala:
Closes #2317 

I did test this!

### DIFF
--- a/visidata/loaders/xlsx.py
+++ b/visidata/loaders/xlsx.py
@@ -10,6 +10,7 @@ from visidata.type_date import date
 
 
 vd.option('xlsx_meta_columns', False, 'include columns for cell objects, font colors, and fill colors', replay=True)
+vd.option('xlsx_color_cells', True, 'color cells based on xlsx source')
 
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Looks good to me.

PR #2459, opened 2024-07-18
[main-] when batch mode is interactive, keep undo, quitguard

midichef:
Undo and quitguard are turned off by batch mode, even in interactive mode. That can be demonstrated by doing:
`vd -b -p tests/fill-zero.vdj -i`
and pressing `U`, to see `options.undo not enabled`.
This PR turns these options off only when batch mode is not interactive.

### DIFF
--- a/visidata/main.py
+++ b/visidata/main.py
@@ -288,8 +288,9 @@ def main_vd():
     vd.domotd()
 
     if args.batch:
-        options.undo = False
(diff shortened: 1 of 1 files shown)


saulpw (approved):
This makes sense to me.

anjakefala:
Thank you!

PR #2461, opened 2024-07-18
[tests] pass tests if tsv keeps hash lines

midichef:
This gets the tests passing again after 6fff636.

### DIFF
--- a/tests/golden/pr2302.tsv
+++ b/tests/golden/pr2302.tsv
@@ -1,4 +1,3 @@
-# comment						
 OrderDate	Region	Rep	Item	Units	Unit_Cost	Total
 2016-01-06	East	Jones	Pencil	95	1.99	189.05
 2016-01-23	Central	Kivell	Binder	50	19.99	999.50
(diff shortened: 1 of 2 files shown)


midichef:
Or should the tests check that the hash line was loaded into the sheet?

saulpw (approved):
This is good, thank you!

anjakefala:
Thank you for fixing the build! =)

tests/golden/pr2302.tsv
```diff
@@ -1,4 +1,3 @@
-# comment
```
anjakefala, line 1: This test was added for testing https://github.com/saulpw/visidata/pull/2302 @midichef. Is the behaviour you addressed in that PR now broken?

midichef: Oh, nice catch. I didn't look into the history of this test. The behavior in that PR is still functional. But it is obsolete after https://github.com/saulpw/visidata/commit/6fff636f33f2ffeaac2e74359f8d199c2358c960.
I'll open a PR removing the remaining uses of `regex_skip` from `tsv.py`.

midichef: On closer inspection, the remaining uses of `regex_skip` in tsv.py look to be intentional, so I'll leave it as is.

PR #2462, opened 2024-07-18
[tests-] add tests extra:  fecfile

midichef:
I forgot to include `fecfile` in the list of libraries installed by `pip install ".[test]"`.


anjakefala (approved):
Thanks! I also added `PyYAML` which was missing. Our CI now uses the `test` extras, so we should be catching this from now on. =)

PR #2484, opened 2024-08-03
[sheets-] prevent offscreen draw for tall col headers

midichef:
A multi-line header will try to draw offscreen when the window is shorter than the header.
To see this, load [tall.json](https://github.com/user-attachments/files/16478458/tall.json) with `vd tall.json`, and shrink the window to be 13 terminal columns or less:

```
  File "/home/midichef/.local/lib/python3.10/site-packages/visidata/sheets.py", line 721, in drawColHeader
    clipdraw(scr, y+i, x, name, hdrcattr, w=colwidth)
...
```
This PR prevents this error, by not drawing the header lines below the last line on screen.

### DIFF
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -718,10 +718,11 @@ class TableSheet(BaseSheet):
             if i == h-1:
                 hdrcattr = update_attr(hdrcattr, colors.color_bottom_hdr, 5)
 
-            clipdraw(scr, y+i, x, name, hdrcattr, w=colwidth)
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Thanks for fixing this!  How do you find these bugs anyway?

midichef:
I was looking into `addstr()` calls trying to find [the bug reported by @frosencrantz](https://github.com/saulpw/visidata/discussions/2480).

PR #2487, opened 2024-08-05
[canvas-] fix alignment of points with axis labels

midichef:
The calculation of the y-coordinate for points was slightly different from the method used for axis labels. The reason was that floating-point coordinates were rounded to integers in two slightly different ways:  for points, it was `round(i+f)`, for labels it was `i+round(f)`.

Points were always plotted correctly relative to each other. But they could be misaligned with labels. As you can see on the left image, the label for y=3.0 was one square higher than the point with y=3.0. The right image shows the new, corrected behavior.
rounding-bad rounding-good

To be more specific about what the error was: if a point was near the edge of a terminal character box, it might not line up properly with a label. It could be drawn one terminal character box higher or lower. However, since visidata draws points with vertical resolution of four points per terminal square (`⡇`), the vertical error of the actual dot drawn, relative to a label for the same coordinate, was at most 1/4th of a terminal character box.

As an aside, what is the proper name for what I have called a "terminal character box"? That is, an 80x25 terminal window has 80*25=2000 places the cursor can be. What are those individual places called?

### DIFF
--- a/visidata/canvas.py
+++ b/visidata/canvas.py
@@ -649,13 +649,13 @@ class Canvas(Plotter):
         else:
             return h
 
-    def scaleX(self, canvasX):
+    def scaleX(self, canvasX) -> int:
         'returns a plotter x coordinate'
-        return round(self.plotviewBox.xmin+(canvasX-self.visibleBox.xmin)*self.xScaler)
+        return self.plotviewBox.xmin+round((canvasX-self.visibleBox.xmin)*self.xScaler)
 
-    def scaleY(self, canvasY):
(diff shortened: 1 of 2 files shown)


saulpw:
> As an aside, what is the proper name for what I have called a "terminal character box"? That is, an 80x25 terminal window has 80*25=2000 places the cursor can be. What are those individual places called?

I would call it a "character cell".  A half-width character takes up one character cell and a full-width character takes up two character cells.

saulpw (approved):
Thanks for fixing this, @midichef!

PR #2489, opened 2024-08-06
[canvas-] draw lines at user-defined x or y

### LINKED ISSUE #2482: [] Feature request: Horizontal guide lines in plots
It would be cool if I could add horizontal guide to plot. I mean that I could give specific Y values where a line would be drawn from the left edge of the screen to the right edge. Whereever there are datapoints, the line won't be drawn. 

Here's a grab-bag of horizontal line characters at various heights:

```
━─═﹉﹊﹋﹌﹍﹎﹏―—–‐⁃_‗⎯⎼⎽⏤▁▔⚊一
```

We could use the different heights to make the lines more accurately-placed.

midichef:
This is a draft of a feature to draw vertical or horizontal lines on a `Canvas`.

For now I used the command names: `add-line-x`, `remove-line-x`, `clear-lines-x`. Any better names I could use instead?

When finalized, this PR closes #2482.

### DIFF
--- a/visidata/canvas.py
+++ b/visidata/canvas.py
@@ -544,13 +544,14 @@ class Canvas(Plotter):
         'adjust visibleBox.xymin so that canvasPoint is plotted at plotterPoint'
         self.visibleBox.xmin = canvasPoint.x - self.canvasW(plotterPoint.x-self.plotviewBox.xmin)
         self.visibleBox.ymin = canvasPoint.y - self.canvasH(plotterPoint.y-self.plotviewBox.ymin)
-        self.refresh()
+        self.resetBounds()
 
     def zoomTo(self, bbox):
         'set visible area to bbox, maintaining aspectRatio if applicable'
         self.fixPoint(self.plotviewBox.xymin, bbox.xymin)
         self.xzoomlevel=bbox.w/self.canvasBox.w
         self.yzoomlevel=bbox.h/self.canvasBox.h
+        self.resetBounds()
 
     def incrZoom(self, incr):
         self.xzoomlevel *= incr
--- a/visidata/canvas.py
+++ b/visidata/canvas.py
@@ -234,12 +234,28 @@ class Plotter(BaseSheet):
 
     def draw(self, scr):
         windowHeight, windowWidth = scr.getmaxyx()
-        disp_canvas_charset = self.options.disp_canvas_charset or ' o'
-        disp_canvas_charset += (256 - len(disp_canvas_charset)) * disp_canvas_charset[-1]
-
         if self.needsRefresh:
             self.render(windowHeight, windowWidth)
 
+        self.draw_pixels(scr)
+        self.draw_labels(scr)
+
+    def draw_empty(self, scr):
+        # use draw_empty() when calling draw_pixels() with clear_empty_squares=False
+        cursorBBox = self.plotterCursorBox
+        for char_y in range(0, self.plotheight//4):
+            for char_x in range(0, self.plotwidth//2):
+                cattr = ColorAttr()
+                ch = ' '
+                # draw cursor
+                if cursorBBox.contains(char_x*2, char_y*4) or \
+                    cursorBBox.contains(char_x*2+1, char_y*4+3):
+                    cattr = update_attr(cattr, colors.color_current_row)
+                scr.addstr(char_y, char_x, ch, cattr.attr)
+
+    def draw_pixels(self, scr, clear_empty_squares=True):
+        disp_canvas_charset = self.options.disp_canvas_charset or ' o'
+        disp_canvas_charset += (256 - len(disp_canvas_charset)) * disp_canvas_charset[-1]
         if self.pixels:
             cursorBBox = self.plotterCursorBox
             getPixelAttr = self.getPixelAttrRandom if self.options.disp_pixel_random else self.getPixelAttrMost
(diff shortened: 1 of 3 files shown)


visidata/canvas.py
```diff
@@ -808,6 +865,13 @@ def deleteSourceRows(self, rows):
 Canvas.addCommand('g'+ENTER, 'dive-visible', 'vs=copy(source); vs.rows=list(rowsWithin(plotterVisibleBox)); vd.push(vs)', 'open sheet of source rows visible on screen')
 Canvas.addCommand('gd', 'delete-visible', 'deleteSourceRows(rowsWithin(plotterVisibleBox))', 'delete rows on source sheet visible on screen')
 
+Canvas.addCommand('', 'add-line-x', 'sheet.gridlines_x += [float(x) for x in input("add line(s) at x = ", value="0.0", defaultLast=True).split()]; sheet.refresh()', 'draw a vertical line at x-values (space-separated)')
```
saulpw: I could see bindings being `gx` and `gy` to add a gridline, and `zx`/`zy` to remove them (and `gzx`/`gzy` to clear them).  Maybe swapping `gx`/`gy` and current `x`/`y`.  What do you think about this?

visidata/canvas.py
```diff
@@ -808,6 +865,13 @@ def deleteSourceRows(self, rows):
...
+Canvas.addCommand('', 'add-line-y', 'sheet.gridlines_y += [float(y) for y in input("add line(s) at y = ", value="0.0", defaultLast=True).split()]; sheet.refresh()', 'draw a horizontal line at y-values (space-separated)')
```
saulpw: These should use `col.type()` (dates are a numeric type).  Ideally the default value should be something reasonable, maybe the median of the data range?  And let's use `type="gridlinex"`/gridliney with both the add and remove, so people can scroll up to remove the most recently added gridline.

For the longnames, let's not use the 'add/remove/clear' verbs since current 'add-' commands are used to modify the data.  What can we use to indicate this is an interface-only command?

saulpw (commented):
Thanks for doing this, @midichef!  I think it looks nice. Maybe we could add the theme elements to the ascii themes with `|` and `-`?

visidata/canvas.py
midichef: Instead of 'add/remove/clear', how about 'draw/erase/erase'?

saulpw: That works for me!  `draw-` especially is very clearly interface only.

visidata/canvas.py
midichef: Those bindings all sound reasonable to me. I don't use current `x` and `y` so I don't have an opinion on those.

midichef:
Okay, the last commit addresses all the feedback on this feature so far.

One choice that I'd like feedback on is the parsing of input for gridline values. It currently allows multiple values as a space-separated string.  E.g. `draw-line-x` `3.0 5.0 7.0`. But that means input can't take strings that are naturally space separated, like dates such as `Jan 1, 1935`. So right now, input when drawing/erasing gridlines requires date strings that contain no spaces. Dates must be input in the form `YYYY-MM-DD` regardless of the setting of `disp_date_fmt`. Similarly, the input strings can't contain commas, because we use commas to separate x-values in the case when the graph uses multiple x-columns.

An alternative design would be to accept input in any form, but restrict the inputs for `draw-line-*` and `erase-line-*` to only take one value at a time. Let me know if that's better.

Another change in this commit, I moved the gridlines out of `Canvas` and into `GraphSheet`, because they depend on things like `xcols` that don't exist for `Canvas`. Grid lines are more naturally part of `GraphSheet` than for Canvas. Just like axis labels are part of `GraphSheet and not Canvas`.

cool-RR:
Hey,

Thanks for implementing this.

1. Looks good overall, including x lines.
2. I do think the term "refline" would be better than "line".
3. Is there a way to change the color of the lines?
4. Check this out: 

I drew a line at 0.4 (I think):

screenshot_HOPPER_2024-08-20_103604

But when I zoom out I don't see it: 

screenshot_HOPPER_2024-08-20_103612

Maybe it's because there are too many data points around that Y value. However, notice how there's clear space both on the left and on the right of the data. Could you extend the lines to there? As long as they don't clash with the tick numbers, of course.

midichef:
To change the color of the line you can add a line to your .visidatarc like `options.color_gridline = 'green'`
We will probably change the name to something more consistent with other color names like `color_graph_refline`.

I'm not sure why the y=0.4 line disappeared on zooming in your images. That looks like a bug to me. The refline should be drawn on the left and right of the data.

I can't reproduce it in my limited testing. Can you find a way to consistently reproduce that, and post a sample data file, command log, and terminal dimensions in characters (such as 120x40)?

cool-RR:
@midichef 

Data file and command log: https://www.dropbox.com/scl/fo/5vpb430tpnyq9iinbcljk/AKPM09_NkdomvwodvrpyFCE?rlkey=ssbhjdqs8w92dblxzkhdmwne1&dl=0

Terminal size:

```
➜  Desktop echo "Rows: $(tput lines), Columns: $(tput cols)"
Rows: 33, Columns: 133
```

One thing that may not be clear in that command log is that the selected rows (i.e. columns) are `reward.a0.coeval` to `reward.a3.coeval`, a total of 4 rows/columns.

midichef:
Thanks for the data files. They reproduced the missing refline on my system.

The problem was that certain kinds of zoom, like the ones used by `set-y` and `set-x`, were not recalculating the view window for the next screen redraw. I've fixed that behavior.

I've also changed the command names to `draw-line-*` to `draw-refline-*`, etc.

I believe that addresses all the feedback so far. Any other issues?

visidata/canvas.py
```diff
@@ -552,11 +554,15 @@ def zoomTo(self, bbox):
         self.xzoomlevel=bbox.w/self.canvasBox.w
         self.yzoomlevel=bbox.h/self.canvasBox.h
 
+        self.resetBounds()
```
saulpw, line 554: Do we ever call resetBounds without refresh, or vice versa?  Should these be combined into one function?

cool-RR:
Great, thank you. Yes, this addresses all my feedback. I'll probably start using this feature when the next release of visidata is made. Thanks for implementing this.

visidata/canvas.py
midichef: They can't be combined, as each is currently called without the other in at least one place.

`resetBounds()` and some of the nearby code would benefit from a refactoring and reordering. That would allow the two lines to be combined. But it'd be a bigger change, and it'd belong in a separate PR.

midichef: I take it back: on thinking about it some more, I think `resetBounds()` and `refresh()` can be combined simply. I'll update this PR to do that.

midichef:
@cool-RR 
I noticed visidata makes a small and subtle error when setting the y range via `y` (`resize-y-input`).

After I do `y` `1 5`, the actual range is 1.02649 to 5.02649, not 1.0 to 5.0. So just be aware that in visidata v3.0 versions, if you set the y axes, the bottom edge isn't quite what it seems. Points that you'd expect to see at the bottom edge of the graph, like y=1.0, would not be shown.

The error has existed since my commit abc9affae4bbe7cd03a1b9b8cff3ebaae9291146 from June 2023, and I've submitted a fix here that will be in the next release of visidata.

Please let me know if you ever notice any similar errors in the edges of ranges/boxes when using graphs.

saulpw (approved):
Thanks @cool-RR for the suggestion and thanks @midichef for the implementation!

cool-RR:
Awesome! @saulpw Any rough idea when the next release will be? A month, a year?

saulpw:
Should be mid-October.

Issue #2382, opened 2024-04-16
vd v3.0.2 performance regression on Android


snooppr:
I use visidata on Android (Termux, Python 3.11) to view csv/xlsx.  Everything works fine with version vd **2.11**.  Updated to the latest version vd **3.0.2**.  It became impossible to use, terrible brakes: the movement of the cursor downwards/upwards/sideways is delayed by seconds.  I tried even small files of 200 lines, it’s impossible to use at the moment, I rolled back to the previous version, everything works.

v3.0.2
![ScreenRec](https://github.com/saulpw/visidata/assets/61022210/0a86512d-7f83-461b-a928-7cc5dc224c3b)

---------

v2.11
![ScreenRecord](https://github.com/saulpw/visidata/assets/61022210/18c52e56-3189-40bf-b1b6-0589f5dc912d)

Simply put, the delay in responding to commands has increased tenfold.

In addition, in v3.0.2 in xlsx the selected line on which the cursor is located is not highlighted.

snooppr:
1) I found what exactly causes the slowdown is the “slidebar”, which turns on automatically all the time; you can turn it off with the “b” key.  I believe that it is problematic and should be disabled by default (enabled from the menu or by key).
2) I could not return in version 3.0.2 the selection of the selected line under the cursor; in the xlsx format it does not work; in the csv format it selects the line, but not in all columns, but selectively.

v2.11.1
xlsx format, cursor on line "55-59", line highlighting is visible, in 3.0.2 it is not there.
![Screenshot_20240416-210415_Termux](https://github.com/saulpw/visidata/assets/61022210/5a8e4ed3-e1e4-4ccf-af93-6db5c4deeb6a)

anjakefala:
Having an option that turns off sidebar by default will be tracked here: https://github.com/saulpw/visidata/issues/2202

Issue #2202, opened 2023-12-30
[help] sidebar help mechanics need work


saulpw:
In any input mode, help can be cycled with `Ctrl+G`.  I don't particularly like this keybinding but `Ctrl+H` is of course problematic.  Ideas welcome!

There also needs to be a way to:

- [x] show `Ctrl+G to cycle help 1/N` on edge of sidebar if N>1
- [x] cycle/choose sidebar help with non-input commands
- [ ] include title on ddw help screens
- [ ] allow templatized ddw with e.g. `{sheet.name}`
- [x] choose an acceptable default for `disp_help`
- [x] turn off sidebar
- [ ] `split-win` mode: sidebar appears in the appropriate pane
- [ ] `__dunder__` should not be markdown  #2211

gennaro-tedesco:
I am adding to this thread to avoid opening a new issue. First of all thank you for all the work on the sidebar, it is awesome!

I have noticed that setting `options.disp_sidebar = False` in the general `.visidatarc` does not have any effect, namely the sidebar is always present (I can however hide it with `b` "toggle sidebar"): perhaps I misunderstood how the option is intended?

anjakefala:
I think the reason options.sidebar is not doing the trick, is because Sheet-options override general options sometimes, and the disp_sidebar check happens for the Sheet-option. 

It could be helpful to have a check that if `options.disp_sidebar` is False, that trumps the sheet-specific option. 

The current workaround for you is to play with `options.disp_help = 0`. I'm not sure if `disp_help` will stick around conceptually, but for 3.0.x, you can use it to turn off the sidebar by default.

gennaro-tedesco:
Commenting from [here](https://github.com/saulpw/visidata/issues/2241#issuecomment-1891118686) (as per suggestion to have this discussion here):

>  Input help can be disabled with vd.options.disp_help=0 with the above change.

I am on the latest release `v3.0.2` but still see no difference with `vd.options.disp_help=0`. For instance when typing `/` for search or `<space>` to open the command palette the display on the right hand side still shows - is my understanding incorrect, should this not be hidden now with such option?

anjakefala:
> I am on the latest release v3.0.2 but still see no difference with vd.options.disp_help=0. For instance when typing / for search or <space> to open the command palette the display on the right hand side still shows - is my understanding incorrect, should this not be hidden now with such option?

[...]

PR #2495, opened 2024-08-15
Updated `disp_sidebar` functionality, it should no longer revert to displaying the sb if `options.disp.sidebar=False`

niv-l:


niv-l:
This PR targets Issue #2202 and #2382.

saulpw:
Hi @ziova, thanks for the suggestion.  If we do this, however, you will never see any status messages when the sidebar is disabled.  This isn't workable, as status messages often have essential feedback for commands.

I have some code I'm working on to handle the sidebar, which I'll hopefully get around to finishing in the next week or so.

niv-l:
Oh @saulpw, thank you this has been noted. I'm looking forward to your updated code :)

Closed.

PR #2499, opened 2024-08-23
[input-] honor record=False for input()

midichef:
`vd.input()` does not seem to use the `record` parameter in interactive mode. It can be seen by using `^X` to execute:
```
vd.input('will be recorded: ', record=True)
vd.input('will be recorded, but should not: ', record=False)
```
and then `tail -n 4 ~/.visidata/input_history.jsonl`.

The expected output should record only what's typed into the first input:
```
...
{"type": "expr", "input": "vd.input('will be recorded: ', record=True)", "n": 2}
{"type": null, "input": "first", "n": 1}
{"type": "expr", "input": "vd.input('will be recorded, but should not: ', record=False)", "n": 2}
```
But the actual output includes the second input:
```
{"type": "expr", "input": "vd.input('will be recorded: ', record=True)", "n": 1}
{"type": null, "input": "first", "n": 1}
{"type": "expr", "input": "vd.input('will be recorded, but should not: ', record=False)", "n": 1}
{"type": null, "input": "second", "n": 1}
```

This commit fixes it.

### DIFF
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -551,7 +551,7 @@ def input(vd, prompt, type=None, defaultLast=False, history=[], dy=0, attr=None,
                         updater=_drawPrompt,
                         **kwargs)
 
-    if ret:
+    if ret and kwargs.get('record', True):
         vd.addInputHistory(ret, type=type)
 
     elif defaultLast:
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -507,7 +507,7 @@ def input(vd, prompt, type=None, defaultLast=False, history=[], dy=0, attr=None,
         - *history*: list of strings to use for input history.
         - *defaultLast*:  on empty input, if True, return last history item.
         - *display*: pass False to not display input (for sensitive input, e.g. a password).
-        - *record*: pass False to not record input on cmdlog (for sensitive or inconsequential input).
+        - *record*: pass False to not record input on cmdlog or input history (for sensitive or inconsequential input).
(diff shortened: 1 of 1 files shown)


saulpw (approved):

anjakefala:
Thanks @midichef!

Originally, `record` was designed to dictate whether the command got logged in the input column of the cmdlog. That's why setting it did not impact whether it got logged in the input history jsonl. However, I think it makes a lot of sense for record=False to also mean, "don't log in input history", considering it is intended for sensitive information. I updated the docstrings to match your adjustment!

frosencrantz:
@midichef and @anjakefala it looks like this PR and line 469 do not work together:

https://github.com/saulpw/visidata/blob/cd0eb7f2cf3476522effed495a0b637826c00b98/visidata/_input.py#L464-L470

I  noticed by trying to select rows based on a regular expression with `|` `select-col-regex` and it was not recording the input.

midichef:
@frosencrantz Thanks for quick feedback on this. It also appears I broke `inputMultiple()` and `select-col-regex`, which now ignores user input and acts like the user typed the last value in the input history. I'll submit a PR fixing it now.

midichef:
> Thanks @midichef!
> 
> Originally, `record` was designed to dictate whether the command got logged in the input column of the cmdlog. That's why setting it did not impact whether it got logged in the input history jsonl. However, I think it makes a lot of sense for record=False to also mean, "don't log in input history", considering it is intended for sensitive information. I updated the docstrings to match your adjustment!

Oh, good to know. I didn't realize the dual use of `record`.

PR #2501, opened 2024-08-27
[input-] fix input(record=False) and inputMultiple

midichef:
My PR in #2499 broke `inputMultiple()`. It currently ignores whatever input the user typed, and uses the last value in the user history from `input_history.jsonl`.

This PR fixes that.

I also allow `display=False` to override `record=True`, to match the current behavior of `editText()`.


saulpw (approved):

frosencrantz:
@midichef after pulling this change, it appears to work.  Thank you!

PR #2504, opened 2024-08-27
[misc-] remove blank lines before elif/else

midichef:
In #2499 I introduced a bug because I didn't look past the line break after the `if:` to see that there was a subsequent `elif:`. I should be more careful. But should we remove any chances of a repeat?

### DIFF
--- a/visidata/column.py
+++ b/visidata/column.py
@@ -384,7 +384,6 @@ class Column(Extensible):
                                             error=['unknown'],
                                             note=f'[:onclick error-cell]{options.disp_note_typeexc}[:]',
                                             notecolor='color_warning')
-
         elif isinstance(typedval, threading.Thread):
             return DisplayWrapper(None,
                                 text=options.disp_pending,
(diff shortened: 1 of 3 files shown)


saulpw:
I prefer a blank line in many of these cases, for instance when there are many elif blocks, and/or the blocks are more than a couple lines of code.  The blank lines break up the vertical visual flow and highlight the possible blocks that could be executed.  But I agree that the one that bit you was unnecessary and I can see how it escaped your attention.  I don't see that one here though; was that fixed in a different PR?

midichef:
I think `menu.py` is a great example of the blank line making the code easier to understand. With that many blocks, I think deleting the visual spaces makes it worse.

Yup, the blank line that started this inquiry was fixed in #2501.

In `main.py` and `loaders/npy.py` the blank spaces are inconsistent across various `else/elif` statements. Some statements follow blank lines, some don't. So we should either add more blank lines or take some away. And `column.py` may be like this too.

saulpw:
Okay.  Let's remove those blank lines in main.py and loaders/npy.py, and any others that still seem confusing, now that we have some rubric for taste.  Would you be willing to modify this PR to do that?

midichef:
Okay, here's that change.

saulpw (approved):

PR #2511, opened 2024-08-30
[sidebar-] turn off sidebar completely if disp_sidebar is off

midichef:
Setting `options.draw_sidebar = False` does not turn off the sidebar, as reported in #2507.

This PR fixes that. And removes an unnecessary return statement. The return value of `drawSidebar()` is never used, and in any case it is always `None`, since `drawSidebarText()` always returns `None`.


midichef:
Oops, I just saw https://github.com/saulpw/visidata/pull/2495, I see this PR is obsolete. I did feel like this issue had come up!

PR #2512, opened 2024-08-31
[sidebar] Ctrl+G to cycle in both main and input modes #2202

saulpw:


visidata/utils.py
```diff
@@ -186,19 +186,28 @@ def __bool__(self):
...
+            return ('{' + field_name + '}', field_name)
 
     def format_field(self, value, format_spec):
         # value is missing
         if value is None:
             return ''
         elif not value:
             return str(value)
-        return super().format_field(value, format_spec)
+        try:
```
saulpw: This try block doesn't do anything and shouldn't be necessary.

anjakefala:
This is part of https://github.com/saulpw/visidata/issues/2202

anjakefala (approved):
With this change, the sidebar guides will not show up by default. Users will be able to cycle through them via `Ctrl+G`.

The necessary next step is to somewhere add a visible message `Ctrl+G for help guides`. Perhaps it alternates with `Alt+H for menu help` in the top menu sidebar. However, this is not a blocker and can go into a second PR.

Issue #2313, opened 2024-02-14
[guides] Updating VisiData Documentation


anjakefala:
[Help us write guides](https://www.visidata.org/blog/2024/introducing-guides/#guides-are-organized-explanations-of-features)!
If someone has committed to working on a guide, they are tagged next to the guide. Checked guides have been written. 

If you want to help out, please comment which guide you would like to write.

[Guide to writing guides](https://www.visidata.org/docs/api/guides)

- [ ] **GuideIndex**: A Guide to VisiData Guides (you are here)
- [ ] **HelpGuide**: Where to Start and How to Quit
- [x] [**MenuGuide**: The VisiData Menu System](https://github.com/saulpw/visidata/pull/2168) (@anjakefala)
- [X] [**CommandsSheet**: How to find the command you want to run](https://github.com/saulpw/visidata/pull/2535) (@thejud)

Barebones Basics
- [ ] **MovementGuide**: Movement and search
- [ ] **InputGuide**: Input keystrokes
- [x] **SortGuide**: Sorting (@saulbert)
- [x] [**ColumnsGuide**: Rename, hide, and resize columns](https://github.com/saulpw/visidata/pull/2527) (@thejud)
- [x] [**TypesSheet**: Column types](https://github.com/saulpw/visidata/pull/2457) (@reagle)

Let's Rev VisiData Up
- [X] [**SelectionGuide**: Selecting and filtering](https://github.com/saulpw/visidata/blob/develop/visidata/guides/SelectionGuide.md) (@anjakefala)
- [ ] **ColumnsSheet**: Columns: the only way to fly
- [ ] **StatusesSheet**: Revisit old status messages
- [ ] **SidebarSheet**: Dive into the sidebar
- [ ] **SaversGuide**: Saving Data
- [X] [**ErrrorsSheet**: What was that error?](https://github.com/saulpw/visidata/blob/develop/visidata/guides/ErrorsSheet.md) (@anjakefala)
- [X] [**ModifyGuide**: Adding, Editing, Deleting Rows](https://github.com/saulpw/visidata/pull/2537) (@thejud)

The Varieties of Data Experience
- [x] [**SlideGuide**: Sliding rows and columns around](https://github.com/saulpw/visidata/pull/2529) (@thejud)
- [ ] **ExprGuide**: Compute Python over every row
- [ ] **JoinGuide**: Joining multiple sheets together
   - [ ] **vd.help_join**: Quick reference help sidebar available during Join command #1961
[...]

PR #2527, opened 2024-09-22
Add a new ColumnsGuide

thejud:
Per issue https://github.com/saulpw/visidata/issues/2313 . The ColumnsGuide describes how to rename, hide and resize columns.

### DIFF
--- a/visidata/guides/ColumnsGuide.md
+++ b/visidata/guides/ColumnsGuide.md
@@ -1,39 +1,4 @@
-# Rename, hide and resize columns
-
-How to rename, hide, and resize columns to focus on key data. These actions
-help organize the sheet, making it easier to prioritize information
-and reduce distractions. 
-
-# Renaming columns
-
-- {help.commands.rename_col}
-
-## Renaming one or more columns using data 
-
-Rename one or more columns using data from the sheet:
-
-- {help.commands.rename_col_selected}
-- {help.commands.rename_cols_row}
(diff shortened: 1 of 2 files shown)


visidata/guides/ColumnsGuide.md
```diff
@@ -0,0 +1,58 @@
...
+
+Rename columns to be valid python identifiers that can be used in column
+expressions.
+
+- {help.commands.normalize_col_names}
+- {help.commands.clean_names}
+
+# Hiding and Unhiding columns
+
+Clean up the sheet by hiding unwanted columns.  Hidden columns are still present, just not visible.
```
saulpw: In the upcoming version, hidden columns are 1 character columns on the right, which can be unhidden with `_`.

thejud: interesting! I see how that would make it much easier to understand what is going on.

anjakefala (approved):
I really appreciate how thoroughly you covered the features related to column visibility! It looks great in VisiData. 

Screenshot from 2024-09-22 18-25-58

I'm wondering if we want to manually type `resize-col-max` so that we can add an `(underscore)` to the right of the keystroke. Does it read clearly as an underscore?

saulpw:
> I'm wondering if we want to manually type `resize-col-max` so that we can add an `(underscore)` to the right of the keystroke. Does it read clearly as an underscore?

I think it's fine for now.

PR #2532, opened 2024-09-23
Add a new guide about the MemorySheet

thejud:
Per issue https://github.com/saulpw/visidata/issues/2313 . Add a guide for the MemorySheet

### DIFF
--- a/visidata/guides/MemorySheet.md
+++ b/visidata/guides/MemorySheet.md
@@ -1,17 +1,7 @@
 # The Memory Sheet 
 
-The MemorySheet provides a place to save cell values for later reference. It is also used by Visidata to store clipboard data when rows, cells or columns are copied.
-
-Create a memo entry from the value of a current cell.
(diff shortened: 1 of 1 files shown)


anjakefala (approved):
Wow, thank you! I'm curious, have you played with Memory Sheet? Do you find it helpful? How do you use it?

thejud:
I actually only used it because I saw it in the guide list and wondered what it did. I initially thought it might be useful to keep track of values as I worked through a set of data transformations, sort of like a little notebook or scratch pad. However, I was thinking that I might drop a few named rows in, etc... However, it probably makes sense just to create new sheets for this use case.

For the individual values, it still might be useful for quick storage. I would potentially like to try out shortcuts that save values to a particular place, like named registers in vim. This could probably be done with some macros and keybindings. 

As I was poking around with the writing the guide, I discovered that the the memory sheet provides a way to introspect the internal clipboard contents.

The short answer is that I haven't really found a way to use it effectively yet, but I produced the guide to document it since I was curious :)

PR #2534, opened 2024-09-23
Add a FrequencyTable Guide

thejud:
Per issue https://github.com/saulpw/visidata/issues/2313 Add a new FrequencyTable guide

For Anja:
- [x] Add numeric binning

### DIFF
--- a/visidata/aggregators.py
+++ b/visidata/aggregators.py
@@ -273,7 +273,7 @@ def chooseAggregators(vd):
             vd.warning(f'aggregator does not exist: {aggr}')
     return aggrs
 
-Sheet.addCommand('+', 'aggregate-col', 'addAggregators([cursorCol], chooseAggregators())', 'Add aggregator to current column')
+Sheet.addCommand('+', 'aggregate-col', 'addAggregators([cursorCol], chooseAggregators())', 'add aggregator to current column')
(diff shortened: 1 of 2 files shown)


thejud:
I'm working on revisions to this based on the PivotTable feedback. I plan to make it more reference-focused, with examples moved to the end.

anjakefala:
@thejud Ah okay! Is this ready for review?

thejud:
yes, but happy to revise per any guidance you have. I tried to make some of the changes you made to other documents, condensing, removing some of the expository text, etc...

anjakefala:
@thejud and I very much appreciate it <3

thejud:
lots of good cleanup. This one was quite challenging!
I thought that some of the usage notes, like extracting unique values, would be very helpful to new users. Also, select first rows is useful for solving other types of problems. Where would it make sense to put usage notes and tips like that? I was thinking that the guides would be a good place to go a bit deeper, and point out some things that might otherwise be missed/forgotten even for non-newbies.

anjakefala:
Hi @thejud!

I wanted to let you know that even if there are a lot of changes, please do not be discouraged. We are so, so grateful for you figuring out the structure and the bones for these guides. Without the structure you create my edits would not exist. And yes, tackling Frequency Table is especially a challenge.

> I thought that some of the usage notes, like extracting unique values, would be very helpful to new users.

It's a conundrum...that was difficult to wrap my head around. I think our original how to guides were this type of content: example workflows. Now that we have the guides, perhaps it makes sense to retire the bulk of the how to guides, and have an indexed page where we collect stuff like this. Let's bring it up in the discord.

I'm not against including them in guides.

Issue #2466, opened 2024-07-20
Guides: URL not working


reagle:
I am working on `TypesSheet.md` with the following content:

```
# Types

Columns usually begin as untyped (`anytype`), but can be set.

- {help.commands.type-any}
- {help.commands.type-currency}
- {help.commands.type-date}
- {help.commands.type-float}
- {help.commands.type-int}
- {help.commands.type-len}
- {help.commands.type-string}

*Note*: `float` uses the decimal separator; `currency` removes non-numeric characters and parses as float; `date` parses dates into ISO8601 format; `len` formats the cell value to the length of the content. See 
[:onclick https://www.visidata.org/docs/columns/]Columns[/].
```

I am not seeing a result: "Columns" is not differentiated, and when I click on it with mouse I get a warning. See [#2456](https://github.com/saulpw/visidata/issues/2456#issuecomment-2239001172) for screenshots.

======

saul.pw/VisiData v3.1dev on macOS 14.5 23F79 arm64

saulpw:
Thanks for reporting this, @reagle.  It should be fixed now.

Issue #2456, opened 2024-07-17
Guides: do guides support markdown tables? If so, which?


reagle:
The following markdown is not rendered in the guide, does it support markdown tables?

```markdown

| Command |   Type   |
|---------|----------|
| `~`     | string   |
| `#`     | int      |
| `%`     | float    |
| `$`     | currency |
| `@`     | date     |
| `z#`    | vlen     |
| `z~`    | anytype  |

```

saulpw:
No; we use simple lists instead.  And there's a helpful macro to turn a longname into a standard form with the keystroke and helpmsg, which is what we should use in this case I think.

reagle:
Thanks, I've realized the markdown is very limited -- not even nested bullets -- but that's probably fine for the task.

reagle:
What about openable URLs? For example, in the `TypesSheet.md` I submitted it could be useful to continue on to <https://www.visidata.org/docs/columns/> for more information.

anjakefala:
Here's an example of openable URLs =) 

> See [:onclick https://docs.python.org/3/library/codecs.html#standard-encodings]https://docs.python.org/3/library/codecs.html#standard-encodings[/]

reagle:
Thank you, I see that's in the [Guide Style Guide](https://www.visidata.org/docs/api/style). I added the following, but am not seeing a result:

> *Note*: `float` uses the decimal separator; `currency` removes non-numeric characters and parses as float; `date` parses dates into ISO8601 format; `len` formats the cell value to the length of the content. See 
[:onclick https://www.visidata.org/docs/columns/]Columns[/].

"Columns" is not differentiated, and when I click on it with mouse I get a warning.

<img width="377" alt="SCR-20240719-hqnj" src="https://github.com/user-attachments/assets/d208bbc6-bed3-4a33-b485-e98698c0535a">
<img width="373" alt="SCR-20240719-hqld" src="https://github.com/user-attachments/assets/7a2938ea-c0d2-41f0-9739-95105fc95f11">

anjakefala:
Could you open a new bug report?

PR #2536, opened 2024-09-24
Add the PivotGuide

thejud:
Per issue https://github.com/saulpw/visidata/issues/2313 Add a new PivotGuide

### DIFF
--- a/visidata/guides/PivotGuide.md
+++ b/visidata/guides/PivotGuide.md
@@ -1,77 +1,86 @@
+---
+sheet: Sheet
+---
 # Pivot Tables are just Frequency Tables with more columns
 
 Pivot Tables group data by two or more fields. They display one set of values as column headings, another set as row headings, and compute aggregates like the sum or total count for each cell.
 
-```
-+------------+-------+-------+
-| date       | color | price |
-+------------+-------+-------+
-| 2024-09-01 | Red   | 30    |
-| 2024-09-02 | Blue  | 28    |
-| 2024-09-02 | Green | 64    |
-| 2024-09-03 | Red   | 100   |
-| 2024-09-03 | Blue  | 33    |
-| 2024-09-03 | Blue  | 99    |
-+------------+-------+-------+
-```
-
-Pivoting on the `date` and `color` columns produces a pivot table that shows how many items of each color were sold each day. Note that here, only the row count, not the sum of prices, is computed for each cell.
-
-```
-+------------+-----+------+-------+
-| date       | Red | Blue | Green |
-+------------+-----+------+-------+
-| 2024-09-01 | 1   | 0    | 0     |
-| 2024-09-02 | 0   | 1    | 1     |
-| 2024-09-03 | 1   | 2    | 0     |
-+------------+-----+------+-------+
-```
-
 ## Create a Pivot Table
 
-To create a pivot table, 
+1. Set the primary column(s) as `key columns`. These values will appear along the left side as row headings.
 
-1. set the primary column(s) as `key columns`. These values will appear along the left side as row headings.
 - {help.commands.key_col}
 
-2. The default aggregation provides a total count of the items in each cell. Create any other aggregation columns before building the pivot table. Each aggregation will add a column for each of the pivot columns, (for example sum Red, mean Red, sum Blue, mean Blue, etc.).  
+2. **OPTIONAL** The default aggregation method for pivot cells is `count`. Create any other aggregation columns ( with [:keys]+[/]) before building the pivot table. Each aggregation adds a column for every pivot column. See **Examples** below.
 
-If custom aggregations have been added, the default 'total count' aggregation will not be computed.
+Note: Custom aggregations replace the default `count` aggregation.
+
 - {help.commands.aggregate_col}
 
-3. Navigate to the column that you want to use as the column headings in your pivot table, and create the pivot table:
+3. Navigate to the column containing the pivot column values. 
+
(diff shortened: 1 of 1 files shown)


saulpw:
Hey @thejud, I don't think we should have these bulky tables in the Guides.  The VisiData renderer can't handle them anyway.  Do you think there's a way to condense this?  This one might be pretty tricky.

thejud:
Hey @saulpw: Yes, I was torn about the tables. However, for melt and pivot in particular, seeing what happens was the best way for me to understand what was going on. I actually previewed them within VD to ensure that they would render.

Do you mean that these guides are simply too text heavy, or that the tables themselves are problematic because of wrappings, etc.., or both :)

Do you have suggestions about what you'd like to see?

image

thejud:
I'll try the following. Please met me know if this makes sense, or if you prefer alternatives:

1. Remove the Melt section... there is a separate guide for that anyway.

2. Remove the tables. 
    - Alternatively, would it be ok to simply condense them, e..g 2 char column names, less space, etc.. so they only take up, say 40 chars of width max?
    - Do you not want to have this sort of visual complexity (text tables) in the guides at all?

3. Condense/clean up some of the explanatory text. Looks like the intent of the guide is more like a quickref: show HOW to do specific things that the user knows about, not explain capabilities or concepts like pivot tables.

saulpw:
Yes, that makes sense.  "See also Melt" to reference the (currently non-existant) page.  And it's true, I don't like the visual clutter of the tables in the Guides, but I don't use the Guides so we're asking on Discord about whether the tables should be condensed or removed; 40 wide by 8 tall would be my suggested max, and using underline for the header and no outer border (borders only between cells).

Also, yes, I think we should only have explanatory text for features that are special/unique to VisiData.  If it's a term that can be easily looked up with a thousand examples on the web, then we don't need to be a tutorial (and that text gets in the way for people who already know about data, which are most VisiData users).

thejud:
Something like:

```
date       | color | price
-----------+-------+------
2024-09-01 | Red   | 30   
2024-09-02 | Blue  | 28   
2024-09-02 | Green | 64   
2024-09-03 | Red   | 100  
2024-09-03 | Blue  | 33   
2024-09-03 | Blue  | 99   
```

or even more minimal like:

```
date       Red Blue Green
2024-09-01 1   0    0
2024-09-02 0   1    1
2024-09-03 1   2    0
```

Note that I chose the original format because it is the most explicit, and because it is both easy to parse and easy to generate using miller (mlr)... I can easily write filters to clean it up to whatever, however.

As a (possible) point of interest, I generated and edited the table in visidata, copied it to the system clipboard as TSV (gs gY), pasted it into vim where I am writing the guide, and then filtered it through `mlr --t2p --barred cat` to transform TSV into pretty format with bars.

saulpw:
Could we do:

```
[:underline]date      | Red | Blue | Green[:]
2024-09-01 | 1 |  0  |  0
2024-09-02 | 0 |  1  |  1
2024-09-03 | 1 |  2  |  0
```

with spaces such that everything lines up?

reagle:
When I was working on a guide I asked about this in #2456 when I realized Guide's didn't support URLs (#2466), nested lists, nor simple tables. Actual markdown support (while adopting a syntax for visidata internals, such as commands) would be ideal, especially if Guides are supposed to be the source of downstream documentation, but that might be a significant effort in itself. (I imagine there are simple libraries in python with permissive licenses.)

With an eye toward markdown compatibility and future support, perhaps use a [standard/extent formats for tables](https://pandoc.org/MANUAL.html#tables) format rather than inventing a new one (e.g., [:underline] for heading row).

thejud:
yes, I definitely prefer a 'plain-text  compatible' format for tables. That's one of the reasons that I avoid the markdown table flavors: columns aren't always visually aligned. The pandoc format seems reasonable and does offer formatting benefits (left, center and right + multi-line cells.) Pandoc also appears to support variants of the "pipe format" that I used.

I wanted something that could be automatically converted back and forth for ease of maintenance, both during initial authoring and for future updates. I'm not quite sure about the tooling flow for converting TSV/CSV <-> pandoc tables, as I haven't worked with pandoc much.

That's part of the reason I shared the mlr workflow, since it seems like a ~standardized~ scriptable way of generating AND parsing the table. I haven't played with VisiData's fixed-width input format, but using VD itself to parse the table is intriguing.

anjakefala:
Hi @thejud Do you have everything you need to update this guide? There seemed to be a definite consensus of tables being welcome in the guides. Have you settled on a style from the conversations?

thejud:
@anjakefala  thanks for putting it ~to the vote~ up for comments :)

I added a simplified version of the tables. Hopefully this is reasonable middle ground. For what it's worth, I ended up running the TSV output through the python tabulate command with `pbpaste | tabulate -s '\t'`. There were some minor formatting cleanups required, but it seems reasonable given the infrequent need to update them.

I also moved the examples to a separate section at the end of the document with the idea that they could be additive, not interfere with quick lookup. This is an approach that pandas takes for their examples, which I find extremely helpful.

LMK if the example table should be added to the example data repository, or if it's fine existing only in the docs.

Thank you and @saulpw for working through these issues with me.

anjakefala:
I loved the example @thejud and the step-by-step guidance! We tried to clarify the description of Pivot Tables (they are so challenging to describe), and restructure the example prose to be `key (longname) to Foo`.

Those were the main things. Thank you so much for tackling these chunkier guides!

PR #2547, opened 2024-10-04
[open-] fix opening a dir with an filetype extension

midichef:
If a directory name looks like a file because of its extension (test.txt, test.tsv), Visidata opens it as if it were a file of that type.

`mkdir test.tsv; vd test.tsv`
```
  File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/loaders/tsv.py", line 89, in iterload
    with self.open_text_source() as fp:
         ^^^^^^^^^^^^^^^^^^^^^^^
[...]
  File "/usr/lib/python3.12/pathlib.py", line 1015, in open
    return io.open(self, mode, buffering, encoding, errors, newline)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
IsADirectoryError: [Errno 21] Is a directory: 'test.tsv'
```

This PR handles that case, including the case where the filename refers to a symlink to a directory.

### DIFF
--- a/visidata/_open.py
+++ b/visidata/_open.py
@@ -81,6 +81,10 @@ def guess_extension(vd, path):
 def openPath(vd, p, filetype=None, create=False):
     '''Call ``open_<filetype>(p)`` or ``openurl_<p.scheme>(p, filetype)``.  Return constructed but unloaded sheet of appropriate type.
     If True, *create* will return a new, blank **Sheet** if file does not exist.'''
+    # allow user to assign a filetype to a pathname:  options.set('filetype', 'csv', '-')
+    filetype = filetype or vd.options.getonly('filetype', str(p), None)  #1710
+    filetype = filetype or vd.options.getonly('filetype', 'global', None)
+
     if p.scheme and not p.has_fp():
         schemes = p.scheme.split('+')
         openfuncname = 'openurl_' + schemes[-1]
@@ -94,8 +98,10 @@ def openPath(vd, p, filetype=None, create=False):
     if not p.exists() and not create:
         return None
 
-    if not filetype:
-        filetype = p.ext or vd.options.filetype
+    # assign filetype from extension, but only for files, not directories
+    if not p.is_dir():  #2547
+        filetype = filetype or p.ext
+    filetype = filetype or vd.options.filetype
 
     filetype = filetype.lower()
 
@@ -146,9 +152,6 @@ def openSource(vd, p, filetype=None, create=False, **kwargs):
 
     if isinstance(p, BaseSheet):
         return p
-
-    filetype = filetype or vd.options.getonly('filetype', str(p), '')  #1710
-    filetype = filetype or vd.options.getonly('filetype', 'global', '')
 
     vs = None
     if isinstance(p, str):
--- a/visidata/shell.py
+++ b/visidata/shell.py
@@ -22,7 +22,7 @@ vd.option('dir_hidden', False, 'load hidden files on DirSheet')
 @VisiData.api
 def guess_dir(vd, p):
     if p.is_dir():
-        return dict(filetype='dir')
+        return dict(filetype='dir', _likelihood=10)
 
 
 @VisiData.lazy_property


saulpw (approved):
Who is even doing this? :)  Thanks for fixing it!

visidata/_open.py
```diff
@@ -108,6 +108,9 @@ def openPath(vd, p, filetype=None, create=False):
         vd.status('creating blank %s' % (p.given))
         return newfunc(p)
 
+    if p.is_dir():  # also matches symlinks to dirs
+        return vd.open_dir(p)
```
anjakefala: This is breaking the vgit tests! They operate on a dir but load them with `--filetype` git. Is there any way to accomodate that usecase?

frosencrantz:
There is also an issue with the filetype guessers and a path that is a directory.  The guessers should not report errors if the path is a directory.

midichef:
@saulpw 
Can you take a careful look at my changes? I don't understand most option handling in visidata, and I don't want to break anything before the 3.1 release.

I've moved two `filetype` assignments into `openPath()` so the logic is all in one place. And I changed the fallback from `''` to `None` to match the default parameter of `openPath()`.

What is `'global'` in `options.getonly`?
https://github.com/saulpw/visidata/blob/d50a1bf9e0f1a65c8cf65165fc2df7dbe7e4c2b3/visidata/_open.py#L151
And how does it differ from vd.options.filetype:
https://github.com/saulpw/visidata/blob/d50a1bf9e0f1a65c8cf65165fc2df7dbe7e4c2b3/visidata/_open.py#L98

Side note, I didn't know we could assign a filetype to a file by its path, like `options.set('filetype', 'csv', '-')`, that's very cool!

midichef:
And what should `guess_*` return to indicate it has no guess for a file: `{}` or `None`? I did `{}` for now.

visidata/_open.py
midichef: Oops! Okay, I've modified the PR to ignore the extension only for directories, and to accept the filetype specified by the user.

midichef:
Oh, I see the existing behavior of `guess_*` in the case of files that did not match the guess, was to return `None`. I've pushed a change to do that instead of return `{}`.

saulpw:
> What is 'global' in options.getonly?

This is if the option has been set, not as a default or on a specific sheet, but globally by the user (visidatarc or cli or from options sheet).  Whereas options.filetype normally looks at a hierarchy of places for the option to be set, either on the specific sheet, or the sheet class, or by the user, and finally falling back to the default.

visidata/loaders/archive.py
```diff
@@ -10,11 +10,13 @@
 
 @VisiData.api
 def guess_zip(vd, p):
+    if p.is_dir(): return
```
saulpw: This shouldn't be required for guessers; we let it raise the exception and it should be silently ignored.

visidata/_open.py
```diff
@@ -81,6 +81,10 @@ def guess_extension(vd, path):
 def openPath(vd, p, filetype=None, create=False):
     '''Call ``open_<filetype>(p)`` or ``openurl_<p.scheme>(p, filetype)``.  Return constructed but unloaded sheet of appropriate type.
     If True, *create* will return a new, blank **Sheet** if file does not exist.'''
+    # allow user to assign a filetype to a pathname:  options.set('filetype', 'csv', '-')
+    filetype = filetype or vd.options.getonly('filetype', str(p), None)  #1710
+    filetype = filetype or vd.options.getonly('filetype', 'global', None)
```
saulpw, line 86: Why were these lines moved?  It seems like code at line 102/103 should fix the issue by itself.

saulpw (changes requested):
The core fix looks reasonable, can we minimize the diff?

visidata/_open.py
midichef: The code at line 102 would fix it by itself. Lines 85, 86, and 104 are all similar in filling in filetype based on options. The intention was to put them together in one function, to make it easier to trace the logic that sets filetype based on options.

visidata/loaders/archive.py
midichef: Okay. My aim was to clean up the `--debug` results. But very few people will ever see those. And I am seeing that visidata style is to take full advantage of ignoring exceptions in such situations, and keep the code lean.

saulpw (approved):

anjakefala:
@midichef Is your expectation for this PR that now `vd test.tsv -f dir` will work, but `vd test.tsv` might still have a `IsADirectoryError: [Errno 21] Is a directory: 'test.tsv'`?

midichef:
> @midichef Is your expectation for this PR that now `vd test.tsv -f dir` will work, but `vd test.tsv` might still have a `IsADirectoryError: [Errno 21] Is a directory: 'test.tsv'`?

Oops, no. I've force-pushed a fix with this diff as the only change:
```
 def guess_dir(vd, p):
     if p.is_dir():
-        return dict(filetype='dir')
+        return dict(filetype='dir', _likelihood=10)
```
In the case of `vd test.tsv`, there were conflicting guesses for file type. `guess_tsv()` was saying `tsv` with likelihood `3`, and `guess_dir()` was saying `dir` with no likelihood, which is interpreted as `1`. Instead, the new version of `guess_dir()` uses a high priority to override. `10` matches the highest priority currently used in visidata, seen in a few loaders:  `api_airtable.py`, `http.py` and `xlsb.py`.

anjakefala:
Thanks @midichef! =)

PR #2552, opened 2024-10-09
[guide] Add a RegexGuide

thejud:
Per issue https://github.com/saulpw/visidata/issues/2313 add a new Regex Guide


anjakefala:
I think your instinct is correct that split and capture could use examples. I'll write some more detailed ones before we merge this!

thejud:
great examples!

halloleo:
This sounds so cool, thanks! 

I gues the RegexGuide is not yet in the [documentation](https://www.visidata.org/docs/), right?

What is a good time to check back for this in the documentation?

saulpw:
@halloleo This is part of the guides in vd itself (`vd --guides`).  We'll be putting them on the website eventually but probably not for awhile.  We could use some help with that if you or anyone is interested!

Issue #2267, opened 2024-01-23
visidata silently fails under Windows with Python 3.12 (3.11 fine)


clach04:
**Small description**

visidata fails to load/display files without error

**Expected result**

visidata to load correctly (or report an error)

**Actual result with screenshot**
If you get an unexpected error, please include the full stack trace that you get with `Ctrl-E`.

**Steps to reproduce with sample data and a .vd**

Either `pip install visidata` or from source code

1. follow https://github.com/saulpw/visidata/blob/6a1f17c961d3a4cc32b198b018169198f7e4e4c3/README.md?plain=1#L30
        `pip3 install git+https://github.com/saulpw/visidata.git@develop`
2. run `visidata`

Will see version information, but process exits without UI showing.

I tried different tags with no success. Works fine with Python 

Calling with a parameter, appears to load as there is a message to that effect but no UI loads up.

First try reproducing without any user configuration by using the flag `-N`.
e.g. `echo "abc" | vd -f txt -N`

Please attach the commandlog (saved with `Ctrl-D`) to show the steps that led to the issue.
See [here](http://visidata.org/docs/save-restore/) for more details.

**Additional context**
Please include the version of VisiData and Python.

any 3.0.x and 3.1

Python 3.12

    
    (py312visidata_pip) C:\tmp>visidata
    saul.pw/VisiData v3.0.2
    
    (py312visidata_pip) C:\tmp>python
    Python 3.12.0 (tags/v3.12.0:0fb18b0, Oct  2 2023, 13:03:39) [MSC v.1935 64 bit (AMD64)] on win32

EDIT See https://github.com/saulpw/visidata/issues/2119#issuecomment-2244526323 - suspected PDCurses bug (note link related to keyboard, unclear if this specific issue).

clach04:
(py3venv) C:\code\py\visidata>visidata --debug
    saul.pw/VisiData v3.0.2

Debug flag has no impact on reporting more details

clach04:
Appears to die inside:

https://github.com/saulpw/visidata/blob/6a1f17c961d3a4cc32b198b018169198f7e4e4c3/visidata/main.py#L341

tahir-hassan:
@clach04 I tried using Visidata on normal Python on Windows and it loads files fine, but there were two issues: 1. the mouse didn't work (I couldn't click on a cell), and 2. pressing double quote (") didn't produce a character.  I ended up installing Visidata via MSYS2, as it works within there.  

[...]

Issue #2119, opened 2023-11-17
Pressing " to filter selected rows does not work. Instead I get the error. No command for CTL_PADENTER


kalyan860:
**Small description**
Pressing " to filter selected rows does not work. Instead I get the error. No command for CTL_PADENTER

**Expected result**
Opening of selected cells as a filter in a duplicate sheet.

**Actual result with screenshot**
![image](https://github.com/saulpw/visidata/assets/95114049/9df2a24c-4a00-4a08-9aa0-5a44ef8484b2)

**Steps to reproduce with sample data and a .vd**
Select a few rows in a sheet with s key and then press " to filter them. 

Please attach the commandlog (saved with `Ctrl-D`) to show the steps that led to the issue.
#!vd -p
{"longname": "open-file", "input": "test2.xlsx", "keystrokes": "o"}
{"sheet": "test2", "col": "", "row": "\u30adtest2", "longname": "open-row", "input": "", "keystrokes": "Enter", "comment": "open current row with sheet-specific dive"}
{"sheet": "test2_test2", "col": "", "row": "1", "longname": "select-row", "input": "", "keystrokes": "s", "comment": "select current row"}
{"sheet": "test2_test2", "col": "", "row": "3", "longname": "select-row", "input": "", "keystrokes": "s", "comment": "select current row"}

**Additional context**
Please include the version of VisiData and Python.
Python 3.11.1
saul.pw/VisiData v2.11.1

saulpw:
What platform, what locale, what keyboard are you using?  This is an issue with your environment and probably not something we can fix.

kalyan860:
What platform? Windows11
What locale?
LCID             Name             DisplayName
----             ----             -----------
1033             en-US            English (United States)
What keyboard?
I have an ASUS laptop and using the keyboard that comes with it. 

Please help how I can fix this, I love visidata but without this functionality it will not be of much use to me.

kalyan860:
This is the list of all my environment variables on windows. Please see if this can help.

ALLUSERSPROFILE=C:\ProgramData
APPDATA=C:\Users\kalya\AppData\Roaming
ChocolateyInstall=C:\ProgramData\chocolatey
ChocolateyLastPathUpdate=133097932560588838
CommonProgramFiles=C:\Program Files\Common Files
CommonProgramFiles(x86)=C:\Program Files (x86)\Common Files
CommonProgramW6432=C:\Program Files\Common Files
[...]

Issue #2062, opened 2023-10-20
[windows] no windows-curses<2.3.1 for Python>=3.11


cool-RR:
This is on Windows 10 using the Bash shell that comes with Git.

```
$ pip install git+https://github.com/saulpw/visidata@bb0c95af89b03d28f79a3103d3d38785b50023a0
Collecting git+https://github.com/saulpw/visidata@bb0c95af89b03d28f79a3103d3d38785b50023a0
  Cloning https://github.com/saulpw/visidata (to revision bb0c95af89b03d28f79a3103d3d38785b50023a0) to c:\users\administrator\appdata\local\temp\pip-req-build-revt5xrz
  Running command git clone --filter=blob:none --quiet https://github.com/saulpw/visidata 'C:\Users\Administrator\AppData\Local\Temp\pip-req-build-revt5xrz'
  Running command git rev-parse -q --verify 'sha^bb0c95af89b03d28f79a3103d3d38785b50023a0'
  Running command git fetch -q https://github.com/saulpw/visidata bb0c95af89b03d28f79a3103d3d38785b50023a0
  Resolved https://github.com/saulpw/visidata to commit bb0c95af89b03d28f79a3103d3d38785b50023a0
  Running command git submodule update --init --recursive -q
  Preparing metadata (setup.py) ... done
Requirement already satisfied: python-dateutil in c:\program files\python311\lib\site-packages (from visidata==2.12.dev0) (2.8.2)
Requirement already satisfied: importlib-metadata>=3.6 in c:\program files\python311\lib\site-packages (from visidata==2.12.dev0) (6.8.0)
INFO: pip is looking at multiple versions of visidata to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement windows-curses<2.3.1; platform_system == "Windows" (from visidata) (from versions: 2.3.1a2, 2.3.1)
ERROR: No matching distribution found for windows-curses<2.3.1; platform_system == "Windows"
```

cool-RR:
I understand why it's happening now. I'm running Python 3.11 on Windows. VisiData requires `windows-curses<2.3.1` on Windows because of #1841 , which practically means `windows-curses==2.3.0`. However, because that version was released in 2021, it has binaries for Python versions 3.6 to 3.10, and no binary for Python 3.11.

saulpw:
[...]

PR #2563, opened 2024-10-14
[windows-] update windows-curses version to 2.4 #2119

### LINKED ISSUE #1841: Keystroke ] not detected on Windows

midichef:
`windows-curses` has released their version 2.4.0. They have updated the version of PDCurses they use, to fix problems inputting some keypresses. This version change should fix https://github.com/saulpw/visidata/issues/1841, https://github.com/saulpw/visidata/issues/2062, https://github.com/saulpw/visidata/issues/2119, and https://github.com/saulpw/visidata/issues/2267.

@daviewales could you see if installing `windows_curses` 2.4.0 works well with visidata? In particular, that visidata starts up as usual, without silently exiting immediately, and that the keypresses `"` and `]` are detected?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ setup(
     py_modules=["visidata"],
     install_requires=[
         "python-dateutil",
-        'windows-curses != 2.3.1; platform_system == "Windows"',  # 1841
+        'windows-curses >= 2.4; platform_system == "Windows"',  # 2119
         "importlib-metadata >= 3.6",
         'importlib_resources; python_version<"3.9"',
     ],


daviewales:
Tested with Python 3.12.
All of the above work. (starts, can use `"` and `]`.)
I noticed that it does not detect mouse clicks however.

anjakefala:
@daviewales Does the trade-off feel worth it to you? Improving support of keystrokes, but losing mouseclicks? Saul and I don't use Windows, so this is up to the community that does.

daviewales:
To add to the tradeoff: https://github.com/saulpw/visidata/issues/2267

I'd say that launching on 3.12 is fairly essential!

anjakefala:
Ah, good call!

midichef:
`windows_curses` 2.4.0 does not support Python 3.13. It's possible that Windows users trying to install the visidata `develop` branch with Python 3.13 will get an error:
```
ERROR: Could not find a version that satisfies the requirement windows-curses (from versions: none)
ERROR: No matching distribution found for windows-curses
```
(seen in [the windows_curses repo](https://github.com/zephyrproject-rtos/windows-curses) as Issues # 69 and # 70)

This will be fixed soon, as a fix has been submitted as PR # 71 in the `windows_curses` repo. I expect a 2.4.1 release of windows-curses to incorporate the fix. When it comes out, we should require it.

Until then, for any affected user, I'd recommend using Python 3.12.

--

EDIT to update, 2025-01-11:
windows-curses has released the version 2.4.1, and it does add support for Python 3.13

mplattner:
> Tested with Python 3.12. All of the above work. (starts, can use `"` and `]`.) I noticed that it does not detect mouse clicks however.

@daviewales, do you remember whether mouse clicks in visidata have ever worked on Windows? (I cannot.) If so, with which versions of visidata, Python, and windows_curses has it worked?

daviewales:
I can't say for sure, but I believe mouse support worked at least with Python 3.10 on a previous version of Visidata in Windows Terminal. If my memory is correct, then the latest version of Visidata prior to this change had working mouse support. (Some v3 version?)

mplattner:
I added a possible fix for the mouse not working in Windows 10: https://github.com/saulpw/visidata/issues/2676#issuecomment-2587884456

daviewales:
Thanks!

PR #2569, opened 2024-10-16
[graph-] fix to implement color_graph_refline option

### LINKED ISSUE #2566: Feature request: Option to customize refline color

midichef:
Closes #2566.


saulpw (approved):

saulpw (changes requested):
Let's either change the option name itself to color_refline, or add an optalias in deprecated.py.  It's minor but I'd like to encourage a habit of maintaining compatibility, at least when it's easy :)

midichef:
> Let's either change the option name itself to color_refline, or add an optalias in deprecated.py.

I added an optalias, and left the preferred name as`color_graph_refline`. That way the prefix is the same as the existing `color_graph_*` option `color_graph_axis`.

saulpw (approved):
Thanks, @midichef!

PR #2578, opened 2024-10-20
[features-] fix cursor index in hide-uniform-cols

### LINKED ISSUE #2577: [hide-uniform-cols] Couple issues.
In https://github.com/saulpw/visidata/pull/2453 @midichef added `hide-uniform-cols`.  Thank you!  I have started using it.  I have a couple suggestions:

1. The `hide-uniform-cols` does not properly handle where the current column is after this command is done.  The code calls column.hide(), which does not change the index of the current column.  So if a column left of the current column is hidden it should move the cursor left by one. If the current column is hidden it should also move the cursor one column left, if possible.

https://github.com/saulpw/visidata/blob/3e10140963a62a5f78fc6b60ebd2ac9264ef2772/visidata/column.py#L289 

2. It would be useful if the command reported (vd.status())when it hides a column the name of the column and its value.  So you know what was hidden.

midichef:
Closes #2577.

Repro case:
`echo "a,a\n1,1\n1,1" |vd -f csv`   then move the cursor to the right column, then `hide-uniform-cols`:
```
File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/sheets.py", line 780, in allAggregators
    col = self.availCols[vcolidx]
          ~~~~~~~~~~~~~~^^^^^^^^^
```

In this PR, I do the cursor move in `hide_uniform_cols()`, not inside `Column.hide()`. That's because I think of `hide()` as a low level width adjuster that does not conceptually involve the cursor, and we don't want to be calling `calcColLayout()` in it unnecessarily. Or does that leave the door open for a repeat of this error with other callers of `Column.hide()`?

### DIFF
--- a/visidata/features/layout.py
+++ b/visidata/features/layout.py
@@ -40,11 +40,16 @@ def hide_col(vd, col):
 def hide_uniform_cols(sheet):
     if len(sheet.rows) < 2:
         return
-    for col in sheet.visibleCols:
+    idx = sheet.cursorVisibleColIndex
+    for i, col in enumerate(sheet.visibleCols):
         vals = (col.getTypedValue(r) for r in sheet.rows)
         first = next(vals)
         if all(v == first for v in vals):
+            vd.status(f'hide {col.name} with value: {col.getDisplayValue(sheet.rows[0])}')
+            if i <= idx:
+                sheet.cursorRight(-1)
             col.hide()
+            Sheet.clear_all_caches()  #2578
 
 Sheet.addCommand('_', 'resize-col-max', 'if cursorCol: cursorCol.toggleWidth(cursorCol.getMaxWidth(visibleRows))', 'toggle width of current column between full and default width')
 Sheet.addCommand('z_', 'resize-col-input', 'width = int(input("set width= ", value=cursorCol.width)); cursorCol.setWidth(width)', 'adjust width of current column to N')


saulpw:
Does it work to just call `checkCursor()` after the columns are hidden?  That would be my preferred method, whether in the execstr or `hide_uniform_cols` or in .hide() itself.

midichef:
It almost works, but the problem is in one of the `checkCursor()` calls. The full stack trace shows the problem is in the new `allAggregators` function from 551b06c332ba40a38b9f1e4cb9569e8a10bddf50.
```
  File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/mainloop.py", line 252, in mainloop
...
  File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/sheets.py", line 643, in checkCursor
...                                            ^^^^^^^^^^^^^^^^
  File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/sheets.py", line 427, in nFooterRows
    return len(self.allAggregators) + 1
               ^^^^^^^^^^^^^^^^^^^
...
  File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/sheets.py", line 780, in allAggregators
    col = self.availCols[vcolidx]
          ~~~~~~~~~~~~~~^^^^^^^^^
```

It seems like `availCols` is using a cached value that is wrong due to the `col.hide()` calls. The `IndexError` goes away if I clear the cache with `Sheet.visibleCols.fget.cache_clear()` in `hide_uniform_cols()`, right after `col.hide()`. But the error still happens if I instead clear the cache on `availCols` with  `Sheet.availCols.fget.cache_clear()`.

What's the right solution here?

saulpw:
Hmmm...does `Sheet.clear_all_caches()` work?  That's the recommended way to clear these caches; instead of trying to find the exact right one to clear, just clear them all.   (Cache invalidation being one of the hardest problems in computer science.)

visidata/features/layout.py
```diff
@@ -40,11 +40,16 @@ def hide_col(vd, col):
 def hide_uniform_cols(sheet):
     if len(sheet.rows) < 2:
         return
-    for col in sheet.visibleCols:
+    idx = sheet.cursorVisibleColIndex
+    for i, col in enumerate(sheet.visibleCols):
         vals = (col.getTypedValue(r) for r in sheet.rows)
         first = next(vals)
         if all(v == first for v in vals):
+            vd.status(f'hid col {col.name} with value: {repr(first)}')
```
saulpw: Let's show the display value instead of the repr of the typed value.  Also I'd like this status to say "hide {col.name} with value {val}", to avoid the irregular past-tense verb and the abbreviated and unnecessary "col".

midichef: Okay, I've fixed that line for display value and wording. The past tense did feel out of step with other status messages.

midichef:
The `allAggregators` `IndexError` error is being triggered in other situations:
```File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/sheets.py", line 780, in allAggregators
    col = self.availCols[vcolidx]
          ~~~~~~~~~~~~~~^^^^^^^^^
```

For example, it can be triggered on a standard TsvSheet by pressing `i`, then `U`. So there is more work to do relating to `allAggregators`. I'll open a separate Issue for that.

saulpw (approved):
Thanks for fixing this!

Issue #2583, opened 2024-10-22
cannot open stdin when it is a tty


tkossak:
**Small description**
can't replay vdj file when data comes from stdin

**Data to reproduce**
create file `test.csv` with content:
```csv
id,Name
1,Ziutek
2,Barbara
3,Hermenegilda
```

create test vdj file (`test.vdj`) - it was generated by visidata with `Ctrl+d`:
```
#!vd -p
{"sheet": "-", "col": null, "row": "csv_doublequote", "longname": "set-option", "input": "False", "keystrokes": "", "comment": null, "replayable": null}
{"sheet": "-", "row": "csv_skipinitialspace", "longname": "set-option", "input": "False", "keystrokes": ""}
{"longname": "open-file", "input": "-", "keystrokes": "o", "replayable": true}
{"sheet": "-", "col": "id", "row": "", "longname": "hide-col", "input": "", "keystrokes": "-", "comment": "hide the current column", "replayable": true}
```

**Steps to reproduce**
Run: `cat test.csv | vd -p test.vdj` and observe the error  `cannot open stdin when it is a tty`. The csv file is not shown, only test.vdj content.

**Expected result**
csv file, without first column (should be hidden by test.vdj file)

**Actual result with screenshot**
![image](https://github.com/user-attachments/assets/3ea2c298-5d94-41e9-9e28-7585077c728e)

**Configuration**

- Does this issue reproduce without any plugins or configuration (using the `-N` CLI flag)? Yes.

- Does this issue reproduce with either the [latest release](https://www.visidata.org/releases/), or with the [develop branch](https://www.visidata.org/install/#update-visidata-from-an-existing-installation)? I have latest release: `saul.pw/VisiData v3.1.1`

**Additional context**

- What platform and version are you using (Linux, MacOS, Windows)?
Linux Manjaro

- Which version of Python? 3.11.9

- Which terminal are you using (for display and input issues)? Konsole

midichef:
Thanks for the report. This will be fixed soon. Until then, as discussed in https://github.com/saulpw/visidata/issues/2582#issuecomment-2430093541 you can for now try commenting out the line that has `open-file` and `"-"`:
```
{"longname": "open-file", "input": "-", "keystrokes": "o", "replayable": true}
```

Kossak-esky:
If it's of any use, I just installed vd from develop branch (3.2dev) and it gives me the same error.
Commenting out the line with "open-file" causes error `no sheet named -`

[...]

PR #2584, opened 2024-10-24
[open-] fix open-file for - in cmdlogs #2582

### LINKED ISSUE #2582: Replaying a cmdlog where input is read from stdin fails
**Small description**
I had written a script which got its input from stdin. It worked fine in v3.0.2, but with v3.1.1 it now fails with a "cannot open stdin when it is a tty" exception.

I traced the issue to these lines
https://github.com/saulpw/visidata/blob/2425a5fa3d9309338f7438c94c5f7002129f86e5/visidata/_open.py#L157-L159

Am I doing something wrong or is it supposed to work this way now?

**Data to reproduce**
Provided inline in next step.

**Steps to reproduce**
This is the minimal reproducible example I came up with.
```bash
echo -e '{"a":1}\n{"a":2}' | vd -f jsonl -p stdin.vdj --batch --debug -N
```
stdin.vdj has the following contents:
```jsonl
#!vd -p
{"sheet": "global", "col": null, "row": "regex_flags", "longname": "set-option", "input": "IS", "keystrokes": "", "comment": null, "replayable": null}
{"sheet": "global", "row": "filetype", "longname": "set-option", "input": "jsonl", "keystrokes": ""}
{"longname": "open-file", "input": "-", "keystrokes": "o", "replayable": true}
{"sheet": "-", "col": "", "row": "", "longname": "transpose", "input": "", "keystrokes": "Shift+T", "comment": "open new sheet with rows and columns transposed", "replayable": true}
```

**Expected result**
Input is read from stdin.

```
saul.pw/VisiData v3.1.1
opening stdin.vdj as vdj [visidata/_open.py:138:openPath]
finished loading stdin_vd [visidata/sheets.py:282:reload]
finished loading F6 [visidata/sheets.py:282:reload]
Your friend is your needs answered. [visidata/motd.py:30:domotd]
cannot open stdin when it is a tty [visidata/_open.py:159:openSource]
vd.push(openSource(inputFilename("open: "), create=True)) [visidata/basesheet.py:217:execCommand]
replay aborted during open-file [visidata/cmdlog.py:314:replayOne]
```

**Configuration**

  - yes

  - latest release

**Additional context**

  - Linux, Ubuntu 23.04

- Which version of Python?
  - Python 3.11.4

  - Alacritty

midichef:
Closes #2582 and #2583, which were bugs I introduced in #2442. The trouble is that the wrong file is tested for being a terminal. When `command | vd` is run in a terminal, `duptty()` eventually turns `sys.stdin` into a terminal. The right file to test is `stdinSource.fptext`.

@anjakefala How can I add some tests that run commands in a shell? I'd like to add tests that pipe to vd, like:
`cat sample.tsv | vd -`
`cat sample.tsv | vd -p -`
`cat sample.tsv | vd -p open-file.vdj`

They'll be good for testing the line of code added in this commit, when I submit a patch to make piped data be binary, instead of text.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -67,6 +67,10 @@ jobs:
         git --no-pager diff --numstat tests/
         git --no-pager diff --exit-code tests/
 
+    - name: Test replaying input from stdin
+      run: |
+          echo -e '{"a":1}\n{"a":2}' | vd -f jsonl -p dev/stdin.vdj --batch --debug -N
+
     - name: Ensure VisiData can create completions
       run: python -v dev/zsh-completion.py
 
--- a/dev/stdin.vdj
+++ b/dev/stdin.vdj
@@ -1,0 +1,5 @@
+#!vd -p
+{"sheet": "global", "col": null, "row": "regex_flags", "longname": "set-option", "input": "IS", "keystrokes": "", "comment": null, "replayable": null}
+{"sheet": "global", "row": "filetype", "longname": "set-option", "input": "jsonl", "keystrokes": ""}
+{"longname": "open-file", "input": "-", "keystrokes": "o", "replayable": true}
+{"sheet": "-", "col": "", "row": "", "longname": "transpose", "input": "", "keystrokes": "Shift+T", "comment": "open new sheet with rows and columns transposed", "replayable": true}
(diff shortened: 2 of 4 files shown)


midichef:
`vd -p -` would lead to a nonresponsive terminal. Similar to the way that `vd -` used to, before 49a1238a89a91af1c9a29bbcac74a84ede26cd81. So I've patched that as well.

saulpw (approved):
Thanks for owning this.

Issue #718, opened 2020-10-11
Error/warning text on manpage


jsvine:
**Small description**

When opening the manpage, I see the following error/warning text:

```
Usage: .No normal_text ... (#659)
Usage: .No normal_text ... (#1126)
mdoc warning: A .Bl directive has no matching .El (#1244)
```

**Expected result**

No error/warning text.

**Actual result with screenshot**

<img width="568" alt="Screen Shot 2020-10-11 at 1 17 55 PM" src="https://user-images.githubusercontent.com/534702/95685167-38d65c00-0bc4-11eb-9b51-265a726e9b6c.png">

**Steps to reproduce with sample data and a .vd**

Either `vd --help` on command line, or `Ctrl-h` within VisiData.

**Additional context**

VisiData 2.-5dev, commit 0b206baa3aa8e0366b339bf7ae97f04b0d6db5bc

Setup: MacOS, iTerm2, `tmux`, `bash`.

frosencrantz:
I guess this is what is a called a man-in-the-middle bug...  🙄

I'll  mention a loosely related issue I saw recently with trying to use VisiData with AJ's  Gitpod integration (from current head), which doesn't have man installed.  Visidata is not able to show the standard help page if the man command is not installed.

anjakefala:
> Visidata is not able to show the standard help page if the man command is not installed.

That is an expected failure! The manpage is written in `groff`, you need `man` installed to parse it. =( That is why we upload a copy of it online, for folks on systems without `man` installed. I will set it up so that you can view a copy of both 1.5.2 and 2.0 manpage online.

Also, thanks for reporting the bug @jsvine!

> I guess this is what is a called a man-in-the-middle bug... roll_eyes

Amusingly, I cannot find it right now, but I think on either the man pages for `man` or `groff` there is a line:

"Man is short for 'manual', not for the 'masculine person man'."

anjakefala:
Thanks again @jsvine! All of the errors should be fixed on the `develop` branch. This will be part of the next release.

PR #2590, opened 2024-10-30
[man-] silence warnings for two normal_text lines

midichef:
There are extra `No` commands at the end of  two lines in `vd.inc`. When `man` runs in `dev/mkman.sh` they cause some error lines:
```
Usage: .No normal_text ... (#785)
Usage: .No normal_text ... (#794)
```
This PR gets gets rid of them, just like 58d6eabec7ec7090acf45c79c836db78508f49e6 did for #718.


saulpw (approved):
Huh, I wonder how those happened...thanks for cleaning them up!

PR #2594, opened 2024-11-04
[keys] use prettykeys internally for input bindings and other getkeystroke uses

saulpw:
Replace legacy ^X with Ctrl+X and other curses KEY_ with internal prettykeys.

Does not replace bindings for commands, but everything else should use be using prettykeys now.


visidata/_input.py
```diff
@@ -227,50 +227,50 @@ def handle_key(self, ch:str, scr) -> bool:
...
+        elif ch == 'Ctrl+G':
             vd.cycleSidebar()
             return False # not considered a first keypress
-        elif ch in ('^H', 'KEY_BACKSPACE', '^?'):  i -= 1; v = delchar(v, i)
-        elif ch == '^I':                           v, i = self.completion(v, i, +1)
-        elif ch == 'KEY_BTAB':                     v, i = self.completion(v, i, -1)
-        elif ch in ['^J', '^M']:                   return True # ENTER to accept value
-        elif ch == '^K':                           v = v[:i]  # ^Kill to end-of-line
-        elif ch == '^N':
+        elif ch in ('Ctrl+H', 'Bksp', 'Ctrl+?'):   i -= 1; v = delchar(v, i)
```
anjakefala, line 267: I got "no command for Ctrl+?" when I typed Backspace

anjakefala:
I also tested `Shift+Z` and `Shift+F` and neither are working! I believe none of the Shifts are working.

Spacebar is also not working.

visidata/mainloop.py
```diff
@@ -212,7 +212,6 @@ def mainloop(vd, scr):
                 vd.warning('duplicate prefix: ' + keystroke)
                 vd.keystrokes = ''
             else:
-                keystroke = vd.prettykeys(keystroke)
```
anjakefala, line 217: It's the removal of this line that broke spacebar (`' '`) and the shift keys!

PR #2595, opened 2024-11-04
Fix ascii8 theme reflines typo

davay:
- [N/A] If contributing a core loader, [the loader checklist](https://visidata.org/docs/contributing#loader) was referenced.
- [N/A] If registering an external plugin, [the plugin checklist](https://visidata.org/docs/contributing#plugins) was referenced.

The Ascii8 theme has a typo, "disp_graph_lines_x_charset" / "disp_graph_lines_y_charset' are invalid options. It should be "reflines" instead of "lines".

### DIFF
--- a/visidata/themes/ascii8.py
+++ b/visidata/themes/ascii8.py
@@ -82,6 +82,6 @@ vd.themes['ascii8'] = dict(
     disp_menu_fmt='7-bit ASCII 3-bit color',
     plot_colors = 'white',
     disp_histogram='*',
-    disp_graph_lines_x_charset='||||',
-    disp_graph_lines_y_charset='----'
+    disp_graph_reflines_x_charset='||||',
+    disp_graph_reflines_y_charset='----'
 )


saulpw (approved):
Thanks for noticing and fixing this, @davay!  Do you use the ascii8 theme yourself?

davay:
Yes, I just started using visidata and I found ascii8 to be softer on the eyes, something about the default theme is a little too contrasty (like black backgrounds in places) -- side note, would you be interested in a test case to make sure all included themes are valid

saulpw:
Yes, a test case for this would be welcome!  Also, would you mind posting a screenshot of your terminal with the default VisiData theme?  I'd love if you pointed out the high-contrast areas you mentioned, it might be an oversight or otherwise fixable.

davay:
Sounds good, I'll get on it some time soon. Thanks for the project, pretty handy way to look at large csvs quickly -- this + xsv is very comfy 😃 

Default theme:

image
image
image

Compare this to: 

image
image
image

The Ascii8 theme is softer to my eyes -- I also use gruvboxdark everywhere so it fits the theme better

PR #2596, opened 2024-11-06
[cmdlog-] save prev replay when starting new replay #2531

### LINKED ISSUE #2531: Macro recordings that include a different macro include its invocation and its steps
The following shows that macro `test` will have **both** the invocation and the steps of the `r1` macro, leading to problems. Being able to nest macros would be more powerful (in case I change `r1` I don't have to change any other macros that depend on it) but I suspect simply not recording the invocation would be easier to implement.

r1
```
{"sheet": "", "col": "", "row": "", "longname": "columns-sheet", "input": "", "keystrokes": "Shift+C", "comment": "open Columns Sheet: edit column properties for current sheet"}
{"sheet": "", "col": "name", "row": "", "longname": "select-col-regex", "input": {"regex": "^(author|created_utc|selftext|title|url)$", "flags": "I"}, "keystrokes": "|", "comment": "select rows matching regex in current column"}
...
```

=======

test
```
{"sheet": "", "col": "", "row": "", "longname": "exec-longname", "input": "exec-r1", "keystrokes": "Space", "comment": "execute command by its longname", "replayable": true}
{"sheet": "", "col": "", "row": "", "longname": "exec-r1", "input": "", "comment": "", "replayable": true}
{"sheet": "", "col": "", "row": "", "longname": "columns-sheet", "input": "", "keystrokes": "Shift+C", "comment": "open Columns Sheet: edit column properties for current sheet", "replayable": true}
...
```

midichef:
This seems to fix [@reagle's nested macro execution test case](https://github.com/saulpw/visidata/issues/2531#issuecomment-2385692036).

This PR saves and restores `vd.currentReplay`. That's enough to fix nested macros.

I also save and restore  `vd.currentReplayRow` although I don't think it currently is necessary. I do that because `replay_cancel()` sets both `currentReplay` and `currentReplayRow`, so it seems like both should be updated at the same time.

### DIFF
--- a/visidata/cmdlog.py
+++ b/visidata/cmdlog.py
@@ -330,6 +330,9 @@ def replay_sync(vd, cmdlog):
     with vd.DisableAsync():
         vd.sync()  #2352 let cmdlog finish loading
         cmdlog.cursorRowIndex = 0
+        # save current replay, for cmdlogs that replay other cmdlogs, such as a macro executing another macro
+        prev_replay = vd.currentReplay
+        prev_replay_row = vd.currentReplayRow
         vd.currentReplay = cmdlog
 
         with Progress(total=len(cmdlog.rows)) as prog:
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Thanks @midichef, this makes sense to me.

Issue #2570, opened 2024-10-16
Feature request: Live updating plots


cool-RR:
It is time for the greatest VisiData feature request of all time. I've wanted this for a long time, and it's pretty ambitious so it might never get implemented, but let's put it on the table anyway.

I want to have live plots. I'm running experiments that take a while to produce output. I'm anxiously awaiting the interesting parts of their results so I could draw conculsions and run the next experiment. I want to see that progress on a plot, live. This means I want to draw a plot of a certain sheet, and then as more rows are written to that sheet, I want the plot to automatically add new points on the screen.

midichef:
Hi @cool-RR, I submitted #2597 and #2598. Once you install both of those, you should be able to monitor your graphs with `reload-every`.

It was not especially complicated to implement as the basic functionality was already implemented. So imo the greatest VisiData feature request of all time is still #2: `make a package for visidata`.

Also, until these patches are incorporated into visidata, you may find the `watch` [shell tool](https://en.wikipedia.org/wiki/Watch_(command)) useful. Something like:  `watch -n 10 vd -p draw_graph.vdj`

cool-RR:
Hmm, this doesn't work for me. I checked out the develop branch and then merged your two branches. I did `pip install -e .` and I ran VisiData on a jsonl file that's getting new data pushed into it continuously. First I tried to use reload-modified on the sheet and draw a plot, and increased the x range expecting to see more points automatically, but I didn't. Then I went back to the sheet and tried reload-every instead which didn't work. Am I doing something wrong.

anjakefala:
Hey @cool-RR!

To make things easier, both commits are now in `develop`.

The steps I would do are:
`pip uninstall visidata` - Make sure this is the only version of VisiData in your environment
`pip install git+https://github.com:saulpw/visidata.git`

And then try it out. I believe the step is to use `reload-every` on the underlying sheet and on the graph sheet. Let us know if you get it working!

midichef:
Hmm, I hadn't actually tested reloading graphs while another sheet reloads from a file. That case is not working. It looks like the source sheet changes are not being reflected on the graph. I'll continue to look into it.

(My test case was a custom sheet that generates more rows over time in memory, not on disk.)

cool-RR:
[...]

PR #2598, opened 2024-11-06
[canvas-] clear box sizes on reset()

midichef:
This change will help support the use case of monitoring graph sheets by repeatedly reloading them.

Currently, when a graph is reloaded, it won't resize itself to show new datapoints that are outside the current plot view box. After this PR, it will recalculate its size just as if the new data had been present during the first graph draw.

If this PR and #2597 are both accepted, #2570 can be closed.


saulpw (approved):
Sure.

PR #2602, opened 2024-11-10
[aggr-] allow undo for aggregate-col/cols

midichef:
After adding an aggregator to a column with `+`, it can't be removed with `undo-last`. This PR adds undo.

It also removes an old check for `hasattr(col, 'aggregators)` that is obsolete, now that `Column.aggregators` has 
 become a property.
https://github.com/saulpw/visidata/blob/bee4116c94f650e6200f5743a551b5a3d59f0058/visidata/aggregators.py#L75

### DIFF
--- a/visidata/aggregators.py
+++ b/visidata/aggregators.py
@@ -3,6 +3,7 @@ import math
 import functools
 import collections
 import statistics
+from copy import copy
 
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Makes sense to me!  Thanks @midichef.

PR #2614, opened 2024-11-29
[sheets-] fix error adding col to sheet with no visible cols

midichef:
Adding columns to an empty sheet gives an error: `echo hi |vd`   then `-` then `i`.

```
  File "addcol-incr", line 1, in <module>
  File "/home/midichef/.venv.vd_develop/lib/python3.12/site-packages/visidata/sheets.py", line 579, in addColumnAtCursor
    index = self.columns.index(ccol)+1
ValueError: <Column: > is not in list
```


saulpw (approved):
Thanks, @midichef !

PR #2615, opened 2024-11-29
[features-] add sysedit-cell command, using external editor

### LINKED ISSUE #2500: Feature request: Text-editing conveniences when editing a cell
I often edit text in cells by pressing `e`. I'm annoyed by how basic the text-editing interface is.  Here are a few things I'd want:

1. Bind both `Ctrl-Backpace` and `Alt-Backspace` to delete a word.
2. Bind both `Ctrl-Delete` and `Alt-Delete` (or `Alt-D`) to delete a word forward.
3. More difficult: Allow selecting text by pressing `Shift-Right`, `Shift-Left`, and selecting whole words by pressing `Shift-Ctrl-Right` etc. Though then we'd have to think how the clipboard would work in that context.

Any other suggestions are appreciated.

midichef:
Addresses a feature requested in the replies to #2500, thanks and credit to @sourissouriante for implementing this.

I could have used `setValue()` which is lower overhead (no `@asyncthread` and no `Progress`). But it does not have undo functionality built in, so I went with `setValues()` that does have undo.

### DIFF
--- a/visidata/features/sysedit.py
+++ b/visidata/features/sysedit.py
@@ -1,5 +1,5 @@
 from copy import copy
-from visidata import vd, asyncthread, Path, Sheet, IndexSheet
+from visidata import vd, asyncthread, Path, Sheet, IndexSheet, TableSheet
 
 
 @Sheet.api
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Works for me!

PR #2617, opened 2024-11-29
[archive,sqlite-] guess sqlite/tar/zip filetypes confidently

midichef:
This increases the likelihoods guessed when using a few of the guessers that inspect binary data. Right now `guess_sqlite/tar/zip` give an inferred likelihood of 1:
https://github.com/saulpw/visidata/blob/7e75dee682b81e7e9206a9b8d0ab4e695cdff8bd/visidata/_open.py#L66

After this PR they would give a likelihood of 10, in line with `guess_xls`, `guessurl_mimetype` and, `guessurl_airtable`.

As Visidata operates now, I don't think this PR makes any difference to filetype determination in practice. But it could if changes are made to the guessing algorithm later.

### DIFF
--- a/visidata/loaders/archive.py
+++ b/visidata/loaders/archive.py
@@ -5,18 +5,21 @@ import datetime
 from visidata.loaders import unzip_http
 
 from visidata import vd, VisiData, asyncthread, Sheet, Progress, Menu, options
-from visidata import ColumnAttr, Column, Path
+from visidata import ColumnAttr, Column, Path, filesize
 from visidata.type_date import date
 
 @VisiData.api
 def guess_zip(vd, p):
     if not p.is_url() and zipfile.is_zipfile(p.open_bytes()):
-        return dict(filetype='zip')
+        return dict(filetype='zip', _likelihood=10)
 
 @VisiData.api
 def guess_tar(vd, p):
+    # an empty file will pass is_tarfile(), but can't be opened by tarfile.open()
+    if filesize(p) == 0:
+        return None
     if tarfile.is_tarfile(p.open_bytes()):
-        return dict(filetype='tar')
+        return dict(filetype='tar', _likelihood=10)
 
 @VisiData.api
 def open_zip(vd, p):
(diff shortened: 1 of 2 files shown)


midichef:
One such change that would make this PR matter, would be lowering the priority of file extensions. Currently, file extensions override filetype guessing completely:
https://github.com/saulpw/visidata/blob/7e75dee682b81e7e9206a9b8d0ab4e695cdff8bd/visidata/_open.py#L121-L124
Because of that override, `guess_extension()` never determines a filetype in practice. It never actually runs when there is a file extension. Its implementation details show it would assign a lower confidence to extension-based filetypes, using the relatively weak likelihood of 3:
https://github.com/saulpw/visidata/blob/7e75dee682b81e7e9206a9b8d0ab4e695cdff8bd/visidata/_open.py#L72-L77
But if that override is ever removed (by removing lines 121-123), then this PR would improve file guessing.

For example, let's say we have a file with a `.txt` extension, containing a SQLite database. Right now such a file is `txt` with no file guessing, because `guess_extension()` never runs. And it's `txt` with or without this PR. But in the hypothetical world where the override is removed, then without this PR, the file would be guessed as `txt` (likelihood: 3), not `sqlite` (likelihood: 1). With this PR, it would be guessed as `sqlite` (likelihood:10).

saulpw (approved):
Notably, a text file that starts with "SQLite format considerations" can't be opened.  But as long as `-f txt` works in that case I guess it's fine.  (Though I'm also wondering how much overhead these guessers will add to the startup time for every file).

midichef:
Oh, good example. False positives would be something to keep in mind.

For sqlite files in particular, once it mattered, at that point we could switch to the more specific search string `'SQLite format 3\000'`. It looks like [3 is the only version number ever used](https://www.sqlite.org/fileformat.html) with this header format, and [versions 2.0 and 2.1](https://stackoverflow.com/questions/54950214/is-sqlite-2-database-header-format-different-from-that-of-sqlite-3/69722897#69722897) had entirely different formats.

But I don't see any need to switch the guessing design right now. Low overhead is good.

If visidata goes in the direction of advanced file guessing, another tool to keep in mind is [python-magic](https://github.com/ahupp/python-magic).

saulpw (approved):

PR #2618, opened 2024-11-30
[errors-] add sysopen-error command to view vd source code in editor

midichef:
This PR adds the ability to press `Enter` in an `ErrorSheet` traceback to open an external editor at the relevant line of Visidata's source code.

### DIFF
--- a/visidata/textsheet.py
+++ b/visidata/textsheet.py
@@ -1,4 +1,5 @@
 import textwrap
+import re
 
 from visidata import vd, BaseSheet, options, Sheet, ColumnItem, asyncthread
 from visidata import Column, vlen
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Do you use this yourself, @midichef ?  (I'm not keen on adding code that seems neat unless it's actually useful for someone).

midichef:
I do use it. I've had this feature in my .visidatarc for about a year. I didn't submit it earlier, because I felt it wouldn't be broadly useful.

But now I've realized it could be helpful for power users who report issues, by making it easy for them to look into the details of errors.

That said, I use a modded version, that directs the edits to my local github repo instead of my pip installation of visidata. And I'll have to keep that version in my .visidatarc whether or not this PR is taken. So it's really up to other users.

PR #2620, opened 2024-12-02
[errors-] fix error-recent always showing "no error"

midichef:
`error-recent` outputs `no error` in the statusbar even when an error exists to be inspected, because [`vd.push()`](https://github.com/saulpw/visidata/blob/c72f696690e54f4e926685c497ca8717e3b920d2/visidata/sheets.py#L1084) only ever returns None.

### DIFF
--- a/visidata/textsheet.py
+++ b/visidata/textsheet.py
@@ -95,7 +95,7 @@ def recentErrorsSheet(self):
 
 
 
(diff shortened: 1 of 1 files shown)


saulpw (approved):
I see now that Python is different than Perl :)

PR #2622, opened 2024-12-03
[sheets-] record key-col toggle for replay as key-col-on/-off

midichef:
I'm following up on @saulpw's [comment about trouble caused by keycol toggling in cmdlog replay](https://github.com/saulpw/visidata/discussions/1877#discussioncomment-5760360).

>it's tricky to change behavior based on existing column state, as this can make cmdlog replay a lot less predictable. We already have this with e.g. key-col and it's caused some problems. Maybe the ultimate answer is to have key-col put a more precise command like key-col-on or key-col-off (please help come up with better longnames if we do this) on the cmdlog.

Does the problem with key-col for cmdlog-replay still exist? If so, this PR should fix it, by recording `key-col-on` or `key-col-off` instead of `key-col`.

I'm preparing to submit the feature proposed in that Discussion:  changing a column's sort direction while preserving other column's sort direction. So if this kind of fix is still needed for `key-col`, then I'll apply the same idea to `sort-asc` and the commands like it.

### DIFF
--- a/visidata/deprecated.py
+++ b/visidata/deprecated.py
@@ -252,3 +252,12 @@ def keystr(sheet, row):
     return sheet.rowname(row)
 
 vd.optalias('color_refline', 'color_graph_refline') # color_refline was used in v3.1 by mistake
+
+@deprecated('3.2', '[self.unsetKeys([c]) if c.keycol else self.setKeys([c]) for c in cols]')
+@visidata.TableSheet.api
+def toggleKeys(self, cols):
+    for col in cols:
+        if col.keycol:
+            self.unsetKeys([col])
+        else:
+            self.setKeys([col])
--- a/visidata/macros.py
+++ b/visidata/macros.py
@@ -107,9 +107,10 @@ def afterExecSheet(cmdlog, sheet, escaped, err):
     if not vd.activeCommand: return
     if vd.activeCommand.longname == 'macro-record': return
 
-    cmd = copy(vd.activeCommand)
-    cmd.sheet = ''
-    vd.macroMode.addRow(cmd)
+    if vd.activeCommand.replayable:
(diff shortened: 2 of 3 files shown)


saulpw (approved):
It hadn't quite occurred to me to do it this way, but you know what, I'll take it.  Nice job.

midichef:
Hmm, this PR still needs work. I hadn't thought about the case of macros.

With this change, when `key-col` is used in a macro, the macro will contain two commands, the original `key-col` and `key-col-on` (or `-off`). So that needs a fix.

But there's also a UI difference between macros and replay logs that makes this approach difficult. Users may expect `!` in a macro to perform toggling. Whereas in a cmdlog replay file that's only replayed once, what is desired is setting the keycol state, not toggling.

midichef:
>With this change, when key-col is used in a macro, the macro will contain two commands, the original key-col and key-col-on (or -off). So that needs a fix.

Okay I've fixed this problem. However, users who use macros to replay movement commands will find that this change breaks that use case. Commands like `go-left` are already excluded from command logs, and this PR now excludes them from macros.

To save movement commands in macros, add this to `.visidatarc`:
```
for longname in ['go-left', 'go-right',
                 'go-down', 'go-up',
                 'go-left-page', 'go-right-page'
                 'go-pagedown', 'go-pageup' ]:
    for sheet_type, cmd in vd.commands.get[longname]:
        cmd.replayable = True
```

midichef:
Oops, I just noticed that when I force-pushed the followup commit for macros, I had deleted my original commit about keycols. I've just brought it back as [7e1ecbd](https://github.com/saulpw/visidata/pull/2622/commits/7e1ecbdf253b10d255b7c167443dbd9ff0374c0d). It is unchanged, except for rebasing it onto the latest develop branch.

Issue #2634, opened 2024-12-13
[question] Replaying command log file is significantly slower in release 3.1.1 compared to 2.11


DmitryAndreyeff:
Replaying command log file is significantly slower in release 3.1.1 compared to 2.11.

Based on [this](https://github.com/saulpw/visidata/discussions/2628#discussioncomment-11489070) discussion, attaching [files.zip](https://github.com/user-attachments/files/18126892/files.zip) to reproduce the issue.

env details:
Mac OS Sonoma 14.6
Python 3.12.7
Visidata 3.1.1

Please advise.

midichef:
Okay, I can reproduce some slowness, it takes about 1.6 seconds longer on v3.1.1 vs 2.11.1.

How fast is it when you add `-b -i` to your replay command:  `vd -b -i -p sheets.vdj`. For me that restores the quick playback that I see in v2.11.1. Does it restore the previous quickness for you?

DmitryAndreyeff:
Hi @midichef , indeed, these flags make the playback in v3 as fast as it was in v2. Thank you very much!

midichef:
Okay, then I know what caused the change. We should be able to figure out a fix for the slowness, so you won't need to do `-b -i` in the future. (And `-b -i` has some definite drawbacks, like not being able to see data on-screen during loading, which could be a long time for a slow sheet.)

DmitryAndreyeff:
@midichef , actually `-b` flag breaks completely the ability to edit cells.

midichef:
Thanks @DmitryAndreyeff for reporting that. I have created an issue #2639 for that problem.

For now, I would recommend applying the fix in #2635, which is this commit: 87c99f215182930e12814b96f5ac35dfa2db078a. It will let you avoid using `-b -i`. It will allow `edit-cell` to work as usual. It will remove some of the visible delay in your replays, though not all of it.

PR #2635, opened 2024-12-15
[mainloop-] wait less time between quick replay commands #2635

midichef:
Addresses #2634.

The problem is new to v3.1. It came from my patch 02432ff86ec91c36ed53c507319402d7d7c054b0, which fixed slowness during replay of long CPU-heavy commands (#2369). Screen redraws happened too frequently, taking CPU time away from command execution. The fix introduced a delay of 0.1 seconds (`nonidle_timeout`) between screen redraws, as intended. But unfortunately, the delay also happens between each command. That slows down files that contain many replay commands. For example, a replay file (from issue #2634, specifically `sheets.vdj` contained in the issue creator's `files.zip`) with 14 lines runs 1.5 seconds slower than it did in v3.0.

This PR gets rid of the 0.1 s wait for commands when no background thread is ongoing. The intention is to slow down redraws/replays only after running a threaded command that may run for a long time, like `open-file`. The change fixes most of the delay in replay, though it does not bring back the full speed of v3.0. 

For the example file `sheets.vdj`, the slowness compared to v3.0 shrinks from +1.5 s to +0.6 s. The remaining +0.6 s occurs because there are 4 commands that create a 0.1 s delay: `open-file` and each of the three `pivot` commands. Each of those creates a new sheet, and each new sheet makes a background thread for `reload()`. Each background thread creates a 0.1s delay during replay.

Overall, the bugs #2369 and #2634 come from the structure of `mainloop()`: every time through the loop, it draws the screen and replays a command, and then waits out the delay. To get the speed back to the level of v3.0, we need the loop to use different delays for replay vs. redraw, simultaneously. So during replay, vd could run the mainloop more often than now, but not draw the screen every time through the loop, and not replay a command every time through. But implementing that change is too complex for me to tackle.

### DIFF
--- a/visidata/mainloop.py
+++ b/visidata/mainloop.py
@@ -251,13 +251,14 @@ def mainloop(vd, scr):
         vd.checkForFinishedThreads()
         vd.callNoExceptions(sheet.checkCursor)
 
-        # no idle redraw unless background threads are running
         time.sleep(0)  # yield to other threads which may not have started yet
         if vd._nextCommands:
-            if vd.options.replay_wait > 0:
-                vd.curses_timeout = int(vd.options.replay_wait*1000)
-            else:
+            if vd.unfinishedThreads:  #2369 #2635
+                # while running a bg thread for a command, schedule infrequent redraws
                 vd.curses_timeout = nonidle_timeout
+            else:
+                # otherwise, schedule the next redraw and command (immediately, for default replay_wait)
+                vd.curses_timeout = int(vd.options.replay_wait*1000)
         elif vd.unfinishedThreads:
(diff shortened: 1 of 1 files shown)


saulpw (approved):
This loop is tricky.  We want to show every step in the replay, which requires at least a redraw (use `-b -i` to not see every step).   But I agree, even with this requirement, the delays after showing add up.

anjakefala:
Thanks for your patience, @midichef, I'm checking all the PRs now. I checked this by replaying with some of my larger cmdlogs, and basic replay functionality still seems fine, and the replays are indeed, much faster. `-w` is also still working. 

Thanks!

PR #2636, opened 2024-12-15
[sheets-] speed up loading check of max_rows

midichef:
Affects vd v3.1. After 7c24ffe7, loading sheets became much slower. A 2 million row CSV sheet with multiple columns went from 10 seconds to 16 seconds.

I didn't look into the root cause of why `vd.options.max_rows` is so slow. Perhaps that's the proper fix instead?

### DIFF
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -1054,7 +1054,7 @@ class SequenceSheet(Sheet):
         # add the rest of the rows
         max_rows = self.options.max_rows
         for i, r in enumerate(vd.Progress(itsource, gerund='loading', total=0)):
-            if self.precious and i == max_rows:
+            if self.precious and i >= max_rows:
                 break
             self.addRow(r)
 
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -303,8 +303,9 @@ class TableSheet(BaseSheet):
         self.rows = []
         try:
             with vd.Progress(gerund='loading', total=0):
+                max_rows = self.options.max_rows
                 for i, r in enumerate(self.iterload()):
-                    if self.precious and i > self.options.max_rows:
+                    if self.precious and i == max_rows:
                         break
                     self.addRow(r)
         except FileNotFoundError:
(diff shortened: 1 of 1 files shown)


visidata/sheets.py
```diff
@@ -303,8 +303,9 @@ def loader(self):
         self.rows = []
         try:
             with vd.Progress(gerund='loading', total=0):
+                m = self.options.max_rows
```
saulpw: prefer to name the local variable `max_rows` instead.

saulpw:
Option lookup is slow.  It should be cached, but still, accessing the cache in an inner loop called millions of times is unnecessary.  So this is the reasonable fix.

visidata/sheets.py
```diff
@@ -303,8 +303,9 @@ def loader(self):
         self.rows = []
         try:
             with vd.Progress(gerund='loading', total=0):
+                max_rows = self.options.max_rows
                 for i, r in enumerate(self.iterload()):
-                    if self.precious and i > self.options.max_rows:
+                    if self.precious and i == max_rows:
```
saulpw: Let's stick with `>`.  This `==` actually stops a row earlier.

midichef: With `>` I'm getting an extra row:
```
seq 5 | vd -f txt --max-rows=3 -b -o -   # with >
saving 1 sheets to - as txt
1
2
3
4
```
But with `==` it seems right to me:
```
seq 5 | vd -f txt --max-rows=3 -b -o -  # with ==
saving 1 sheets to - as txt
1
2
3
```

```
seq 5 | vd --header=0 -f tsv --max-rows=3 -b -o - # with ==
saving 1 sheets to - as tsv

1
2
3
```

saulpw (approved):
Okay, thanks for your diligence.  I made it >= so we can have the right number of rows and also stop in case it goes over for some reason.  (@anjakefala please squash)

midichef:
Oh, good idea. I'll remember to consider `>=` instead of `==` for termination comparisons.

Issue #2645, opened 2024-12-18
command palette ranks results in unexpected ways


midichef:
**Small description**
See #2643 for spaces in queries not being handled well:  a search for `sele arou` does not find `select-around-n`, but a search for `selearou` does.
A search for `EDITOR` does not rank commands like `sysedit-cell` at the top. Neither does a search for `$EDITOR`. But a search for `EDIT` does.

**Steps to reproduce**
`exec-longname` then `sele arou` vs. `selearou`
`exec-longname` then `EDIT` and `EDITOR` and `$EDITOR`

**Configuration**
vd v3.2dev

PR #2646, opened 2024-12-20
[cmdpalette] fix scoring of space-separated search terms

midichef:
Fixes part of #2645.

When multiple search terms are used in `fuzzymatch()`, not all the terms are used for a match. Only the last matching term is used to score each part of the match. That is, for `longname`, only the last matching term's score is used, and for `description`, only the last matching term's score is used.

To see this, hit `Space`, then type `select a`. `select-after` should be the top result, but it's not shown at all. (The proper rankings should look more like what you see searching for `selecta` with no space.)

That's because `select-after` is incorrectly scored as a match for `a` only, not as a match for `select` and a match for `a`. So it is incorrectly pushed way down the ranking. That leaves `select-error-col` as the top result, because it does not have any match for `a` dragging it down. It gets the highest score by matching solely for `select` and not for `a`.

This PR lets each term contribute to a combined score. With it, `select a` properly ranks `select-after` and `select-around-n` as the top two results. (There are remaining differences in the rankings with `selecta` vs `select a`, because the `select a` search adds a bonus to results where `description` has `a` at the start of a word instead of just in the middle.)

### DIFF
--- a/visidata/fuzzymatch.py
+++ b/visidata/fuzzymatch.py
@@ -375,15 +375,17 @@ def fuzzymatch(vd, haystack:"list[dict[str, str]]", needles:"list[str]) -> list[
         formatted_hay = {}
         for k, v in h.items():
             if k[0] == '_': continue
+            positions = set()
             for p in needles:
                 mr = _fuzzymatch(v, p)
                 if mr.score > 0:
-                    match[k] = mr
-                    formatted_hay[k] = _format_match(v, mr.positions)
+                    match.setdefault(k, []).append(mr)
+                    positions |= set(mr.positions)
+            formatted_hay[k] = _format_match(v, positions)
 
         if match:
             # square to prefer larger scores in a single haystack
-            score = int(sum(mr.score**2 for mr in match.values()))
+            score = int(sum([mr.score**2 for mrs in match.values() for mr in mrs]))
             matches.append(CombinedMatch(score=score, formatted=formatted_hay, match=h))
 
     return sorted(matches, key=lambda m: -m.score)


aborruso:
Thank you very much

saulpw (approved):

aborruso:
Hi @midichef what should I do if I want to use it?

Do I must install the develop version running `pip3 install git+https://github.com/saulpw/visidata.git@develop`?

Thak you to you, @saulpw and @anjakefala

midichef:
I pushed a new commit that didn't change any code, only the description, to insert a `-` into `[cmdpalette]`.

midichef:
@aborruso yes, once the PR is accepted into the `develop` branch, you can use that command, or the shorter form `pip3 install git+https://github.com/saulpw/visidata.git`. But the current `develop` has a separate bug that slows down the loading of large files, which is not present in vd v3.1.1. That bug will be fixed by #2636, so I suggest you wait till that is in `develop` too.
> Do I must install the develop version running `pip3 install git+https://github.com/saulpw/visidata.git@develop`?

midichef:
@aborruso
Until both those fixes are up, I merged both those fixes into [a branch on my fork of visidata](https://github.com/midichef/visidata/tree/tmp.fuzzymatch_space_and_max_rows). You can install it with: `pip3 install git+https://github.com/midichef/visidata.git@tmp.fuzzymatch_space_and_max_rows`
Feel free to try it if you'd like. I'll keep it available until both the fixes have made it into `develop`, then I'll take it down.

aborruso:
Thank you @midichef a stupid question: is it better first to uninstall the version I have and then install the new?

midichef:
I uninstall first. But I don't actually know if that is better.

aborruso:
@midichef wow, it works.

I had noticed this for a very long time. I had never written it down, because it seemed strange to me that no one had ever noticed it, and then I thought there must be some reason why it must be malfunctioning :)

Thank you again

image

anjakefala:
Thanks for your patience @aborruso! I was away for the holidays. =) It is now merged into develop.

You can do `pip3 install git+https://github.com:saulpw/visidata.git`

aborruso:
Thank you very much @anjakefala

PR #2651, opened 2024-12-24
stacktrace: fix missing caller frames and remove carets from vd_cli() traces

midichef:
https://github.com/saulpw/visidata/commit/4df730808958677562070bc593323e5069aba62b fixes a bug in #2586. The caller of `stacktrace()` was being removed in too many cases, like in `getCell()`. Now the only caller of stacktrace that is removed is `exceptionCaught()`.

Test case: `vd` `=` `1+[]` `z^E` and note how there is no line that ends with `in getCell` until after applying https://github.com/saulpw/visidata/commit/4df730808958677562070bc593323e5069aba62b.

https://github.com/saulpw/visidata/commit/aa4285bfa1db110fdf88073d83dba9a4fda3af4a cleans up the carets in tracebacks from `vd_cli` (seen in Python 3.11 and up). Like in the traceback when an HTTP connection closes:
```
  File "/usr/lib/python3.12/http/client.py", line 331, in begin
    version, status, reason = self._read_status()
                              ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/http/client.py", line 300, in _read_status
    raise RemoteDisconnected("Remote end closed connection without"
http.client.RemoteDisconnected: Remote end closed connection without response
```
You can test this with `netcat`: `nc -l -p 2000 -w 1& vd http://localhost:2000/`

I tested https://github.com/saulpw/visidata/commit/aa4285bfa1db110fdf88073d83dba9a4fda3af4a in the most recent Python major versions from 3.8-3.12.

Note that the new exception formatting will not happen for events like `KeyboardInterrupt`. So the carets will be seen if you do `vd /dev/zero` then hit `^C`.

### DIFF
--- a/visidata/errors.py
+++ b/visidata/errors.py
@@ -1,4 +1,5 @@
 import traceback
+import sys
 import re
 
 from visidata import vd, VisiData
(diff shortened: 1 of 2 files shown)


saulpw (approved):
Man, why is it so complicated!  We just need a decent stacktrace.  (Not blaming you, I'm mad at Python on this one.)

anjakefala:
Thank you always including test cases for your PRs!

PR #2656, opened 2024-12-28
[features-] sysedit: modify cell only if editor changes value

midichef:
A followup to #2615. Cells are unnecessarily modified upon `^O` then exiting the editor with no changes. That marks the sheet as modified.

With this PR, vd leaves the cell unchanged if the editor did not change the display value.


saulpw (approved):
Thanks @midichef, this makes sense to me.

PR #2663, opened 2025-01-03
[input-] fix word locator for Ctrl+Right motion

midichef:
In phrases like "word&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;next", moving one word right went to the incorrect location, to 1 character after the end of the word, which is in the empty space if the word is followed by several spaces:
```
echo 'word     second---third   ' |vd -
```

This PR also renames `find_nonword()` to `find_word()`.


saulpw (approved):
I'm surprised I never noticed this before.  Nice catch and fix.

PR #2664, opened 2025-01-04
[disp] Rename disp_pixel_random to disp_graph_pixel_random

### LINKED ISSUE #2661: Rename options.disp_pixel_random to include graph
Rename `options.disp_pixel_random` to include the word `graph`. (Mentioned in #2657)

cool-RR:
Fixes #2661 

Do I need to add anything to make the old name still work?


cool-RR:
Bump.

saulpw:
Thanks for pulling this together, @cool-RR.  We do need to add a line to the end of deprecated.py:

```
vd.optalias('disp_pixel_random', 'disp_graph_pixel_random')  #2661
```

But if that's the only `disp_pixel_random` remaining in the codebase after this PR, then we're good.

cool-RR:
Done.

saulpw (approved):
Thanks for noticing and fixing an inconsistency, @cool-RR !

cool-RR:
How do I do 

```
options.disp_pixel_random = True
```

In my `.visidatarc` such that it works both before the change and after it, without showing any warning that the option doesn't exist?

cool-RR:
@saulpw Any idea about the question above?

saulpw:
I have some thoughts but nothing immediate, and my dayjob means that I don't really have time to respond during the week.  Please be patient.

cool-RR:
🥺

cool-RR:
I ended up going with this: 

```
try:
    options.disp_pixel_random
except ValueError:
    options.disp_graph_pixel_random = True
else:
    options.disp_pixel_random = True
```

I figured there would be a more elegant solution because I imagine this isn't the first time a name is changed, but I guess there isn't.

saulpw:
That's about as good as I was going to do.  The other possibility is to check the `visidata.__version__` string.  Glad you found something that worked for you!  (and you might be the only person using that feature anyway :)

Issue #2416, opened 2024-05-30
[multiline] wordwrap for non-single-width characters (CJK)


duofinal:
**Small description**
  Only english text works correctly when using multiline command (wrap text)

**Expected result**
  Wrap text should work for all languages

**Actual result with screenshot**
- [sample.csv](https://github.com/saulpw/visidata/files/15496265/sample.csv)
![image](https://github.com/saulpw/visidata/assets/1624712/9147269e-05ea-4f50-a668-bb7b7c6b0eb5)

**Steps to reproduce**
  open a file -> execute multiline command (v or View -> toggle display -> multiline)

**Configuration**

- Does this issue reproduce without any plugins or configuration (using the `-N` CLI flag)?
  Yes
- Does this issue reproduce with either the [latest release](https://www.visidata.org/releases/), or with the [develop branch](https://www.visidata.org/install/#update-visidata-from-an-existing-installation)?
Latest release v3.0.2 

**Additional context**

- What platform and version are you using (Linux, MacOS, Windows)?
 Ubuntu 22.04
- Which version of Python?
 3.11.5
- Which terminal are you using (for display and input issues)?
  iTerm2 with macOS and use the ubuntu remotly

saulpw:
Can you provide sample data like that used in the screenshot?

duofinal:
yes i've updated issue

midichef:
I looked into this a bit, for Python 3.10.12. The trouble seems to be the use of `textwrap.TextWrapper._wrap_chunks()`. The wrapping calculations there use `len()` for strings, where visidata would use their display width `dispwidth(s)`. As a quick check, I seem to get proper wrapping when I use a modified version of `_wrap_chunks()` for `sheets.py`.

```
def dispwidth_wrap_chunks(self, chunks):
    len = dispwidth
    # the rest of the function is unmodified from the original _wrap_chunks()
    ...
textwrap.TextWrapper._wrap_chunks = dispwidth_wrap_chunks
```
So one solution would be to subclass `textwrap.TextWrapper` to override `_wrap_chunks()` with a version that uses `dispwidth()` instead of `len()`. Then replace `textwrap.wrap()` with `wtw = WideTextWrapper(); wtw.wrap()`. Overriding like that is [a solution that textwrap puts forward](https://github.com/python/cpython/blob/a8f1152b70d707340b394689cd09aa0831da3601/Lib/textwrap.py#L22-L23):
>  If you want to completely replace the main wrapping algorithm,
[...]

PR #2666, opened 2025-01-06
[misc-] handle chars with screen width != 1, in unimportant cases

midichef:
472953e8a6dc54518cdef14527b79a67731eaf46 - fixes behavior in several files. The effects are minor.
And I added this comment:
`# textwrap.wrap does not handle variable-width characters  #2416`
because it seems I need the reminder, as I just looked into `textwrap`'s handling of such characters, eventually discovering [I looked into it several months ago](https://github.com/saulpw/visidata/issues/2416#issuecomment-2144064190) but forgot.

5ae669b6b06cc5d07ba5933994d239bbbf05b797 - should not change behavior of current visidata in any visible way. But in the future, if menus use variable-width characters, it would matter.
cb2bad37e84cd2c683bc9021db54e2377a445be6 - `man/parse_options.py`, it should not change current behavior.
1c707a1ba47c85e2d59aefbb9637a0255c793670 - `ddwplay.py` for `Animation.load_from()`. I did not test this change, as I do not know how to test `Animation`.
066db54241b401f8c93908ad35cba9764375bd7d - These changes are not strictly necessary, as the `prompt` strings are standard ASCII text. The fix is only for completeness, useful only if the code is copy/pasted or altered.

With this PR, [my audit of `len()` vs `dispwidth()`](https://github.com/saulpw/visidata/pull/1958#issuecomment-1920773423) is finished.

### DIFF
--- a/visidata/aggregators.py
+++ b/visidata/aggregators.py
@@ -6,7 +6,7 @@ import statistics
 from copy import copy
 
 from visidata import Progress, Sheet, Column, ColumnsSheet, VisiData
-from visidata import vd, anytype, vlen, asyncthread, wrapply, AttrDict, date, INPROGRESS
+from visidata import vd, anytype, vlen, asyncthread, wrapply, AttrDict, date, INPROGRESS, dispwidth
 
 vd.help_aggregators = '''# Choose Aggregators
 Start typing an aggregator name or description.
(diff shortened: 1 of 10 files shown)


visidata/statusbar.py
```diff
@@ -32,7 +32,7 @@
 BaseSheet.init('longname', lambda: '')
 
 def fitWithin(s, n=10):
-    if len(s) > n:
+    if dispwidth(s) > n:
         return s[:n//2-1] + '…' + s[-n//2+1:]
```
saulpw: This would need to use the same kind of `partition()` function that was mentioned in #2662.

midichef: I've removed the change to `fitWithin()` and kept the others. I'll make the `fitWithin` change in a different PR.

saulpw (approved):

anjakefala:
I tested the animation! Everything looked fine.

PR #2667, opened 2025-01-07
[cliptext-] clipstr() to width 1: truncate chars having width > 1

midichef:
Let's say we have a sheet with full-width characters, like あ,
normal
`clipstr()` currently causes a problem: if we change a column to width 1, each cell's first character is drawn with its full width, causing columns to be misaligned.
width1-current

This PR changes `clipstr()` to trim the fullwidth character to `''`, so that the column will stay aligned:
width1-new

Note that this affects the behavior of the headers of hidden columns. They currently show the first character even when it's full-width:
hidden-current
but after this PR they will become blank:
hidden-new

I'm fine with blank headers for hidden columns. If we want to solve that, it can be fixed in `drawColHeader()`:
https://github.com/saulpw/visidata/blob/6b0a78a7b62fcc75fde37377143a0866aa114e7a/visidata/sheets.py#L741

An alternate design for this PR would have `clipstr()` replace the single chars with `trunch`:
```
        if w_s > 1:
            #add these 2 lines to replace the char with trunch:
            if trunch:
                return trunch, dispwidth(trunch)
```
But then we'd want to disable the use of the truncator-symbol  in certain places, like the top of hidden columns. To do that is easy enough, just call the `clipstr()` or `clipdraw()` with `truncator=''`. But locating all those places would be more effort. So I'm submitting this PR, which is a less invasive change.

### DIFF
--- a/visidata/cliptext.py
+++ b/visidata/cliptext.py
@@ -148,7 +148,7 @@ def _clipstr(s, dispw, trunch='', oddspacech='', combch='', modch=''):
     Note: width may differ from len(s) if chars are 'fullwidth'.
     If *dispw* is None, no clipping occurs.
     If *trunch* has a width greater than *dispw*, the empty string
-    will be used as a truncator instead, and a warning will be shown.'''
+    will be used as a truncator instead.'''
     if not s or (dispw is not None and dispw < 1): #iterator s would be truthy
         return '', 0
 
@@ -165,7 +165,6 @@ def _clipstr(s, dispw, trunch='', oddspacech='', combch='', modch=''):
         #if the next character will not fit
         if dispw and w+chlen > dispw:
             if trunchlen > dispw:
-                vd.warning(f'truncator {repr(trunch)} is too big to fit in dispw of {dispw}')
                 return _clipstr(s, dispw, trunch='', oddspacech=oddspacech, combch=combch, modch=modch)
             # if the trunch by itself can fit
             if trunchlen and dispw >= trunchlen:
--- a/visidata/cliptext.py
+++ b/visidata/cliptext.py
@@ -142,13 +142,15 @@ def iterchars(x):
 
 @functools.lru_cache(maxsize=100000)
 def _clipstr(s, dispw, trunch='', oddspacech='', combch='', modch=''):
-    '''Return clipped string and width in terminal display characters.
-    Note: width may differ from len(s) if East Asian chars are 'fullwidth'.'''
-    if not s:
+    ''' *s* is a string or an iterator that contains characters.
+    *dispw* is the integer screen width that the clipped string will fit inside, or None.
+    Return clipped string and width in terminal display characters.
+    Note: width may differ from len(s) if chars are 'fullwidth'.
+    If *dispw* is None, no clipping occurs.
+    If *trunch* has a width greater than *dispw*, the empty string
+    will be used as a truncator instead, and a warning will be shown.'''
+    if not s or (dispw is not None and dispw < 1): #iterator s would be truthy
         return '', 0
-
-    if dispw == 1:
-        return s[0], 1
 
     w = 0
     ret = ''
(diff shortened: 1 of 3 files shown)


visidata/cliptext.py
```diff
@@ -148,7 +148,16 @@ def _clipstr(s, dispw, trunch='', oddspacech='', combch='', modch=''):
         return '', 0
 
     if dispw == 1:
-        return s[0], 1
+        w_s = dispwidth(s)
+        if w_s > 1:
+            ret = ''
+            for c in s:
```
saulpw: Okay.  Do we need a for-loop here?  Aren't there just two cases?

```
        if dispwidth(s[0]) == 1:
            return s[0], 1
        else:
            return '', 0
```

midichef: I don't know if there are more cases, as I'm not sure which characters get a width of 0 from `wcwidth()`. I put the for-loop there to handle a situation where a string starts with multiple characters with `wcwidth`/`dispwidth` of `0`, followed by a final character that has a `dispwidth` of `1`.

An example of a character that might get a `wcwidth()` of `0` is [u"\u0ccd"    # Joiner, Category 'Lo', East Asian Width property 'N' -- KANNADA SIGN VIRAMA](https://github.com/jquast/wcwidth/blob/57cfbda2a87d9d6c5a2ba7b5018f3ce2c04b6383/tests/test_core.py#L375), taken from a test in [jquast/wcwidth](https://github.com/jquast/wcwidth). Visidata's `wcwidth()` returns `1` for this character. But `wcwidth()` from `jquast/wcwidth` returns `0`. Measuring a longer similar string with jquast's `wcswidth()` function:  `wcswidth("\u0CCD\u0CCDj)"` gives a width of `1`, though Visidata's `dispwidth` gives a width of `3`.

So even though I don't know offhand of a case where Visidata's `wcwidth()` gives 0 for multiple characters in a row, I figured that it could happen.

saulpw: Hm, I guess VisiData's `wcwidth` is wrong then.  We should be able to come up with some basic strings that use a zero-width joiner like this one and watch them clobber some aspects of the VisiData interface.  We may want/have to vendorize `jquast/wcwidth` and extend it for our purposes.

midichef: Okay, I opened an issue for the `wcwidth` implementation. Meanwhile this PR stands.

midichef:
I changed `_clipstr()` to use the alternate design I discussed above:  to have clipstr() replace the single chars with `trunch`. It gives `clipstr` a default truncator behavior that will prove useful in my proposed fix for #2662, where I'll introduce two functions similar to `clipstr`.

The commit I have just force-pushed handles more corner cases, and also adds tests.

midichef:
Two tests are failing, with something broken in cells holding lists. I'll look into it soon.

midichef:
Ah the problem was `dispw < 1`. Sometimes `dispw` is `None`, and the comparison would raise an exception. I've fixed that and documented that it can occur. And added tests for the `dispw=None` case.

I also added https://github.com/saulpw/visidata/pull/2667/commits/b99f8c5dffc6d7209b8d5ed7c11a3d0adfef0f2a which bypasses the `clipstr()` call for `dispw=None` in `formatValue()`.

I'd appreciate a close look at these changes. Since I didn't know about `dispw=None`, I don't know if I've forgotten any other corner cases.

visidata/cliptext.py
```diff
@@ -160,12 +162,17 @@ def _clipstr(s, dispw, trunch='', oddspacech='', combch='', modch=''):
             newc = c
             chlen = dispwidth(c)
 
+        #if the next character will not fit
         if dispw and w+chlen > dispw:
-            if trunchlen and dispw > trunchlen:
-                lastchlen = _dispch(ret[-1])[1]
-                if w+trunchlen > dispw:
+            if trunchlen > dispw:
+                vd.warning(f'truncator {repr(trunch)} is too big to fit in dispw of {dispw}')
```
saulpw: This is a very hot/inner routine, so we shouldn't show a warning here, nor even a debug.  I removed it.

visidata/tests/test_cliptext.py
```diff
@@ -19,12 +20,87 @@ def test_dispwidth(self, s, dispw):
         (' jsonl', 5, ' jso…', 5),
         ('abcdで', 6, 'abcdで', 6),
         ('abcdで', 5, 'abcd…', 5),
+        ('a', 1, 'a', 1),
```
saulpw, line 23: Thanks for adding these tests!

saulpw (approved):
Thanks for seeing this through!

PR #2668, opened 2025-01-07
[form-] fix underlining of substrings by FormCanvas

midichef:
`FormCanvas` would underline an incorrect number of characters, unless `index` of the underlined string happened to be 1. The underlining feature fixed here is not actually unused anywhere. So I created a form that used underlining to test the fix.


saulpw (approved):
I can never remember whether slices are `[start:length]` or `[start:end]`.  Thanks for fixing.

Issue #2573, opened 2024-10-17
Can't click or switch to errors_recent tabs


daviewales:
**Small description**

If I press `Ctrl+E` to view an error, a tab appears in the status bar called `errors_recent`. However, if I switch to a different tab, I can't click on the `errors_recent` tab to return to it. I also can't switch to it with `Alt+NUMBER`.
I have to use the Sheets sheet to interact with it.

**Data to reproduce**

Open Visidata in a way which creates a Python error. (Put `1/0` in your `.visidatarc`)

**Steps to reproduce**

Press `Ctrl+E` in Visidata after a Python error is reported.

**Expected result**

Expect to be able to interact with `errors_recent` tabs the same as other tabs.

**Configuration**

Reproduces with current release `saul.pw/VisiData v3.1.1`

**Additional context**

- What platform and version are you using (Linux, MacOS, Windows)? Linux (WSL)

- Which version of Python? 3.10.12

- Which terminal are you using (for display and input issues)? Windows Terminal (WSL)

midichef:
I can reproduce this behavior on Ubuntu. I also see the problem with `StatusSheet`, not just `ErrorSheet`. And more generally, any sheet that satisfies `not sheet.precious`:
https://github.com/saulpw/visidata/blob/3e10140963a62a5f78fc6b60ebd2ac9264ef2772/visidata/sheets.py#L1104
Because `ErrorSheet` has `self.precious = False`, it doesn't get added to `vd.allSheets` during `vd.push()`. A sheet needs its position in `allSheets` to calculate its `shortcut`. The `shortcut` is needed for a mouse-click to switch tabs, by running the command `jump-sheet-{shortcut}`.  For these sheets, that turns into `jump-sheet-`, which fails.

I'm not sure what the right fix is.

saulpw:
What if we didn't show non-precious sheets on the sheets tab?  The whole point is that they're disposable.

daviewales:
Would that mean they also don't get a named tab? I think that's less confusing. If they don't appear as a tab, then I won't be confused when I can't click on them. Another interesting thing I noticed was that if I keep pressing `Ctrl+E`, duplicate error sheets are created, rather than just jumping me back to the original one.

saulpw:
[...]

PR #2671, opened 2025-01-08
[errors-] create Error(s)Sheet only if it is not already open

midichef:
Fixes an inconvenience noted by @daviewales as part of #2573:
>  if I keep pressing Ctrl+E, duplicate error sheets are created, rather than just jumping me back to the original one.

### DIFF
--- a/visidata/textsheet.py
+++ b/visidata/textsheet.py
@@ -83,6 +83,9 @@ This is the traceback for an error.
                 vd.launchEditor(match.group(1), f'+{match.group(2)}')
                 return
 
+    def reload(self):
+        src = self.source or (vd.lastErrors[-1] if vd.lastErrors else [])
+        self.rows = list(enumerate(src))
 
 class ErrorCellSheet(ErrorSheet):
     columns = [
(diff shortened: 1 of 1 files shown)


visidata/textsheet.py
```diff
@@ -116,10 +116,33 @@ def recentErrorsSheet(self):
...
+            return
+    vd.push(vd.recentErrorsSheet)
+
+@VisiData.api
+def errors_all(vd):
+    if not vd.lastErrors:
+        vd.status("no errors")
+        return
+    for vs in vd.sheets:
+        if vs.name == 'errors_all' and isinstance(vs, ErrorsSheet) and vs.source == vd.lastErrors:
```
saulpw: If we make the two errorSheet properties return a fixed instance, then this behavior will happen automagically if it's already pushed.

We can do this by having each `reload()` use its specific error instead of passing in a `source` attribute, and then marking both as `@lazy_property`. I don't think these Error Sheets are being used in other contexts.

midichef: Oh, that would be a cleaner change. It works for `errors-all`, but how can it work for the `recentErrorsSheet` in `error-recent`? If I do:
```
class ErrorSheet(TextSheet):
    ...
    def reload(self):
        self.rows = list(self.readlines(vd.lastErrors[-1] if vd.lastErrors else []))
@VisiData.lazy_property
def recentErrorsSheet(self):
    return ErrorSheet("errors_recent")
```
the `ErrorSheet` will not know to update/reload itself when new errors are appended to `vd.lastErrors`.

midichef: Oh right, I should call `reload` manually.

midichef:
Should I use an explicit `source=None` parameter when creating the sheets, in `ErrorsSheet("errors_all")` and `ErrorSheet("errors_recent")`? They're getting it initialized like that from `BaseSheet` already.

saulpw:
> Should I use an explicit source=None parameter when creating the sheets, in ErrorsSheet("errors_all") and ErrorSheet("errors_recent")? They're getting it initialized like that from BaseSheet already.

No, let them use the default.

visidata/textsheet.py
```diff
@@ -83,6 +83,9 @@ def sysopen_error(self, col, row):
                 vd.launchEditor(match.group(1), f'+{match.group(2)}')
                 return
 
+    def reload(self):
+        if vd.lastErrors:
+            self.rows = list(self.readlines(vd.lastErrors[-1]))
```
saulpw: Do we want `readlines` (i.e. wordwrap) for errors?

midichef: Oops, no, we don't want `readlines`. That would make the stack trace harder to understand. And totally break the regexp search used by `sysopen-error`. So I've taken `readlines` out.

saulpw (approved):

anjakefala:
I found a bug with this fix!

To reproduce, put `1/0` in your `~/.visidatarc`.
Press `^E`, and it will correctly open the ZeroDivisionError.

Then type a column without numerical values with `#`. Use `z^E` on one of the cells, and it will incorrectly load the ZeroDivisionError as the cell error.

midichef:
Oops, nice catch. I fixed it by letting `self.source` override the default sources of `vd.lastErrors[-1]` (in `ErrorSheet`) and `vd.lastErrors` (in `ErrorsSheet`). But note [my style question here](https://github.com/saulpw/visidata/commit/5a02cdd630ea1bb157c11d96e3f15d0d5631e005#r151492029).

midichef:
Based on feedback by @daviewales I replaced the [short-circuiting `and`](https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not) with the more common `if`/`else`. This PR is complete now.

anjakefala (approved):
Yup, everything works!

PR #2688, opened 2025-01-22
[sort-] allow z[ and z] to reverse col sort dir or ignore col

midichef:
This PR is a modified version of a feature I prototyped in https://github.com/saulpw/visidata/discussions/1877.

It lets `z[` and `z]` modify a column for sorting, while keeping the existing sort criteria if possible. This is very convenient when exploring data.

There are three cases for these commands:
1. add column to sort criteria (this is the only use for `z[` and `z]` in current visidata)
This PR adds new behavior only for the other two cases, where the column is already in the sort criteria:
2. toggle off the column's sorting
When the column is used for sorting, and the command sort direction matches the existing direction. Like `z[` for a column that is already in ascending order due to a previous `z[`. The column will be removed from the sort criteria. Any existing sort criteria on columns that are higher in priority, is kept. Lower priority columns are dropped from the sort criteria.
3. reverse the column's sort direction
When the column is used for sorting, and the new direction is the opposite of the existing direction. Like `z[` for a column that is in descending order from a previous `z]`. The column will reverse its sort direction. Existing sort criteria are handled the same as in case 2.

(In current visidata, for cases 2 and 3, the column is added to the sort criteria again, uselessly, which is a bug that throws off the sort arrow indicators for future columns used for sorting.)


midichef:
However, in the original discussion of this feature, [@saulpw noted a complication in replay](https://github.com/saulpw/visidata/discussions/1877#discussioncomment-5760360) when commands depend on state in the sheet:

> However, it's tricky to change behavior based on existing column state, as this can make cmdlog replay a lot less predictable. We already have this with e.g. key-col and it's caused some problems. Maybe the ultimate answer is to have key-col put a more precise command like key-col-on or key-col-off (please help come up with better longnames if we do this) on the cmdlog. These sort commands could then do the same.

I implemented this suggestion for `key-col` in https://github.com/saulpw/visidata/pull/2622. That PR converts `!` into the state-independent `key-col-on` or `key-col-off` for the command log. But that method won't work for `z[`, because one `z[` command can modify several columns at once. In that situation, `z[` would need to generate multiple `sort-asc` and `sort-asc-add`/`sort-desc-add` commands. Multiple commands from a single command would cause inconsistent state in `undo`, because `undo` assumes that commands are undone one at a time:
https://github.com/saulpw/visidata/blob/8f8aaecffb045ba31393eaac1d2c76b1206d8b7e/visidata/undo.py#L52

So how should we handle state-dependent commands here? What specific problems were caused for cmdlog replay? Are these kinds of problems still present? I know some replay behavior and timing has been changed since v2.11.

I find this feature convenient enough that I'd rather have it, even if it does complicate cmdlog replay.

saulpw:
We could turn the `sort-` commands into `sort-order` and have the input be a list of column names, each prefixed by `<` or `>`. The sort-order command would use `re` to split the list, find the columns, and call orderBy.  Would you want to implement this on this PR, or should we merge it as it already is?

midichef:
> each prefixed by `<` or `>`. The sort-order command would use `re` to split the list

I think `<` and `>` are too common to rely on. I often create expression columns with names like `value < 5`, which would break replay.

Could we separate the column names by adding [some type of Unit Separator](https://stackoverflow.com/a/18782271)? Like the printable one in Unicode:  `␟`.  So we would split at `␟<` and `␟>`?

midichef:
Can we just use the column indices instead of the column names? Names are tricky to parse, and can be ambiguous when two columns have the same name. Using indices, the input strings in the cmdlog would look like:  `"input": "<1>2"`. Is there a drawback to using indices?

saulpw:
> Can we just use the column indices instead of the column names?

I agree with all your points, but this would make the cmdlog even more brittle; for instance if a .jsonl file happens to have the fields in a different order.  I would feel more comfortable with using column indices (perhaps more generally instead of just the sorting commands) if we converted batches of `slide-` and `hide-col` commands (maybe others) to a more general "set column order" command.

midichef:
>this would make the cmdlog even more brittle; for instance if a .jsonl file happens to have the fields in a different order

Ah, right. Columns are likely to be presented in different order, and it's important that visidata handles that robustly. I see now. I'll go with column names then. A more general "set column order" command is more than I can tackle at the moment.

midichef:
Alright, I've submitted a draft that saves the full  sort `ordering` to the cmdlog `input` parameter, but only for the `sort-*-add` and `sort-*-change` commands, not for `sort-asc` and `sort-desc`.

I didn't know what special string to use to separate the columns in the cmdlog, so the strings contain `\u0000` as a placeholder, not meant to be permanent. `␟` would show up as `\u241f`, which is not great either.

midichef:
>  `␟` would show up as `\u241f`, which is not great either.

It would show up as `\u241f` in the .vdj file, which is not great, but as `␟` when viewing the .vdj file with visidata, which is nice visually and does a good job of explaining its purpose in the string.

visidata/sort.py
```diff
@@ -34,7 +61,46 @@ def __eq__(self, other):
...
+    return ordering
+
+def edit_ordering(ordering, col, reverse):
+    '''Return a modified ordering based on editing a single column *col*:   add it, remove it, or flip its direction.
+    Columns after *col* in the ordering (with lower sort priority) are also removed from the ordering.
+    *ordering* is a list of tuples:  (Column, boolean), where the boolean defines the sort direction.
+    '''
+    new_ordering = []
+    # handle changes to status of columns that are already in the ordering:  add/remove/flip
+    changed = False
```
saulpw, line 87: If we never want to preserve lower-priority sort columns (even if the direction remains the same), then could we simplify this to:
```
for c, old_reverse in ordering:
    if c is col:
        break
    else:
        new_ordering.append((c, old_reverse))
new_ordering.append((col, reverse))
```

visidata/sort.py
```diff
@@ -84,10 +150,12 @@ def _sortkey(r):
 Sheet.addCommand('g]', 'sort-keys-desc', 'orderBy(None, *keyCols, reverse=True)', 'sort descending by all key columns; replace any existing sort criteria')
 
 # add to existing sort criteria
-Sheet.addCommand('z[', 'sort-asc-add', 'orderBy(cursorCol)', 'sort ascending by current column; add to existing sort criteria')
-Sheet.addCommand('z]', 'sort-desc-add', 'orderBy(cursorCol, reverse=True)', 'sort descending by current column; add to existing sort criteria')
-Sheet.addCommand('gz[', 'sort-keys-asc-add', 'orderBy(*keyCols)', 'sort ascending by all key columns; add to existing sort criteria')
-Sheet.addCommand('gz]', 'sort-keys-desc-add', 'orderBy(*keyCols, reverse=True)', 'sort descending by all key columns; add to existing sort criteria')
+Sheet.addCommand('', 'sort-asc-add', 'orderBy(cursorCol, save_cmd_input=True)', 'sort ascending by current column; add to existing sort criteria')
```
saulpw: Let's make the default for `save_cmd_input` to be True, so we don't have to pass it in these execstrs.  There are only 3 other places that call `orderBy` and they are either unused or very infrequently used; we can pass save_cmd_input=False there instead.

saulpw:
> > `␟` would show up as `\u241f`, which is not great either.
> 
> It would show up as `\u241f` in the .vdj file, which is not great, but as `␟` when viewing the .vdj file with visidata, which is nice visually and does a good job of explaining its purpose in the string.

I'm fine with this separator.

visidata/sort.py
```diff
@@ -1,28 +1,55 @@
 from copy import copy
-from visidata import vd, asyncthread, Progress, Sheet, options, UNLOADED
+from visidata import vd, asyncthread, Progress, Sheet, Column, options, UNLOADED
+import re
+
+cmdlog_col_prefix='\x00'  #string to mark the start of column info in an ordering string
```
saulpw: Please change to `\u241f` as you mentioned.

visidata/sort.py
```diff
@@ -34,7 +61,46 @@ def __eq__(self, other):
     def __lt__(self, other):
         return other.obj < self.obj
 
+def order_string(sheet):
+    # replace ambiguous colname strings with unambiguous Column objects
+    sheet._ordering = sheet.ordering
+    ret = ''.join([cmdlog_col_prefix+('>' if reverse else '<') + str(sheet.columns.index(col)) for col, reverse in sheet._ordering])
```
saulpw: I think it's better to use the column name than the column index.  Use `sheet.column(colname)` below to get the column (it should search in availCols order, which should virtually always be what's desired).

saulpw (changes requested):
A few changes but I think we're on track to get this for v3.2.

visidata/sort.py
midichef: That does not implement the "remove" case in "add/remove/flip": when `z[` is applied to a column that is already ascending, the column should get taken out of the sort ordering. It does seem like this loop is unnecessarily complex but I don't quite see a way to simplify it.

I will change the comparison to `c is col` instead of `c == col` though.

visidata/sort.py
midichef: I see 5 places:  `[`, `]`, `g[`, `g]` and the column sorting in `threads.py`. I've just pushed a fix to remove the one in `threads.py`.

For the remaining 4, how about we just make them use the new default of `save_cmd_input=True`? Then all calls to `orderBy` would be the same.

Advantages:
The default would be backwards compatible with old .vdj files. That is, replay would still work for those 4 commands, because they had `input=''`.
It would make it easier to speed up replayed sort commands, by hand-editing a .vdj file. Right now, replaying multiple `sort-*-add` commands in a row is slow, because the sheet is sorted many times needlessly. With the new default, a user who needs speed could edit the .vdj file to consolidate multiple `sort-*-add` commands with `col='col1'`, `col='col2'` etc, into a single `sort-asc` with an `input` field with all columns specified at the same time:  `input='␟<col1␟<col2␟>col3␟>col4'`. (If `save_cmd_input` defaults to `False` for these 4 commands, then those hand-edited input fields would be lost if the replayed cmdlog is ever saved a second time.)
Disadvantage:
Anyone who used to edit `sort-*` commands using the `col` field in a .vdj file will have to learn to edit the `input` field, because `input` overrides `col`, and `input` would always be present in new cmdlogs.

I've submitted 14d5039 to demonstrate this proposed default.

visidata/sort.py
midichef: Right, yes, I thought I'd implemented that based on earlier feedback. I suspect I lost the commit making that change.

midichef:
@anjakefala 
Can you please merge #2746 (or `develop`) into this branch? I just said I'd rebase this set of commits myself, to incorporate that fix in #2746, but I worry that may obscure [the existing unresolved conversation about what the default value of `save_cmd_input` should be](https://github.com/saulpw/visidata/pull/2688#discussion_r2103823683).

saulpw (approved):
Okay, this looks good to me.  Thanks @midichef !

Issue #2689, opened 2025-01-23
display issue for double quotes


need47:
I have a CSV file generated by using the Python csv module, in which some cells contain double quotes, e.g., `"""Velders"""`. The display `"Velders"""` in visidata (v3.1.1) seems unexpected.

![Image](https://github.com/user-attachments/assets/39df902b-131f-4e89-bafa-b85d2e8a430f)

midichef:
Can you attach the CSV file, and the exact command line you're running `vd` with?

I can reproduce this behavior under visidata v3.1.1 (and the latest `develop`), but only when visidata has to guess the filetype:
```
echo 'column,value\nvalueShape,""\nvalueExamples,"""Velder"""\nnote,' > data.csv
cp data.csv data
# shows "Velders""" when filetype is guessed due to lack of extension
vd data
# shows "Velders" when filetype is present in extension or specified with -f
vd data.csv
vd -f csv data
```

When visidata has to guess the filetype, it has to guess [what csv parsing options to use](https://docs.python.org/3/library/csv.html#dialects-and-formatting-parameters), using Python's `csv.Sniffer`. The behavior you saw happens when `csv.Sniffer` guesses `doublequote=False`. You can also override the guessed option:
`vd --csv_doublequote=True data`

If this doesn't address your problem, then giving more information would be helpful:  the data file, full vd command as run in the shell, as well as OS version, and Python version.

need47:
You are absolutely right. My file has no .csv extension. visidata has to guess in this case. Didn't even notice it.

visidata is amazing. I use it almost on a daily basis.

midichef:
I've submitted a PR to show a warning on the status bar when the CSV parser is guessing settings that aren't the default. Thanks for pointing out this surprising behavior.

PR #2690, opened 2025-01-23
[csv-] warn when guessed option differs from default

midichef:
When csv is guessed without a filetype, visidata uses Python's `csv` library to guess [the parsing options](https://docs.python.org/3/library/csv.html#dialects-and-formatting-parameters) using [csv.Sniffer](https://docs.python.org/3/library/csv.html#csv.Sniffer). For the simplest csv files, it seems that `doublequote` is `False`, whereas the default in visidata is `True`. That can surprise users, as it did in #2689.

This PR makes the guessing process more obvious. It outputs a warning for all guessed CSV options that differ from visidata's defaults:

```
echo a,b |vd
...
guessed option csv_doublequote=False
guessed option csv_skipinitialspace=False
```

(versions that do the silent guessing are 3.0 and up)

Or should we override the `Sniffer` guesses to use visidata's defaults?

### DIFF
--- a/visidata/loaders/csv.py
+++ b/visidata/loaders/csv.py
@@ -26,7 +26,11 @@ def guess_csv(vd, p):
 
         for csvopt in dir(dialect):
             if not csvopt.startswith('_'):
-                r['csv_'+csvopt] = getattr(dialect, csvopt)
+                v = getattr(dialect, csvopt)
+                optname = 'csv_'+csvopt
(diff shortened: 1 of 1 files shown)


saulpw (approved):

saulpw:
Thanks @midichef, this will be especially helpful for diagnosing when csv sniffing goes awry.

PR #2704, opened 2025-02-11
[vdsql-] ensure each thread in vdsql tracks its single connection

anjakefala:
Uses [threading.local](https://docs.python.org/3/library/threading.html#threading.local) to keep track of a thread's connection.

Otherwise, you occassionally get this error when opening an SQLite table with `vdsql`:

```
  File "/home/anja/anaconda3/envs/visidata/lib/python3.11/threading.py", line 995, in _bootstrap
    self._bootstrap_inner()
  File "/home/anja/anaconda3/envs/visidata/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "/home/anja/anaconda3/envs/visidata/lib/python3.11/threading.py", line 975, in run
    self._target(*self._args, **self._kwargs)
  File "/home/anja/git/visidata/visidata/threads.py", line 209, in _with_active_cmd
    _toplevelTryFunc(func, *args, **kwargs)
  File "/home/anja/git/visidata/visidata/threads.py", line 227, in _toplevelTryFunc
  File "/home/anja/git/visidata/visidata/sheets.py", line 281, in reload
    self.loader()
  File "/home/anja/git/visidata/visidata/sheets.py", line 307, in loader
    for i, r in enumerate(self.iterload()):
  File "/home/anja/git/visidata/visidata/apps/vdsql/_ibis.py", line 357, in iterload
    self.query = self.baseQuery(con)
  File "/home/anja/git/visidata/visidata/apps/vdsql/_ibis.py", line 345, in baseQuery
    tbl = con.table(self.table_name)
  File "/home/anja/anaconda3/envs/visidata/lib/python3.11/site-packages/ibis/backends/sql/__init__.py", line 100, in table
    table_schema = self.get_schema(name, catalog=catalog, database=database)
  File "/home/anja/anaconda3/envs/visidata/lib/python3.11/site-packages/ibis/backends/sqlite/__init__.py", line 298, in get_schema
    with self.begin() as cur:
  File "/home/anja/anaconda3/envs/visidata/lib/python3.11/contextlib.py", line 137, in __enter__
    return next(self.gen)
  File "/home/anja/anaconda3/envs/visidata/lib/python3.11/site-packages/ibis/backends/sqlite/__init__.py", line 147, in begin
    cur = self.con.cursor()
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 139669398107904 and this is thread id 139668276848384.
```

### DIFF
--- a/visidata/apps/vdsql/_ibis.py
+++ b/visidata/apps/vdsql/_ibis.py
@@ -1,4 +1,5 @@
 from copy import copy
+import threading
 import functools
 import operator
 import re
@@ -80,24 +81,18 @@ vd.openurl_sqlite = vd.open_vdsql
 class IbisConnectionPool:
     def __init__(self, source, pool=None, total=0):
         self.source = source
-        self.pool = pool if pool is not None else []
-        self.total = total
+        self._local = threading.local()
+        self._local.connection = None
 
     def __copy__(self):
-        return IbisConnectionPool(self.source, pool=self.pool, total=self.total)
+        return IbisConnectionPool(self.source)
 
     @contextmanager
     def get_conn(self):
-        if not self.pool:
-            import ibis
-            r = ibis.connect(str(self.source))
-        else:
-            r = self.pool.pop(0)
-
-        try:
-            yield r
-        finally:
-            self.pool.append(r)
+        import ibis
+        if not hasattr(self._local, 'connection') or not self._local.connection:
+            self._local.connection = ibis.connect(str(self.source))
+        yield self._local.connection
 
 
 class IbisTableIndexSheet(IndexSheet):


visidata/apps/vdsql/_ibis.py
```diff
@@ -80,24 +82,27 @@ def open_vdsql(vd, p, filetype=None):
...
     @contextmanager
     def get_conn(self):
-        if not self.pool:
+        if not hasattr(self._local, 'connection'):
             import ibis
-            r = ibis.connect(str(self.source))
-        else:
-            r = self.pool.pop(0)
-
+            with self._lock:
```
saulpw: Do we need this lock if we're just putting things on threading.local() anyway?  Do we have to even have a `_local` member or can we just use `threading.local()` each time?

visidata/apps/vdsql/_ibis.py
```diff
@@ -80,24 +82,27 @@ def open_vdsql(vd, p, filetype=None):
...
+                self._local.connection = ibis.connect(str(self.source))
         try:
-            yield r
+            yield self._local.connection
         finally:
-            self.pool.append(r)
+            pass
+
+    def close_all(self):
+        if hassattr(self._local, 'connection'):
```
saulpw: This must not have been tested because `hassattr` is mispelled.

visidata/apps/vdsql/_ibis.py
```diff
@@ -80,24 +82,27 @@ def open_vdsql(vd, p, filetype=None):
...
+        self._lock = Lock()
 
     def __copy__(self):
-        return IbisConnectionPool(self.source, pool=self.pool, total=self.total)
+        return IbisConnectionPool(self.source)
 
     @contextmanager
     def get_conn(self):
-        if not self.pool:
+        if not hasattr(self._local, 'connection'):
```
saulpw: Prefer `getattr` to `hasattr`.

visidata/apps/vdsql/_ibis.py
anjakefala: The examples I found involved the lock, but for our needs it seems to work without it!

visidata/apps/vdsql/_ibis.py
```diff
@@ -80,24 +81,19 @@ def open_vdsql(vd, p, filetype=None):
 class IbisConnectionPool:
     def __init__(self, source, pool=None, total=0):
         self.source = source
-        self.pool = pool if pool is not None else []
-        self.total = total
+        self._local = threading.local()
```
saulpw, line 84: There's no need to store this on a member variable, you can just use `threading.local()` where it's needed.

visidata/apps/vdsql/_ibis.py
```diff
@@ -80,24 +81,19 @@ def open_vdsql(vd, p, filetype=None):
...
-            r = self.pool.pop(0)
-
+        import ibis
+        self._local.connection = getattr(self._local, 'connection', ibis.connect(str(self.source)))
         try:
-            yield r
+            yield self._local.connection
         finally:
-            self.pool.append(r)
+            pass
```
saulpw: `try: X finally: pass` is a no-op, it should become just `X`.

visidata/apps/vdsql/_ibis.py
```diff
@@ -80,24 +81,19 @@ def open_vdsql(vd, p, filetype=None):
...
     @contextmanager
     def get_conn(self):
-        if not self.pool:
-            import ibis
-            r = ibis.connect(str(self.source))
-        else:
-            r = self.pool.pop(0)
-
+        import ibis
+        self._local.connection = getattr(self._local, 'connection', ibis.connect(str(self.source)))
```
saulpw: `ibis.connect` will be called even if the attribute exists.  So we need to do `if not threading.local().connection: threading.local().connection =  ibis.connect(...)` in order to not create extra connections only to discard them immediately.

visidata/apps/vdsql/_ibis.py
saulpw: But this would be a good place to initialize `local().connection = None`.

anjakefala: It seems `local()` creates a new variable each time? 

```
threading.local().connection = None
status(threading.local().connection)
```

gets `AttributeError: '_thread._local' object has no attribute 'connection'`

So we won't be able to take this approach. We'll need to set a member variable.

anjakefala: Hm, it doesn't. I'm not sure why `.connection` isn't sticky.:

```
<_thread._local object at 0x7f6c5852c090>
<_thread._local object at 0x7f6c5852c090>

```

anjakefala: Scratch that, I think I know what's going on.

anjakefala: Okay, so my guess was right. When we go into `getconn`, `threading.local()` will actually create a new thread-local object. So we'll need to assign it to a member variable.

anjakefala: Sorry, I'm still trying to wrap my head around how thread-local storage works. It seems when we initialise self._local.connection = None in the constructor, that only sets the connection attribute for the main thread that created the IbisConnectionPool instance.
When a different thread tries to access `self._local.connection`, it's looking at its own thread-specific storage, which doesn't have the connection attribute initialized yet, and we get an `AttributeError`.

My instinct is to use `hasattr` here.

anjakefala: `threading.local()` did create new instances. =( So we'll need to use member variables.

saulpw (approved):

PR #2707, opened 2025-02-20
Tune up codespell config and fix more typos

yarikoptic:
GitHub workflow was removed in 8accaba3df2b69abc84ca8e132eac64f9a7fa399 due to 
```
`F`requency
```
and such.  I have added ignores for those, and fixed other typos now detected (potentially simply due to newer codespell or may be introduced since then).

Let me know if you would like to bring CI back.

Cheers


saulpw (approved):
Thanks @yarikoptic.  We're not going to put this into CI but it's good to fix public-facing spelling errors.  I'm less keen on putting energy into fixing spelling errors in comments and internal docs like CHANGELOG but I've looked them over and they all seem reasonable.  Thanks for reviewing the ignore-words-list and adding real words to it.

PR #2718, opened 2025-03-10
Add OFX loader

daviewales:
Add OFX loader.
This is a common format for exporting transactions from banks.

Note: Work in progress, I haven't completed the checklist yet.

Note: For testing, we need some sample OFX files. ofxtools is licensed under GPLv3, and has [some test files](https://github.com/csingley/ofxtools/tree/master/tests/data). VisiData is also GPLv3, so the licenses are compatible.

Can we just copy the files, or do we need to attribute the original project? (**Edit**: Most of the files have errors, which I guess makes sense for testing a library... Maybe we can roll our own OFX test files.)


daviewales:
Sorry, I'm going to close this as I haven't had a chance to finish it, and I couldn't find an easy way to yield rows one at a time with the ofxtools library. As it doesn't meet the requirements for a core loader, I should probably just make it a plugin.

Issue #2719, opened 2025-03-10
Easy fix for missing `vd` on Windows


ptyork:
Revisiting a closed issue (#2262) since I ran into it and was annoyed enough to make my own quick fix. But it's a VERY, VERY simple fix that I think you can add in under a minute. You simply add a one-line file called `vd.cmd` to your /bin folder. Its contents are:
```
visidata %*
```
Then modify setup.py:
```
    scripts=["bin/vd", "bin/vd2to3.vdx", "bin/vd.cmd"],
```
Windows (both cmd.exe and powershell) searches the PATH for and executes *.exe, *.bat and *.cmd files if you enter them with or without the file extension. So users will be able to type `vd` and it will find and execute `vd.cmd`, which in turn does nothing but call `viewdata` (the .exe version) and pass along any additional arguments. Just an alias in batch file form.

For all other users, the file is simply ignored....nothing but a few wasted bytes.

anjakefala:
Thanks for the Windows insight! If that helps solve this issue for others, that'd be great.

@ptyork Could you open a PR, please?

ptyork:
Probably more work to merge a PR than to just add it yourself, but I'll do it. 😉 

Did anyone submit a PR for #2676? If not, I can try to incorporate that one, as well.

anjakefala:
Nope, no PR for that one yet!

ptyork:
Created PR #2721 for this.

Note that there is a bug on my system that prevents redirection to visidata. It's something in curses. Maybe just me. I don't know. But this `test.py` script:
``` py
import curses

print("BEFORE")
curses.initscr()
print("AFTER")
```
run using:
``` cmd
echo HELLO | python test.py
```
outputs:
```
BEFORE
Redirection is not supported.
```
Prints both BEFORE and AFTER without the redirection. Assuming it's an issue with windows-curses, I don't think there's much that can be done about it.

anjakefala:
Fixed by @ptyork!

Issue #2676, opened 2025-01-11
No mouse support on Windows


mplattner:
I run visidata v3.1.1 on Windows 10 with Windows Terminal 1.21.3231.0. (Python 3.8.5 and Python 3.13.1)
Mouse clicks are not registered in the application. They are on Linux. It also works when I use `ssh` to connect to a Linux server that runs visidata v3.1.1 - so it shouldn't be the terminal's fault (I think).

Is there a way to fix this?

midichef:
I do not know if this can be fixed. The lack of mouse clicks [was reported a few months ago by @daviewales](https://github.com/saulpw/visidata/pull/2563#issuecomment-2412621467). It was reported when `windows-curses` released version 2.4.* (which fixed bugs making some keyboard keys unusable). So it may be a problem in `windows-curses`. However, I have not seen any reports on [the `windows-curses` repo](https://github.com/zephyrproject-rtos/windows-curses) mentioning mouse clicks being a problem.

[The dev team does not use Windows](https://github.com/saulpw/visidata/issues/2172#issuecomment-1863705680) so the recommendation for running Windows is to use visidata inside WSL. Fixes to support Windows without WSL are welcome. They're just difficult to test and maintain.

If I had a Windows machine right now, I'd check whether `handleMouse()` is even being called when a click happens:
https://github.com/saulpw/visidata/blob/72b003a08b8e9f3b38e5c4354f11ecaa4ecb1ccc/visidata/mainloop.py#L205-L207

And if no mouse event is being generated, maybe try looking at `initCurses`:
https://github.com/saulpw/visidata/blob/72b003a08b8e9f3b38e5c4354f11ecaa4ecb1ccc/visidata/mouse.py#L15

mplattner:
Thanks @midichef. I did some testing and found a fix:

Adding `curses.def_prog_mode()` after the call to `mousemask` in `mouse.py:17` https://github.com/saulpw/visidata/blob/72b003a08b8e9f3b38e5c4354f11ecaa4ecb1ccc/visidata/mouse.py#L17 enables mouse support on Windows 10 with Windows Terminal.

Result:
```
    curses.mousemask(curses.MOUSE_ALL if vd.options.mouse_interval else 0)
    curses.def_prog_mode()
    curses.mouseinterval(vd.options.mouse_interval)
```

I found it by trial-and-error and honestly I have no idea what `curses.def_prog_mode()` is doing ;-)
[...]

PR #2721, opened 2025-03-13
Changes referenced in #2719 and #2676

ptyork:
For #2719 :
 - Add bin/vd.cmd as an alias for visidata.exe
 - Modify setup.py to include the new alias script
 - Change README.md to remove distinction between Windows and others

For #2676 :
 - Modify mouse.py to add curses.def_prog_mode() after curses.mousemask() call

### DIFF
--- a/bin/vd.cmd
+++ b/bin/vd.cmd
@@ -1,0 +1,1 @@
+visidata %*
--- a/bin/vd.cmd
+++ b/bin/vd.cmd
@@ -1,1 +1,0 @@
-visidata %*
--- a/setup.py
+++ b/setup.py
@@ -1,10 +1,14 @@
 #!/usr/bin/env python3
 
 from setuptools import setup
+import platform
+
 
 # tox can't actually run python3 setup.py: https://github.com/tox-dev/tox/issues/96
 # from visidata import __version__
 __version__ = "3.2dev"
+
+vd_bin = "bin/vd" if platform.system() != "Windows" else "bin/vd.cmd"
 
 setup(
     name="visidata",
@@ -17,7 +21,7 @@ setup(
     author_email="visidata@saul.pw",
     url="https://visidata.org",
     download_url="https://github.com/saulpw/visidata/tarball/" + __version__,
-    scripts=["bin/vd", "bin/vd2to3.vdx", "bin/vd.cmd"],
+    scripts=[vd_bin, "bin/vd2to3.vdx"],
     entry_points={
         "console_scripts": ["visidata=visidata.main:vd_cli"],
     },
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ setup(
     author_email="visidata@saul.pw",
     url="https://visidata.org",
     download_url="https://github.com/saulpw/visidata/tarball/" + __version__,
-    scripts=["bin/vd", "bin/vd2to3.vdx"],
+    scripts=["bin/vd", "bin/vd2to3.vdx", "bin/vd.cmd"],
     entry_points={
         "console_scripts": ["visidata=visidata.main:vd_cli"],
     },
--- a/setup.py
+++ b/setup.py
@@ -8,8 +8,6 @@ import platform
 # from visidata import __version__
 __version__ = "3.2dev"
 
-vd_bin = "bin/vd" if platform.system() != "Windows" else "bin/vd.cmd"
-
 setup(
     name="visidata",
     version=__version__,
@@ -21,9 +19,10 @@ setup(
     author_email="visidata@saul.pw",
     url="https://visidata.org",
     download_url="https://github.com/saulpw/visidata/tarball/" + __version__,
-    scripts=[vd_bin, "bin/vd2to3.vdx"],
+    scripts=["bin/vd2to3.vdx"],
     entry_points={
-        "console_scripts": ["visidata=visidata.main:vd_cli"],
+        "console_scripts": ["vd=visidata.main:vd_cli",
+                            "visidata=visidata.main:vd_cli"],
     },
     py_modules=["visidata"],
     install_requires=[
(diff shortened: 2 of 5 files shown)


bin/vd.cmd
```diff
@@ -0,0 +1 @@
+visidata %*
```
anjakefala: Peek 2025-03-13 20-26
Hey! 

It's possible that this doesn't work as expected on a Linux machine, but for me VisiData launches, but doesn't actually open file passed through the CLI.

anjakefala:
I really appreciate you opening this PR @ptyork! Thank you. ^^

@daviewales Are you open to helping test this PR?

visidata/mouse.py
```diff
@@ -15,6 +15,7 @@
 def initCurses(vd):
     curses.MOUSE_ALL = 0xffffffff
     curses.mousemask(curses.MOUSE_ALL if vd.options.mouse_interval else 0)
+    curses.def_prog_mode()
```
anjakefala, line 18: I confirmed no mouse functionality got borked on a Linux machine!

bin/vd.cmd
ptyork: `vd.cmd` is not intended to work on Linux. On Windows, the ".cmd" extension basically means "batch script". Though 100% harmless, it isn't supposed to be used by Linux folks. Python's setuptools automatically makes it "executable" on Linux, but really it's meant only to work on Windows. It does _kind of_ work because Linux is treating it as a shell script. So it runs `visidata` but instead of passing along the remainder of the arguments (like file names) it just tries and fails to find a file named `%*`.

But it's useful for Windows users since they can now ALSO just type `vd` (or `vd <filename>` or whatever). Here's output from my Windows Terminal:
``` cmd
C:\Users\paul ❯ vd --version

visidata --version
saul.pw/VisiData v3.2dev
```

anjakefala: Ah, thank you! I've pinged some Windows users that have contributed to see if they can help test.

setup.py
```diff
@@ -17,7 +17,7 @@
     author_email="visidata@saul.pw",
     url="https://visidata.org",
     download_url="https://github.com/saulpw/visidata/tarball/" + __version__,
-    scripts=["bin/vd", "bin/vd2to3.vdx"],
+    scripts=["bin/vd", "bin/vd2to3.vdx", "bin/vd.cmd"],
```
daviewales: Would it make sense to check the platform before appending `vd.cmd`?

ptyork: Not sure if that's a question to me or not, but I'd think it'd be more trouble than it's worth. To do it right you'd include either `vd` or `vd.cmd`. It's just a few wasted bytes on disk that 99.9% of folks would never even know existed. And it's a possible point of failure.

But I guess if wanted, it'd be something like:
```
import platform
vd_bin = "bin/vd" if platform.system() != "Windows" else "bin/vd.cmd"
```
and then
```
scripts=[vd_bin, "bin/vd2to3.vdx"],
```

anjakefala: It would fix the tests if we did it this way. I'll push up an update later today. Thanks for the suggestion @daviewales

setup.py
```diff
@@ -17,7 +21,7 @@
     author_email="visidata@saul.pw",
     url="https://visidata.org",
     download_url="https://github.com/saulpw/visidata/tarball/" + __version__,
-    scripts=["bin/vd", "bin/vd2to3.vdx"],
+    scripts=[vd_bin, "bin/vd2to3.vdx"],
     entry_points={
         "console_scripts": ["visidata=visidata.main:vd_cli"],
```
daviewales: I've been doing some more research.
Is there a reason we don't just make the following change?

```suggestion
        "console_scripts": ["vd=visidata.main:vd_cli"],
```

This will automatically create `vd.exe` on Windows, and `vd` on everything else.
We can then remove `vd_bin` from `scripts` entirely.

Additionally, I'm not sure what `vd2to3.vdx` does, but it may also make sense to add it to `entry_points["console_scripts"]` too, to ensure it automatically generates a Windows compatible executable wrapper as required.

**Edit**: If there is a desire to retain the long form `visidata`, this can just be added to the `entry_points["console_scripts"]` list:

```suggestion
        "console_scripts": ["vd=visidata.main:vd_cli", "visidata=visidata.main:vd_cli"],
```

This is also officially recommended over `scripts`:
https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#scripts

> Although setup() supports a [scripts](https://setuptools.pypa.io/en/latest/deprecated/distutils/setupscript.html#distutils-installing-scripts) keyword for pointing to pre-made scripts to install, the recommended approach to achieve cross-platform compatibility is to use [Creating executable scripts](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#console-scripts) entry points (see below).

daviewales:
I've tested this change on Windows.

I can confirm that `vd` correctly launches Visidata, and that mouse support works.

However, see my note above regarding an alternative way to get a `vd` command on Windows.

setup.py
ptyork: I think it's likely wise to keep both `vd` and `visidata` just so you don't break anything that may assume the full name. The only negative of the two executables is size. `vd.cmd` is 11 bytes. `vd.exe` would be 108,393 bytes. 108K isn't a big deal, though I'm old enough to have had floppy disks with less capacity. :D

anjakefala:
Thank you so much @daviewales for taking the time to double-check this!

setup.py
anjakefala: Excellent suggestion @daviewales!

@ptyork I just pushed an update with this suggestion. Could you install it and see if things still work on Windows?

anjakefala:
Thank you so much @ptyork and @daviewales! It's been merged and is now in develop. =)

PR #2733, opened 2025-04-21
[cliptext-] clipstr: fix handling of iterators, prepare for grapheme cluster changes

midichef:
`clipstr(iterchars('abc'),1,truncator='あ')` should return `'a', 1` but instead it returns `'c', 1`.
That's because my patch d1fb2354 introduced a bug in how `clipstr` handles iterators. It is triggered in the rare situation where the truncator is wider than the non-zero width passed to clipstr. This should not ever happen unless someone runs visidata with altered `disp_truncator`.

164bb7e14b0868d9d5d14978fc83b512df69fb05 fixes that iterator handling.

6599e5c41441871d153f2dbec1743e07d0b4aa92 is a cleanup patch. It shouldn't change the results of any `clipstr` call. But it cleans up the handling of the truncator. And it partly prepares `_clipstr` to handle grapheme clusters properly, and makes future improvement simpler. I've already got an exploratory commit ready, that builds on it. That commit isn't ready to be part of a PR, but once this PR is accepted, I'll open an issue so we can discuss the general approach.

### DIFF
--- a/visidata/cliptext.py
+++ b/visidata/cliptext.py
@@ -156,6 +156,7 @@ def _clipstr(s, dispw, trunch='', oddspacech='', combch='', modch=''):
     ret = ''
 
     trunchlen = dispwidth(trunch)
+    s = ''.join(s)
     for c in s:
         newc, chlen = _dispch(c, oddspacech=oddspacech, combch=combch, modch=modch)
         if not newc:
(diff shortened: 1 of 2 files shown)


saulpw (approved):
It's amazing how much care and work goes into truncating unicode strings for display in a fixed width.

anjakefala:
Great! I also checked that this PR doesn't revert your previous fix. Thanks for including a reference to it.

PR #2741, opened 2025-04-28
[sidebar- mainloop-] put bounds on coords for scr height 1

### LINKED ISSUE #2740: exits 255 when resized to height=1
**Small description**

When the terminal VisiData is running in is resized to a height of 1, it immediately exits with a code of 255.

**Steps to reproduce**

1. Run `vd`
2. Resize height of terminal window to 1

**Expected result**

VisiData shouldn't crash and when the terminal is resized to its original size it should continue running just like other programs likes `less` or `top` or `nvim`.

(nothing interesting here, no error messages are printed or anything)

**Configuration**

Yes.

I am using version `saul.pw/VisiData v3.1.1`

**Additional context**

Linux and MacOS.

- Which version of Python?

3.10.12

`iTerm2` with `tmux` running in it. I constantly resize tmux panes down to a height of 1 and then restore the size.

midichef:
Fixes #2740.

### DIFF
--- a/visidata/mainloop.py
+++ b/visidata/mainloop.py
@@ -64,6 +64,7 @@ def setWindows(vd, scr, pct=None):
     disp_menu = getattr(vd, 'menuRunning', None) or vd.options.disp_menu
     topmenulines = 1 if disp_menu else 0
     h, w = scr.getmaxyx()
+    if h == 1: topmenulines = 0
 
     n = 0
     if pct:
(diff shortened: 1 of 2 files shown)


midichef:
I tested this with tmux panes of height 1.

There probably needs to be similar error checking for narrow screens of width 1, but I ran vd in tmux and then shrank its pane to width 1, and didn't see any immediate crash.

saulpw (approved):

midichef:
FYI, for testing, the tmux keypresses I used to make a pane of height 1 were: `Ctrl+b` `"` to split into panes, then `Ctrl+b` `:` `resize-pane -y 1`

PR #2745, opened 2025-05-05
[input-] speed up pasting long strings into line editor

midichef:
Pasting long strings into the line editor is slow, because the sheet is redrawn after each character. What if we skip the redraw when several keys are pending? On my system, this seems to make pasting 80 characters about 2x faster.

This is a quick hack, I haven't thought about this change much.

### DIFF
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -173,14 +173,13 @@ class InputWidget:
         self.former_i = None
         self.just_completed = False
 
-    def editline(self, scr, y, x, w, attr=ColorAttr(), updater=None, bindings={}, clear=True) -> str:
+    def editline(self, scr, y, x, w, attr=ColorAttr(), updater=lambda val:None, bindings={}, clear=True) -> str:
         'If *clear* is True, clear whole editing area before displaying.'
         with EnableCursor():
             while True:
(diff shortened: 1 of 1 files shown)


saulpw (approved):

frosencrantz:
Can this PR be merged?  I've noticed this problem, and would like to see the get this change.  Thank you @midichef for providing a fix.

midichef:
On looking closer, the patch was incomplete. I expected a larger speedup than my first version gave. In my primary test case the speedup was 2x, though it gives a nearly instant paste in other cases. It turns out that `vd.drawInputHelp()` needs the same optimization. Once I stop calling that so often, pasting is nearly instant, in general.

However, that complicated the fix. When the optimization is just on `drawSheet` and `drawInputHelp`, it leads to visual artifacts when pasting is handled slowly by `updater()`. This happens only when pasting into the cmd palette. (Or perhaps in `experimental/live_search.py`). The palette processes the paste slowly, as it is compute-heavy, and until it's nearly done handling the pasted string, the Help sidebar is drawn under the updating palette, instead of on top of it. So for that reason I've decided to put the call to `updater()` under the same check.

saulpw (approved):

PR #2746, opened 2025-05-06
[sort-] fix undo when sheet has a previous ordering

midichef:
When a secondary sort column is added, and undone, the previous sort is reimposed. That can leave rows in a different order than they actually were.

For example, on this sheet:
```
x,y
0,5
0,4
0,3
```
`sort-asc` on column x, then `slide-down` on the first row, puts y values in the order:  `4, 5, 3`.
Then `sort-asc-add` on column y makes them `3, 4, 5`. `undo-last` leaves them as `3, 4, 5` when they should instead go back to `4, 5, 3`. Another `undo-last` (of `slide-down`) scrambles the rows more:  `4, 3, 5` when they should be back to their original order of `5, 4, 3`.

### DIFF
--- a/visidata/sort.py
+++ b/visidata/sort.py
@@ -6,10 +6,7 @@ def orderBy(sheet, *cols, reverse=False):
     'Add *cols* to internal ordering and re-sort the rows accordingly.  Pass *reverse* as True to order these *cols* descending.  Pass empty *cols* (or cols[0] of None) to clear internal ordering.'
     if options.undo:
         vd.addUndo(setattr, sheet, '_ordering', copy(sheet._ordering))
-        if sheet._ordering:
-            vd.addUndo(sheet.sort)
-        else:
-            vd.addUndo(setattr, sheet, 'rows', copy(sheet.rows))
+        vd.addUndo(setattr, sheet, 'rows', copy(sheet.rows))
 
     do_sort = False
     if not cols or cols[0] is None:


saulpw (approved):
Okay, this makes sense.  Less code is better code anyway =)

midichef:
@saulpw @anjakefala 
It turns out this PR accidentally fixed another bug I have been seeing intermittently for a long time. Open a medium-sized sheet (over 10k rows), and do two consecutive sort operations in opposite order:
`seq 111222 |vd -` then hit `[` `]` `U`
```
  File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/sort.py", line 74, in sort
    self.rows.sort(key=_sortkey)
  File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/sort.py", line 32, in __eq__
    return other.obj == self.obj
AttributeError: 'str' object has no attribute 'obj'
```
Now I understand what causes this error. When a sheet has an ordering, and then gets sorted in the reverse direction, and then that sort is undone, the following operations were called to undo it:
```
#from orderBy():  vd.addUndo(sheet.sort)
sheet.sort()       #gets its own thread from @asyncthread
#from orderBy():  vd.addUndo(setattr, sheet, '_ordering', copy(sheet._ordering))
sheet._ordering = previous_ordering  
```
`Sheet.sort()` has the `@asyncthread` decorator. So when it's called in `undo()`, the sort begins but will pause at some point. The first slow task it does is to call `sortkey()` on each row in the sheet. Before it has finished all the rows, it has to pause. Now the second undo instruction has time to run:  `sheet._ordering` is restored to its previous direction. That is, it reverses, from descending to ascending. The next time `sortkey()` runs, it sees the reversal in `_ordering`, so it stops returning `Reversor` objects and starts returning string objects. Eventually `sort()` tries to sort the resulting mix of `Reversor` objects and `string` objects. When it tries to compare them, we get the `AttributeError` above.

So the general rule is that **functions called as undo-helpers currently must not run in their own thread.**

I think there is only one more undo operation in the codebase that has its own thread. It's `_removeRows()`, in `modify.py`, for `addRows()`. I don't think it causes any bugs, but I'll open a separate issue for how to handle the general case of this bug for the future.

You could still run into this bug while testing #2688, so I'll prevent that by rebasing that branch to the current `develop`. EDIT: (update, actually, I won't rebase, I'll ask for a merge in that thread)

Issue #2649, opened 2024-12-22
[sort-order] Add sort order details in the ColumnSheet


frosencrantz:
It would be helpful to see the sort order in the ColumnSheet. It would be nice if sort order details could be modified from the ColumnSheet.  Another option would be to have a separate sheet for showing/modifying the sort order of the columns in a sheet.

There are 2 bits that are useful for sort order, the order of the columns when performing the sort, and whether the column should be sorted, and the direction of the sort (ascending/descending).

midichef:
I submitted a draft PR to handle the display aspect. I don't have a suggestion to handle changing the sort direction.

It'd be hard to set aside keystrokes on the `ColumnsSheet` that match all the column-sorting keys: `[`, `z[`, `]`, and `z]`. Maybe there could be a special key prefix for the `ColumnsSheet`, similar to `g` and `z` and `gz`, that would apply the keystrokes to the underlying column? So you'd go to a row that represents a `Column`, then press `prefix` + `[` to sort the column represented by that row.

frosencrantz:
Thank you for working on this, @midichef!

I have not had a chance to try the PR.  I don't remember if I had any good ideas for how to edit the sort values from the column sheet.  I guess if there was a sheet only for sort details the ordering of the rows would be the order of the sorting, and there would be a column that was for the direction of the sort, but I don't have a good key stroke for that.  Maybe I was thinking it would be something like the options sheet with boolean values.

Another prefix is a reasonable suggestion for the ColumnSheet so we can keep the '[', ']' characters. I'm not sure what that prefix would be, and does it have a related mnemonic like 'g' and 'z', something "same/remote scope" (vs wider/narrower)?  

For some sheets like ColumnSheet, I guess I would be ok with some semantic changes to the keystrokes if there is a logic to it. 
So replace the defaults for some of the keystrokes.  All the commands would still exist, and are available via menu or long commands, but key binding is specific to the column sheet.  There are different things I want to do on a ColumnSheet vs a typical data sheet.  I don't typically sort on the ColumnSheet, at times the column names, but I haven't done a secondary sort.

midichef:
[...]

PR #2748, opened 2025-05-10
[columns-] show sort direction on ColumnsSheet

midichef:
In response to @frosencrantz's [suggestion](https://github.com/saulpw/visidata/issues/2649), here's a draft to start off a discussion. It makes column names one character wider to make room for the same sort symbol used for the column headers.


midichef:
The sort arrows look like this, to the left of the column names:
columns_sort_arrows

visidata/metasheets.py
```diff
@@ -43,7 +48,7 @@ def setValue(self, srcCol, val):
 
     columns = [
             ColumnAttr('sheet', type=str),
-            ColumnAttr('name', help='rename the column on the source sheet'),
+            Column('name', getter=lambda c, r: (sort_disp_sym(c,r) or ' ') + r.name, help='rename the column on the source sheet'),
```
saulpw, line 51: I see what you're going for here, it's clever.  But the column value can't include the sort symbol, or it will make things difficult (for example editing the value).  I think the idea from #2649 for a separate sort order column is the most workable.

midichef: Oops yes, this breaks editing.

PR #2749, opened 2025-05-11
[input-] allow edit of cell in hidden column

midichef:
If you navigate to a hidden column and do `edit-cell` or `rename-col`, the edit fails with an error:
```
  File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/_input.py", line 613, in editCell
    col = self.visibleCols[vcolidx]
```
This PR stops the error by changing `visibleCols` to `availCols`.


midichef:
It also updates the `Tab` motion to move from hidden column to hidden column to eventually visible columbs. But moving from a visible column only ever moves to another visible column.

visidata/_input.py
```diff
@@ -598,19 +598,19 @@ def __call__(self, val, state):
 
 
 @Sheet.api
-def editCell(self, vcolidx=None, rowidx=None, value=None, **kwargs):
-    '''Call vd.editText for the cell at (*rowidx*, *vcolidx*).  Return the new value, properly typed.
-
+def editCell(self, colidx=None, rowidx=None, value=None, **kwargs):
```
saulpw: I use the `colidx` variable/argument name to mean an index into `columns`, and `vcolidx` to mean an index into `visibleCols`, which is always at the beginning of `availCols`.  So `vcolidx` here is kind of more appropriate still, though maybe we should invent a new `acolidx` name to indicate this.  But the ordering implied by `colidx` is incorrect here.

saulpw:
This makes sense, @midichef.  To be honest I'm not 100% sold on how hidden columns work currently, but my complaints are still hazy, so it'll probably stay for awhile.  I'll merge it once we talk about `vcolidx` vs `acolidx`.

visidata/_input.py
midichef: Okay, I changed it back to `vcolidx`. I'll leave it as `vcolidx` since it sounds like the workings of hidden columns may change in the future.

midichef:
This PR is also resolved and ready to go.

saulpw (approved):

PR #2750, opened 2025-05-11
[column-] fix rjust padding for wide chars in numeric col

midichef:
The right justification of strings in a numeric column uses Python's built in `rjust()`. `rjust` does not give correct answers for full-width characters. It inserts too many spaces, which pushes cell contents off the edge of the cell, even when they can fit on screen.

This sheet will show the behavior, when columns are made numeric:
```
a,b
１２３４,1234
/１２３４/,1234
```
It currently looks like this:
padding-wrong
With this PR, the contents fit properly:
padding-right

### DIFF
--- a/visidata/column.py
+++ b/visidata/column.py
@@ -267,7 +267,8 @@ class Column(Extensible):
            The 'generic' displayer does not do any formatting.
         '''
         if width is not None and width > 1 and vd.isNumeric(self):
-            yield ('', dw.text.rjust(width-2))
+            padding = max((width-2) - dispwidth(dw.text), 0) * ' '
+            yield ('', padding + dw.text)
         else:
             yield ('', dw.text)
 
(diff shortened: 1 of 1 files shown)


midichef:
Or maybe `displayer_generic` should follow the approach taken by `displayer_currency`, and skip right-justifying strings for cells that aren't numbers? 

`displayer_currency()` checks the typed value of the cell for the case of `int`/`float`:
https://github.com/saulpw/visidata/blob/bc37a26baccfbe792dd0771dec1ed8caffb89dab/visidata/type_currency.py#L21
and returns the rest without right-justifying them:
https://github.com/saulpw/visidata/blob/bc37a26baccfbe792dd0771dec1ed8caffb89dab/visidata/type_currency.py#L28

saulpw:
This is an interesting case.  I think I prefer that it skips right-justifying cells that aren't numbers, as per your second comment.   This seems like it's the more consistent solution and possibly more performant.  I've added it in 56a8edfdc456b3a03b45b22ed452f1989df6dfae, which I believe should supercede this PR.  Thanks for drawing our attention to this, @midichef.

PR #2755, opened 2025-05-20
[config] XDG for default visidata_dir #2716

### LINKED ISSUE #2716: input_history ends up in non-XDG location
**Small description**

vd insists on creating `~/.visidata/input_history.jsonl`
but it should use ~/.config/visidata (which is already supported for config at least) or ~/.local/share/visidata or ~/.local/state/visidata

**Steps to reproduce**

```
rm -rf ~/.visidata
mkdir ~/.config/visidata
vd <anything>
# so some stuff in vd
ls ~/.visidata
```

**Expected result**

~/.visidata should not exist

**Configuration**

visidata 3.1.1

saulpw:
Fixes #2716.  See #2754 for discussion before merging.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -63,7 +63,7 @@ jobs:
       run: |
         mkdir -p ~/.visidata
         cp tests/macros/macros.jsonl tests/macros/1.vdj ~/.visidata/
-        vd --batch -p tests/macros/test_macro.vd --output tests/macros/golden/test_macro.tsv
+        vd --batch --visidata-dir ~/.visidata -p tests/macros/test_macro.vd --output tests/macros/golden/test_macro.tsv
         git --no-pager diff --numstat tests/
         git --no-pager diff --exit-code tests/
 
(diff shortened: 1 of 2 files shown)


gennaro-tedesco:
I am trying out the new `XDG` default dir in the latest (3.2) release, however - unless I have misunderstood the purpose - it does not seem to work for configuration. I am expecting that I can move the `~/.visidatarc` to `${XDG_CONFIG_HOME}/visidata/.visidatarc`: this however seems not to read the config nevertheless.

> Use XDG for default visidata_dir https://github.com/saulpw/visidata/issues/2716 (https://github.com/saulpw/visidata/pull/2755)
> Note that if you want to keep your macros, input history, or plugins, you need to move the $HOME/.visidata directory to its XDG location; e.g. $HOME/.config/visidata on Linux.

...or is that just for the input history?

I am on a macOS 13.6.1 Ventura if it matters.

Thank you for the awesome work as always and up we go with the new release!

anjakefala:
@gennaro-tedesco It's only for the `.visidata` repo! Too many folks, like me, expect their dotfiles to be in their home directory. I think that change would've been much more disruptive.

PR #2758, opened 2025-05-24
[threads-] catch error trying to start a second profiler

midichef:
When I press '^_' for `toggle-profile`, then start any `@asyncthread` command, like `sort-asc`, an exception is raised, and drawn over the screen making it unreadable:
```
  File "/usr/lib/python3.12/threading.py", line 1030, in _bootstrap
    self._bootstrap_inner()
  File "/usr/lib/python3.12/threading.py", line 1073, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.12/threading.py", line 1010, in run
    self._target(*self._args, **self._kwargs)
  File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/threads.py", line 217, in _with_active_cmd
    _toplevelTryFunc(func, *args, **kwargs)
  File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/threads.py", line 237, in _toplevelTryFunc
    with ThreadProfiler(t) as prof:
  File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/threads.py", line 362, in __enter__
    self.thread.profile.enable()
ValueError: Another profiling tool is already active
```

It only happens at Python 3.12.0 and up, not in Python 3.11.12. More detail on that profiler error change is given here:
https://github.com/python/cpython/issues/110770#issuecomment-1759986100

I'm not sure if this PR is a full fix for the profiler. Perhaps the profiler time totals are failing to cover some times since each `ThreadProfiler.__exit__()` disables the profiler? But if the numbers are incorrect, it's not new incorrectness from Python v3.12; it's just more explicit about an error for behavior that's been incorrect.

Hiding the stacktraces is a definite UI improvement. For example, it would help in case any user accidentally turns on profiling.

### DIFF
--- a/visidata/threads.py
+++ b/visidata/threads.py
@@ -319,13 +319,14 @@ def open_pyprof(vd, p):
 @VisiData.api
 def toggleProfiling(vd):
     t = threading.current_thread()
-    if not t.profile:
-        t.profile = cProfile.Profile()
+    if not vd.options.profile:
+        if not t.profile:
+            t.profile = cProfile.Profile()
         t.profile.enable()
-        if not vd.options.profile:
-            vd.options.set('profile', True)
+        vd.options.set('profile', True)
     else:
-        t.profile.disable()
+        if t.profile:
+            t.profile.disable()
         vd.options.set('profile', False)
     vd.status('profiling ' + ('ON' if vd.options.profile else 'OFF'))
 
--- a/visidata/threads.py
+++ b/visidata/threads.py
@@ -438,6 +438,10 @@ def codestr(code):
     return code.co_name
 
 
+@VisiData.lazy_property
+def allThreadsSheet(self):
+    return ThreadsSheet("threads", source=vd.threads)
+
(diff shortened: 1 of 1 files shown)


midichef:
I added a couple more fixes.

Repeated presses of `^T` make a new sheet every time. After this PR, it only makes one sheet, and subsequent `^T` presses switch to the existing sheet.
And `toggle-profile` currently can only toggle profiling on and off once. After that, it can never turn profiling back on. I've fixed that too.

midichef:
Hmm actually the last 2 commits interact badly with each other. The existence of `allThreadsSheet` seems to stop `toggle-profile` from working.

anjakefala:
(We fixed the tests, so I updated the build!)

saulpw:
Thanks for doing this, @midichef.  It looks good to me; I don't know why allThreadsSheet would interfere with `toggle-profile`.  I can investigate later if you can't figure it out.

midichef:
I was mistaken, there is no problem interaction between the two commits. When I thought there was one, I was testing the wrong version of the code. So this PR is complete.

saulpw (approved):

PR #2759, opened 2025-05-26
[freqtbl-] fix excessive memory use for undo of selections

midichef:
On a `FreqTableSheet`, the memory use when selecting thousands of bins at once, can quickly exceed the total memory of the system. For example, on my system, let's say I select 25,000 bins by doing:
`seq 25000 |vd` then pressing `F` `Right` `,`
The memory used is about 12 GB.

The excessive use comes from storing a history of selected rows. `undoAddSelection()` makes a copy of all previously selected rows, for every bin that is selected. So for the first bin, it copies no selected rows. For the second bin it copies the 1 row previously selected for the previous bin, for the next bin 2 rows, etc. till for each the last several bins, it is copying around 25,000 rows each time. For a total of `1+2+3+...+24999` rows, and that works out to `n*(n+1)/2`, or around 300 million.

This PR changes the `FreqTableSheet` to only save the selected rows once per `FreqTableSheet` command, at the very start, instead of once for every bin selected by the command.

I'm not sure if we want to make a change to `FreqTableSheet` selection so close to the v3.2 release. This is a brand new fix and I have not tested it extensively.

### DIFF
--- a/visidata/freqtbl.py
+++ b/visidata/freqtbl.py
@@ -1,7 +1,7 @@
 from copy import copy
 import itertools
 
-from visidata import vd, asyncthread, vlen, VisiData, Column, AttrColumn, Sheet, ColumnsSheet, ENTER, Fanout
+from visidata import vd, vlen, VisiData, Column, AttrColumn, Sheet, ColumnsSheet, ENTER, Fanout
 from visidata.pivot import PivotSheet, PivotGroupRow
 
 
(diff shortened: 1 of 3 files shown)


visidata/freqtbl.py
```diff
@@ -60,13 +60,48 @@ def groupByColsName(self):
...
-        self.source.unselect(row.sourcerows)
+        self.source.unselect(row.sourcerows, add_undo=False)
         return super().unselectRow(row)
 
+    # override Sheet operations that handle multiple rows:
+    #     select(), unselect(), and toggle()
+    # to make undo more efficient. Without this optimization, the memory
+    # use for the undo-tracking on the source sheet is O(n^2) in the number
+    # of bins selected, which can easily exceed all available memory.
+    def select(self, rows, status=True, progress=True, add_undo=None):
```
saulpw: This ignores `add_undo` (which should have a bool default anyway).  Shouldn't this be passed through, in the case of a freq of a freq?  (I freq the freq count sometimes to select the singles on the original source sheet).

midichef: Oh right, that's an important case, a FreqTable that is doing operations on another FreqTable.

By looking closer at [904dc76](https://github.com/saulpw/visidata/pull/2759/commits/904dc76a26b995205b7495bfa82082ade7f182d3), I can see it definitely gets the nested FreqTable case wrong, because `select-row` on the final FreqTableSheet undoes the selections on the original data Sheet twice, instead of once.

Passing through `add_undo` helps and seems correct, but other cases are broken so this PR needs at least a few more lines of changes.

I have a draft of a fix that seems to work, but I can't be sure it's right, because the call stack gets quite complicated:
select_row -> FreqTable.select_row -> Sheet.select_row -> FreqTable2.selectRow -> FreqTable1.select (as a FreqTable) -> FreqTable1.select (as a Sheet) -> FreqTable1.selectRow -> TsvSheet.select -> TsvSheet.selectRow
I'll think about it a bit longer.

midichef:
Okay, this is a cleaner fix that definitely fixes the nested `FreqTableSheet` case. As you pointed out, passing through `add_undo` is essential.

The other part of the solution is to give `FreqTableSheet` its own version of `Sheet.addUndoSelection()` that calls `addUndoSelection()` on its source.


@saulpw
It looks like the tests are breaking at `append.vd` in `dev/test.sh` due to 37494ba2.

midichef:
I just added a commit (72ca8f03c63e548ccdb2812f642722abb39f20e8) that isn't about memory use, but still relates to frequency table selection. After making a selection in `PandasFreqTable`, undo was failing to undo the selection on the source data. To see the problem:

`vd sample_data/co3.stata`
Then press: `F` `s` `U` `q` and see that rows are still selected.

I should mention that I don't know how Pandas works. I do see from the Pandas-related code that apparently there are reasons to implement selection differently there. I think it's still similar enough that this commit should be taken as-is. But I bring up the difference, just in case there's some reason not to handle undo this way.

midichef:
This PR is complete, by the way. I had forgotten to mark part of the conversation as resolved.

saulpw (approved):

anjakefala:
This made such a big difference! Wow!

midichef:
Yes, it's not just a memory use difference, it's a speed improvement.

I've also got another optimization to propose that builds on this restructuring, I'll make a draft PR for that soon as it requires some design discussion.

PR #2761, opened 2025-05-29
Fix formatting of examples

seanf:


saulpw:
Thanks, @seanf, I think we can take this anyway :)

saulpw:
Fixing small details like this is much appreciated!

frosencrantz:
Thank you, @seanf. I had noticed this recently but hadn't reported it.  Thank you for contributing!

Issue #2763, opened 2025-05-31
[menu&mouse] Not able to activate menu by mouse click


frosencrantz:
**Small description**
Not able to activate menu by mouse click. 

**Data to reproduce**

None needed.

**Steps to reproduce**

Try clicking on the menu with a mouse.  If you click on an item right after VisiData starts, it opens for a second, and then closes, and it is not possible to open again.  If you first click somewhere else, and then try to click on the menu, it does not open the menu, but does allow you to switch the current column depending on where you click in the menu, as if the menu was not there.

Mouse clicks appear to work elsewhere.

**Expected result**

Can click on menu, and navigate menu with mouse and keyboard after that.

**Actual result with screenshot**

Sorry not screenshot.

**Configuration**

- Does this issue reproduce without any plugins or configuration (using the `-N` CLI flag)?

- Does this issue reproduce with either the [latest release](https://www.visidata.org/releases/), or with the [develop branch](https://www.visidata.org/install/#update-visidata-from-an-existing-installation)?
- 
It does not happen with the latest release, it does happen with the latest from develop

**Additional context**

- What platform and version are you using (Linux, MacOS, Windows)? Linux & MacOS

- Which version of Python? 3.13

- Which terminal are you using (for display and input issues)?
with and without tmux on multiple terminal emulators.

midichef:
Some more information. I can reproduce this on Python 3.12.3 on Ubuntu 24.04.

bisect traces the first instance of this problem behavior to a4f47913. To fix it requires reverting that commit and also the later 9d2d8674.

saulpw:
Thanks for reporting, @frosencrantz, and for investigating, @midichef.  Those commits were related to the fix for #2607.  They seemed safe and innocuous enough, perhaps a bit heavy-handed but they made sense to me.  It doesn't make sense that overclearing the drawcache would be detrimental to mouseclicks; I'm guessing it's related to https://github.com/saulpw/visidata/blob/develop/visidata/mouse.py#L29.  I'm not going to dive deep into this one right now though, as reverting the two commits you called out does not seem to regress #2607.

frosencrantz:
Thank you, @saulpw and @midichef!  The menu is back to working.

PR #2765, opened 2025-06-04
[misc-] fix undo editing cells in unusual columns

midichef:
For several specialized column types, an error is seen after editing a cell with `e`, and trying to undo it with `U`:
```
  File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/undo.py", line 109, in _undo
    c.setValue(r, v, setModified=False)
TypeError: ColumnsSheet.ValueColumn.setValue() got an unexpected keyword argument 'setModified'
```

To test the fix for `ValueColumn` (ae7de1fbe438cb24ecb50b884d63f292c670a080):
`echo a |vd -`
`C` `zc4` `Enter` `e` `asdf` `Enter` `U`

To test the fix for `ExpandedColumn` (cf124f492b93b02389200a39548d6f0597b58d23):
`echo '{"col": {"k":"val"}}' |vd -f json`
`(` `e` `asdf` `Enter` `U`

for `ColumnSourceAttr` (33113186183dd22ce0f133802e155134c7b96b61):
`echo a |vd -`
`^X` `sheet.cursorCol` `Enter` then `/` `name` `Enter` then `zc1` `Enter` then `e` `asdf` `Enter` then `U`

for org-mode file orig_contents (33fb6c862808f13d63e599d904eb76a54da90d0e):
`touch testfile.org; vd testfile.org`
`zc9` `Enter` then `_` then `zc5` `Enter` then `e` `asdf` `Enter` then `U`

I believe these commits fix all the `setValue` variants that could trigger this type of error.

There  is one remaining variant in `loaders/_pandas.py`, but it is not called during undo of an edit. That's because `PandasSheet` creates columns with a setter function that gets called for undo:
https://github.com/saulpw/visidata/blob/49f0e533f261d9f4db67f3167b547f1d8ff8b39d/visidata/loaders/_pandas.py#L191
So `setValue(..., setModified=True)` in the undo eventually calls `putValue` -> the `setter` function -> `setValue()` where `setModified` argument isn't supplied, so this `setValue()` variant does not need to be fixed.

### DIFF
--- a/visidata/loaders/orgmode.py
+++ b/visidata/loaders/orgmode.py
@@ -55,9 +55,11 @@ def encode_date(dt=None):
 
 
 class OrgContentsColumn(Column):
-    def setValue(self, row, v):
+    def setValue(self, row, v, setModified=True):
         super().setValue(row, v)
         orgmode_parse_into(row, v)
+        if setModified:
+            self.sheet.setModified()
 
     def putValue(self, row, v):
         self.sheet.save(row)
--- a/visidata/loaders/orgmode.py
+++ b/visidata/loaders/orgmode.py
@@ -56,10 +56,8 @@ def encode_date(dt=None):
 
 class OrgContentsColumn(Column):
     def setValue(self, row, v, setModified=True):
-        super().setValue(row, v)
+        super().setValue(row, v, setModified=setModified)
         orgmode_parse_into(row, v)
-        if setModified:
-            self.sheet.setModified()
 
     def putValue(self, row, v):
         self.sheet.save(row)
@@ -90,6 +88,7 @@ def sectionize(lines):
 
 def orgmode_parse(all_lines):
     root = parent = OrgSheet().newRow()
+    root.orig_contents = ''
     for linenum, lines in sectionize(all_lines):
         section = OrgSheet().newRow()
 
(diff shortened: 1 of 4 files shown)


midichef:
The error in the tests right now is unrelated to these commits. It has to do with the changes in PR # [2763](https://github.com/saulpw/visidata/issues/2763).

Of note, for this PR, I used a base commit other than the most recent `develop`, because that is currently failing tests in a different way. I was trying to use an earlier known-good commit that wouldn't fail that specific way. I'm not sure why the test breaks here, to be honest. The tests pass on my system, both `cd visidata; pytest -xsv tests` and `dev/test.sh`.

visidata/loaders/orgmode.py
```diff
@@ -55,9 +55,11 @@ def encode_date(dt=None):
 
 
 class OrgContentsColumn(Column):
-    def setValue(self, row, v):
+    def setValue(self, row, v, setModified=True):
         super().setValue(row, v)
```
saulpw: I think it would be better to pass `setModified=setModified` here, instead of duplicating the code below.

midichef: Oh, right, okay, I added a commit that handles that. And initializes the first row `orig_contents` to `''` instead of `None` because `setValue` wasn't handling the restoration of `None` for the root node, it expected a string.

saulpw (approved):

PR #2766, opened 2025-06-07
[input, sheets] fix off-by-1 when drawing at right edge of screen

midichef:
This PR fixes errors in v3.2dev that were not visible in v3.1.1 and earlier versions. They were uncovered by my changes to v3.2dev in 0ec05d3.

de9b70b92b4318d06fb52d322ee0cb57b38b3efa is a necessary fix to a bug I introduced in 0ec05d3. When there is only one very wide cell, editing tries to draw past the edge of the screen:
`python -c "print('12345'*100)" |vd -` then press `e`
```
File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/_input.py", line 242, in draw
  clipdraw(scr, y, x+w, ' ', attr, 1, clear=clear, literal=True)
...
```
So this commit adds a bounds check.

But there is another lower-level reason for the above bug, which is that the cell's calculated width is too large by 1. The calculation does not take right-side margin into account. 0e53e2fae5a5826842471e957fa1b9fb50256996 fixes that. Specifically, the problem is in `calcSingleColLayout()`. The specific case where it goes astray is if the cell starts at x=0 and is as wide as the screen, or wider. It gives a width that is too large by 1.

This part of the column layout calculation has been essentially the same since around 2019. I'm a bit hesitant in editing code that's been fine that long. It makes me wonder if there are other off-by-1 adjustments that have compensating for the off-by-1 error, that might be broken by this change (the 3rd commit in this PR, dcca752bf5d989664e3d2583669a1be7715f0511, fixes just such an adjustment). Logic tells me the only situations affected by this fix are when columns satisfy (EDITED) `x == self.windowWidth-width-1`. So that's what to look for, in the future, when seeing if this fix has caused other problems. Two particular special cases that are affected are: 1) `x == 0` and `width == self.windowWidth-1`, e.g. the single wide cell for the `edit-cell` bug above, and 2) `x+width == self.windowWidth-1`, a column whose last character is the last column of the sheet. The next commit relates to a change in column separators caused by exactly that second special case:

dcca752bf5d989664e3d2583669a1be7715f0511 fixes a drawing anomaly that occurs after 0e53e2fae5a5826842471e957fa1b9fb50256996 is applied:  for a sheet wider than the screen, try resizing the screen smaller, dragging it through a series of decreasing widths. Most layouts cause column separators to be drawn in the right-side margin.

### DIFF
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -239,7 +239,9 @@ class InputWidget:
 
         #clipdraw will truncate the right side of dispval with trunch as needed
         clipdraw(scr, y, x, dispval, attr, w, clear=clear, literal=True)
-        clipdraw(scr, y, x+w, ' ', attr, 1, clear=clear, literal=True)
+        if x+w < scr.getmaxyx()[1]:
+            #draw a space to indicate that the user can scroll right of the cell's final char
+            clipdraw(scr, y, x+w, ' ', attr, 1, clear=False, literal=True)
         if scr:
             prew = dispwidth(dispval[:i])
             scr.move(y, x+prew)
--- a/visidata/cliptext.py
+++ b/visidata/cliptext.py
@@ -215,8 +215,6 @@ def clipdraw(scr, y, x, s, attr, w=None, clear=True, literal=False, **kwargs):
 
     x = max(0, x)
     y = max(0, y)
-    assert x >= 0, x
-    assert y >= 0, y
 
     return clipdraw_chunks(scr, y, x, chunks, attr, w=w, clear=clear, **kwargs)
 
--- a/visidata/column.py
+++ b/visidata/column.py
@@ -474,7 +474,7 @@ class Column(Extensible):
                     break  #1747  early out to speed up wide columns
             w = w_max
         w = max(w, nlen)+2
-        w = min(w, self.sheet.windowWidth)
+        w = min(w, self.sheet.windowWidth-1)
         return w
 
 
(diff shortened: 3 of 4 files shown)


anjakefala:
Test failures are now real!

saulpw:
The current test failure is that, in a test that saves a markdown file, the final column is some 10 characters wider.  I would accept a single character width change given the markdown format's sensitivity to column size, but it seems odd that the width has increased so much.

I think the intent of the fix is fine otherwise; @midichef can you look into the markdown saving difference?

midichef:
Okay, I fixed it so the tests pass. There was another `windowWidth` that needed to be changed to `windowWidth - 1`.

`checkCursor` runs a loop to layout columns using `calcColLayout()`/`calcSingleColLayout()`. The earlier commit changed `calcSingleColLayout()` to avoid the right margin, but the test in `checkCursor` needed to be updated to match.

I audited the rest of the lines in the codebase that use `windowWidth`. No important lines need the same kind of `-1` adjustment, though a few unimportant lines could use it for the sake of technical correctness.

So this PR is done. And in my opinion, it should be part of v3.2, to avoid breaking the edit of cells in wide columns.

midichef:
Oops, looking at the leftover instances of `windowWidth`, there is actually one more boundary check that does make a visible difference, `getMaxWidth()`, so I've updated that one too.

saulpw (approved):

PR #2783, opened 2025-06-15
[npy-] fix bug truncating cols wider than window

midichef:
When saving .npy files, any column wider than the window has its data truncated to fit in the window width.

To test the bug and its fix, see that saving a batch file as a .npy file truncates anything after the 80th character:
```
py -c "print(79*'-'+'>🦜')" |vd --batch -o truncated.npy; vd --batch truncated.npy -o -
[...]
------------------------------------------------------------------------------->
```
After the fix the output is no longer truncated:
```
[...]
------------------------------------------------------------------------------->🦜
```

Thanks to @daviewales for the utility function `getMaxDataWidth()` from #2255, fixing a similar truncation error in #1849. It made this PR very simple.


midichef:
I updated the test `save-benchmarks.npy` to use the output of the new `save_npy()`. The old test file didn't have truncation, but it did have different field widths for each column of data.

saulpw (approved):

anjakefala:
Very cute test!

PR #2789, opened 2025-06-16
fix: include missing files in MANIFEST.in

chenrui333:
- [ ] If contributing a core loader, [the loader checklist](https://visidata.org/docs/contributing#loader) was referenced.
- [ ] If registering an external plugin, [the plugin checklist](https://visidata.org/docs/contributing#plugins) was referenced.

---

seeing some build errors as 

```
    File "/private/var/folders/4w/t9h6qm850g395cdb8js574nc0000gn/T/pip-req-build-9p78t7gz/setup.py", line 10, in all_requirements
      with open('requirements.txt', 'r', encoding='utf-8') as f:
           ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  FileNotFoundError: [Errno 2] No such file or directory: 'requirements.txt'
```

relates to https://github.com/Homebrew/homebrew-core/pull/226979

PR #2792, opened 2025-06-17
add requirements.txt to sdist

sthen:
The sdist tar.gz files on pypi don't currently include requirements.txt. As of visidata 3.2 this is now required by setup.py when installing (or creating a wheel or sdist) so the current sdist is not usable.


saulpw:
Thanks for the PR, @sthen.  It looks like this is a subset of #2789, which I just approved.  So we'll take that one and make a patch release soon.

PR #2805, opened 2025-07-02
[editor-] sysedit-cell: edit unprintable values as binary

midichef:
I'm trying to allow `sysedit-cell` (`^O`) to edit cells as binary, if they contain unprintable values.

It'd be useful to allow me to save binary content, such as PNG files that are cells in SQLite database sheet, by opening the cell in an editor for saving.

But I'm not sure about how to properly get the data in a way that handles all the possible kinds of cells. Right now this draft PR uses `str(sheet.cursorValue)`. What's the proper way to do it?
```
cv = str(sheet.cursorValue)
edit = vd.launchExternalEditor(bytes(cv, encoding=sheet.options.encoding), binary=True)
```

### DIFF
--- a/visidata/editor.py
+++ b/visidata/editor.py
@@ -43,27 +43,33 @@ def launchBrowser(vd, *args):
 
 
 @visidata.VisiData.api
-def launchExternalEditor(vd, v, linenum=0):
-    'Launch $EDITOR to edit string *v* starting on line *linenum*.'
+def launchExternalEditor(vd, v, linenum=0, binary=False):
+    'Launch $EDITOR to edit bytes *v* starting on line *linenum*.'
     import tempfile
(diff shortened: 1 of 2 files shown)


midichef:
@saulpw I'm pinging you to make my question easier to notice: How should I process a cell to extract its contents as an unformatted Python value, then convert it to a string for external processing outside Python?

saulpw:
Thanks for the ping, sorry I missed this (I skipped "Drafts" in this recent set of reviews).

This is tricky to do in the general case.  Maybe use `cursorTypedValue` and then change `launchExternalEditor` to do something different in the `bytes` case, and let it stringify everything else?

midichef:
I'm closing this draft PR, as I'm not sure how to make the editing work robustly for all data types.

Issue #2806, opened 2025-07-04
getDisplayValue() gives incorrect widths for collections


midichef:
**Small description**
For cells that have dict/list/tuple contents, `getDisplayValue()` produces unexpected results that depend on the column width.

**Steps to reproduce**
`echo '{"string": "[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]", "list": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]}' |vd -f jsonl`

**Expected result**
There are a variety of anomalies with the list column. 
1. I would expect both columns to be equally wide. But the second column, holding a list, is only wide enough to fit its name `list`.
2. In a wide window, resizing a narrow list column with `_` causes it to repeatedly increase its width. Each time it more than doubles its width, to 2*`col.width`+2, until it reaches about half the window width and can no longer double.
3. `syscopy` results are truncated for list cells, but not string cells. For example, start by narrowing the `string` column to width `6` with `z_` to match the `list` column. `zY` on `string` takes the full cell contents: `[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]`, but on the too-narrow `list` column it takes only `[10] 1; 2; …`. The difference in truncation is surprising. And the width of that truncated value is surprising too. It is actually wider than the value shown on screen, which is `[10]…`.

**Configuration**
Anomaly #1 happens in visidata 3.2 but not 3.1.1 due to d1fb2354f7cdb65ac7f93ae2c5aefc18743422b3.
Anomaly #2 is a longstanding visidata behavior.
Anomaly #3 is longstanding too.

**Additional context**
The problem behavior here is for dict/list/tuple columns. It manifests in a variety of ways. I'll use this as a central thread for discussing the `getDisplayValue` width-related behavior. And I'll make separate PRs that address the different manifestations.

saulpw:
Thanks for logging this and addressing some of the issues already.  It looks like you've fixed anomaly 1 and anomaly 2.  I just tested anomaly 3 and I agree, that is a bug; I would expect syscopy to receive the entire contents of the cell no matter how wide the column is.

saulpw:
Please verify that my commit fixes the issue for you @midichef

midichef:
Yes, this works. We should also make the same change to use `cursorFullDisplay` in `sysedit-cell`.

[...]

PR #2807, opened 2025-07-04
[column-] for collection cols, improve getMaxWidth and getDisplayValue

midichef:
This PR addresses part of #2806. It is for columns that hold collections, specifically dicts/lists/tuples. They have problems setting their width.

For example, the default layout will only let a dictionary column be as wide as its column name.
before
After this PR, the layout is back to normal:
after

This layout problem is specific to visdata v3.2. It was introduced by https://github.com/saulpw/visidata/commit/d1fb2354f7cdb65ac7f93ae2c5aefc18743422b3. Before, `clipstr(s, 0)` returned the original string. Now it returns `''`.

A related problem: if you have a column containing only dictionaries/tuples/lists, and its width is say, `10`, hitting `_` for `resize-col-max` will not make it take up all free width on the screen. Instead, its new width is `2*width+2`. In this case, `22`. Hitting `_` again more than doubles its width to `46`. This behavior is longstanding and not specific to v3.2.

This PR also fixes that size-doubling behavior. The new behavior is as expected: `_` makes the column take up all free space, and hitting it again decreases it to `default_width`.

**The cause of these problems**
When `getMaxWidth()` runs on dict/list/tuple cells, it uses the current width of the column.

For initial layout, when the column width is `None`, the width used is `0`. That leads function calls of `getMaxWidth` to eventually pass `0` to `clipstr()`: `getmaxWidth()` ->->-> `format(typedval, self.width or 0)` ->-> `clipstr(iterchars(typedval), dispw=0)`. The result of that clipping is `''`, as of visidata v3.2. So every cell looks like it has width 0, and the column's width is set by its header name.

For `_` causing a column's width to roughly double, the ultimate cause is a different line <s>not changed by this PR</s>: (EDIT: I subsequently pushed a commit that does change this line)
https://github.com/saulpw/visidata/blob/654eb8707961f35b48a8caf7019180417a880ec5/visidata/column.py#L401
<s>However, the doubling is fixed by this PR without needing to change that line, because `getMaxWidth()` constrains the final width to be the size of the drawable window</s>.

### DIFF
--- a/visidata/column.py
+++ b/visidata/column.py
@@ -401,7 +401,7 @@ class Column(Extensible):
         dw.typedval = typedval
 
         try:
-            dw.text = self.format(typedval, width=(self.width or 0)*2) or ''
+            dw.text = self.format(typedval, width=self.width) or ''
 
             # annotate cells with raw value type in anytype columns, except for strings
             if self.type is anytype and type(cellval) is not str:
--- a/visidata/column.py
+++ b/visidata/column.py
@@ -465,21 +465,17 @@ class Column(Extensible):
         return vd.status('set %d cells to %d values' % (len(rows), len(values)))
 
     def getMaxWidth(self, rows):
-        'Return the maximum length of any cell in column or its header (up to window width).'
-        w = 0
+        'Return the maximum length of any cell in column or its header (up to drawable window width).'
+        drawable_width = self.sheet.windowWidth-1
         nlen = dispwidth(self.name)
-        if len(rows) > 0:
-            w_max = 0
-            for r in rows:
-                row_w = dispwidth(self.getDisplayValue(r), maxwidth=self.sheet.windowWidth)
-                if w_max < row_w:
-                    w_max = row_w
-                if w_max >= self.sheet.windowWidth:
-                    break  #1747  early out to speed up wide columns
-            w = w_max
-        w = max(w, nlen)+2
-        w = min(w, self.sheet.windowWidth-1)
-        return w
+        w_max = nlen
+        for r in rows:
+            row_w = dispwidth(self.getDisplayValue(r), maxwidth=drawable_width)
+            if w_max < row_w:
+                w_max = row_w
+            if w_max >= self.sheet.windowWidth:
+                break  #1747  early out to speed up wide columns
+        return min(w_max+2, drawable_width)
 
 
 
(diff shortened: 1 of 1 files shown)


midichef:
After thinking some more, I see that 4b80657 is only a workaround and does not get at the fundamental problem. I'll propose an alternate solution here soon.

midichef:
Okay I tried a different approach. The fundamental problem is that `getMaxWidth()` shouldn't actually use `getDisplayValue()`. The display value for a dict/list/tuple is clipped to the column's current width. What is needed by `getMaxWidth()` is the width of the cell values, independent of the current column width. So I made a new function purely for measuring the width of data in cells.

For speed, it stops measuring at the right edge of the screen. For large data, it will be significantly faster than the previous implementation. For a single wide dict column (40 rows of 450KB wide cells), it cuts the time to run`getMaxWidth()` from ~170ms to ~2ms, a factor of 85x (using v3.1 as a reference; v3.2 is fast but wrong). (We previously discussed this kind of optimization in https://github.com/saulpw/visidata/issues/1728#issuecomment-1421680748, but at that time I was not knowledgeable enough to implement it.)

midichef:
I also added 222a563, which removes a width doubling that has been present since 58159d91642fad62569e5f1d622e179c5cc14f57. This fixes a bug where the `syscopy`/`zY` value of a dict/list/tuple is usually 2x wider than the value on screen, when the data is wider than the column.

By removing `self.width or 0`, this commit also resolves a new inconsistency, as of v3.2, between `getCell()` and `formatValue()`, in how they handled `col.width` of `None` vs `0`.  `getCell()` treated `col.width` of `None` just like `0`. But `formatValue()` used `width` of `None` to mean "return the full-width value" and `0` to mean "truncate the value to the empty string". This inconsistency was created by https://github.com/saulpw/visidata/commit/d1fb2354f7cdb65ac7f93ae2c5aefc18743422b3, my change to `clipstr()`, which changed `0` to mean "truncate" instead of "return full value".

visidata/column.py
```diff
@@ -461,22 +465,36 @@ def setValuesTyped(self, rows, *values):
...
+        typedval = wrapply(self.type, cellval)
+        if isinstance(typedval, (TypedWrapper, threading.Thread)):
+            return dispwidth(self.getCell(row).text, maxwidth=maxwidth)
+        try:
+            text = self.format(typedval, width=maxwidth) or ''
+        except Exception as e:  # formatting failure
+            try:
+                text = str(cellval)
+            except Exception as e:
+                text = str(e)
```
saulpw, line 496: Should the error text here be the outer Exception?  (I'm not sure in what case this would happen anyway).

saulpw (approved):
Thanks again, @midichef!

Issue #2794, opened 2025-06-19
[layout] hide-uniform-cols raises KeyError drawing a hidden column


midichef:
**Small description**
`hide-uniform-cols` raises `KeyError` for hidden column during `draw_all()`

**Data to reproduce**
I cannot reproduce the error. I have only had it once.

**Steps to reproduce**
I believe the commands were: `go-right` in a sheet with a few dozen columns; perhaps move left a few times, keeping the sheet's rightmost column visible; then `hide-uniform-cols`. The rightmost column became hidden because it was uniform, and so did most of the other columns.

The file that had this error was a JSON file, 50MB in size, 20k rows, 44 columns. But I haven't been able to reproduce the error with that file, in several dozen attempts. It may be relevant that the final column contained a small JSON dictionary.

**Actual result with screenshot**
```
    File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/mainloop.py", line 192, in mainloop
        vd.draw_all()
    File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/mainloop.py", line 119, in draw_all
        vd.drawSheet(vd.win1, ss1[0])
    File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/mainloop.py", line 47, in drawSheet
        vd.callNoExceptions(sheet.draw, scr)
    File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/mainloop.py", line 28, in callNoExceptions
        return func(*args, **kwargs)
    File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/sheets.py", line 861, in draw
        rightx, rightw = self._visibleColLayout[self.rightVisibleColIndex]
    KeyError: 43
```

**Configuration**
visidata 3.2 initial release, commit 654eb870
Ubuntu 24.04.2, Python 3.12.3

midichef:
@frosencrantz I know you use `hide-uniform-cols` too, perhaps you can keep an eye out for this error?

frosencrantz:
@midichef I have not noticed this error.  I have seen other issues with hidden columns, like running a frequency sheet on a hidden column, and the column is not shown.

saulpw:
Seems like it was a spurious error.  If it happens again, please document here and reopen.

PR #2810, opened 2025-07-18
fix availCols IndexError in nScreenRows

### LINKED ISSUE #2607: undoing column additions causes `IndexError` in `allAggregators`
**Small description**
After adding a column, undoing the add causes an error.

**Steps to reproduce**
Open any visidata sheet, then `i` then `U`:
`vd sample_data/sample.tsv`

**Expected result**
The column insertion is undone, with normal status.

The column insertion is undone, but an error appears:
```
  File "/home/j/.venv/lib/python3.10/site-packages/visidata/mainloop.py", line 252, in mainloop
...
  File "/home/midichef/.venv/lib/python3.10/site-packages/visidata/sheets.py", line 643, in checkCursor
...
  File "/home/midichef/.venv/lib/python3.10/site-packages/visidata/sheets.py", line 780, in allAggregators
    col = self.availCols[vcolidx]
```

The error is probably related to caching. Clearing caches prevented [the same traceback after hiding a column](https://github.com/saulpw/visidata/pull/2578#issuecomment-2436796516).

**Configuration**
visidata v3.2dev
Python 3.10.12

midichef:
Fixes #2607. Most likely prevents the never-reproduced #2794. Changes the fix for #2578. 

tldr: this PR makes sure that `calcColLayout()` happens every time `checkCursor` runs. 

Here's how the error happens in #2578 and #2607 . An `IndexError` happens in `checkCursor()`, when deciding whether to adjust the view vertically, due to the use of `nScreenRows`. `nScreenRows` has to access the list of which columns are visible on screen, to set aside rows for their aggregators. But after any command that makes columns disappear, the indexes into that list become inaccurate. They stay inaccurate until the next time `calcColLayout()` is called. That happens in every `draw()`, and it also happens at the end of `checkCursor()`, but only in some situations.

The simplest fix is to change `checkCursor()` to make sure `calcColLayout()` runs every time, and that it runs before the use of `nScreenRows` in `bottomRowIndex`. `calcColLayout()` was already being run there if the cursor was right of the leftmost visible column. But now it's being run in the other two situations:
1) when the cursor is in the leftmost visible column, or left of it
2) when the cursor is in a key column

And when the layout moves the cursor column as the first visible column, `calcColLayout()` is also running one more time, added where the loop used to exit:
https://github.com/saulpw/visidata/blob/aad595dc5dceaabfd9ec716428c6c00ffe91e625/visidata/sheets.py#L672-L673
That removes a small period of inconsistency where `_visibleColLayout` could differ from the first run of `checkCursor()` after a command, vs the next time it ran in `draw()`. (I forget the details, but I think the difference was slight and never visible.)

The benefits of this change: because `execCommand` runs `checkCursor` for every command, we do not need the `calcColLayout()` in `undo-last` and `contract-cols`, or any future commands. And we can take it out of `cursorRight()`, which may make this PR a net speedup.

The error was not related to caching. [I once claimed that cache clearing fixed this error](https://github.com/saulpw/visidata/pull/2578#issuecomment-2436796516) for `hide-uniform-cols`. But now when I try to see the effect of cache clearing in past visdata versions, I don't see it fixing any error. So I've removed that code from `hide-uniform-cols`. If it ever was beneficial, it should definitely not be needed after this PR.

I've also got a few more improvements to `calcColLayout()`. But since they do not relate to the `IndexError`, I'll submit them as a separate PR that depends on this PR.

### DIFF
--- a/tests/golden/bulk-rename-cols.tsv
+++ b/tests/golden/bulk-rename-cols.tsv
@@ -1,8 +1,8 @@
 name	width	type	fmtstr	value	aggregators	sortorder
-feature_OrderDate				2016-01-06		
-feature_Region				East		
-feature_Rep				Jones		
-feature_Item				Pencil		
-feature_Units				95		
-feature_Unit_Cost				1.99		
-feature_Total				189.05		
+feature_OrderDate	12			2016-01-06		
+feature_Region	9			East		
+feature_Rep	10			Jones		
+feature_Item	9			Pencil		
+feature_Units	7			95		
+feature_Unit_Cost	11			1.99		
+feature_Total	9			189.05		
--- a/tests/golden/load-jsonla.tsv
+++ b/tests/golden/load-jsonla.tsv
@@ -1,4 +1,4 @@
 name	width	type	fmtstr	value	aggregators	sortorder
-Name				Daniel		
-Number		int		1		
-Exile		bool		True		
+Name	16			Daniel		
+Number	8	int		1		
+Exile	7	bool		True		
--- a/visidata/features/expand_cols.py
+++ b/visidata/features/expand_cols.py
@@ -143,7 +143,6 @@ def contract_cols(sheet, cols, depth=1):  # depth == 0 means contract all the wa
             col.width = sheet.options.default_width
 
         sheet.columns = [col for col in sheet.columns if getattr(col, 'origCol', None) not in origCols]
-        sheet.calcColLayout()
 
 
 @Sheet.api
--- a/visidata/features/layout.py
+++ b/visidata/features/layout.py
@@ -49,7 +49,6 @@ def hide_uniform_cols(sheet):
             if i <= idx:
                 sheet.cursorRight(-1)
             col.hide()
-            Sheet.clear_all_caches()  #2578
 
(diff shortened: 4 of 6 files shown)


saulpw:
Aye, that's a deep dive, @midichef!  This has been plaguing us for some time now.  Thanks for putting it all together.  

@anjakefala Please test thoroughly.  Make sure we don't regress #2763 in addition to fixing all the issues mentioned in the description.

saulpw (approved):

midichef:
I wanted to open a PR that builds on this PR, but I couldn't see a way to submit it to this repo as dependent. I ended up submitting it to my own repo:  https://github.com/midichef/visidata/pull/1.

I mention it for guiding testing. It may be more efficient to test just the final commit in that PR, because both PRs modify the inner workings of `calcColLayout`/`calcSingleColLayout`.

anjakefala:
Hey @midichef, so sorry for the delay. Could you please clarify which set of commits I should test?

midichef:
The commits to test are the 9 commits in my branch [midichef/calc_col_layout](https://github.com/midichef/visidata/tree/calc_col_layout). That's the 6 commits in this PR, plus 3 more building on those 6.

anjakefala:
@midichef Hey midifchef! We're going to take this. Please open a new PR with the updated commits!

PR #2811, opened 2025-07-18
streamline the testing process

midichef:
`dev/test.sh` used to exit with a non-zero status when the `tests/golden` have diffs. But now it doesn't. This PR restores that exit status.

I also pushed a commit to skip the test of `sysedit-cell`, to avoid the blocking step of having to quit out of vim.

fyi as a general tip, I also avoid the blocking effect of browser tests by running the tests as `BROWSER=~/bin/browser.bg pytest -xsv tests`, where `browser.bg` is just a shell that opens a browser in the background:
```
#!/bin/sh
firefox "$@" < /dev/null >/dev/null 2>/dev/null &
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,12 +40,14 @@ jobs:
     - name: Ensure VisiData can load a dir
       run: vd . --batch
 
-    - name: Install lxml dependencies
-      run: sudo apt install -y libxml2-dev libxslt-dev
-
-    - name: Install ping dependencies
-      run: sudo apt install -y traceroute
-
+# Apt packages are slow to install, often taking up to a minute.
+# lxml dependencies are unneeded when installing lxml via pip
+#    - name: Install lxml dependencies
+#      run: sudo apt install -y libxml2-dev libxslt-dev
+# only needed for testing open-ping, which isn't tested currently anyway
+#    - name: Install ping dependencies
+#      run: sudo apt install -y traceroute
+#
     - name: Install optional dependencies
       run: |
         pip3 install ".[test]"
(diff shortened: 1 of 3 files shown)


midichef:
I pushed a change to `dev/test.sh` to silence warnings and errors from [shellcheck](https://www.shellcheck.net/).

The errors it flagged caused no problems in practice. The most helpful error msg was this one, on why not to mix eval with arrays:
https://www.shellcheck.net/wiki/SC2294

The other catch was that nested quotes actually led to the `$i` argument to be unquoted:
https://github.com/saulpw/visidata/blob/aad595dc5dceaabfd9ec716428c6c00ffe91e625/dev/test.sh#L59

midichef:
I pushed a change to the build process to skip all the apt packages. It saves 18-60 seconds per visidata-ci-build.

The apt packages weren't needed for the current testing setup, because:
1)  `pip install lxml` sets up lxml fine, so we don't need the libxml and libxslt packages
2) `traceroute` is only needed to test vd's `open-ping` command, but that's not getting tested anyway since it requires the pip `sh` package.

I don't know why the apt install was so inconsistent in its length. In any given workflow, the 6 runs (for 6 Python versions) would not take the same amount of time. It seems 4 of the visidata-ci-builds would only need 15-20 seconds for the apt step, but another 2 seemingly identical builds would take over a minute.

In my experiments, disabling just the libxml/xslt install, or just the traceroute install wouldn't solve the delay. It would just push the 60 second apt delay into the remaining apt install. It wasn't specific to the package. So that's why I've disabled both.

saulpw (approved):

PR #2813, opened 2025-07-22
[basesheet] start new column names fresh for each sheet

### LINKED ISSUE #2796: Auto-named column name counter never resets back to 'A'
Test on current dev...

The auto-named column name counter never resets back to 'A' ... 
Creating the first new sheet, col 0 will be named 'A', and any other auto-name col add action (e.g. `addcol-incr`) will step the letters up as expected, but start a new sheet and it won't start from 'A' for that sheet.

To observe:
Run `open-new` multiple times, only the first one will have its column named as 'A'

midichef:
Closes #2796.

### DIFF
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -290,7 +290,7 @@ class TableSheet(BaseSheet):
         pass
 
     def resetCols(self):
-        'Reset columns to class settings'
+        'Reset columns to class settings or constructor settings'
         self.columns = []
         for c in self.initialCols:
             self.addColumn(deepcopy(c))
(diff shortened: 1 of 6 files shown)


visidata/sheets.py
```diff
@@ -1169,8 +1169,15 @@ def preloadHook(sheet):
 
 @VisiData.api
 def newSheet(vd, name, ncols, **kwargs):
-    return Sheet(name, columns=[SettableColumn(width=vd.options.default_width) for i in range(ncols)], **kwargs)
-
+    #use temporary column names until we have the sheet object that can assign names and keep track of them
+    rename_cols = [SettableColumn(width=vd.options.default_width, name=f'name{i}') for i in range(ncols)]
+    vs = Sheet(name, columns=rename_cols, **kwargs)
+    # For TableSheet, we cannot change colnames using setName() on vs.columns[i],
```
saulpw, line 1175: This smells like unnecessary complexity. Ideally this would all work smoothly so that you wouldn't have to dig into it and have to explain why it's done in a funny way.  But I'm unmotivated to untangle it myself and I would be loath to inflict it on you, so we'll just take this as-is.  Thanks for your attention to the details!

saulpw (approved):

visidata/sheets.py
midichef: You are right, the complexity here is unnecessary. I'll submit a cleanup PR.

PR #2814, opened 2025-07-22
[sheets-] assign col width on sheets with no rows

### LINKED ISSUE #2795: resize-col-half throws error under certain conditions
Just an odd bug I noticed (tested with dev version):

On a new sheet, `resize-col-half` throws this error:
TypeError: unsupported operand type(s) for //: 'NoneType' and 'int'

Notes:
- This only happens for the second column onwards (col 1+ using zero index)
- A column resized using `resize-col-input` will stop throwing this error
- Adding a row to the sheet stops the error completely for all (existing) columns

midichef:
Closes #2795.

### DIFF
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -706,12 +706,15 @@ class TableSheet(BaseSheet):
         self.rightVisibleColIndex = vcolidx
 
     def calcSingleColLayout(self, col:Column, vcolidx:int, x:int=0, minColWidth:int=4):
-            if col.width is None and len(self.visibleRows) > 0:
+            if col.width is None:
+              if len(self.visibleRows) > 0:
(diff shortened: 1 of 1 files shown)


anjakefala:
Hey @midichef!

Thanks for taking a stab at this. I actually am planning on pushing a fix on this. :sweat_smile: Sorry about that. Tried to communicate I was working on it by assigning myself to the issue.

The problem with this approach is if the first rows in an async loader take awhile, all the columns will be default width. So the approach we need to take is for `None` to be interpreted as `default_width` within the commands itself. I'm going to push up this commit in a bit!

midichef:
Ah I see. The proper behavior is more subtle than I realized.

And I'll note that meaning of assignment in the future. I didn't realize work was underway. I had attempted this fix a while ago, and never got around to submitting it, your recent comment on repro prompted me to submit it to try to avoid us duplicating effort.

PR #2817, opened 2025-07-24
[test-] allow parallel tests in test.sh with -j

midichef:
This PR speeds up the tests in `dev/test.sh` by running them in parallel.

People testing on their own machines would have to opt-in to parallelization by using the -j flag: `dev/test.sh -j 4`. On my local dev environment, using 4-5 jobs speeds up `dev/test.sh` by a factor of about 3, going from 1.5 mins to 0.5 mins.

But I'm not sure how many jobs to use on Github. I ran several tests in my fork of the visidata repo. My tests showed a clear 30% improvement to the point when test.sh ran 4-5 jobs at once, going from 95 seconds to about 65 seconds. Further increases gave unclear results.

Github's action runners for open source repos have ["4-vCPUs, 16GiB of memory" since Jan 2024](https://github.blog/news-insights/product-news/github-hosted-runners-double-the-power-for-open-source/). So I'll just try pushing commits here that tests different numbers of jobs in this PR, and see what looks best.

Parallelizing the tests is a bit of extra complexity, but in my opinion, the convenience is worth the speedup. It should be safe on Github, as long as no data files or config files are edited at the same time. The visidata cache, or macros.jsonl or input_history.jsonl could end up in an odd state. That's why local parallelization would have to be opt-in.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -57,8 +57,8 @@ jobs:
       run: |
         pytest -sv visidata/tests/
 
-    - name: Run Cmdlog Tests
-      run: dev/test.sh
+    - name: Run Cmdlog Tests in parallel
+      run: dev/test.sh -j 4
       shell: bash
 
     - name: Test macros
--- a/dev/test.sh
+++ b/dev/test.sh
@@ -1,18 +1,30 @@
 #!/usr/bin/env bash
 
-# Usage: test.sh [testname]
+# Usage: test.sh [-j n_jobs] [testname]
 
 #set -e
 shopt -s failglob
 
 trap "echo aborted; exit;" SIGINT SIGTERM
 
+MAX_PARALLEL_JOBS=1
+while getopts "j:" opt; do
+    case "$opt" in
+        j)
+            MAX_PARALLEL_JOBS="$OPTARG"
+            ;;
+    esac;
+done
+shift $((OPTIND - 1))
+echo "Using MAX_PARALLEL_JOBS=$MAX_PARALLEL_JOBS"
+
 run_silent_unless_error() {
   output=$(PYTHONPATH="$PYTHONPATH" "$@" 2>&1)  # Captures ALL stdout and stderr
   exit_code=$?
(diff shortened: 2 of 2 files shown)


midichef:
Okay, I tested n=2 to n=7 here and the best seems to be using 4 jobs. The results here are similar to what I got on my fork of the visidata repo.

On this public repo, the runtime of `dev/test.sh` is roughly 54 seconds as a baseline with no parallelization. With 2 jobs it's about 29s, with 3 jobs it's about 26.6 secs (averaged over 2 runs), with 4 jobs it's about 24.6 seconds, averaged over 3 runs:  (24.33, 24.5, 25.0).

That's not a lot of statistical power, but it aligns with my own repo tests showing n=4 was slightly better than n=3. So the final commit I've pushed for this PR uses n=4, and should result in the tests finishing 30 seconds faster, for a total time savings of about 30%.

midichef:
There are also a couple of optimizations I tried and rejected:
1. Turning on pip caching by adding "cache: 'pip'" to the `setup-python` action in `main.yml`, `vdsql.yml`, and `vgit.yml`. This cut the pip install times by only a couple of seconds (very roughly 10%). I don't find that worth maintainers having to think about any cache problems, and having to perhaps manually flush the cache.
2. Updating the Github actions, [checkout](https://github.com/actions/checkout) from v2 to v4; and [setup-python](https://github.com/actions/setup-python) from the <s>deprecated</s> v2 to v5. (EDIT: I now think v2 is not deprecated. My vague understanding is that v2 can produce warnings when it uses deprecated features, but is not itself deprecated.)
Updating these actions may be a good idea in general, but not for speed, as updating didn't make them faster (or slower).

saulpw (approved):

anjakefala:
I did a basic test to see if the test outputs will interleave in confusing ways. It seems to be legible for now. If we have trouble parsing logs, we can revert, but let's take this for now!

(Logs being confusing to interpret bit me when I ran tests with parallelisation at my last job.)

PR #2818, opened 2025-07-25
[reload] run reload_rows one at a time

### LINKED ISSUE #2808: VisiData doesn't scroll horizontally
This bug has been happening since I upgraded to VisiData 3.2. I haven't been able to reproduce it but I thought it might be useful to share it anyway.

Sometimes I use the right arrow key to scroll through the columns. When I go to the rightmost visible column and keep going right, VisiData probably does select the column to the right of it, but it doesn't scroll horizonally so I can't see what's going on there.

This might have something to do with `reload-modified` since it usually happens after running that command.

**Configuration**

- Does this issue reproduce without any plugins or configuration (using the `-N` CLI flag)? **Can't reproduce so I don't know.** 

- Does this issue reproduce with either the [latest release](https://www.visidata.org/releases/), or with the [develop branch](https://www.visidata.org/install/#update-visidata-from-an-existing-installation)? **Can't reproduce so I don't know.** 

**Additional context**

- What platform and version are you using (Linux, MacOS, Windows)? **Linux**

- Which version of Python? **3.13**

- Which terminal are you using (for display and input issues)? **Windows Terminal**

midichef:
Fixes #2808.

Existing problem in `reload_rows`: inside it, `ScopedSetattr` sets `checkCursor` to a no-op function. If another `reload_rows()` starts then, before the first one has finished, it will eventually use that no-op as the final version of `checkCursor`, when the cleanup of the `with` statement restores `checkCursor`. That destroys `checkCursor`, which disables scrolling the sheet outside the current view.

This PR makes it if `reload_rows` is running, a second call to `reload_rows()` will wait for the first one to finish.

### DIFF
--- a/visidata/threads.py
+++ b/visidata/threads.py
@@ -270,6 +270,27 @@ def asyncsingle(func):
     _execAsync.searchThread = None
     return _execAsync
 
+def asyncsingle_queue(func):
+    '''Function decorator like `@asyncthread` but as a singleton.  When called, `func(...)` spawns a new thread, and waits for the end of any previous thread still running *func*.
+    ``vd.sync()`` does wait for unfinished asyncsingle_queue threads, which is an important difference from asyncsingle.
+    '''
+    @functools.wraps(func)
+    def _execAsync(*args, **kwargs):
+        def _func(*args, **kwargs):
+            func(*args, **kwargs)
+            _execAsync.searchThread = None
+            # end of thread
+
+        # cancel previous thread if running
+        if _execAsync.searchThread:
+            vd.sync(_execAsync.searchThread)
+
+        _func.__name__ = func.__name__ # otherwise, the the thread's name is '_func'
+
+        _execAsync.searchThread = vd.execAsync(_func, *args, **kwargs)
+    _execAsync.searchThread = None
+    return _execAsync
+
 @VisiData.property
 def unfinishedThreads(self):
     'A list of unfinished threads (those without a recorded `endTime`).'
@@ -465,6 +486,7 @@ vd.addGlobals({
     'Progress': Progress,
     'asynccache': asynccache,
     'asyncsingle': asyncsingle,
+    'asyncsingle_queue': asyncsingle_queue,
     'asyncignore': asyncignore,
 })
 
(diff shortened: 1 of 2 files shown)


midichef:
Here is some demo code to test the new @asyncsingle_queue decorator,  in .visidatarc:
```
## for testing reload_rows() #2818
# run:  vd sample_data/benchmark.csv 
# then run `reload-rows-several`
vd.load_counter = 0
@Sheet.api
def beforeLoad(sheet):
    if sheet.name == 'benchmark':
        vd.status(f'sheet.rows id: {id(sheet.rows)%100_000:05d}, beforeLoad()')
@Sheet.api
def afterLoad(sheet):
    vd.load_counter += 1
    if sheet.name == 'benchmark' and vd.load_counter > 1:
        sleep(1)
    if sheet.name == 'benchmark':
        vd.status(f'sheet.rows id: {id(sheet.rows)%100_000:05d}, afterLoad() ended')

Sheet.addCommand(
    '',
    'reload-rows-several',
    'reload_rows(); reload_rows(); reload_rows(); reload_rows(); ',
    'reload sheet 4x to trigger bug'
)
```
Note that without the new decorator, running `reload-rows-several` takes only about a second, and that the sheet.rows object has the same ID when all the `afterLoad()` calls finish around the same time. But with the new decorator, it takes 4x1 seconds, and the sheet.rows have a different ID at the end of each call to `reload_rows()`.

midichef:
The test failure is unrelated to the PR. It's due to a timeout when fetching [usage.tsv via plain http](http://visidata.org/usage.tsv). I've noticed similar failures for this URL a few times this week, so maybe it's worth looking into if anything unusual is happening with the visidata.org server.

Otherwise, I can submit a PR changing this test to use a local file instead of a remote one.

visidata/threads.py
```diff
@@ -465,6 +486,7 @@ def allThreadsSheet(self):
     'Progress': Progress,
     'asynccache': asynccache,
     'asyncsingle': asyncsingle,
```
saulpw, line 488: I like your new name.  We should think about renaming this to `asyncsingle_cancel`.

saulpw (approved):

visidata/threads.py
midichef: `asyncsingle_cancel` makes sense to me.

PR #2819, opened 2025-07-26
[macro] fix macro corruption when saving via putChanges

### LINKED ISSUE #2787: [macros] Deleting macros causes corruption to the macro.jsonl file
https://github.com/saulpw/visidata/issues/2784
**Small description**
Deleting macros causes corruption to the macro.jsonl file.  When trying to delete a macro by removing rows from the macro sheet, it seems to corrupt the macros.jsonl file, leading to potential data integrity issues and failure to load or use macros properly.

**Data to reproduce**

**Steps to reproduce**

Create macros, exit, try to delete the rows in the macro sheet. The macros.jsonl file is not updated properly.

**Expected result**

[If there is an error message, please include the full stack trace shown with `Ctrl+E` or `vd --debug`.]

**Configuration**

**Additional context**

- Which version of Python?

midichef:
Closes #2787.

The problem is that the sheet was being reloaded, before saving the macros was done:
https://github.com/saulpw/visidata/blob/d5c082f0c618b12db4b5ab3c0ff8fb41409b1e6d/visidata/macros.py#L43-L45
Both `saveSheets()` and `reload()` have the `@asyncthread` decorator, so there is a race condition that causes corruption, but only sometimes, depending on which function finishes first.

The corruption appeared 60% of the time on my system using the repro in #2787. With the fix, it happens 0 times in 100 automated runs, so this is a high-confidence fix.

### DIFF
--- a/visidata/macros.py
+++ b/visidata/macros.py
@@ -40,7 +40,7 @@ class MacroSheet(IndexSheet):
     def putChanges(self):
         self.commitDeletes()  #1569  apply deletes early for saveSheets below
 
-        vd.saveSheets(self.source, self, confirm_overwrite=False)
+        vd.sync(vd.saveSheets(self.source, self, confirm_overwrite=False))
         self._deferredDels.clear()
         self.reload()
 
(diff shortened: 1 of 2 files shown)


saulpw (approved):
Thanks for fixing this!

frosencrantz:
Thank you for fixing this @midichef!  Also, thank you, @anjakefala for all the work to get this and the other recent changes merged safely into the repo!

Issue #2670, opened 2025-01-08
[fuzzymatch] fuzzymatch shows matched items in lowercase


midichef:
**Small description**
Fuzzymatch shows matches in lowercase.

**Steps to reproduce**
`vd sample_data/benchmark.csv`
`Space` `go-col-name` `date`
As soon as the first letter (`d`) of the search pattern is typed:  the match for `Date` is shown in lowercase: `date`.

**Expected result**
I expect the match to preserve its case:  `Date`.

**Configuration**
vd v3.2dev

**Additional context**
This was introduced by me in #2658.

PR #2822, opened 2025-07-28
[palette-] scroll matches with PageUp/PageDown

midichef:
This PR has two changes to improve the input palette.

The first change is to display only the items that match user input, and no longer show unmatched items. For example, when typing `Space` then `freeze`, there's a clear visual change from 10 results to 3, when the `z` is typed. As part of this change, note the deliberate design choice to display blank lines when there are no results, as after `Space` `zz`.

The second change allows scrolling through the palette using `PgUp` and `PgDn`, and adds on-screen instructions.

For reviewing this PR, there are 2 lines I moved that may need explanation: `unfinished_words = words` and `bindings[' '] = partial(add_to_input, value=topitem[value_key])`. The reason I moved them was to align the order of lines in the block with other similar blocks nearby.

### DIFF
--- a/visidata/features/cmdpalette.py
+++ b/visidata/features/cmdpalette.py
@@ -184,7 +184,7 @@ def inputPalette(sheet, prompt, items,
 
             clipdraw(sheet._scr, h-nitems-2+i, 0, match_summary, attr, w=w)
         attr = colors.color_cmdpalette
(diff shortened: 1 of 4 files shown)


midichef:
I pushed an extra commit that closes #2670. It fixes the lowercasing of palette results, which is not strictly part of this PR. But putting it here instead of a separate PR avoids a merge conflict later.

saulpw (approved):
Nice, this is definitely seems like it will be a quality of life improvement.

PR #2823, opened 2025-07-28
[input-] handle resizes during input or cell edit

midichef:
In the middle of `edit-cell` or `exec-longname`, if you resize the terminal pushing the current cell offscreen, you get a curses error:
```
    File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/_input.py", line 396, in editText
        v = widget.editline(vd.activeSheet._scr, y, x, w, attr=attr, updater=updater, bindings=bindings, clear=clear)
...
    File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/_input.py", line 581, in _drawPrompt
        clipdraw(sheet._scr, y, 0, prompt, attr, w=sheet.windowWidth-rstatuslen-1)
...
    File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/cliptext.py", line 242, in clipdraw_chunks
        scr.addstr(y, x, disp_column_fill*actualw, cattr.attr)  # clear whole area before displaying
```
This PR catches the error gracefully.

It was hard to connect this error to the resize event, because the stack trace would go to the terminal, and only be noticed much later on exiting visidata.

### DIFF
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -684,7 +684,10 @@ def editCell(self, vcolidx=None, rowidx=None, value=None, **kwargs):
     editargs = dict(value=value, options=self.options)
 
     editargs.update(kwargs)  # update with user-specified args
-    r = vd.editText(y, x, w, attr=colors.color_edit_cell, **editargs)
+    try:
+        r = vd.editText(y, x, w, attr=colors.color_edit_cell, **editargs)
+    except curses.error:
+        vd.fail(f'aborting edit due to resize')
 
     if rowidx >= 0:  # if not header
         r = col.type(r)  # convert input to column type, let exceptions be raised
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -245,7 +245,8 @@ class InputWidget:
             clipdraw(scr, y, x+w, ' ', attr, 1, clear=False, literal=True)
         if scr:
             prew = dispwidth(dispval[:i])
-            scr.move(y, x+prew)
+            if x+prew < scr.getmaxyx()[1]: #move cursor back to where the user is editing
+                scr.move(y, x+prew)
 
     def handle_key(self, ch:str, scr) -> bool:
         'Return True to accept current input.  Raise EscapeException on Ctrl+C, Ctrl+Q, or ESC.'
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -586,7 +586,7 @@ def input(vd, prompt, type=None, defaultLast=False, history=[], dy=0, attr=None,
 
     w = kwargs.pop('w', _drawPrompt())
     restarts = 0
-    while True:
+    while restarts < 100:
         #recalculate y to handle resize events
         y = sheet.windowHeight-dy-1
         try:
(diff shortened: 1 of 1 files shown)


midichef:
The test here did not really fail, it's the timeout trying to fetch http://visidata.org/usage.tsv that keeps happening, this time it had this error:
`cannot open URL: [SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol (_ssl.c:1010)`

midichef:
For testing, the most consistent way I have found to trigger the curses error (for gnome-terminal on Ubuntu) is: press `Space` to run `exec-longname` and then increase the font size with `Ctrl+Shift+=` aka `Ctrl +`

visidata/_input.py
```diff
@@ -583,21 +584,32 @@ def _drawPrompt(val=''):
...
+                ret = history[-1]
 
-    return ret
+            return ret
+        except curses.error:
+            vd.warning('restarting input due to resize')
+            restarts += 1
+            # if it keeps happening, it's probably not resize events, so give some debug output
+            if restarts >= 100:
+                vd.fail(f'aborting input:  y={y}, w={w}, windowHeight={sheet.windowHeight}, windowWidth={sheet.windowWidth}')
```
saulpw: Use `vd.error` for internal error conditions like this.  (`fail` is for "user error", i.e. things the user can fix)

saulpw: did you experience this for some reason in your testing?

midichef: No, I only put it in as a general alert. But I since have discovered it does happen when the screen shrinks while `inputMultiple()` is going. I'll look into it.

midichef: Okay, I've switched to `vd.error`, and handled the `inputMultiple()` case as well

visidata/_input.py
```diff
@@ -583,21 +585,32 @@ def _drawPrompt(val=''):
...
-                        **kwargs)
-
-    if ret:
-        if kwargs.get('record', True) and kwargs.get('display', True):
-            vd.addInputHistory(ret, type=type)
-    elif defaultLast:
-        history or vd.fail("no previous input")
-        ret = history[-1]
+    restarts = 0
+    while True:
```
saulpw: Can we structure this as `while restarts < 100`, and then move the `error` outside the while loop?

visidata/_input.py
```diff
@@ -583,21 +585,32 @@ def _drawPrompt(val=''):
...
 
-    return ret
+            return ret
+        except curses.error:
+            vd.warning('restarting input due to resize')
+            restarts += 1
+            # if it keeps happening, it's probably not resize events, so give some debug output
+            if restarts >= 100:
+                vd.error(f'aborting input:  y={y}, w={w}, windowHeight={sheet.windowHeight}, windowWidth={sheet.windowWidth}')
+            continue
```
saulpw: This `continue` at the bottom of the loop is unnecessary.

visidata/_input.py
midichef: That's more concise, okay.

saulpw (approved):
Another wart removed, thanks @midichef!

PR #2830, opened 2025-08-02
speed up display by caching formatter_helpstr

midichef:
`formatter_helpstr` is slow, and runs once per command, including all movements like `go-down`. On my system it adds an average delay of about 7 ms. Every so often it adds 20-30 ms.

The code that calls this is `drawRightStatus()`. The use of `formatter_helpstr` was already being cached, but the cache is cleared between commands. This PR caches the result permanently, once per sheet.

The 7 ms delay is in the creation of `HelpSheet`. I don't know why the delays are irregular, spiking up to 20-30ms about 2% of the time. I suspect it's due to the `vd.sync()` call in `vd.sync(self.helpsheet.ensureLoaded())`.

In my testing, the net effect of this PR is about a 40% speedup for `draw_all()`.

### DIFF
--- a/visidata/sidebar.py
+++ b/visidata/sidebar.py
@@ -33,7 +33,7 @@ class AddedHelp:
             vd.clearCaches()
 
 
-@BaseSheet.property
+@BaseSheet.lazy_property
 def formatter_helpstr(sheet):
     return AttrDict(commands=CommandHelpGetter(type(sheet)),
                     options=OptionHelpGetter())


saulpw (approved):
Wow, thanks for drilling down into this!  A 40% speedup is nothing to sneeze at, for one token change.  How long did this take you to figure out?

midichef:
I overstated the speedup, it's probably not 40%, it's more like 25% (on `sample_data/benchmark.csv`). And I should have said it is a speedup to the overall `draw_all() + drawRightStatus()` pair (20ms + 7ms). Not to `draw_all`, which is unaffected by this change since it doesn't call `drawRightStatus()`.

It took about 2 hours. I started out trying to understand why `addUndo()` is called ~20 times for any keypress, even a prefix like `z`, and found that it was due to creation of a `HelpSheet` inside `formatter_helpstr`.

Issue #2831, opened 2025-08-03
Auto-detect CSV format when `--csv-delimiter` is specified


dennisangemi:
## Problem Description

Currently, when using `--csv-delimiter` with a non-default delimiter (e.g., semicolon), VisiData doesn't automatically recognize the file as CSV format. This results in:

1. **"unknown filetype" message** for files without `.csv` extension
2. **File opened as plain text** instead of CSV
3. **User must manually specify `-f csv`** in addition to `--csv-delimiter`

## Current Behavior

```bash
# This shows "unknown filetype" and opens as text
echo "Name;Age;City
John;25;New York
Alice;30;London" | vd --csv-delimiter ";" -
```

<img width="433" height="269" alt="Image" src="https://github.com/user-attachments/assets/370eaf38-505a-4e33-839b-0b30b47e4a09" />

```
# User must specify both flags
echo "Name;Age;City
John;25;New York  
Alice;30;London" | vd --csv-delimiter ";" -f csv -
```

<img width="432" height="270" alt="Image" src="https://github.com/user-attachments/assets/532eece4-6552-4a0a-b694-86b4947bca2d" />

## Expected Behavior
When `--csv-delimiter` is specified with a non-default value, VisiData should automatically assume CSV format without requiring the `-f csv` flag.

## Use Case
This is particularly useful for:

- CSV files using semicolons as delimiters
- Data files without .csv extension but with CSV structure
- Simplified command-line usage when working with non-comma CSV data

## Environment
VisiData version: 3.2+
Files affected: Any text file with CSV structure using non-comma delimiters

## Pull request
I'm opening a pull request to implement this feature

saulpw:
This looks to be fixed thanks to your PR, thanks @dennisangemi!

PR #2832, opened 2025-08-03
feat: auto-detect CSV format when csv_delimiter is specified

dennisangemi:
Related issue: #2831

## Summary

Automatically detect CSV file format when user specifies `--csv-delimiter` with a non-default value, eliminating the need to also specify `-f csv`.

## Problem Solved

Previously, users had to specify both `--csv-delimiter ";"` AND `-f csv` to properly load CSV files with non-comma delimiters. Without the `-f csv` flag, VisiData would show "unknown filetype" and open the file as plain text.

## Implementation

Added `guess_csv_delimiter()` function to `visidata/loaders/csv.py` that:

1. **Checks if `csv_delimiter` option differs from default** (",")
2. **Returns CSV filetype with likelihood=2** when delimiter is modified
3. **Integrates seamlessly** with VisiData's existing auto-detection system

## Code Changes

- **File modified:** `visidata/loaders/csv.py`
- **Lines added:** 10 (surgical precision implementation)
- **No breaking changes**

### Before (shows "unknown filetype"):
```sh
echo "Name;Age;City
John;25;New York
Alice;30;London" | vd --csv-delimiter ";" -
# Output: unknown "" filetype, opening as txt
```

After (correctly detects CSV):

```sh
echo "Name;Age;City
John;25;New York
Alice;30;London" | python3 -m visidata --csv-delimiter ";" -
# Output: opening as csv
```

## Implementation Details
The solution leverages VisiData's existing `guessFiletype()` system by adding a new guess function that runs when the CSV loader is imported. It has a likelihood score of 2, which is higher than content-based guessing (0) but lower than extension-based detection (3).

```python
@VisiData.api
def guess_csv_delimiter(vd, p):
    'If csv_delimiter option has been modified from default, assume CSV format.'
    default_delimiter = ','
    current_delimiter = vd.options.csv_delimiter
    
    if current_delimiter != default_delimiter:
        return dict(filetype='csv', _likelihood=2)
```

### DIFF
--- a/visidata/loaders/csv.py
+++ b/visidata/loaders/csv.py
@@ -15,11 +15,8 @@ vd.option('safety_first', False, 'sanitize input/output to handle edge cases, wi
 @VisiData.api
 def guess_csv_delimiter(vd, p):
     'If csv_delimiter option has been modified from default, assume CSV format.'
-    default_delimiter = ','
-    current_delimiter = vd.options.csv_delimiter
     
-    # If user has explicitly set a different delimiter, assume CSV format
-    if current_delimiter != default_delimiter:
+    if vd.options.csv_delimiter != vd.options.getdefault('csv_delimiter'):
         return dict(filetype='csv', _likelihood=2)
 
 @VisiData.api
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Good idea, @dennisangemi.  I made a change so that it checks the actual default instead of hard-coding it.

dennisangemi:
Wow! Thank u so much @saulpw

PR #2835, opened 2025-08-06
[sheets-] on reload, fetch fresh data for URLs

### LINKED ISSUE #2825: Data loaded from http is not refreshed  by Ctrl-R
**Small description**
Data loaded from http is not refreshed  by Ctrl-R

**Data to reproduce**

https://api.github.com/events

**Steps to reproduce**

1. Start `vd https://api.github.com/events`
2. Press Ctrl-R
3. Content of the sheet has not changed, though  response of https://api.github.com/events changes all the time

**Expected result**
Data from http links should be refreshed just like the data from local files.

**Actual result**
Content of the sheet has not changed, though  response of https://api.github.com/events changes all the time.

**Configuration**

Yes.

Observed with visidata 3.0 and confirmed in 3.2 from current arch linux repos.

**Additional context**

Linux. Though should be irrelevant

- Which version of Python?
3.13.5-1

midichef:
Closes #2825.

The problem was that for URLs, the initial content of the URL is saved in in `sheet.source` (a `visidata.Path` object) inside a `RepeatFile`. Reloading the sheet merely reread the saved contents.

### DIFF
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -1208,6 +1208,21 @@ def async_deepcopy(sheet, rowlist):
     _async_deepcopy(ret, rowlist)
     return ret
 
+@Sheet.api
+def reload_or_replace(sheet):
+    sheet.preloadHook()
+    if isinstance(sheet.source, visidata.Path) and \
+       sheet.source.is_url() and sheet.source.scheme != 'file':  #2825
+        #retrieve data again, because the earlier data saved in sheet.source may be outdated
+        vs = vd.openSource(visidata.Path(sheet.source.given))
+        if type(vs) != type(sheet):  #new data may have a different filetype
+            vd.push(vs)
+            vd.remove(sheet)
+            #user needs feedback that sheet changed, since the new sheet has a different shortcut
+            vd.status('replaced sheet due to changed filetype')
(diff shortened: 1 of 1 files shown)


visidata/sheets.py
```diff
@@ -1208,6 +1208,21 @@ def _async_deepcopy(newlist, oldlist):
...
     return ret
 
+@Sheet.api
+def reload_or_replace(sheet):
+    sheet.preloadHook()
+    if isinstance(sheet.source, visidata.Path) and \
+       sheet.source.is_url() and sheet.source.scheme != 'file':  #2825
+        #retrieve data again, because the earlier data saved in sheet.source may be outdated
+        vs = vd.openSource(Path(sheet.source.given))
+        if type(vs) != type(sheet):  #new data may have a different filetype
```
saulpw, line 1218: I think there's a more Pythonic way to do this.  `is` or `isinstance`?  But I'm not sure, and I don't think it's a big deal.

saulpw (approved):

midichef:
@anjakefala Thanks for catching my error writing `Path` instead of `visidata.Path`. I'm not sure how I let that through!

PR #2641, opened 2024-12-17
[main-] enable cell editing for interactive batch mode

midichef:
Closes #2639 and #2640. And reverts d9fa61bd6f6e405da23df192dda9eb25a189add5.

PR #2840, opened 2025-08-12
fix editText for files on command line with -b -i -p

### LINKED ISSUE #2639: edit-cell fails in interactive batch mode
**Small description**
With vd in interactive batch mode, `edit-cell` will set each cell to `''`

**Steps to reproduce**
`vd -b -i -p cmdlog.vdj` then press `e`

**Expected result**
The usual edit-cell mode.

**Configuration**
vd v3.2dev

**Additional context**
The relevant line is in `editText()`:
https://github.com/saulpw/visidata/blob/ff489d3717a817a0a04a9cc0b48d5c19ea5e540c/visidata/_input.py#L355-L356

midichef:
This is a continuation of #2641, which fixes #2639. The previous fix failed when a file was present on the command line. That is, it fixed `vd -b -i -p cmd.vdj`, but not `vd -b -i -p cmd.vdj data.tsv`.

demo: `vd -b -i -p tests/fill.vdj sample_data/a.tsv`, then switch to the sheet `a` and hit `e` on any cell and see that it goes blank.

In exploring this fix, I was caught out by the behavior of `vd.options`. Due to the sheet-specific overrides, right after assigning to `vd.options.batch`:
https://github.com/saulpw/visidata/blob/6ac1bd3ba9dcc441f20fa081016a731d67063dfb/visidata/main.py#L362
it was still the case that `vd.options.batch == True`! I'll keep that possibility in mind in the future.

Also, just to double-check: in this context of `main.py`, is there any semantic difference in using `vd.options` vs `options`? There they have the same Python id as given by `id()`, so I know they're the same object.

### DIFF
--- a/dev/checklists/manual-tests.md
+++ b/dev/checklists/manual-tests.md
@@ -96,3 +96,5 @@
 24. Test adding multiple aggregators via palette (+)
 25. time vd -p tests/quit-nosave.vdj  - note down the time. compare to PR #2369
 26. Use the z; command. Then type in a command line like echo "| Ceci n'est pas une pipe"
+27. vd -b -i -p tests/fill.vdj sample_data/a.tsv 
+Check that both benchmark and a.tsv are edittable.
--- a/visidata/main.py
+++ b/visidata/main.py
@@ -255,7 +255,11 @@ def main_vd():
                         optval = sys.argv[i+1]
                         i += 1
 
-            current_args[optname] = optval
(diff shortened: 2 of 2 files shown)


saulpw:
> Also, just to double-check: in this context of main.py, is there any semantic difference in using vd.options vs options? 

They are the same.  `options` is a deprecated global that's still used in some places internally, but should be phased out in favor of `vd.options`.

saulpw (approved):

Issue #1529, opened 2022-09-16
Exception with `entry_points().get('visidata.plugins', [])` upon startup


anjakefala:
### Discussed in https://github.com/saulpw/visidata/discussions/1528

<div type='discussions-op-text'>

<sup>Originally posted by **jsdealy** September 15, 2022</sup>
I may be doing something stupid, but my attempts to install have continually just produced this: 

```
Traceback (most recent call last):
  File "/home/justin/.local/bin/visidata", line 8, in <module>
    sys.exit(vd_cli())
  File "/home/justin/.local/lib/python3.10/site-packages/visidata/main.py", line 350, in vd_cli
    rc = main_vd()
  File "/home/justin/.local/lib/python3.10/site-packages/visidata/main.py", line 209, in main_vd
    vd.loadConfigAndPlugins(args)
  File "/home/justin/.local/lib/python3.10/site-packages/visidata/settings.py", line 378, in loadConfigAndPlugins
    for ep in entry_points().get('visidata.plugins', []):
  File "/usr/lib/python3.10/site-packages/importlib_metadata/__init__.py", line 1015, in entry_points
    return SelectableGroups.load(eps).select(**params)
  File "/usr/lib/python3.10/site-packages/importlib_metadata/__init__.py", line 453, in load
    ordered = sorted(eps, key=by_group)
  File "/usr/lib/python3.10/site-packages/importlib_metadata/__init__.py", line 1012, in <genexpr>
    eps = itertools.chain.from_iterable(
  File "/usr/lib/python3.10/site-packages/importlib_metadata/_itertools.py", line 16, in unique_everseen
    k = key(element)
  File "/usr/lib/python3.10/site-packages/importlib_metadata/__init__.py", line 947, in _normalized_name
    return self._name_from_stem(stem) or super()._normalized_name
  File "/usr/lib/python3.10/site-packages/importlib_metadata/__init__.py", line 616, in _normalized_name
    return Prepared.normalize(self.name)
  File "/usr/lib/python3.10/site-packages/importlib_metadata/__init__.py", line 871, in normalize
    return re.sub(r"[-_.]+", "-", name).lower().replace('-', '_')
  File "/usr/lib/python3.10/re.py", line 209, in sub
    return _compile(pattern, flags).sub(repl, string, count)
TypeError: expected string or bytes-like object
```
I'm running Arch Linux. 

Assistance would be greatly appreciated, thank you! </div>

anjakefala:
[...]

PR #2843, opened 2025-08-13
fix plugin autoload for Python versions < 3.10

### LINKED ISSUE #2841: Receiving plugin autoload failed message
I recently upgraded from 3.1.1 to 3.2 and I've started receiving the following message when I load files:

> plugin autoload failed; see issue #1529

Outside of the new message, the file(s) appear to load normally.

I've reviewed the issue, but I'm not able to find a solution.

I've ran the following, but there is no output.

> import importlib_metadata as md
> dists = md.distributions()
> broken = [dist for dist in dists if dist.name is None]
> for dist in broken:
> ... print(dist._path)
> ... print('metadata length', len(dist.metadata()))

I've tried opening files with the -N option and this removes the message.

In case it helps, I've added a line to my settings.py file to print the exception to see if there is any additional information.

>     # autoload installed plugins first
>     args_plugins_autoload = args.plugins_autoload if 'plugins_autoload' in args else True
>     if not args.nothing and args_plugins_autoload and vd.options.plugins_autoload:
>         from importlib_metadata import entry_points 
>         try:
>             eps = entry_points()
>             eps_visidata = eps.select(group='visidata.plugins') if 'visidata.plugins' in eps.groups else []
>         except Exception as e:
>             eps_visidata = []
>             vd.warning(e) # ---------New Line
>             vd.warning('plugin autoload failed; see issue #1529')

With this, I get the following message

> 'dict' object has no attribute 'groups'

I've tried looking at the entry_points to see if there is an issue with the groups using the following, but there doesn't seem to be anything missing.

> from importlib.metadata import entry_points
> from pprint import pprint
> pprint(entry_points())

I'm using python 3.9.18 on Rocky Linux 9.4.

I show 8.7.0 for importlib_metadata using both

> python -m pip freeze

and

> python -c "import importlib_metadata; print(importlib_metadata.version('importlib-metadata'))"

I might be missing something, but I'm not sure what else to check.  Any help would be appreciated.

midichef:
Closes #2841. It broke for Python 3.8 and 3.9 in visidata v3.2 with 7e75dee682b81e7e9206a9b8d0ab4e695cdff8bd, giving the warning `plugin autoload failed; see issue #1529`.

I tested the new code paths on Python v3.8 through v3.13. For each version I tested it two ways:  1) without any autoloading plugins installed, and 2) with 1 autoloading plugin, using `darkdraw`. For my test, I checked that startup finished without any warning, and that the `eps_visidata` list had no entries if `darkdraw` wasn't installed, and 1 entry if it was.

The specific Python versions I tested are: 3.8.10/3.8.20, 3.9.10/3.9.19, 3.10.17, 3.11.12, 3.12.10, 3,13.3.

### DIFF
--- a/visidata/settings.py
+++ b/visidata/settings.py
@@ -447,12 +447,16 @@ def loadConfigAndPlugins(vd, args=AttrDict()):
     args_plugins_autoload = args.plugins_autoload if 'plugins_autoload' in args else True
     if not args.nothing and args_plugins_autoload and vd.options.plugins_autoload:
         from importlib.metadata import entry_points
+        eps_visidata = []
         try:
             eps = entry_points()
-            eps_visidata = eps.select(group='visidata.plugins') if 'visidata.plugins' in eps.groups else []
+            vp = 'visidata.plugins'
+            if hasattr(eps, 'groups'): #Python >= 3.10
+                eps_visidata = eps.select(group=vp)
+            else:                      #Python <  3.10
+                eps_visidata = eps.get(vp, [])
         except Exception as e:
-            eps_visidata = []
-            vd.warning('plugin autoload failed; see issue #1529')
+            vd.warning(f'plugin autoload failed; see issue #1529:  {e}')
 
         for ep in eps_visidata:
             try:


saulpw (approved):

anjakefala:
Oh strange. I thought we only needed that for Python 3.7 and less. Thanks for catching that behaviour!

PR #2851, opened 2025-08-31
Bugfix to fix Visidata crash on startup on NetBSD (and probably other *BSD) wscons tty consoles, which do not expose "mousemask" in their system curses implementations.

rsmirnov90:
Signed-off-by: Russ <russ@sentibyte.com>

### DIFF
--- a/visidata/mouse.py
+++ b/visidata/mouse.py
@@ -13,6 +13,8 @@ BaseSheet.init('mouseY', int)
 
 @VisiData.after
 def initCurses(vd):
+    if not getattr(curses, 'mousemask', None):
+      return
     curses.MOUSE_ALL = 0xffffffff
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Thanks @rsmirnov90!  Weird that BSDs don't at least mock that function out.

rsmirnov90:
> Thanks @rsmirnov90! Weird that BSDs don't at least mock that function out.

Yeah I was really surprised too! There's probably a "proper" way to fix this by making it an application-wide flag to disable the mouse altogether, but unfortunately my knowledge of Python falls short to attempt such an undertaking. 

However, even with just that fix, Visidata has been running perfectly on NetBSD wscons ever since. So, whatever gets us closer to more people using this amazing project! Which is really a masterpiece, you have done a fantastic job on it. It's become an irreplaceable tool for me (I don't think I've opened Excel since discovering this 😁)

PR #2852, opened 2025-09-01
fix minor display bugs in calcColLayout

midichef:
The commit messages have more detail on each. I believe the only visible consequences of the bugs were minor variations in the # of rows reserved for aggregators. But fixing them improves the overall correctness in case there are future modifications to `calColLayout()` and `calcSingleColLayout()`.


saulpw (approved):
Sure, okay.  Thanks for the bit of featherdusting :)

PR #2853, opened 2025-09-02
[disp_help_flags] added options.disp_help_flags; deprecated options.disp_expert

### LINKED ISSUE #2771: Turn sidebar off by default
Features needed:

- [ ] Have "off-by-default" still allow the sidebar to be toggled on.
- [ ] Sidebar guides are referenced in the Help tab ("Get help for this sheet")
- [ ] The default Sheet will have a brief sidebar written, so if people click in the Help Tab, something shows up for them.

anjakefala:
- these feature flags are supported: cmdpalette hints nometacols guides inputkeys inputfield all
- see helpstr for more details
- set option to '' to turn off

Closes #2771

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # VisiData version history
 
+# v3.3 (2025-09-xx)
+
+- added options.disp_help_flags; deprecated options.disp_expert
+   - these feature flags are supported: cmdpalette hints nometacols guides inputkeys inputfield all
+   - see helpstr for more details
+   - set option to '' to turn off all help overlays
+
 # v3.2 (2025-06-15)
 
 Thanks to @midichef for many bugfixes and improvements.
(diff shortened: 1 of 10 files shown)


anjakefala:
@midichef @frosencrantz @geekscrapy Thoughts?

midichef:
This seems better. Having feature flags is more intuitive for me. I never attempted to change the `disp_help`/`disp_expert` values because I didn't know what the different numeric values did.

A minor suggestion is to take out the leading newline in the `'''\nlist of helper features` string. As it's distracting in the `Shift+O` options sheet. And to match the order of the default feature flags with the order in the bulleted descriptions, so `cmdpalette` is first in the string and first in the bulleted list.
https://github.com/saulpw/visidata/blob/73b7e798106e32542e146695fa46acf0261718fa/visidata/help.py#L8-L10
As an aside, cmdpalette is so helpful that I wonder if anyone will actually turn it off.

saulpw:
Thanks for the feedback.  I put the options in alphabetical order in both the defaults and the helpstr.

frosencrantz:
I haven't tried it, but I like the way it allows per context control. This is more understandable.  Thank you!

PR #2855, opened 2025-09-06
[basesheet-] simplify column autoname during newSheet

midichef:
This is a follow up to #2813. It simplifies my unnecessarily complex column renaming in https://github.com/saulpw/visidata/pull/2813/files/a2c40e58e76af939d5b57125a9aa9a85a06f814e#r2312807995

I also moved the column counter into `TableSheet`. I'm reasoning that `columns` are a feature of `TableSheet` but not `BaseSheet`, because I see handling `columns=` init parameters happens only in `TableSheet`:
https://github.com/saulpw/visidata/blob/95434243bddcbb09a1307a7bbd0d99c72569c4a1/visidata/sheets.py#L192


saulpw (approved):
This is great, thank you! Much cleaner and clearer.  and the locations for things make more sense too.

PR #2856, opened 2025-09-07
[graph-] accept refline input for case of multiple xcols

### LINKED ISSUE #2752: Feature request: Support refline for categorical plots
I did a plot where I have two key columns: One numerical and one string. When trying to add an x refline on 120, I got `must have 2 x values, had 1 values: 120`. I'm guessing it's confused because there are two x columns. Out of curiosity I also tried `120 greedy`, `greedy 120` and `120,greedy` and it failed on all of these.

@midichef

midichef:
Closes #2752.

To demonstrate, make a file `xcols.csv`:
```
x1,x2,y
0,2,1
0,4,2
4,1,3
5,1,5
5,1,3
```
Use `!` to make `x1` and `x2` keycols . Make all columns numeric with `#`. Then plot `y` with `.`
Add a refline with:  `gx` `0` `Enter`. On current visidata you'll get the error `must have 2 x values, had 1 values: 0`.

One commit in this PR turns off the check that gave that error.

The other commit changes the error checking when erasing x and y reflines. In current visidata, each axis was missing a check that was performed for the other one.

### DIFF
--- a/visidata/graph.py
+++ b/visidata/graph.py
@@ -319,10 +319,9 @@ class GraphSheet(InvertedCanvas):
         xstrs = vd.input("add line(s) at x = ", type="reflinex", value=suggested, defaultLast=True).split()
 
         for xstr in xstrs:
-            vals = [ v.strip() for v in xstr.split(',') ]
-            if len(vals) != len(self.xcols):
-                vd.fail(f'must have {len(self.xcols)} x values, had {len(vals)} values: {xstr}')
-            self.reflines_x += [xtype(val) for xcol, val in zip(self.xcols, vals) if xtype(val) not in self.reflines_x ]
+            refval = xtype(xstr.strip())
+            if refval not in self.reflines_x:
+                self.reflines_x.append(refval)
         self.refresh()
 
     def draw_refline_y(self):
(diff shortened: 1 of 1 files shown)


visidata/graph.py
```diff
@@ -341,11 +340,16 @@ def erase_refline_x(self):
         suggested = format_input_value(self.reflines_x[0], xtype)
 
         xstrs = vd.input('remove line(s) at x = ', value=suggested, type='reflinex', defaultLast=True).split()
-        for input_x in xstrs:
-            self.reflines_x.remove(xtype(input_x))
+        for x in xstrs:
+            try:
+                self.reflines_x.remove(xtype(x))
+            except ValueError:
+                vd.fail(f'value {x} not in reflines_x')
```
saulpw: This `fail` should be a `warning`, so that subsequent reflines will be removed anyway.

I wonder also if anyone (@cool-RR ?) uses the "erase specific refline" commands; maybe the "erase all reflines" commands are sufficient and these other ones could be deprecated?

cool-RR: I think I only used the command for removing all reflines. I wouldn't want to have to type in the refline's number. If the command was able to display all reflines and let me choose, I might. I think that visidata commands should be able to display dialogs rather than asking questions one by one... This would help for my command to export plots. But this is a tangent.

cool-RR:
I'm happy to see this PR!

visidata/graph.py
midichef: Okay, I've changed it to warnings.

PR #2857, opened 2025-09-08
[main-] fix undo of first user alteration to a replayed cmdlog

midichef:
When I replay a command log, like with `echo |vd -p -` then modify the sheet `i`, and try to undo it with `U`, I get a fail error: `nothing to undo on current sheet`.

That's because undo skips the first command in the log (after skipping any earlier options-setting commands), expecting it to be the command that created the sheet:
https://github.com/saulpw/visidata/blob/641a5eaef1e3ae747aeaec83df0cff5739dbef7b/visidata/undo.py#L44-L46

That's true for most sheets, but the replayed cmdlog sheet does not have a sheet creation command, and it breaks the expectation.

This PR fixes the undo by inserting a `no-op` as the first command.

### DIFF
--- a/visidata/main.py
+++ b/visidata/main.py
@@ -49,6 +49,8 @@ def eval_vd(logpath, *args, **kwargs):
         vs = vd.openSource(src, filetype='vdj')
     else:
         vs = vd.openSource(src, filetype=src.ext)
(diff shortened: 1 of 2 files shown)


saulpw (approved):

saulpw:
Thanks @midichef!

PR #2858, opened 2025-09-10
[palette-] when no choices match, show warning on Enter

midichef:
The input palette has a problem after hitting `Enter` or `Space` when the typed input has narrowed the matches to nothing. What gets chosen is the previous best match, shown earlier when there were matches.

For example, hit `Space` and hit `zz`, there are no matches displayed. When you hit `Enter`, what gets input is `freeze-col`, the top match for `z`.

This PR changes `Enter` to reject the command with the warning: `no choice matching zz`.
It changes `Space` in the input palette (when matching multiple inputs with `multiple=True`, like for choosing aggregators after `+`) to keep the input and to let the user keep typing.

### DIFF
--- a/visidata/features/cmdpalette.py
+++ b/visidata/features/cmdpalette.py
@@ -1,7 +1,7 @@
 import collections
 import math
 from functools import partial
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Good catch, this makes sense.

PR #2859, opened 2025-09-11
[mainloop-] for bindkey cmd, catch fail() from execCommand 

midichef:
This PR is for the situation when `bindkey` names a command that does not exist, like from this .visidatarc line:  `Sheet.bindkey('2', 'not-a-command')`
When the user presses 2, visidata immediately exits with status -1, with no other output.

That's because `fail()` in `execCommand()` raises `ExpectedException` which bubbles up past `run()` and `main_vd()` to `vd_cli()`:
https://github.com/saulpw/visidata/blob/d4999149259ddf7544ec272757a018525f897104/visidata/basesheet.py#L203-L204
https://github.com/saulpw/visidata/blob/d4999149259ddf7544ec272757a018525f897104/visidata/main.py#L386-L394

This PR prevents the abrupt exit and shows the fail message:  `no command for not-a-command`

### DIFF
--- a/visidata/settings.py
+++ b/visidata/settings.py
@@ -383,8 +383,12 @@ def getCommand(sheet, cmd):
         return cmd
 
     longname = cmd
+    seen = []
     while vd.bindkeys._get(longname, obj=sheet) is not None:
         longname = vd.bindkeys._get(longname, obj=sheet)
+        if longname in seen:
+            vd.fail(f'keystroke/command definitions form a cycle: {longname}')
+        seen.append(longname)
 
     return vd.commands._get(longname, obj=sheet)
 
--- a/visidata/mainloop.py
+++ b/visidata/mainloop.py
@@ -225,7 +225,7 @@ def mainloop(vd, scr):
             return vd.lastErrors and '\n'.join(vd.lastErrors[-1])
         elif vd.bindkeys._get(vd.keystrokes) is not None:
             try:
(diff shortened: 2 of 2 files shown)


midichef:
The commit message records why I catch `Exception` instead of the more specific `ExpectedException`. I tried for about 20 minutes to import `ExpectedException` in mainloop.py, but I could not figure out a way to untangle the resulting circular imports. Using `Exception` with `exceptionCaught` gets the job done, and matches the error handling elsewhere in the same function:
https://github.com/saulpw/visidata/blob/d4999149259ddf7544ec272757a018525f897104/visidata/mainloop.py#L239-L243

visidata/settings.py
```diff
@@ -383,8 +383,13 @@ def getCommand(sheet, cmd):
         return cmd
 
     longname = cmd
+    seen = []
     while vd.bindkeys._get(longname, obj=sheet) is not None:
         longname = vd.bindkeys._get(longname, obj=sheet)
+        if longname in seen:
+            vd.warning(f'keystroke/command definitions form a cycle: {longname}')
```
saulpw: can we use `vd.error` or `vd.fail` instead of `vd.warning` and `break`?

midichef: Oops, yes. I'll push a replacement commit.

saulpw (approved):

PR #2863, opened 2025-09-15
Fixed bug when connecting to MySQL with no password

pequiste:
Visidata currently crashes when trying to connect to a MySQL database with no password because `unquote(self.url.password)` crashes on None. And `mysql.connect` cannot take `password=None` either.

Fixed the issue by conditionnally adding the password to the params if present.


saulpw (approved):
Thanks, @pequiste !

PR #2865, opened 2025-09-17
Update clipboard commands to specify internal clipboard

### LINKED ISSUE #2864: clarify the difference between internal and system clipboad in the command names
While using visidata, I tried to copy the content of a cell. I tried it a few times, but it did not work. I thought that the cause could be my terminal, or the fact that I run wayland.
I though about opening an issue for that, but since nobody had the same problem, I went to look at the code.
Only at this point did I realize that "clipboad" meant **internal** clipboard and not system clipboard, just like in vim. It did not cross my mind it could be the case.

For that reason, the docs should clarify the distinction between both, in particular in the command descriptions.

rambip:
closes #2864

There might be another way of improving the docs regarding clipboard, but it's the most straightforward solution I found.

### DIFF
--- a/visidata/clipboard.py
+++ b/visidata/clipboard.py
@@ -176,15 +176,15 @@ def paste_after(sheet, rowidx):
     sheet.addRows(addedRows, index=rowidx)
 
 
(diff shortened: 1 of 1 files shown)


saulpw:
I made the changes manually in another commit.  Thanks for the feedback @rambip!

PR #2867, opened 2025-09-20
Setting the locale to en_US before running the tests.

pequiste:
Several tests fail if the locale is not set to en_US before running, mainly around date formatting.

This path sets this locale in environment variables before launching the tests.

### DIFF
--- a/dev/test.sh
+++ b/dev/test.sh
@@ -7,7 +7,7 @@ shopt -s failglob
 
 trap "echo aborted; exit;" SIGINT SIGTERM
 
-export LC_NUMERIC="en_US.UTF-8" # 2867
+export LC_NUMERIC="en_US.UTF-8" #2867
 export LC_TIME="en_US.UTF-8"
 
 MAX_PARALLEL_JOBS=1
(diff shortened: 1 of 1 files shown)


dev/test.sh
```diff
@@ -7,6 +7,15 @@ shopt -s failglob
...
 trap "echo aborted; exit;" SIGINT SIGTERM
 
+export LANG="en_US.UTF-8"
+export LC_COLLATE="en_US.UTF-8"
+export LC_CTYPE="en_US.UTF-8"
+export LC_MESSAGES="en_US.UTF-8"
+export LC_MONETARY="en_US.UTF-8"
+export LC_NUMERIC="en_US.UTF-8"
+export LC_TIME="en_US.UTF-8"
+export LC_ALL=
```
saulpw: could we just set `LC_ALL="en_US.UTF-8"`?  I'd much prefer 1 line to 8.

pequiste: I tried it and the failing tests are still failing. I think that LC_ALL is just the default value if another value is not set but they are all set to whatever the default value is for the locale the OS was installed in.

However, we can get away with only two lines, LC_NUMERIC and LC_TIME. If at some point we add something like money formatting, then we’d need to add another line but right now every test passes on my machine with those two.

saulpw: Okay, then let's do the minimum.   Please reference this PR with a comment (`#2867`) at the change.

saulpw (approved):

PR #2868, opened 2025-09-20
[Proposal] Preserving the type returned by the aggregator for mean, avg, and median instead of coercing to an int.

pequiste:
visidata’s current behavior for average, mean, and median is to coerce the result to a float which works fine for the builtin types but doesn’t play nice when adding custom precision preserving types, like Python’s `Fraction`, and `Decimal` types.

It would also be possible to replace `sum(vals) / len(val)` by Python’s native `statistics.mean` which also preserves the int type when possible:

- `statistics.mean([1, 2]) # 1.5 (float)`
- `statistics.mean([1, 2, 3]) # 2 (int)` 

But that would mean that two golden files have to change since they would lose some `.00`.

### DIFF
--- a/visidata/aggregators.py
+++ b/visidata/aggregators.py
@@ -141,7 +141,7 @@ def aggregator_list(vd, name, helpstr='', type=anytype, listtype=anytype):
 def mean(vals):
     vals = list(vals)
     if vals:
-        return float(sum(vals))/len(vals)
+        return sum(vals)/len(vals)
 
 def vsum(vals):
     return sum(vals, start=type(vals[0] if len(vals) else 0)())  #1996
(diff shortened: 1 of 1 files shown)


visidata/aggregators.py
```diff
@@ -233,9 +233,9 @@ def identity(val):
 
 vd.aggregator('min', min, 'minimum value')
 vd.aggregator('max', max, 'maximum value')
-vd.aggregator('avg', mean, 'arithmetic mean of values', type=float)
-vd.aggregator('mean', mean, 'arithmetic mean of values', type=float)
-vd.aggregator('median', statistics.median, 'median of values')
+vd.aggregator('avg', mean, 'arithmetic mean of values', type=lambda x: x)
```
saulpw, line 236: Use `anytype` for the identity function as a type.

pequiste: Using anytype instead of lambda x:x fails the following tests:

- tests/golden/aggregators-set.tsv
- tests/golden/avg-nulls.tsv
- tests/golden/freq-error.tsv
- tests/golden/freq-summary.tsv
- tests/golden/pivot.tsv

Lots of n.00 that turn into n.0 but also a 624.00 that turns into a 623.9999999999999

saulpw: Huh!  That's weird to me.  I guess the TSV saver must do something special for anytype columns?

saulpw: Okay, I did a deeper dive on this today.  We actually want a generic "numtype" to use instead of anytype, which will preserve the expected precision and formatting.  In the case of this PR it's mostly equivalent to `lambda x:x`, but numtype also allows us to preserve numeric-ness on those columns (for plotting or numeric binning, for example).  We'll take this PR and then I'll apply another change on top of it, to use a new `numtype` instead of the `lambda x: x`.

PR #2870, opened 2025-09-22
[cmdlog-] make vdj shebang work on macOS

### LINKED ISSUE #2869: Shebang in `vdj` files doesn't work on macOS, easy to fix
**Small description**

This is similar to, but should work better than the now-reverted
https://github.com/saulpw/visidata/commit/226d428c8f91f32d4a3dede657d0bc5dfdd24d78

Currently, I get this error if I try to execute a vdj file on my Mac:

```console
# Create any vdj file with visidata on macOS
$ chmod +x test.vdj
$ ./test.vdj
exec: Failed to execute process './test.vdj': The file specified the interpreter 'vd -p', which is not an executable command.
```

The problem is that the shebang currently says `#!vd -p`.

I've tested a fix, changing the shebang to:

```
#!/usr/bin/env -S vd -p
```

fixes the problem and should be cross-platform. See [`man env`](https://www.mankier.com/1/env#Options--S/--split-string_usage_in_scripts), there's also a good example
of using such shebangs in  https://docs.astral.sh/uv/guides/scripts/#using-a-shebang-to-create-an-executable-file.

This should work better than the (now-reverted) https://github.com/saulpw/visidata/commit/226d428c8f91f32d4a3dede657d0bc5dfdd24d78.

**Expected result**

Executing the vdj file should work

See above.

-----

I considered creating a quick PR for this, but then I read contributing docs, and I'm not sure that I can transfer copyright for my contributions. (Such a small patch is probably not copyrightable anyway, but it's simpler if I just make it an issue instead).

ilyagr:
This is similar to, but should work better than the now-reverted https://github.com/saulpw/visidata/commit/226d428c8f91f32d4a3dede657d0bc5dfdd24d78

Currently, I get this error if I try to execute a vdj file on my Mac:

```console
$ chmod +x test.vdj
$ ./test.vdj
exec: Failed to execute process './test.vdj': The file specified the interpreter 'vd -p', which is not an executable command.
```

With this change, it works.

-----------------

Some references for the `-S` option:

- `man env`
- https://docs.astral.sh/uv/guides/scripts/#using-a-shebang-to-create-an-executable-file is another example of using it.

Fixes #2869

### DIFF
--- a/visidata/cmdlog.py
+++ b/visidata/cmdlog.py
@@ -42,7 +42,7 @@ VisiData.save_vd = VisiData.save_tsv
 @VisiData.api
 def save_vdj(vd, p, *vsheets):
     with p.open(mode='w', encoding=vsheets[0].options.save_encoding) as fp:
-        fp.write("#!vd -p\n")
+        fp.write("#!/usr/bin/env -S vd -p\n")
         for vs in vsheets:
             vs.write_jsonl(fp)
 


saulpw (approved):
Thanks @ilyagr!

ilyagr:
Thanks for making visidata! :)

PR #2875, opened 2025-09-25
[aggr-] catch nonexistent aggr in addcol-aggregate

midichef:
A fix to add error checking for `addcol-aggregate`. If I run `addcol-aggregate`, then enter any string that is not an aggregator, like `noop`, I get the appropriate warning:  `aggregator does not exist: noop` but I also get a traceback:
```
File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/aggregators.py", line 226, in aggregate_groups
    aggr_val = aggr.aggregate(col, [rows[rownum] for _, _, rownum in group])
AttributeError: 'NoneType' object has no attribute 'aggregate'
```


saulpw (approved):
Thanks for catching this, @midichef.

PR #2877, opened 2025-10-06
[status-] with lstatus_max, fix sheet list substr dupe

midichef:
The string length calculation in `truncateMiddle()` is off by a factor of 2 which leads to sheetlist duplication in the statusbar. You can see it by running:
`vd --disp_lstatus_max=50 sample_data/benchmark.csv`
The status bar shows `1›benchmark…u_active]1›benchmark|`, repeating`1›benchmark`.

The `…u_active]` comes from `middleTruncate()` clipping in the middle of internal visidata markup. I'll open a separate issue for that, as that is a tougher problem.

### DIFF
--- a/visidata/statusbar.py
+++ b/visidata/statusbar.py
@@ -147,7 +147,7 @@ def debug(vd, *args, **kwargs):
         return vd.status(*args, **kwargs)
 
 def middleTruncate(s, w):
-    if len(s) <= w:
+    if len(s) <= 2*w:
         return s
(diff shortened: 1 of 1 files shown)


saulpw (commented):
Thanks for finding and fixing this! I think that this would be cleaner if we pass the field width into `middleTruncate`, and then do the division inside that function.

midichef:
I have a more complete fix that completely replaces `middleTruncate()` so I'll close this PR.

PR #2881, opened 2025-10-10
Native Graph: Fixed axes, Seaborn Graph: added Title, axes

meestahp:
Hello, I have noticed that on the native graph the x graph is always truncated with '>>' and that the Y axis is missing, not sure if its a resolution issue. Anyway the seaborn graph is missing axes as well as a title. I have gone ahead made a few changes for your review. 
2025-10-10_16-19_original
2025-10-10_16-20_vdv
Figure_1
Figure_1_vdv

### DIFF
--- a/visidata/canvas.py
+++ b/visidata/canvas.py
@@ -355,7 +355,7 @@ class Canvas(Plotter):
     leftMarginPixels = 10*2
     rightMarginPixels = 4*2
     topMarginPixels = 0*4
-    bottomMarginPixels = 1*4  # reserve bottom line for x axis
+    bottomMarginPixels = 2*4  # reserve bottom line for x axis
     guide = '# Canvas\n'
 
     def __init__(self, *names, **kwargs):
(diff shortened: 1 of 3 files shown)


midichef:
I like this change. Having fully readable labels on the axis is a very helpful improvement. Thanks!

saulpw:
Hi @meestahp, thanks for the PR.  The seaborn changes are very welcome.  I'm not sure about adding labels to the TUI graph, however; in particular the left/y axis label seems awkward to manage--for instance, when multiple numeric columns are graphed with `g.`, or if the axis name is longer than the height.  I'm wondering if we could show "<y-axis-name> vs <x-axis-name>" on the top of the graph like a title, and avoid at least the y-axis label, if not the x-axis label too? This would save that screen real-estate, and they could get turned off with `v` like the other legends.  What do you think?

meestahp:
Ok, what I will do start another PR with just the seaborn changes and send the TUI graph changes separately as the TUI requires more careful attention.

Issue #1673, opened 2023-01-13
[graph] failure when there is only one value in the x or y column


midichef:
**Small description**
plot-column() fails when the x values are all the same, or when the y values are all the same.

**Expected result**
Some kind of plot should appear.

**Actual result with screenshot**
```
vd -p graph_one_x.vdj

Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/visidata/threads.py", line 198, in _toplevelTryFunc
    t.status = func(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/visidata/graph.py", line 74, in reload
    self.resetBounds()
  File "/usr/local/lib/python3.10/dist-packages/visidata/graph.py", line 78, in resetBounds
    super().resetBounds()
  File "/usr/local/lib/python3.10/dist-packages/visidata/canvas.py", line 489, in resetBounds
    self.visibleBox = Box(0, 0, self.plotviewBox.w/self.xScaler, self.plotviewBox.h/self.yScaler)
  File "/usr/local/lib/python3.10/dist-packages/visidata/canvas.py", line 521, in xScaler
    xratio = self.plotviewBox.w/(self.canvasBox.w*self.xzoomlevel)
ZeroDivisionError: float division by zero
```

```
vd -p graph_one_y.vdj

Traceback (most recent call last):
...
  File "/usr/local/lib/python3.10/dist-packages/visidata/canvas.py", line 489, in resetBounds
    self.visibleBox = Box(0, 0, self.plotviewBox.w/self.xScaler, self.plotviewBox.h/self.yScaler)
  File "/usr/local/lib/python3.10/dist-packages/visidata/canvas.py", line 530, in yScaler
    yratio = self.plotviewBox.h/(self.canvasBox.h*self.yzoomlevel)
ZeroDivisionError: float division by zero
```

**Steps to reproduce with sample data and a .vd**
[graph_one_x.vdj.txt](https://github.com/saulpw/visidata/files/10407511/graph_one_x.vdj.txt)
[graph_one_y.vdj.txt](https://github.com/saulpw/visidata/files/10407512/graph_one_y.vdj.txt)
[one_x_val.csv](https://github.com/saulpw/visidata/files/10407513/one_x_val.csv)
[one_y_val.csv](https://github.com/saulpw/visidata/files/10407515/one_y_val.csv)

**Additional context**
visidata-2.10.2 or latest develop branch

[...]

PR #2884, opened 2025-10-13
[canvas-] fix division by zero when col's only value is a large float

midichef:
When plotting a graph, the plotter tries to prevent `xmin == xmax` by expanding the range. It adds 1 to `xmax` in that situation. (We ran into this problem before in #1673.) But that fails for large floats, because adding 1 leaves the number unchanged, as in `x=2**53; float(x) == float(x + 1)`.  (Because `math.ulp(2**53)` is 2.0.)

Example, when plotting the first 2 or last 2 lines of this data with `.`:
```
x,y
9007199254740992,1
9007199254740992,2
1,9007199254740992
2,9007199254740992
```
visidata 3.3 gives `ZeroDivisionError: float division by zero` because the width or height of the box is zero when `xmin == xmax` or `ymin == ymax`:
```
File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/canvas.py", line 648, in xScaler
    xratio = self.plotviewBox.w/(self.canvasBox.w*self.xzoomlevel)
```
and
```
File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/canvas.py", line 657, in yScaler
    yratio = self.plotviewBox.h/(self.canvasBox.h*self.yzoomlevel)
```

This PR handles the case of such large floats.

### DIFF
--- a/visidata/canvas.py
+++ b/visidata/canvas.py
@@ -587,8 +587,12 @@ class Canvas(Plotter):
             ymax = ymax or 0
             if xmin == xmax:
                 xmax += 1
+                if xmin == xmax:  #handle large floats that were unchanged by += 1
+                    xmin = xmin * 0.99  #the alternative of increasing xmax could hit infinity
             if ymin == ymax:
                 ymax += 1
+                if ymin == ymax:
+                    ymin = ymin * 0.99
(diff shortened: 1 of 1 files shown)


saulpw (approved):
I don't know how you find these edge cases, @midichef!  But thanks for your diligence.

PR #2887, opened 2025-10-18
[path-] handle filenames ending in . in Python 3.14

### LINKED ISSUE #2883: [python] New Python version 3.14 released
New Python version 3.14 released. And 3.9 is now EOL.

See https://devguide.python.org/versions/

It might be useful to run tests on the new release.

(Similar to Python 3.13 release https://github.com/saulpw/visidata/issues/2550)

midichef:
`_path.suffix` for the path `'foo.'` is `'.'` in Python 3.14. In previous versions it was `''`

This fixes a test I noted in #2883 that fails only in Python 3.14.

EDITED to add: now that wheels are out to allow testing with Python 3.14, this PR closes #2883.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.8, 3.9, "3.10", "3.11", "3.12", "3.13"]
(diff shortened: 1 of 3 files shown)


saulpw (approved):
Thanks @midichef !

PR #2891, opened 2025-10-30
feat(ui): confirm on exit when quitting last sheet (q) and quit-all (gq)

denisotree:
Summary
- Add exit confirmation prompt when:
  - pressing `q` on the last stacked sheet (app exit)
  - running `gq` (quit all sheets)
- Prompt text: "Are you sure you want to exit?"

Rationale
- Prevents accidental app exits, consistent with existing quit guards for modified/guarded sheets.

User-facing behavior
- `q`:
  - normal: quits one sheet
  - last sheet: shows confirmation; only exits on confirmation
- `gq`: shows confirmation before quitting all sheets
- All prompts are in English.

Implementation
- visidata/sheets.py:
  - BaseSheet.api: [quit_with_exit_confirm(sheet)](cci:1://file:///Users/denisotree/code/visidata/visidata/sheets.py:1192:0-1198:18)
  - VisiData.api: [quit_all_with_exit_confirm(vd)](cci:1://file:///Users/denisotree/code/visidata/visidata/sheets.py:1201:0-1207:23)
  - Rebound:
    - `q` → [quit_with_exit_confirm()](cci:1://file:///Users/denisotree/code/visidata/visidata/sheets.py:1192:0-1198:18)
    - `gq` → [quit_all_with_exit_confirm()](cci:1://file:///Users/denisotree/code/visidata/visidata/sheets.py:1201:0-1207:23)

CAA
- I will e-sign the VisiData CAA (https://visidata.org/caa).

### DIFF
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -1191,6 +1191,24 @@ def confirmQuit(vs, verb='quit'):
 
 
 @BaseSheet.api
+def quit_with_exit_confirm(sheet):
+    # confirm when quitting the last stacked sheet (app exit)
+    if len(vd.stackedSheets) <= 1 and not vd._nextCommands:
+        vd.draw_all()
+        vd.confirm('Are you sure you want to exit? ')
+    vd.quit(sheet)
+
+
+@VisiData.api
+def quit_all_with_exit_confirm(vd):
+    # confirm when quitting all sheets (app exit)
+    if not vd._nextCommands:
+        vd.draw_all()
+        vd.confirm('Are you sure you want to exit? ')
+    vd.quit(*vd.sheets)
+
+
+@BaseSheet.api
 def preloadHook(sheet):
     'Override to setup for reload().'
     sheet.confirmQuit('reload')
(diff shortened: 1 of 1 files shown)


saulpw:
Hi @denisotree, thanks for the PR.  Doesn't the `quitguard` option do exactly this?

denisotree:
Hi @saulpw, I understand this seems like a duplication, but I've often encountered this situation: I'm exploring a CSV in Visidata, I've already labeled the columns, I've done groupings, and sometimes I accidentally press q too many times and exit the program. It's very frustrating. I'd be happy if the program would always warn me that pressing q would kick me out to the console.

saulpw:
Do you have `options.quitguard = True` in your .visidatarc though?  I find that VisiData is often too aggressive about asking me to confirm quitting sheets that are only minimally changed (e.g. columns renamed or hidden).

denisotree:
Hi!
Yes, I’ve double-checked — options.quitguard = True is set in my .visidatarc file. A warning does appear when exiting a modified document as expected.

My change handles the specific case where there’s only one sheet left before returning to the console, or when quitting the entire console with g + q. It prevents the full column frame and computed values from being reset when q is pressed accidentally. In this situation, options.quitguard = True doesn’t affected.

An alternative (and probably cleaner) approach would be to adjust confirmQuit to trigger whenever len(vd.stackedSheets) <= 1. What do you think?

denisotree:
I’ve updated the confirmQuit function to what I think is the ideal flow. You’re right — VisiData tends to be a bit too aggressive about prompting for quit confirmations when options.quitguard = True, even for minimal changes.

The new implementation keeps your original logic intact, but now a quit confirmation is only triggered when pressing q would actually exit the console.

visidata/sheets.py
```diff
@@ -1182,11 +1182,14 @@ def quit(vd, *sheets):
 
 @BaseSheet.api
 def confirmQuit(vs, verb='quit'):
-    if vs.options.quitguard and vs.precious and vs.hasBeenModified and not vd._nextCommands:
-        vd.draw_all()
+    if vd._nextCommands:
+        return
+    vd.draw_all()
+    if len(vd.stackedSheets) <= 1:
+        vd.confirm(f'Are you sure you want to {verb}?')
```
saulpw, line 1189: This will always ask for confirmation before quitting.  I can accept that if the sheet has been modified OR if options.quitguard is True, but we need to launch vd and press `gq` immediately to exit without confirmation.  If you want, lobby for options.quitguard to be set to True by default, but we can't hardcode it like this.

visidata/sheets.py
```diff
@@ -1182,11 +1182,14 @@ def quit(vd, *sheets):
 
 @BaseSheet.api
 def confirmQuit(vs, verb='quit'):
-    if vs.options.quitguard and vs.precious and vs.hasBeenModified and not vd._nextCommands:
-        vd.draw_all()
+    if vd._nextCommands:
+        return
```
saulpw, line 1186: This means that having commands queued up means the quit is auto-confirmed, since vd.confirm raises an Exception to indicate disconfirmation.

saulbert:
Implemented a different approach in 9196601a on develop: when `quitguard` is True, `vd.quit()` now confirms before exiting the program, with a message showing how many sheets are modified. This replaces the per-sheet confirmations when exiting, so `gq` only prompts once.

Thanks @denisotree for raising the use case!

[Claude Opus 4.6]

PR #2892, opened 2025-11-06
[movement-] fix desc of scroll-up/down/left/right amount

midichef:
A small correction to four command descriptions. I phrased them like "scroll up by one row increment" because the alternative "scroll one row increment up" felt clunky.


saulpw (approved):
Ah yes, it wasn't always using `scroll_incr`.  Thanks for updating it.

PR #2893, opened 2025-11-07
[macro-] fix vd attempts to use deleted macros

### LINKED ISSUE #2778: VD attempts to use a deleted macro
Just a small macro oddity, tested in dev version.

Deleting a macro will not free that hotkey up properly.

- Create a macro and bind it to an as-yet-unbound key (e.g. 'r')
- Delete that macro
- Press 'r' again

The status msg `KeyError: 'r'` shows, and it can be seen that it attempts `exec-r`.
Expected would be a `no command for "r"` msg, as would show before macro was recorded.

haoyeau:
Fix #2778

- add `removeCommand` api on BaseSheet.
- remove command registration while deleting row from MacroSheet.

### DIFF
--- a/visidata/macros.py
+++ b/visidata/macros.py
@@ -31,9 +31,20 @@ class MacroSheet(IndexSheet):
 
     def iterload(self):
         yield from vd.macrobindings.values()
-
+            
     def commitDeleteRow(self, row):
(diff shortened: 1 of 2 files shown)


saulpw (approved):
Thanks for fixing this, @haoyeau!

PR #2894, opened 2025-11-08
[column-] fix width in formatValue for list/dict/tuple

midichef:
For lists, dicts, and tuples, the results of `cursorDisplay` are 1 character too long, because `formatValue()` uses a width that's 1 too large.

Tos ee this in action: `echo '{"list": [1, 2, 3, 4]}' |vd -f jsonl` then resize the column to width 10:  `z_` `10`. What's visible becomes `[4] 1; 2…`, with no semicolon after the 2. But upon `zy`, the clipboard value shown in the statusbar has the extra semicolon:  `[4] 1; 2;…`.

This PR fixes the results of `formatValue()`/`cursorDisplay` to match what's drawn on screen.

### DIFF
--- a/visidata/column.py
+++ b/visidata/column.py
@@ -246,7 +246,7 @@ class Column(Extensible):
             if isinstance(typedval, (dict, list, tuple)):
                 if width is None:
                     return ''.join(iterchars(typedval))
-                dispval, dispw = clipstr(iterchars(typedval), width)
+                dispval, dispw = clipstr(iterchars(typedval), width-1) #subtract 1 for left-side margin
                 return dispval
 
         if isinstance(typedval, bytes):


visidata/column.py
```diff
@@ -246,7 +246,7 @@ def formatValue(self, typedval, width=None):
             if isinstance(typedval, (dict, list, tuple)):
                 if width is None:
                     return ''.join(iterchars(typedval))
-                dispval, dispw = clipstr(iterchars(typedval), width)
+                dispval, dispw = clipstr(iterchars(typedval), max(width-1, 0)) #subtract 1 for left-side margin
```
saulpw: `width` should never be 0, because that means a hidden column.  So `width-1` should be sufficient.

midichef: Okay, true, and it turns out that  `_clipstr()` treats all widths < 1 the same too, so it's definitely not needed. I've pushed a version with `width-1`.

saulpw (approved):

PR #2898, opened 2025-11-14
fix 3.13 histogram

haininhhoang94:

### DIFF
--- a/visidata/column.py
+++ b/visidata/column.py
@@ -10,29 +10,65 @@ import json
 from visidata import options, anytype, stacktrace, vd, drawcache
 from visidata import asyncthread, dispwidth, clipstr, iterchars
 from visidata import wrapply, TypedWrapper, TypedExceptionWrapper
-from visidata import Extensible, AttrDict, undoAttrFunc, ExplodingMock, MissingAttrFormatter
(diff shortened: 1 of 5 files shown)


saulpw:
Hi @haininhhoang94, I can't see what's changed because of all the formatting changes.  Can you remove them, and only include the changes necessary to fix the issue you found?  Also it would be great if you would link to the issue you are fixing (use `#1234` for the github issue in the commit message).

haininhhoang94:
Sure! I was using visidata and saw no one mention the histogram breaking in 3.13 when press F, so I fast fixing those

saulbert:
Thanks for the contribution @haininhhoang94.

We weren't able to reproduce a histogram bug on Python 3.13 — the frequency table works correctly on 3.13.9 with the current code. Without a specific issue to reference or steps to reproduce, we can't merge a fix for something that doesn't appear to be broken.

The PR also includes unrelated changes (str-to-int coercion, pandas StringDtype handling) and large-scale formatting changes that make it difficult to review. As mentioned in the earlier comment, VisiData uses single-quoted strings and its own formatting conventions.

If you do encounter a specific reproducible bug with the histogram on 3.13, please open an issue with steps to reproduce and we can look into it.

PR #2901, opened 2025-11-17
merge deferred mods with existing deferred add

justin2004:
i think we have a little bug wrt deferred modifications.

demonstration

put this in the CsvSheet class (just for this demo)
```
defer = True
```

currently we have this behavior:

vd sample_data/benchmark.csv
add-row
edit the customer column

Screenshot 2025-11-17 at 3 10 14 PM

then in the [vd repl](https://github.com/saulpw/visidata/discussions/2290)

vd.sheet.getDeferredChanges()
({140131282503616: [None, None, None, None, None, None, None]}, {}, {})

note that there is a deferred addition but has no column values (all `None`) even though we modified 1 cell.
that is, the deferred modifications are lost if applied to a new row.

with this PR we have this behavior:

vd sample_data/benchmark.csv
add-row

Screenshot 2025-11-17 at 3 12 58 PM

then confirm it is an empty row at the vd repl
vd.sheet.getDeferredChanges()
({139923036424208: [None, None, None, None, None, None, None]}, {}, {})

note empty row -- as expected

now edit the name column and another column if you like

Screenshot 2025-11-17 at 3 14 21 PM

and at the vd repl
vd.sheet.getDeferredChanges()
({139923036424208: [None, 'fred', None, 'chickpea', None, None, None]}, {}, {})

note that the deferred addition now includes all modifications -- so we effectively rolled the modifications into the addition (since it was a deferred addition anyway)

### DIFF
--- a/dev/GIT.md
+++ b/dev/GIT.md
@@ -78,6 +78,8 @@ We've started doing this for PRs also. It's a bit of clutter but it helps:
 
 Commit before posting GitHub comments/replies. Let Saul push first, then post the comment. Don't post replies to issues before the relevant code is committed and pushed.
 
+AI-generated comments should have the model attribution at the bottom (e.g. `[Claude Opus 4.6]`), not the top.
+
 ## Branch and Merge Workflow
 
 In general:
--- a/dev/STYLE.md
+++ b/dev/STYLE.md
@@ -333,6 +333,10 @@ api_key = vd.options.my_api_key or vd.fail(...)  # Don't do this
 
 Use **module name** or abbrevation as prefix for options used exclusively by that module.
 
+### Options Defined in Feature Files
+
+When core code (e.g. `sheets.py`) needs to check an option defined by a feature file, use `options.get()` with a default so it works even if the feature isn't loaded.  Don't define the option in two places with different defaults.
+
 ## Best Practices
 
 ### Documentation
--- a/dev/test.sh
+++ b/dev/test.sh
@@ -62,11 +62,9 @@ declare -A FLAKY_TESTS
 # Clean output directory before running tests
 rm -f tests/output/*
 
-# Build batch: all tests in a single vd process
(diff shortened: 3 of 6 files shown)


justin2004:
just to note the reason...

i am extending the Imap loader to support sending emails (with smtp) and using deferred changes is useful for that:

image

saulpw:
Hm, I thought this is how it already worked.  When I view a sqlite database with `--debug` (which shows the SQL statements executed), do a similar set of add/edit commands and then commit, it only shows the one INSERT, and no UPDATEs.  Are there different code paths on this for some reason?

justin2004:
hey @saulpw , i think i see what is happening.

since the deferred mods don't get added to the deferred adds (in the develop branch -- like my example above shows)  
the sqlite loader, upon commit, when it has a deferred add has to reach back into the sheet to get the values itself using: 
`col.getTypedValue(row)` or `col.getDisplayValue(r)`

notice how the parameters to the execute sql function [are built](https://github.com/saulpw/visidata/blob/develop/visidata/loaders/sqlite.py#L167) by calling 2 helper functions that reach back into the sheet (instead of using the values in the deferred adds (because they get discarded if done after the initial add-row command)).

i think i can simplify the sqlite loader on top of this PR. let me see...

saulpw:
Oh, that would be great if you can simplify the sqlite with this!

saulbert:
Rebased onto develop HEAD as 2901-unfurl branch — only the modify.py fix and test remain, test infrastructure noise removed.

saulbert:
Merged onto develop via rebase (8e12bcd1). Thanks @justin2004!

saulpw:
Thanks @justin2004!  I got it fixed up and added a test.  (and then got sidetracked to fixing the test framework, as usual :).  I couldn't update this PR so I merged it separately.

PR #2904, opened 2025-11-22
[graph-][seaborn-] Add title and axis labels

meestahp:
Added the seaborn graph title and axis labels to plots called by plot-ext

### DIFF
--- a/visidata/features/graph_seaborn.py
+++ b/visidata/features/graph_seaborn.py
@@ -4,16 +4,17 @@ Add plot-column-ext and plot-numerics-ext to Sheet, and plot-ext to GraphSheet, 
 
 from visidata import vd, VisiData, Sheet, GraphSheet, Progress, asyncthread
 
+
 @VisiData.api
 @asyncthread
 def plot_seaborn(vd, rows, xcols, ycols):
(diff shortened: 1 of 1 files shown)


saulpw (approved):
Thanks, @meestahp !

saulpw (approved):

midichef:
Repeating my question here, which got lost in the transition from #2881:
Why is the `vd` argument newly added to `ext_plot_seaborn()`? Is the purpose to speed up future loads, by allowing the imported modules from `importExternal` to stay in the `vd` object?

https://github.com/saulpw/visidata/blob/9c6e901bca8c1ab5281dae95d91dac80ba8426bd/visidata/features/graph_seaborn.py#L17-L20

meestahp:
> Repeating my question here, which got lost in the transition from #2881: Why is the `vd` argument newly added to `ext_plot_seaborn()`? Is the purpose to speed up future loads, by allowing the imported modules from `importExternal` to stay in the `vd` object?

I think it was because the plotting runs as a separate process, for the numeric checking and the vd.options.debug call. Since process is discarded there is no chachinig speed afaik.

PR #1998, opened 2023-08-17
[graph] widen left margin to hold y-axis labels

midichef:
This expands the left margin when the y-axis has really big numbers. It's necessary because Apple makes so much money that it was ruining the Visidata graph layout in [@cool-RR's talk at 25:11](https://www.youtube.com/watch?v=gdvanybQn8Y&t=1511s)

PR #2905, opened 2025-11-22
[graph-][TUI-] Moved axis labels to top, as title

meestahp:
TUI Graph: Moved existing truncated axis labels to top to serve as title, since they are part of the canvas, pressing v will make them dissapear along with the UI

### DIFF
--- a/visidata/graph.py
+++ b/visidata/graph.py
@@ -87,8 +87,8 @@ class GraphSheet(InvertedCanvas):
         self.ycols or vd.fail('%s is non-numeric' % '/'.join(yc.name for yc in kwargs.get('ycols')))
 
     def resetCanvasDimensions(self, windowHeight, windowWidth):
-        if self.left_margin < self.ylabel_maxw:
-            self.left_margin = self.ylabel_maxw
+        # 1 char for y-name, 1 char for blank, 10 chars for y-tick-labels
+        self.left_margin = 4 + 20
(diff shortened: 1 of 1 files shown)


visidata/graph.py
```diff
@@ -87,8 +87,8 @@ def __init__(self, *names, **kwargs):
         self.ycols or vd.fail('%s is non-numeric' % '/'.join(yc.name for yc in kwargs.get('ycols')))
 
     def resetCanvasDimensions(self, windowHeight, windowWidth):
-        if self.left_margin < self.ylabel_maxw:
-            self.left_margin = self.ylabel_maxw
+        # 1 char for y-name, 1 char for blank, 10 chars for y-tick-labels
```
midichef: (repeating my comment here from #2881)
The existing code that widens the left margin to be wider than the widest y-label is necessary. It avoids drawing labels over the leftmost data points. This does happen for real data, as can be seen in the video linked in https://github.com/saulpw/visidata/pull/1998 where such drawing-over happened in a real presentation.

What considerations are driving the new choice of a fixed width of 10 chars (i.e. `4 + 20` "plotter pixels") for the y-tick-labels? If I understand that, maybe we can find a solution that avoids drawing-over data. For example, you could start by setting `left_margin` to `4 + 20`, and then use the existing code to widen the margin to fit the labels.

PR #2907, opened 2025-11-25
[config-] respect XDG_CONFIG_HOME and XDG_CACHE_HOME on macOS

maxim-uvarov-ai-assistant:
I (@maxim-uvarov) discovered that XDG settings are not respected on my macOS. I prompted `claude code` to  fix it. Since the change is minimal, I kindly ask you to check it. Though please feel  free to discard it if it is by any means inappropriate.

## Summary
- Respect `XDG_CONFIG_HOME` and `XDG_CACHE_HOME` environment variables on macOS
- Mirrors existing Linux behavior in the vendored appdirs.py
- Falls back to native macOS paths (`~/Library/Preferences`, `~/Library/Caches`) when XDG vars are not set

## Changes
Minimal 2-line change in `visidata/vendor/appdirs.py`:
- `user_config_dir()`: Check `XDG_CONFIG_HOME` before defaulting to `~/Library/Preferences`
- `user_cache_dir()`: Check `XDG_CACHE_HOME` before defaulting to `~/Library/Caches`

## Test plan
- [ ] Verify default behavior unchanged when XDG vars are unset
- [ ] Verify `XDG_CONFIG_HOME` is respected when set
- [ ] Verify `XDG_CACHE_HOME` is respected when set
- [ ] Verify `--visidata-dir` flag still overrides all defaults

### DIFF
--- a/visidata/vendor/appdirs.py
+++ b/visidata/vendor/appdirs.py
@@ -194,7 +194,7 @@ def user_config_dir(appname=None, appauthor=None, version=None, roaming=False):
     if system == "win32":
         path = user_data_dir(appname, appauthor, None, roaming)
     elif system == 'darwin':
-        path = os.path.expanduser('~/Library/Preferences/')
+        path = os.getenv('XDG_CONFIG_HOME', os.path.expanduser('~/Library/Preferences'))
         if appname:
             path = os.path.join(path, appname)
     else:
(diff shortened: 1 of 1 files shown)


maxim-uvarov-ai-assistant:
I signed it, yet the `recheck` does nothing. 

image

saulpw:
Thanks @maxim-uvarov, I appreciate your candor in identifying the provenance of this change.  The actual changes seem quite reasonable, and I only ask that you explicitly verify that this code is tested according to the test plan.  That is, on MacOS, it works for you when the variables are set, and continues to work as it currently does when the variables are unset.  If that's the case then we can merge this.

Also, I don't know why the CLA bot is asking for your signature; we have a "minimum" change size to require it, something like 3-5 lines, which your code clearly falls under.  So there's no actual copyright to assign.

maxim-uvarov:
thank you, @saulpw! 

> That is, on MacOS, it works for you when the variables are set, and continues to work as it currently does when the variables are unset. If that's the case then we can merge this.

I tested it, and on my machine, it appears to be working!

PR #2908, opened 2025-11-27
[statusbar-] make lstatus_max truncation preserve vd markup

midichef:
This PR supersedes #2877. The problem described there is that `truncateMiddle()` duplicates substrings, and does not properly handle visidata's internal markup. To demonstrate, do `vd --disp_lstatus_max=50 sample_data/benchmark.csv`.
The status bar shows `1›benchmark…u_active]1›benchmark|`, repeating`1›benchmark`. The `…u_active]` comes from `middleTruncate()` clipping in the middle of internal visidata markup.

This PR creates `clip_markup_middle()`, which removes entire substrings between markup sections, which could be inside markup, as in `'[:markup_start]removed_string[:]'` or `'[:markup_start]removed_string[/markup_end]'`, or outside markup sections, like `'[:]removed_string[:next_markup_start]'`.

### DIFF
--- a/visidata/cliptext.py
+++ b/visidata/cliptext.py
@@ -391,6 +391,66 @@ def clipstr_middle(s, n=10, truncator='…'):
         return res, dispwidth(res)
     return s, dispwidth(s)
 
+def clip_markup_middle(s:str, w:int):
+    '''takes a string *s* containing optional visidata markup, and returns a string
+    truncated to have a display width less than or equal to *w*, while preserving markup
+    for the remaining text. When text with markup is clipped, what is omitted is one or
+    more entire markup sections, between markup start and end delimiters:
+    [:markup] [:] [/markup] [/]
+    The dropped text is replaced with a single disp_truncator.
+    When text without markup is clipped, *clipstr_middle()* is used.
+    '''
+    trunch = options.disp_truncator
+
+    if w <= 0: return ''
+    if dispwidth(s) <= w:
+        return s
+    if w < dispwidth(trunch): return ''
+
+    markup_section_re = r'(\[.*?\].*?\[[/:].*?\])'  # [:whatever]text[:] or [:whatever]text[/anything]
+    if not re.match(internal_markup_re, s):
+        return clipstr_middle(s, w, truncator=options.disp_truncator)
+    # build the front half of the string
+    output = []
+    chunks_w = 0
+    truncated = False
+    chunks = re.split(markup_section_re, s)
+    for i, chunk in enumerate(chunks):  #chunks are either regular text, or marked up section:  start, text, end
+        parts = re.split(internal_markup_re, chunk)
+        if len(parts) == 1:      #text with no markup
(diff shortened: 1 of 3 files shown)


saulpw (approved):
Another banger, @midichef!  It's a lot of code but I can see that dealing with the markup is finicky.  Thanks for putting in the time (and the tests).

visidata/statusbar.py
```diff
@@ -165,6 +160,7 @@ def leftStatus(sheet):
     return sheet.formatString(sheet.options.disp_status_fmt)
 
 
+import re
```
saulpw: extra import

visidata/statusbar.py
```diff
@@ -8,7 +8,7 @@
 import sys
 
 import visidata
-from visidata import vd, VisiData, BaseSheet, Sheet, ColumnItem, Column, RowColorizer, options, colors, wrmap, clipdraw, ExpectedException, update_attr, dispwidth, ColorAttr, clipstr_middle
+from visidata import vd, VisiData, BaseSheet, Sheet, ColumnItem, Column, RowColorizer, options, colors, wrmap, clipdraw, ExpectedException, update_attr, dispwidth, ColorAttr, clipstr_middle, iterchunks, clip_markup_middle
```
saulpw: do we still need iterchunks?

visidata/cliptext.py
```diff
@@ -391,6 +391,66 @@ def clipstr_middle(s, n=10, truncator='…'):
...
+
+    markup_section_re = r'(\[.*?\].*?\[[/:].*?\])'  # [:whatever]text[:] or [:whatever]text[/anything]
+    if not re.match(internal_markup_re, s):
+        return clipstr_middle(s, w, truncator=options.disp_truncator)
+    # build the front half of the string
+    output = []
+    chunks_w = 0
+    truncated = False
+    chunks = re.split(markup_section_re, s)
+    for i, chunk in enumerate(chunks):  #chunks are either regular text, or marked up section:  start, text, end
```
saulpw, line 419: can we use iterchunks for this?  So the parsing all happens in one place.

visidata/statusbar.py
midichef: no, I'll take out the imports.

visidata/cliptext.py
midichef: no, because `iterchunks` chops the string up at every markup boundary `'[:markup]`  `[:]`  `[/]`, and for the statusbar we try a bit to preserve the structure of sections.

This only works now because tags are not currently nested in the statusbar. If they become nested, like `'[:tag1]text[:tag2]and[:]more text[:]', `clip_markup-middle()` will produce invalid markup, by treating `'[:tag1]text[:tag2]` as a section.

My main goal here is to make the minimal needed patch to the statusbar so I can open a dozen sheets and see the newest one in the statusbar. Right now, in that scenario the rightmost sheet titles don't fit on screen.

PR #2924, opened 2025-12-27
[join-] warn on concat of sheets with unequal # of cols

midichef:
This will warn users when they do `concat` joins of sheets with different numbers of columns. I just caught myself making such a mistake when assembling a dataset, which caused my column data to get mixed up.


saulpw (approved):
Okay, it's simple enough, if it helps someone not be surprised.  Should it be a `fail` though?

midichef:
Okay I changed it to `fail` instead of `warning`. If the `fail` message isn't right, please feel free to change it.

PR #2926, opened 2025-12-28
[loaders-] update conll loader to work with pyconll v4.0

### LINKED ISSUE #2916: [conllu] load test broken on python 3.14
**Small description**

The Python 3.14 build is broken with apparently no error output: https://github.com/saulpw/visidata/actions/runs/20319739634/job/58372499003

Running the `load-conllu.vdj` test with Python 3.14 locally gets an error:

`AttributeError: module 'pyconll' has no attribute 'load'`

It looks like the 4.0 release a few weeks ago is for Python 3.14, and possibly the interface has changed with this new major version.

midichef:
Fixes #2916. Tested CoNNL loading using `tests/load-conllu.vdj` in Python 3.14.2, with the older pyconll 3.3.1 and the new pyconll 4.0.0.

### DIFF
--- a/visidata/loaders/conll.py
+++ b/visidata/loaders/conll.py
@@ -36,14 +36,27 @@ class ConllSheet(TableSheet):
         ItemColumn('misc', 10, type=dict),
     ]
     def iterload(self):
+        # relied on official pyconll guidance for migrating v3 to v4
+        # https://pyconll.readthedocs.io/en/stable/migration.html
         pyconll = vd.importExternal('pyconll')
+        try: #succeeds for pyconll version >= 4
+            import pyconll.conllu
+            v4 = True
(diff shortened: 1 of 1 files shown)


midichef:
I changed the use of `fp` to `self.source.given`, because in pyconll v4.0, `iter_from_file` would not take `fp`, a `FileProgress` object. It only takes `str`, `bytes`, or `PathLike`, and `self.source.given` is a string (EDITED to add:) or `PathLike`.

saulpw:
Thanks for fixing this (hopefully).  I was going to move the loader into experimental as a first deprecation step.  I wonder if anyone besides @polm uses this?  In any event I think we can take your changes to stave off the bitrot awhile longer.

PR #2927, opened 2025-12-29
Jump to first row and column with 'go'

daviewales:
Same as Vim.

Related to #2686.

### DIFF
--- a/visidata/experimental/vimcompat.py
+++ b/visidata/experimental/vimcompat.py
@@ -1,0 +1,9 @@
+"""Additional vim-like bindings for Visidata
+
+Enable in .visidatarc with `import visidata.experimental.vimcompat`.
+"""
+
+from visidata import BaseSheet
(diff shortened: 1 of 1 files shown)


saulpw:
Thanks for the suggestion, @daviewales.  There is already an unbound `go-home` command which I believe is what you're looking for.  I'm not sure we should bind it to `go` by default, since `o` is bound to `open` commands (and `go` is used for `open-rows` in the scraper, at least).

I propose that we add a `experimental/vimcompat.py` with this binding and a few others that are off by default. Then you could add `import visidata.experimental.vimcompat` to your visidatarc and get this behavior.

daviewales:
Thanks @saulpw. I like the `vimcompat` idea. I've pushed a new version.

saulpw (approved):
Thanks for putting this together, @daviewales !

PR #2929, opened 2026-01-02
[join-] concat: store src sheet names in a hidden col

midichef:
Here's a prototype of a feature for convenience. When joining sheets with `concat`, add a hidden column with the name of the source sheet.

Should we do the same for `full`, `append`, and `merge`?

### DIFF
--- a/visidata/features/join.py
+++ b/visidata/features/join.py
@@ -3,7 +3,7 @@ import itertools
 import functools
 from copy import copy
 
-from visidata import vd, VisiData, asyncthread, Sheet, Progress, IndexSheet, Column, CellColorizer, ColumnItem, SubColumnItem, TypedWrapper, ColumnsSheet, AttrDict, dispwidth
+from visidata import vd, VisiData, asyncthread, Sheet, Progress, IndexSheet, Column, CellColorizer, ColumnItem, SubColumnItem, TypedWrapper, ColumnsSheet, AttrDict, dispwidth, SettableColumn
 
 vd.help_join = '# Join Help\nHELPTODO'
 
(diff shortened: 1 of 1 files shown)


saulpw:
Isn't there already a hidden column `origin_sheet` on ConcatSheet with the source sheet in it?

frosencrantz:
It looks like `append` has it, and I use it all the time with `append` ConcatSheets.  It looks like `full` and `merge` sheets do not.

midichef:
Okay, so `append` gets a `origin_sheet`, but `concat`, `full`, `merge` and `diff` don't. I think that column should be added to all of them.

Also, I just noticed that joining a sheet with `full` or `diff` fails when there are no key columns on any sheet:
https://github.com/saulpw/visidata/blob/57ff6af16f5dcf67889a46d7cd8b3b274fc9a35e/visidata/features/join.py#L75-L76
I'd like that join to succeed when all the joined sheets have 0 key columns.

Should I implement both of these?

saulpw:
Thanks, @midichef, an implementation would be appreciated.  Ideally it would work like it does for concat, with the sheet object being stored in the underlying row data structure, and a hidden column providing the sheet object itself (which will display as its name).

I'm not sure what origin_sheet would mean for those other join types (like `full`), which pull data from multiple sheets.  Do you think it should be `origin_sheets` in those cases, and expose an array of all sheets the data came from?

saulbert:
Thanks for the prototype, @midichef! We've implemented this on develop with a slightly different approach — wrapping concat rows as `(srcSheet, srcRow)` tuples with `ConcatColumn`, same pattern as `ConcatSheet` (append). The hidden column is named `origin_sheet` for consistency.

For the other join types (`full`, `merge`, `diff`), the design is less clear since rows can come from multiple sheets, so we're leaving that for a follow-up.

Closing as the feature is now on develop. Thanks for the idea!

[comment by Claude Opus 4.6 and approved by @saulpw]

PR #2941, opened 2026-01-12
fix width and coordinate calculations for data containing markup

midichef:
There are a lot of problems in using `dispwidth()` with strings that contain text like `[:text]` because it gets interpreted as visidata markup. This PR fixes a few issues:

1d9338b75226c0cc6415ecbc2f212461eff86bd5 Fixes a bug I introduced starting in visidata 3.2:  the cursor goes to the wrong location in a pseudo-markup string. Like if you input an expression with `=` `colname[:5]`, the cursor goes to the right of the `e`, as if the `[:5]` does not exist.

7f1762157ddb7fbfcc6f4e8bd2c689533e40e04b Fixes a longstanding problem showing any column name with `[:text]`. To demonstrate, `vd sample_data/benchmark.csv`:  press `=` `Customer[:5]`, the new column's name is shown as `Customer`.

00f2e2ed1560937f74b15c8a8e7920ed4f7fa69d Fixes data truncation in fixed width files containing `[:text]`
For example, if you load this data as `vd -f fixed`
```
reactor_instructions
if temp sensors show 1500 K[:IMPORTANT:for over 10 MINUTES total in any 24 HOUR period]
continue operating reactors[:IMPORTANT:with POWER LIMITED to 5% USING DANGER PROTOCOLS]
```
the data shown on screen hides the `[` and what follows it, and if you save the data as `instructions.fixed`, you get:
```
reactor_instructions        
if temp sensors show 1500 K 
continue operating reactors 
```

These are probably not the only such problems. We should do an audit of all uses of `dispwidth`, and `clip*`.

### DIFF
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -196,32 +196,32 @@ class InputWidget:
             '''Return a formatted substring of *dispval* that fills the on-screen width *w*.'''
             if i == len(dispval): # add a fillchar so the user perceives room to type
                 dispval += self.fillchar
-            dw = dispwidth(dispval)
+            dw = dispwidth(dispval, literal=True)
             if dw <= w:  # entire value fits
                 dispval += self.fillchar*(w-dw)
                 return dispval, i
             if w <= tr_w: # column is too narrow to hold a left and right truncation
                 return trunch, 0
 
-            dw = dispwidth(dispval[i:])
+            dw = dispwidth(dispval[i:], literal=True)
             if dw + tr_w <= w and dw <= w//2: #cursor is within half-colwidth of end
                 #truncate the left and show the end
-                frag, n = clipstr_start(dispval, w-tr_w)
+                frag, n = clipstr_start(dispval, w-tr_w, literal=True)
                 offset = len(dispval) - i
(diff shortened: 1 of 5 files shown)


midichef:
The `edit-cell` change is still a work in progress, because I just noticed [1d9338b](https://github.com/saulpw/visidata/pull/2941/commits/1d9338b75226c0cc6415ecbc2f212461eff86bd5) introduced a new display bug. So I'll keep working on it.

midichef:
Okay, this PR is ready. The previous bug just fixed, was due to adding `literal=True` to `dispwidth()` while forgetting to also add it to `clipstr_start()`.

saulpw:
Thanks for fixing these, @midichef.

midichef:
With #3074, I believe all the remaining uses of `dispwidth` and `clipdraw*` handle markup properly. They all leave the value of `literal` unspecified, falling back to the default of `literal=False`. In each case, it is fine because either the input has markup, and `literal=False` is necessary, or the input definitely cannot have markup, and `literal=True` is unnecessary because it won't change the output. Though `literal=True` will generally be slightly faster.

PR #2942, opened 2026-01-13
[loaders-] update airtable loader to work with latest pyairtable library

mplattner:
The existing airtable loaders seems to be compatible with pyairtable@1.x only.
This PR supports the latest pyairtable version.


saulpw:
Thanks @mplattner !  This seems good to me.  Do you use the airtable loader yourself?

mplattner:
Only very infrequently, like two times during the last 12 months.

PR #2953, opened 2026-01-26
[canvas-] delete in linear time instead of quadratic

midichef:
Deleting large numbers of points takes significant time. On my slow system, deleting 30k points takes 18 seconds, deleting 10k points takes about 2. To test: `yes '5.0' | head -n 30000 |vd` and plot with `%` `!` `.` then delete the single point with `d`

I considered caching `self.source` as a local variable `src` but I suspect it probably doesn't make much difference.


saulpw (approved):

saulpw:
Thanks @midichef !  This makes sense to me.

PR #2955, opened 2026-01-28
[dedupe-] enable custom sheet name suffixes

midichef:
This would let me override the default `dedupe-rows`, to give the resulting sheet a suffix other than `_deduped`. Like I did in  https://github.com/saulpw/visidata/discussions/2836 for `dup-selected`, `transpose`, and `dup-rows`.

I know I may be the only person who will ever use this feature...


saulpw:
That's fine, it doesn't even add a single line of code :)

PR #2959, opened 2026-01-31
[palette-] prevent interpretation of markup in go-col-name

midichef:
The palette choosers for `&` `+` and `gc` have some minor display inconsistencies with the way longnames are displayed in `exec-longname`. The first two commits in this PR fix those inconsistencies.

The third commit fixes a bug in `gc` where a column name had markup interpreted, so 'Customer[:5]' was shown as `Customer`, and cell contents had markup interpreted, so a cell of `[:bold]asdf[/]` was drawn as `asdf` in bold. However, to properly display the column name as a literal with no markup, I had to turn off the bolding of the column name that was just added in the first commit (https://github.com/saulpw/visidata/commit/40d4243b5706013685a19f2c6be4e25e6681720c).

### DIFF
--- a/visidata/aggregators.py
+++ b/visidata/aggregators.py
@@ -365,7 +365,7 @@ def chooseAggregators(vd, prompt = 'choose aggregators: '):
     def _fmt_aggr_summary(match, row, trigger_key):
         formatted_aggrname = match.formatted.get('key', row.key) if match else row.key
         r = ' '*(dispwidth(prompt)-3)
-        r += f'[:keystrokes]{trigger_key}[/]  ' if trigger_key else '   '
+        r += f' [:keystrokes]{trigger_key}[/] ' if trigger_key else '   '
         r += f'[:bold]{formatted_aggrname}[/]'
         if row.desc:
             r += ' - '
--- a/visidata/aggregators.py
+++ b/visidata/aggregators.py
@@ -366,7 +366,7 @@ def chooseAggregators(vd, prompt = 'choose aggregators: '):
         formatted_aggrname = match.formatted.get('key', row.key) if match else row.key
         r = ' '*(dispwidth(prompt)-3)
         r += f'[:keystrokes]{trigger_key}[/]  ' if trigger_key else '   '
-        r += formatted_aggrname
+        r += f'[:bold]{formatted_aggrname}[/]'
         if row.desc:
             r += ' - '
             r += match.formatted.get('desc', row.desc) if match else row.desc
--- a/visidata/cliptext.py
+++ b/visidata/cliptext.py
@@ -354,44 +354,46 @@ def wraptext(text, width=80, indent=''):
         active_tags = line_tags
 
         textchunks = [x for x in chunks if not is_vdcode(x)]
-        if ''.join(textchunks) == '':  #for markup with no contents, like '[:tag][/]' or '[:]' or '[/]'
+        linetext = ''.join(textchunks)
+        if linetext == '':  #for markup with no contents, like '[:tag][/]' or '[:]' or '[/]'
             yield '', ''
             continue
         # textwrap.wrap does not handle variable-width characters  #2416
-        for linenum, textline in enumerate(textwrap.wrap(''.join(textchunks), width=width, drop_whitespace=False)):
-            txt = textline
-            r = ''
+        for linenum, wrapped_textline in enumerate(textwrap.wrap(linetext, width=width, drop_whitespace=False)):
+            wrapped_remainder = wrapped_textline
+            marked_up = ''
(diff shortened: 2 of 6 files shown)


midichef:
Hmm. Turning off formatting in `go-col-name` was a bad idea. Now it fails to interpret the coloring of the match. So for a colname like `Customer`. When I type `C` I see:  `[:match]C[/]ustomer`. So we have to trade off between:  coloring matches, vs. correctness for column names that contain pseudomarkup like `[:text]`.

What would be a good way to escape visidata markup?

midichef:
Perhaps we can escape vd markup by wrapping every open bracket `[` between some no-op markup that has no color bound to it, like: `[:literal]` and `[:]`. I think it should work, judging by the working of `iterchunks()` and `internal_markup_re`. But the idea needs to be double-checked for validity, in case I have missed any important case.

```
def escape_vdcode(s:str) -> str:
    return re.sub(r'\[', '[:literal][[:]', s)

def unescape_vdcode(s:str) -> str:
    return re.sub(r'\[:literal\]\[\[:\]', '[', s)
```

If this works, I would discard the changes in https://github.com/saulpw/visidata/pull/2959/commits/f19d7ba722627706548a7cf50c0e66824c4e38aa and change `go-col-name` to escape the colname and the cell contents.

saulpw:
Oof, that escape might work, but it feels like it's overmuch for the task at hand.

What if we injected non-displaying ASCII (or unicode) characters to mark the start/end of literal text?  Like this from
https://en.wikipedia.org/wiki/Unicode_control_characters:

> Three formatting characters provide support for [interlinear annotation](https://en.wikipedia.org/wiki/Interlinear_gloss) (U+FFF9 INTERLINEAR ANNOTATION ANCHOR, U+FFFA INTERLINEAR ANNOTATION SEPARATOR, U+FFFB INTERLINEAR ANNOTATION TERMINATOR). This may be used for providing notes that would typically be displayed between the lines of other text.

This 'literal' usage would be the exact opposite of its intent, but otherwise these characters seem to have the right vibe.  Otherwise we could coopt some underused ASCII control characters like 14/15 (Shift Out/Shift In).  What do you think about this approach?

midichef:
Sure, using the Unicode interlinear annotation characters seems like a good solution. The resulting escaped strings will be shorter.

However, since it requires changing `internal_markup_re`, as well as `wraptext()` and `clipstr_middle()` that make use of that regex. So I'll put together a set of changes. For the `wraptext()` change in particular, I may wait till https://github.com/saulpw/visidata/pull/2991 is merged, to avoid a conflict.

midichef:
What's the definition of which tags close markup? Is it that `[/]` closes the last tag, and `[:]` closes all the tags, as in `iterchunks()`?

https://github.com/saulpw/visidata/blob/d1345bc970c1c18863be64d4fe9543478353891d/visidata/cliptext.py#L76
https://github.com/saulpw/visidata/blob/d1345bc970c1c18863be64d4fe9543478353891d/visidata/cliptext.py#L80

Because a2b64568 only considers `[/]` but it doesn't realize that `[:]` closes tags:
https://github.com/saulpw/visidata/blob/d1345bc970c1c18863be64d4fe9543478353891d/visidata/cliptext.py#L334
https://github.com/saulpw/visidata/blob/d1345bc970c1c18863be64d4fe9543478353891d/visidata/cliptext.py#L371

It seems the AI needs this markup convention spelled out for it more explicitly.

midichef:
I'm going to take the `iterchunks()` behavior as authoritative, and patch `wraptext()` to match it in handling `[:]`.

midichef:
I implemented markup escaping, using the Unicode interlinear annotation characters `'\uFFF9[:markup]...[/]\uFFFb'`.

`wraptext()` was too opaque for me, so I renamed a lot of variables and added clarifying comments. 

One missing feature in the implementation of `wraptext()` is in wrapping text across multiple lines, it won't reopen markup-escaping tags like `\uFFF9` the way it does for tags like `[:bold]` . That seemed too complicated to implement.

saulpw:
Thank you, @midichef, I'm sorry I missed your question here.  I think you chose correctly.  The clanker found a few small issues which it cleaned up, and I'm ready to merge it if you think it's ready too.  Thanks again for putting this together, it's a good find.

saulpw (approved):

midichef:
Nice catches by the AI! That last commit, about not clearing the attr stack, shows good understanding of what this change is intended to do.

I forgot to modify `clipstr_middle()` to handle escaped markup, so I'll do that very soon.

midichef:
Okay this is ready to go.

saulpw:
Thanks for doing this!

PR #2961, opened 2026-02-02
[freqtbl-] add commands to delete rows on source sheet

midichef:
This was a feature request in https://github.com/saulpw/visidata/discussions/2950 by @Ankit-Kulkarni.

For deleting rows I used `self.deleteBy(lambda r: self.rowid(r) == rid)`. Can I replace it with the slightly faster `self.deleteBy(lambda r, freqtbl_row=freqtbl_row: r is freqtblq_row)`? I'm not sure.

### DIFF
--- a/visidata/freqtbl.py
+++ b/visidata/freqtbl.py
@@ -162,6 +162,21 @@ Each row on this sheet corresponds to a *bin* of rows on the source sheet that h
     def openCell(self, col, row):
         return Sheet.openCell(self, col, row)
 
+    @asyncthread
+    def deleteSourceRows(self, freqtbl_rows):
+        src = self.source
+        src.addUndoSelection()
+        for freqtbl_row in freqtbl_rows:
+            srows = freqtbl_row.sourcerows
+            srowids = dict.fromkeys([src.rowid(r) for r in srows], True)
+            src.deleteBy(lambda r,srowids=srowids,src=src: src.rowid(r) in srowids)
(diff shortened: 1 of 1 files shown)


saulpw:
I proposed [a workflow in #2950](https://github.com/saulpw/visidata/discussions/2950#discussioncomment-15688192) that has a couple more steps, but doesn't require additional commands or override existing bindings.  Do you think we need/want specialized commands to do this?

Ankit-Kulkarni:
@saulpw  as a everyday regular user of vd, I have been using these commands from a few days now and honestly while the other flows still work(i was using the same earlier ) this approach is just blazing fast (i.e less commands to remember and navigate through).

Mentally its just super easy zd and thats it , no inverse selection , no select and delete. The problem with such select flows is that if you end up selecting wrong you will have to undo or carefully unselect , basically chances of having more operations. 

Unless adding commands is not an overhead or can create an issue , I really thing having this as default would definitely help. 

@midichef

saulbert:
Thanks for working on this! The inverse-filter use case from #2950 is a good one.

This has been addressed differently on develop — instead of deletion (which is destructive and has tricky undo semantics from a derived sheet), `zEnter` and `gzEnter` on FreqTableSheet now open a copy of the source sheet with all rows *except* those in the current/selected bins. This mirrors the existing `Enter`/`gEnter` pattern and is non-destructive.

Closing in favor of that approach. Thanks again for the motivation!

PR #2963, opened 2026-02-03
Fix three bugs in pandas loader (#2960)

### LINKED ISSUE #2960: Pandas loader error when infering column type
**DISCLAIMER**
After filing the 1st bug, I found two others. As they are all related to `_pandas.py` I amended my report.
`Codex` AI provided a fixes for all three problems that I encounterd. The fixes seem reasonable to me, but I am far from understanding the whole problems.

I have a bad feeling fixing code errors with AI and not completely understanding the solutions, but maybe this is how life is these days... BTW, all three fixes suggested by the AI worked.

I am not really into git pull/push workflow, so I hope you do not mind me pasting code here. I tried to be concise in my description.

My original idea was to write a loader for a ctx-file (`AqBanking` file format): file should be parsed into a pandas DataFrame, then I would construct a `Sheet` from DataFrame.

Having said this, here is my description:
**Small description**

I have this minimal config file `config_bug_PandasSheet.py`:
```python
import pandas as pd
vd.addGlobals({'pd': pd})


@VisiData.api
def open_ctx(vd, p):
    df = pd.DataFrame({"a":[1,2], "b":[3,4], "c":["x","y"]})
    from visidata.loaders._pandas import PandasSheet
    return PandasSheet(p.base_stem, source=df)
```

which implements a loader for filetype `ctx` (filetype does not matter for this purpose). In shell, I run

```bash
$ touch a.ctx
$ ./bin/venv/bin/vd --config=./mybanking/visidata_config/config_bug_PandasSheet.py a.ctx
```

I get this error:

Image

**FIX**

Consider this code snippet from `loaders/_pandas.py` in method `reload(self)` around line 189.
`type` is infered from `df[col]` (a `Series`), while it should be infered from `df[col].dtype` (?).
I added the two comments, the latter of which might provide a fix for the problem:

```python
for col in (c for c in df.columns if not c.startswith("__vd_")):
            self.addColumn(Column(
                col,
                # FIX: infer type from column of df, not from Series object, like this
                # type=self.dtype_to_type(df[col].dtype),
                type=self.dtype_to_type(df[col]),
                getter=self.getValue,
                setter=self.setValue,
                expr=col
            ))
```

The error disappears, if infering the type from `df[col].dtype` instead of `df[col]`:

Image

**Another bug: rows don't get selected by pressing `s`**

Moreover, you cannot select rows in this Sheet by pressing `s` (before and after the above fix), so I consider this another bug... Again, `Codex` provided a fix with another explanation, which again is some distance away from my understanding:

Image

**Yet another bug: only working duplication command is `dup-selected`**

After selection worked, I tried to duplicate the selection, but I didn't work, only for `dup-selected`. `Codex` provided yet another fix (?), here is the conversation:

Image

Image

```python
diff --git a//home/info/98_banking/bin/visidata/visidata/loaders/_pandas.py b//home/info/98_banking/bin/visidata/visidata/loaders/_pandas.py
--- a//home/info/98_banking/bin/visidata/visidata/loaders/_pandas.py
+++ b//home/info/98_banking/bin/visidata/visidata/loaders/_pandas.py
@@ -391,2 +391,5 @@
 PandasSheet.addCommand(None, 'unselect-rows', 'unselectByIndex()', 'unselect all rows')
+PandasSheet.addCommand('g"', 'dup-rows', 'vs=PandasSheet(sheet.name+"_copy", "copy", source=sheet.df); vd.push(vs)', 'open a duplicate sheet with all rows')
+PandasSheet.addCommand('z"', 'dup-selected-deep', 'vs=PandasSheet(sheet.name+"_selecteddeepcopy", "selecteddeepcopy", source=selectedRows.df.copy(deep=True)); vd.push(vs); status("pushed sheet with deepcopy of selected rows")', 'open duplicate sheet with deepcopy of selected rows')
+PandasSheet.addCommand('gz"', 'dup-rows-deep', 'vs=PandasSheet(sheet.name+"_deepcopy", "deepcopy", source=sheet.df.copy(deep=True)); vd.push(vs); status("pushed sheet with deepcopy of all rows")', 'open duplicate sheet with deepcopy of all rows')
```

Image

After this `Codex`-suggested fix, I all duplication worked.

Thank you!

saulpw:
## Summary
This PR fixes three bugs in the pandas loader (`_pandas.py`) as reported in issue #2960:

1. **Column type inference bug**: Fixed incorrect type inference by passing `df[col].dtype` instead of `df[col]` (the entire Series object) to `dtype_to_type()` (line 189)

2. **Row selection failure**: Added command overrides for basic selection operations (`s`, `u`, `t`) to ensure they work properly with PandasSheet's custom selection mechanism and include undo support

3. **Missing duplication commands**: Added three new duplication commands that were previously missing:
   - `g"` - duplicate all rows
   - `z"` - deep copy selected rows  
   - `gz"` - deep copy all rows

## Test plan
- [ ] Verify column types are correctly inferred from pandas DataFrames
- [ ] Test that `s` key now works to select rows in PandasSheet
- [ ] Test that `u` and `t` keys work for unselecting and toggling row selection
- [ ] Test all duplication commands: `"`, `g"`, `z"`, `gz"`

## Related issue
Fixes #2960  
Reported by @mweich

### DIFF
--- a/visidata/loaders/_pandas.py
+++ b/visidata/loaders/_pandas.py
@@ -186,7 +186,7 @@ class PandasSheet(Sheet):
         for col in (c for c in df.columns if not c.startswith("__vd_")):
             self.addColumn(Column(
                 col,
-                type=self.dtype_to_type(df[col]),
+                type=self.dtype_to_type(df[col].dtype),
                 getter=self.getValue,
                 setter=self.setValue,
                 expr=col
@@ -384,6 +384,11 @@ def view_pandas(vd, df):
     run(PandasSheet('', source=df))
 
 
+# Override basic selection commands to work with PandasSheet's selection mechanism
+PandasSheet.addCommand('s', 'select-row', 'addUndoSelection(); selectRow(cursorRow)', 'select current row')
+PandasSheet.addCommand('u', 'unselect-row', 'addUndoSelection(); unselectRow(cursorRow)', 'unselect current row')
+PandasSheet.addCommand('t', 'stoggle-row', 'addUndoSelection(); toggle([cursorRow])', 'toggle selection of current row')
+
 # Override with vectorized implementations
 PandasSheet.addCommand(None, 'stoggle-rows', 'toggleByIndex()', 'toggle selection of all rows')
 PandasSheet.addCommand(None, 'select-rows', 'selectByIndex()', 'select all rows')
@@ -406,6 +411,9 @@ PandasSheet.addCommand('g\\', 'unselect-cols-regex', 'selectByRegex(regex=inputR
 
 # Override with a pandas/dataframe-aware implementation
 PandasSheet.addCommand('"', 'dup-selected', 'vs=PandasSheet(sheet.name, "selectedref", source=selectedRows.df); vd.push(vs)', 'open duplicate sheet with only selected rows')
+PandasSheet.addCommand('g"', 'dup-rows', 'vs=PandasSheet(sheet.name+"_copy", "copy", source=sheet.df); vd.push(vs)', 'open duplicate sheet with all rows')
+PandasSheet.addCommand('z"', 'dup-selected-deep', 'vs=PandasSheet(sheet.name+"_selecteddeepcopy", "selecteddeepcopy", source=selectedRows.df.copy(deep=True)); vd.push(vs)', 'open duplicate sheet with deepcopy of selected rows')
+PandasSheet.addCommand('gz"', 'dup-rows-deep', 'vs=PandasSheet(sheet.name+"_deepcopy", "deepcopy", source=sheet.df.copy(deep=True)); vd.push(vs)', 'open duplicate sheet with deepcopy of all rows')
 
 vd.addGlobals({
     'PandasSheet': PandasSheet,
--- a/visidata/loaders/_pandas.py
+++ b/visidata/loaders/_pandas.py
@@ -411,9 +411,9 @@ PandasSheet.addCommand('g\\', 'unselect-cols-regex', 'selectByRegex(regex=inputR
 
 # Override with a pandas/dataframe-aware implementation
 PandasSheet.addCommand('"', 'dup-selected', 'vs=PandasSheet(sheet.name, "selectedref", source=selectedRows.df); vd.push(vs)', 'open duplicate sheet with only selected rows')
-PandasSheet.addCommand('g"', 'dup-rows', 'vs=PandasSheet(sheet.name+"_copy", "copy", source=sheet.df); vd.push(vs)', 'open duplicate sheet with all rows')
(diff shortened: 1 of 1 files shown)


saulpw:
@weichm This is the PR that Claude pulled together from your issue.  I made a slight modification to the sheet names which had duplicate parts, but otherwise it looks reasonable to me.  To be completely honest, I've spent less time in `_pandas.py` than other parts of the code, so since it seems to work for you, we'll merge it without too much ado.

saulpw:
It is failing a test though (ignore the xpt test, which was failing previously already): https://github.com/saulpw/visidata/actions/runs/21650204358/job/62412564603?pr=2963

weichm:
Thanks for the quick reply. I think I found a fix and created a pull request (my first one ever): https://github.com/saulpw/visidata/pull/2966

If you don't mind:
I am planning to create another pull request for a feature I would like to have (editing db views in visidata). Should I do something different next time?

visidata/loaders/_pandas.py
```diff
@@ -384,6 +384,11 @@ def view_pandas(vd, df):
     run(PandasSheet('', source=df))
 
 
+# Override basic selection commands to work with PandasSheet's selection mechanism
+PandasSheet.addCommand('s', 'select-row', 'addUndoSelection(); selectRow(cursorRow)', 'select current row')
+PandasSheet.addCommand('u', 'unselect-row', 'addUndoSelection(); unselectRow(cursorRow)', 'unselect current row')
+PandasSheet.addCommand('t', 'stoggle-row', 'addUndoSelection(); toggle([cursorRow])', 'toggle selection of current row')
```
midichef: (I am redoing this comment to apply it to a group of lines instead of a single line.)

The AI's implementation of these sheet selection commands for `PandasSheet` is:
1) unnecessary, because they only do what the default `Sheet` commands already do; they are also
2) broken, because the cursor doesn't move afterward. The new overriding commands are missing the `cursorDown(1)` in the default versions.

I can't recommend better changes to the current Pandas row selection mechanism, because I don't understand it, and I don't use Pandas, so I can't test changes thoroughly. And the current version of Claude seems to share my limitations.

To better get a sense of the kinds of errors it may make in Visidata: a third flaw in these commands is the use of `toggle([cursorRow])`, which is a worse choice than `toggle_row(cursorRow)` used by the existing `Sheet` default `stoggle-row`. The new `toggle()` version unnecessarily starts a new `@asyncthread`.

Row selection is tricky code, with a history of subtle flaws in non-standard sheets like `FreqTableSheet` (#2346 #2759), so I think we should be careful in taking AI changes to it. Though `PandasSheet` is probably not as tricky as `FreqTableSheet`.

saulpw:
> I am planning to create another pull request for a feature I would like to have (editing db views in visidata). Should I do something different next time?

Well I created this PR just to get a visual reference and comment on the actual changes. In the future, create a proper PR rather than an issue with code blocks (and make sure you review the PR before submitting it).  But from more recent PR activity I think you're doing this!  So no worries.

Issue #2947, opened 2026-01-17
Feature request: querying LLMs with visidata cells input


GnarlyMshtep:
### Feature request: querying LLMs with visidata cells as input  
I often work with visidata to look at sheets which have lots of unstructured data (text). Currently, if I try to find pattern in the text, I would try to narrow my search with a combination of Regex search and manual inspection. 

I think could be really useful if I could pick some number of cells e.g. within a certain column (the usual way they are selected in `visidata`) and forward them to an LLM (e.g. rendered as `json` and preceded by the additional query).  

I am not sure how ambitious this but would be curious to hear from the team! To me this seems like a versatile command that will only be getting more useful.   

Huge `visidata` fan! Thank you for the consideration. 
Matan

saulpw:
Thanks  @GnarlyMshtep!  I've been thinking about something like this idea myself.  Can you provide a bit more context for the exact workflow?  Like what would the "additional query" look like, and how would you read/parse the results?

yarikoptic:
"Exactly the same" question came up in the scope of our fresh https://github.com/ReproNim/ABCD-visidata . In our case we have a column (or columns) with meaningful information which we would love to make LLM access/query to answer the question on which of those rows out of 90k of them are of interest to answer the question (e.g. "What is the relationship between volume of amygdala and how much youth was smoking on mondays after having a morning jin & tonic?") -- then we would have those rows selected, I click 'zm' to get the table presenting the data and potentially LLM also takes an initial stab at analytics there

GnarlyMshtep:
Thanks @saulpw ! Sorry for the late response -- was working for a paper deadline. Let me try to throw some thoughts down: 

### Use case
The following is my most common `visidata` workflow: 

I just generated a bunch of LM trajectories, along with traces I generated some summary statistics, and something feels unexpected (e.g. the task success rate is unexpectedly low). I logged all the results in a structured `.jsonl` file (`dataclass -> json`) and I am using `vd` to navigate the trace and 'vibe check' what could be going on. To do that, I might filter / sort by a certain statistic and then read a few trajectories. 

This workflow is nice because I am reading exactly the text inputted and outputted by the LMs, but its not scalable. Sometime I would have a specific question which I would want to ask on all the trajectories, like, "did the LM do X?", especially where the behavior X is rare.  It would be nice if I could just select to run all the text from a column on this question from inside visidata and get an answer. 

[...]

PR #2964, opened 2026-02-04
[llm] add Anthropic Claude integration for column analysis  #2947

saulpw:
Implements #2947 .     
    
    Implements LLM-powered analysis of column data with natural language queries.
    Users can select cells/columns and analyze them using Claude, with results
    returned as score, analysis text, raw response, and token count columns.
    
    - Sequential processing with configurable rate limiting
    - Async execution with progress tracking
    - Custom template support via file or built-in default
    - Comprehensive error handling and statistics reporting
    - API key via ANTHROPIC_API_KEY environment variable only

### DIFF
--- a/visidata/experimental/llm.py
+++ b/visidata/experimental/llm.py
@@ -1,0 +1,211 @@
+'''
+# LLM Analysis Feature
+
+Analyze column data using Anthropic Claude LLM with natural language queries.
+
+## Usage
+
+1. Position cursor on a column containing text data
+2. Run `llm-analyze-column` command (or `llm-analyze-selected` for selected rows only)
+3. Enter your analysis query (e.g. "Did the model exhibit reasoning?")
+4. Four new columns will be added with results:
+   - `{colname}_llm_score` (0-1 float score)
+   - `{colname}_llm_analysis` (text explanation)
+   - `{colname}_llm_raw` (hidden raw JSON response)
+   - `{colname}_llm_tokens` (token count for cost tracking)
+
+## Configuration
+
+Set in ~/.visidatarc or via environment variables:
+
+    options.llm_model = "claude-haiku-4-5"
+    options.llm_max_tokens = 1024
+    options.llm_temperature = 1.0
+    options.llm_rate_delay = 0.5  # seconds between requests
+    options.llm_template_path = ""  # path to custom template file
+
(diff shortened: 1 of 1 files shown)


dev/GIT.md
```diff
@@ -0,0 +1,82 @@
+# Git and Version Control Practices
```
yarikoptic, line 1: I am nobody to visidata (besides an excited and grateful user) but my 1c that for this PR to be reviewed/considered, worth cleaning it up to have **only** changes relevant to the PR. I would have jep that `visidata/experimental/llm.py` and likely some documentation to describe it -- the rest keep locally or propose as separate PRs

saulpw: They were already pushed, I just had to rebase the branch.  Good call keeping PRs clean though (even they are from me).

visidata/experimental/llm.py
```diff
@@ -0,0 +1,211 @@
...
+
+Provide:
+1. A score from 0 to 1 indicating how well the text matches or satisfies the query
+2. A brief analysis explaining your reasoning
+
+Response format: {{"score": <float 0-1>, "analysis": "<text>"}}'''
+
+
+@VisiData.lazy_property
+def anthropic_client(vd):
```
yarikoptic, line 67: oh -- even though I am an anthropic user, I wonder if it would make sense to centralize access via some other API/binding which would allow not only for anthropic?

one particular aspect I am thinking about is working with sensitive data which **must not leave the premises**. For that I would use some local models interfaced via ollama. 

and typically I see people connecting tools via https://openrouter.ai/ which I believe uses OpenAI API, (which is also the API exposed by open webui which e.g. my institution hosts as well)

So I think it would be great to support openAI API (while allowing for alternative server URL) and ollama interfaces in addition of Anthropic one.  May be smth like https://github.com/mozilla-ai/any-llm ?

saulpw: Sure, we can do that, but it should be a separate feature request after this is merged.  (And it should be made by someone who cares about that feature!  If we're all using Anthropic anyway, then I don't want to waste time or code or tokens on a hypothetical use-case.)   PRs (IMO) are for review to catch bugs, flaws, and to find a clean implementation of the proposed feature/fix.  They shouldn't be use to expand the scope of the feature, or we may debate about it endlessly :smiley:

PR #2966, opened 2026-02-04
Fix pandas loader selection/dup behavior

weichm:
A test did not run for a test on https://github.com/saulpw/visidata/pull/2963 for bug reported in #2960 by me.
This is pull request fixes the bug, I hope (a local test did not report the same error).

### DIFF
--- a/visidata/loaders/_pandas.py
+++ b/visidata/loaders/_pandas.py
@@ -212,7 +212,7 @@ class PandasSheet(Sheet):
 
     def _checkSelectedIndex(self):
         pd = vd.importExternal('pandas')
-        if self._selectedMask.index is not self.df.index:
+        if not self._selectedMask.index.equals(self.df.index):
             # DataFrame was modified inplace, so the selection is no longer valid
             vd.status('pd.DataFrame.index updated, clearing {} selected rows'
                       .format(self._selectedMask.sum()))
@@ -384,10 +384,38 @@ def view_pandas(vd, df):
     run(PandasSheet('', source=df))
 
 
-# Override basic selection commands to work with PandasSheet's selection mechanism
-PandasSheet.addCommand('s', 'select-row', 'addUndoSelection(); selectRow(cursorRow)', 'select current row')
-PandasSheet.addCommand('u', 'unselect-row', 'addUndoSelection(); unselectRow(cursorRow)', 'unselect current row')
-PandasSheet.addCommand('t', 'stoggle-row', 'addUndoSelection(); toggle([cursorRow])', 'toggle selection of current row')
+def _pandas_dup_with_selection(sheet, df, name_suffix, kind, select_all=False, status_msg=None):
+    vs = PandasSheet(sheet.name + name_suffix, kind, source=df)
+    positions = []
+    if not select_all:
+        try:
+            positions = [i for i, v in enumerate(sheet._selectedMask.values) if v]
+        except Exception:
+            positions = []
+
+    orig_reload = getattr(PandasSheet.reload, "__wrapped__", PandasSheet.reload)
+
+    def _reload_with_selection(vs=vs, orig_reload=orig_reload, positions=positions, select_all=select_all):
+        orig_reload(vs)
+        try:
+            if select_all:
+                if getattr(vs, '_selectedMask', None) is not None:
+                    vs._selectedMask.iloc[:] = True
+                return
+            if positions:
+                vs._selectedMask.iloc[positions] = True
+        except Exception:
+            try:
+                if positions:
+                    vs.selectByIdx(positions)
+            except Exception:
+                pass
+
+    vs.reload = asyncthread(_reload_with_selection)
+    if status_msg:
+        vd.status(status_msg)
+    return vs
+
 
 # Override with vectorized implementations
 PandasSheet.addCommand(None, 'stoggle-rows', 'toggleByIndex()', 'toggle selection of all rows')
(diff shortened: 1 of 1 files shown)


saulpw:
I assume this is vibe-coded?  I don't understand what the bug was, and what this fixes.  There's a lot of try/except blocks without any indication of which exceptions they're trying to avoid.

weichm:
If you mean by vibe-coded "stopped thinking too early and blindly posted what AI suggested as solution" you are right.
I have to apologize for this.

If I may take another try:
## New analysis
To analyze the problem by myself, I executed the failing test for dup-selected:
```bash
$ ../venv/bin/vd -N --overwrite=False --play tests/pandas_loader_dup_selected.vd --batch --output ./tests/golden/pandas_loader_dup_selected_output.tsv
```
There are no lines in `..._output.tsv`. That is why the test fails.

However, if I skip `--batch` and do:
```bash
$ ../venv/bin/vd -N --overwrite=False --play tests/pandas_loader_dup_selected.vd
```
then I get a `ValueError: The truth value of a Series is ambiguous. Use a.empty(), a.bool(), a.item(), a.any() or a.all()`

Yet another however:
If I try to mimick `pandas_loader_dup_selected.vd` and manually execute the commands by hand that are given in `pandas_loader_dup_selected.vd`:
- `$ ../venv/bin/vd -N sample_data/benchmark.csv`
- selecting three rows,
- doing a "dup-selected"

everything works as expected.

## Possible explanation and fix

Using AI, I could both track down the error to line 54 in `modify.py` and get a possible explanation of the different behaviour "executing test with --batch" vs. "doing it manually".
This is line 54 in `modify.py`:
```python
ColumnColorizer(9, 'color_readonly', lambda s,c,r,v: c and not r and c.readonly)
```
AI says:
"On a PandasSheet, row is a pandas Series, and pandas forbids boolean evaluation. ... r is whatever the sheet uses as its row object. For non‑pandas sheets that’s usually a plain Python object: a list, tuple, dict, namedlist, or a custom row type defined by the sheet. 
When drawing headers, r is None (that’s why that not r check existed)."

This reasoning results in this fix:
```python
ColumnColorizer(9, 'color_readonly', lambda s,c,r,v: c and (r is None) and c.readonly)
```

The failure does not happen when doing it manually because this seems to avoid the path because "the draw state and the command sequence differ". Sounds logic to me.
With or without an AI, I consider line 54 to be a bug and the fix valid.

## To summarize:

- Thanks for reading until here.
- Sorry for the hasty pull request. Please ignore it and maybe consider the fix from above.

midichef:
Congrats on opening this PR, which you said is your first ever! And signing the CLA via CLAassistant. I remember my puzzlement in my first PR here trying to work through that process.

The fix you have suggested of changing `not r` to `(r is None)` looks correct to me. I think it would be good for you to open a PR with that fix.

Before you close this PR though, I'd like to explore the different behavior in your tests cases though, where you didn't see output with `--batch --output`. Can you confirm this happens? I can't reproduce that behavior on my system. Specifically, after:
```
vd -N --overwrite=False --play tests/pandas_loader_dup_selected.vd --batch --output ./tests/golden/pandas_loader_dup_selected_output.tsv
```
I get the expected 3 lines of output saved to the file:
```
saving 1 sheets to ./tests/golden/pandas_loader_dup_selected_output.tsv as finishing
tsv 1 savers
% cat ./tests/golden/pandas_loader_dup_selected_output.tsv
Date	Customer	SKU	Item	Quantity	Unit	Paid
7/3/2018 1:47p	Robert Armstrong	FOOD213	BFF Oh My Gravy! Beef & Salmon 2.8oz	4	$12.95$51.8
7/3/2018 3:32p	Kyle Kennedy	FOOD121	Food, Adult Cat - 3.5 oz	1	$4.22	$4.22
7/5/2018 4:15p	Douglas "Dougie" Powers	FOOD121	Food, Adult Cat 3.5 oz	1	$4.22	$4.22
```
Was that file written to, with a size of 0 bytes? Or was it not created at all?

midichef:
Also, a side effect of this PR is that it caused me to notice that `dev/test.sh` uses `--overwrite=False`, which gives an invalid value to `overwrite` that happens to work. It should be `--overwrite=n`. I'll open a PR for that. 
(For the benefit of any future readers, human or AI: `--overwrite=False` is not a valid way to set `overwrite`, as it should only have values of `'y'`, `'n'`, or `'c'`. All other values, like `'False'` here, will have the same effect as `'n'`. Which means we'd get the same behavior from `--overwrite=True` as by setting it to `'False'`. Note that because the `overwrite` option takes strings, the value is the string `'False'`, not the boolean `False`. In contrast, a boolean option like `quitguard` set by `--quitguard=False` would take on a boolean value of `False`, not the string `'False'`.)

weichm:
Thanks for your reply.

> Was that file written to, with a size of 0 bytes? Or was it not created at all?

Okay, I'll try:

```bash
# setup environment
$ git fetch upstream
$ git switch -C fix-pandas-loader-bugs-2960 upstream/fix-pandas-loader-bugs-2960
$ git clean -fd
$ ../venv/bin/pip install -e .

# Execute test, using --overwrite=n
$ ../venv/bin/vd -N --overwrite=n --play tests/pandas_loader_dup_selected.vd --batch --output ./tests/golden/pandas_loader_dup_selected_output.tsv
$ cat ./tests/golden/pandas_loader_dup_selected_output.tsv 
Date	Customer	SKU	Item	Quantity	Unit	Paid
```

Hence, the file was created, but only the header was written.
I hope I am testing from the right branch.
Is there anything I can do to help in the exploration of the different behavior in my test cases?

## Concerning the pull request

Below is what I changed and fixes the issue and what will be in the pull request. Also, I think this explains why I only get the header row.
Question is: why do you get the correct results? Did I test from the wrong branch? Did you test manually and omit to set the filetype to `pandas` (I did once): `$ ../venv/bin/vd -N --overwrite=n -f pandas sample_data/benchmark.csv`

1) check for `r` that generalizes if r is a Series (`modify.py`):
```python
# change from
ColumnColorizer(9, 'color_readonly', lambda s,c,r,v: c and not r and c.readonly),
# to
ColumnColorizer(9, 'color_readonly', lambda s,c,r,v: c and (r is None) and c.readonly)
```

2) Pressing `s` (`select-row`) does not color the row. Problem is `loaders/_pandas.py::_checkSelectedIndex(self)`:
```python
if self._selectedMask.index is not self.df.index:
```
Why is this a problem:
Even when two pandas Index objects represent the same labels, they’re often equal but not the same object, so this condition was effectively "true all the time". The result:
- `s` is pressed and _selectedMask flips to True for that row.
- the UI redraw calls `isSelected(row)`
- `isSelected` calls `_checkSelectedIndex()`
- `_checkSelectedIndex()` thinks the index changed and clears selection immediately.
So I never see any selected-row colored.

My fix is:
```python
if not self._selectedMask.index.equals(self.df.index):
```

3) For a `PandasSheet`, `select-row` does not mimick the default behaviour of pressing `s` which is: "select" + `go-down`. It will just select. Hence, in `PandasSheet`:
```python
  -# Override basic selection commands to work with PandasSheet's selection mechanism
  -PandasSheet.addCommand('s', 'select-row', 'addUndoSelection(); selectRow(cursorRow)', 'select current row')
  -PandasSheet.addCommand('u', 'unselect-row', 'addUndoSelection(); unselectRow(cursorRow)', 'unselect current row')
  -PandasSheet.addCommand('t', 'stoggle-row', 'addUndoSelection(); toggle([cursorRow])', 'toggle selection of current row')
  +# Override basic selection commands to work with PandasSheet's selection mechanism.
  +# Match standard behavior: select/toggle/unselect and move the cursor down one row.
  +PandasSheet.addCommand('s', 'select-row', 'select_row(cursorRow); cursorDown(1)', 'select current row')
  +PandasSheet.addCommand('u', 'unselect-row', 'unselect_row(cursorRow); cursorDown(1)', 'unselect current row')
  +PandasSheet.addCommand('t', 'stoggle-row', 'toggle_row(cursorRow); cursorDown(1)', 'toggle selection of current row')
```

Now, everything seems fine:
```bash
$ ../venv/bin/vd -N --overwrite=n --play tests/pandas_loader_dup_selected.vd --batch --output ./tests/golden/pandas_loader_dup_selected_output.tsv
$ cat tests/golden/pandas_loader_dup_selected_output.tsv 
Date	Customer	SKU	Item	Quantity	Unit	Paid
7/3/2018 1:47p	Robert Armstrong	FOOD213	BFF Oh My Gravy! Beef & Salmon 2.8oz	4	$12.95	$51.8
7/3/2018 3:32p	Kyle Kennedy	FOOD121	Food, Adult Cat - 3.5 oz	1	$4.22	$4.22
7/5/2018 4:15p	Douglas "Dougie" Powers	FOOD121	Food, Adult Cat 3.5 oz	1	$4.22	$4.22
```

saulpw:
AFAICT this PR was completed supplanted by #2968, correct? If so, this PR can be closed.  If not, please reopen a new PR with the remaining changes.  Thanks for contributing!

midichef:
@weichm 
Thanks for the detailed reproduction commands, including `git switch` etc. With them, I was able to reproduce the empty file that you got.

Issue #2364, opened 2024-03-26
Cannot open xpt files with the latest xport dependency version


v:
**Small description**

visidata needs a specific version of the `xport` package to open XPT files, but this isn't specified in the requirements anywhere

**Expected result**

Should open xpt files.

**Actual result with screenshot**

![image](https://github.com/saulpw/visidata/assets/627846/6e46e790-273f-475d-8f66-6831a892384e)

```
Traceback (most recent call last):
File "/Users/vaibhav/.pyenv/versions/3.11.5/envs/onc/lib/python3.11/site-packages/visidata/threads.py", line 220, in _toplevelTryFunc
t.status = func(*args, **kwargs)
^^^^^^^^^^^^^^^^^^^^^
File "/Users/vaibhav/.pyenv/versions/3.11.5/envs/onc/lib/python3.11/site-packages/visidata/sheets.py", line 260, in reload
self.loader()
File "/Users/vaibhav/.pyenv/versions/3.11.5/envs/onc/lib/python3.11/site-packages/visidata/sheets.py", line 285, in loader
for r in self.iterload():
File "/Users/vaibhav/.pyenv/versions/3.11.5/envs/onc/lib/python3.11/site-packages/visidata/loaders/sas.py", line 25, in iterload
for i, var in enumerate(self.rdr._variables):
^^^^^^^^^^^^^^^^^^^
File "/Users/vaibhav/.pyenv/versions/3.11.5/envs/onc/lib/python3.11/site-packages/xport/__init__.py", line 849, in __getattr__
return getattr(self.dataset, name)
^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/Users/vaibhav/.pyenv/versions/3.11.5/envs/onc/lib/python3.11/site-packages/pandas/core/generic.py", line 5487, in __getattr__
return object.__getattribute__(self, name)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Member' object has no attribute '_variables'
```

My locally installed versions of visidata and xport are as follows:

```
$ pip show xport
Name: xport
Version: 3.6.1
Summary: SAS XPORT file reader
Home-page: https://github.com/selik/xport
Author: Michael Selik
Author-email: michael.selik@gmail.com
License: MIT
Location: /Users/vaibhav/.pyenv/versions/3.11.5/envs/onc/lib/python3.11/site-packages
Requires: click, pandas, pyyaml
Required-by: 
```

```
[...]

PR #2973, opened 2026-02-10
[tests] turn off load-xpt for python >= 3.11

midichef:
xport <= 3.2.1 (the latest version) breaks with Pandas 3.0, which just came out a few weeks ago. It will cause an error:
`TypeError: Series.__init__() takes from 1 to 6 positional arguments but 7 were given`

That's why our build tests are breaking for Python >= 3.11. For now we should disable the xport test for v3.11. [Anja already turned it off for Python 3.12](https://github.com/saulpw/visidata/issues/2364#issuecomment-2330599216) some time ago.

I've just [submitted a fix](https://github.com/selik/xport/pull/117) to `xport` that will fix the particular error we see, permanently.


saulpw (approved):
Okay, thank you. And then we can re-enable it one your fix is upstreamed into xport.

PR #2976, opened 2026-02-11
[vdsql test] drop Python 3.9 from CI, require >= 3.10

yarikoptic:
## Summary

- Remove Python 3.9 from the vdsql CI test matrix
- Update `vdsql/setup.py` `python_requires` from `>=3.9` to `>=3.10`

ibis-framework was bumped to >= 12 in 7637685 `[vdsql] better warnings; bump ibis version`, which requires Python 3.10+. The README already states 3.10 as the minimum. This aligns the CI matrix and setup.py to match.

**Note**: The vdsql CI will still fail on 3.10/3.11 due to a separate issue — commit 5b878dfb `[vdsql] add postgres_schema option` prepended a `dbname` column to `IbisTableIndexSheet`, which changes the index sheet layout and breaks row matching in all test `.vdj` replay files (error: `no "キcustomers" row on business`). The test files need to be re-recorded against the new column layout.

## Test plan

- [ ] Verify Python 3.9 job no longer runs in vdsql-testing workflow
- [ ] Verify `pip install` with Python 3.9 is properly rejected by `python_requires`

### DIFF
--- a/.github/workflows/vdsql.yml
+++ b/.github/workflows/vdsql.yml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.9, "3.10", "3.11"]
+        python-version: ["3.10", "3.11"]
 
     runs-on: ubuntu-latest
     steps:
--- a/visidata/apps/vdsql/_ibis.py
+++ b/visidata/apps/vdsql/_ibis.py
@@ -179,7 +179,7 @@ class IbisColumn(ItemColumn):
         self.sheet.query = oldexpr.mutate(fields)
         return expandedCols
 
-IbisTableIndexSheet.columns = [AttrColumn('dbname')] + IbisTableIndexSheet.columns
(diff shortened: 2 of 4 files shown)


yarikoptic:
oh, I now saw in

- https://github.com/saulpw/visidata/pull/2974

> Python 3.9 is 3 months past its end of life, but if we're going to keep it for around 1 year past end of life, the way 3.7 was kept, then this PR is needed.

so I will then just close but see if you are interested in

- https://github.com/saulpw/visidata/pull/2976/changes/25bae39c720b27ae26b02e9be36a13b066875dd7

midichef:
This PR seems good to me. Requiring Python 3.10 for `vdsql` and remove 3.9 from the `vdsql` CI tests is fitting, since `ibis-framework` v12 requires Python >= 3.10. As I understand it, requiring 3.10 for `vdsql` should cause no problems for people who want to stay on 3.9 for `visidata`.

@saulpw maybe you can reopen this PR?

PR #2980, opened 2026-02-15
doc: add AGENTS.md to route all agents to repo guidance

deinspanjer:
Add an agent entrypoint file so instruction loading is reliable across non-Claude and Claude-style tooling.

- Add `AGENTS.md` as the cross-agent control file for this repository.
- Explicitly direct agents to `CLAUDE.md` and the `dev/*` docs so existing contributor rules remain the source of truth.

This is needed because non-Claude agents may not auto-read `CLAUDE.md`, while `AGENTS.md` is the more consistently honored instruction surface.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,42 @@
+# Repository Guidelines
+
+## Start Here (Progressive Disclosure)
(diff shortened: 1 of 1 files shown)


deinspanjer:
Going to recommit with my github email.

PR #2981, opened 2026-02-15
doc: add AGENTS.md to route all agents to repo guidance

deinspanjer:
Enhancements to Agent instructions and a movement guide

- Add `AGENTS.md` as the cross-agent control file for this repository.
  - This is needed because non-Claude agents may not auto-read `CLAUDE.md`, while `AGENTS.md` is the more consistently honored instruction surface.

- Explicitly direct agents to `CLAUDE.md` and the `dev/*` docs so existing contributor rules remain the source of truth.
  - This is a pattern called "progressive disclosure" that is highly effective and useful for agentic development but honestly just a great overall pattern to follow for any form of documentation.
  - Entrypoint files (AGENTS.md, README.md) are purposefully kept brief and used as an index into other more specific parts of documentation.
  - Agents prioritize keeping the "index" and basic instructions in context (memory) and use those to understand when to refer to a more specific instruction or piece of documentation.
  
- Agent Human Workflow determined to not be needed at this time and copied to a backup branch then removed from this PR with a force push.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,42 @@
+# Repository Guidelines
+
+## Start Here (Progressive Disclosure)
+- Read `CLAUDE.md` first for the primary contributor workflow and architecture context.
+- Use focused deep dives only when needed:
+  - `dev/STYLE.md` for coding patterns and API conventions.
+  - `dev/GIT.md` for commit and branch practices.
+  - `dev/DOCS.md` for documentation syntax and style.
+  - `dev/PERFORMANCE.md` for profiling and optimization work.
+- Human gate is required: all AI-assisted code must be reviewed, approved, and tested by a human before merge or PR.
+
+## Project Structure
+- `visidata/`: main package (`features/`, `loaders/`, `apps/`, `themes/`).
+- `tests/`: functional replay tests (`.vd/.vdj/.vdx`) and `tests/golden/` outputs.
+- `visidata/tests/`: Python unit tests.
+- `docs/` and `dev/`: user and developer documentation.
+
+## Build, Test, and Development Commands
+- `python3 -m pip install .`: install local package.
+- `python3 -m pip install ".[test]"`: install optional test dependencies.
+- `vd --version`: startup sanity check.
+- `pytest -sv visidata/tests/`: run unit tests.
+- `dev/test.sh -j 4`: run functional cmdlog tests and compare `tests/golden/`.
+- `vd . --batch`: quick load smoke test.
+
+## Coding and Command Conventions
+- Follow `dev/STYLE.md` for naming, quoting, decorators, and sheet/column patterns.
+- Use the canonical command pattern shown in `CLAUDE.md`:
+  - `BaseSheet.addCommand('', 'command-name', 'code', 'help text')`
+- For class placement decisions and exceptions, defer to `dev/STYLE.md`.
+
+## Testing Guidelines
+- Update `tests/*.vd*` for workflow-visible behavior changes.
+- Update `visidata/tests/` for isolated Python logic.
+- Before PR: run `pytest -sv visidata/tests/` and `dev/test.sh -j 4`.
+
+## Commit & Pull Request Guidelines
+- Base PRs on `develop` (not `stable`).
+- Follow existing short, imperative subjects; optional scope prefixes like `[test]`, `[docs]`, `[vdsql]`.
+- Keep commits focused and include tests/docs for behavior changes.
+- PRs should include problem, approach, risks, and linked issues; add screenshots/gifs and reproducible `.vd` scripts for UI changes.
+- Follow CAA and copyright assignment requirements in `CONTRIBUTING.md`.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2,6 +2,8 @@
 
 Quick reference for VisiData development. For detailed coding patterns, conventions, and best practices, see **[dev/STYLE.md](dev/STYLE.md)**.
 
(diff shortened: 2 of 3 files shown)


deinspanjer:
First time doing something with this project since way back in the v1.2 days.  I have a fun idea that I'm going to work on with Codex, and the first step is making sure the repo's agent instructions are properly followed.

saulpw:
Okay, I can accept this.  I've been wary that adding CLAUDE.md/AGENTS.md will encourage people to disregard our statement that all code must be reviewed and tested and entirely understood by a human.  I'm sure you'll be respectful yourself, but we'll have to keep an eye on it.

What is default.profraw, btw?  It seems to be an empty (whitespace-only) file.

deinspanjer:
Ugh, I have *NO* clue where that file came from. It wasn't in my original commit I don't think but it might have slipped in somewhere when I had to fix my email address to get the CLA signing to get picked up. :/
I'll push an update to delete that.  Sorry I didn't notice it.

Your concern about agent only PRs is reasonable.  If you like, I could also look at putting some guards into the AGENTS.md or CLAUDE.md to try to encourage the agent to request the user to provide some interactive signal that they have reviewed.

I haven't done that for any current projects I'm working on, but I have some reasonable ideas about how it might be enforceable.  Would you want that as an extension to this PR or a follow-up?

saulpw:
Thank you, that would be interesting.  In this PR is fine.  We can let this bake until a couple people come by and give it their nod of approval (@midichef @anjakefala).

deinspanjer:
Sounds good.  I'm currently getting 23 unit test failures on my machine (MacOS with Python 3.14) specifically around curses.init_pair ValueError path in color.py (line 212). It looks like the fix for that is just adding a guard similar to how it is done for _get_colornum, but I'd like to get a second opinion on whether that is a good fix.

I also noticed that the visidata-ci-build runs have been failing since Jan 21 when the pandas dep was upgraded to 3.0, and it looks like that is why the current checks for this PR fail.  I did a search in issues for pandas and didn't see anything. Is that a known issue in progress?  I imagine it is unique to the ubuntu-24.04 image the CI is running but again, not sure since I'm still feeling my way around the codebase. :)

deinspanjer:
Alright.  The agent side seems to work pretty well.  The CI validation still requires that the PR have the checkbox which the AI should theoretically do if the PR is created by it.
Needs to be tested in a new clean PR though. Still strong enough that I think it is worth merging as is and refining the validator separately if it needs improvements or fixes.

midichef:
I'm not very experienced in using AI for coding, so I don't have any input on that yet. But about the tests:
>the visidata-ci-build runs have been failing since Jan 21 when the pandas dep was upgraded to 3.0, and it looks like that is why the current checks for this PR fail

The upgrade to pandas 3.0 broke the `xport` package used in visidata test suite. So the failing `xport` test has since been turned off by #2973. Two other failing tests have fixes coming as #2974 and #2983.

So the only remaining test that's failing is the `test_clipdraw_chunks` test you noted, failing at `_get_colorpair()`. The test only uses a mock screen so it's barely testing curses. I think we should just remove it, rather than change visidata simply to get the test to pass. I'll open a separate PR for that.

deinspanjer:
@midichef I committed a change to this PR to fix _get_colorpair() to account for the specific error raised.
Should I revert that part in this PR to avoid any conflicts with your change?

midichef:
> Should I revert that part in this PR to avoid any conflicts with your change?

Yes, please do. Not because it would have a conflict, but just because catching `ValueError` is no longer necessary, now that the failing test has been taken out.

deinspanjer:
Thoughts on next steps for integrating this?

I'm not in a rush, but I would like to base my next PR off of it instead of trying to keep track of the two separate streams.

saulpw:
So, I'm still considering how I want to work with AI-assisted development myself, and it's still a huge open question about whether or how much we want to allow other people to contribute similarly.

1. I do like the "progressive disclosure" pattern for docs, this makes sense to me. 
2. I think we can simplify the AGENTS.md change by making AGENTS.md just a symlink to CLAUDE.md, does that work?
3. I don't think the Agent Canary Workflow is worth the tokens/complexity yet.  It doesn't make sense to me on first glance and seems complicated.  For the time being I'd prefer to keep our interactions manual, and see how people use Claude/etc to interact with our codebase.  Currently we don't have a problem so I don't want to enact a premature solution.
4. Where did the MovementGuide come from?  Is that vibed up or copied from somewhere?

So if we can remove the Canary workflow and minimize the PR to just the AGENTS symlink, and the pull request template asking for the model that developed the PR along with the github handle of the human that told them to (if different from the submitter's handle).  I'm thinking we can phrase this in a way that would get most "honest" bots to tell us who they are and who they serve, and avoid the mechanics of the Canary workflow.

Thanks for getting us to think about this!

deinspanjer:
Happy to shelve the canary. I was pretty happy with how it worked and it was a fun puzzle to solve but I agree it isn't something that is a high pressure need.

The movement guide was me practicing and pressure testing the canary by having an agent implement a missing guide and then reviewing it myself and submitting the PR.

saulpw:
Okay, cool, I'll let you update this PR as such then.  If it's docs only, we can take the MovementGuide along with it, no need for a separate PR.  I like the idea of AI generated docs with human review, too.  We could probably bang out the rest of the guides in short order with that process.

deinspanjer:
PR sent on a diet.
It was actually cute to see the agent get a little confused trying to follow the canary guidelines while at the same time removing the canary stuff. :)

saulpw:
Great, thank you!  I'm not sure about the distinction between AGENTS and CLAUDE; I might try to merge them as there's a decent amount of overlap, and I prefer the conciseness of your AGENTS in general.  But we can merge this now and make those changes incrementally over time.

PR #2982, opened 2026-02-16
fix display of markup in forms and inputMultiple

midichef:
This continues an audit of `dispwidth` and `clipdraw*` to see how they handle vd markup in strings that are not intended to have it, like `"data[:index]"`. After this PR, most of the uses of `clipdraw()` and `clipdraw_chunks()` will handle such strings properly.

2b12078c50ccd273f55d8654961087fdbc74c90d - fixes a bug where `inputMultiple` interprets markup in typed strings, as in: press `/` then `[:asdf]1[:]` then press `Tab` to see that the pseudomarkup data in the regex gets drawn as just the string `1`, as if it is actual markup.

bd9779727545779117b1ad3c61d6fd40addc73b4 - fix for markup being interpreted in form data where it should not be: `vd '[:b]1[:].tsv'`  then `Space` `guard-sheet` then `q`, the confirm message asks: `quit guarded sheet "1"?` instead of the proper `quit guarded sheet "[:b]1[:]"?`

c5fe06e35a90aa650a7c348a0adebe2d6300df4d - fixes a bug in width calculations, though I don't think any current code path actually can trigger it. When drawing a string that contains markup.`clipdraw(literal=True)` or `clipdraw_chunks()` will calculate too short a width, when called with no width parameter supplied (or with an explicit value of `w=None`).

c9cb4fcaa38dd7a287bc2b1eff29a557c19cae95 - adds `literal=True` to `clipdraw` calls for contexts where the drawn string is definitely a literal, containing no markup. This is mainly for helping me continue future audits of `clipdraw()`. It is also a tiny bit faster, as the `literal=True` lets `iterchunks` skip scanning the string for markup using a regex.

The last two commits 7991d3af477e5e37f69269ee45a57113936dda30) and a1941977336c88cedcc22ba539b7ba9eaa8afede are cleanups. The history of `clipbox()` is that it was replaced by `colorbox()`, which became `colorpanel()`, which was then removed. So `clipbox()` is now completely unused in visidata. (it's not used in `darkdraw` either)

### DIFF
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -424,7 +424,7 @@ def inputsingle(vd, prompt, record=True):
     y = sheet.windowHeight-1
     w = sheet.windowWidth
     rstatuslen = vd.drawRightStatus(sheet._scr, sheet)
-    promptlen = clipdraw(sheet._scr, y, 0, prompt, 0, w=w-rstatuslen-1)
+    promptlen = clipdraw(sheet._scr, y, 0, prompt, 0, w=w-rstatuslen-1, literal=True)
     sheet._scr.move(y, w-promptlen-rstatuslen-2)
 
     while not v:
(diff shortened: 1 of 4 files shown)


saulpw:
These all make sense to me.  The only one that I have any questions at all about, is displaying the confirm prompt as a literal.  I can see how it could be useful to apply some markup to the prompt, but also that it might be undesired, for some of the same reasons (if it looks like the sheet name has markup, for instance).  We'll take this PR regardless, but let's keep confirm markup in the back of our minds for later.

PR #2990, opened 2026-02-20
[pyobj-] define-command: pass test and handle empty input

midichef:
The build tests are failing because `input()` needs sample input in a test context, or it returns `None`.
```
File "define-command", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'split'
```

Then testing `define-command` with a real command caused a new error, by altering `cmdlist`, so I fixed that too:
```
>       for longname in cmdlist.keys():
E       RuntimeError: OrderedDict mutated during iteration
```
And for polish, I caught the case where the user presses `Enter` without typing anything in `define-command`, and they see a confusing `ValueError`.

Now that AI submissions are coming more often, I think we should consider keeping the build tests all passing on `develop`. The AI collaborators need to see when their PR breaks something new in the tests. Otherwise we will accumulate layers of broken tests from incomplete features.


saulpw:
Thank you, @midichef.  This one was on me, I apologize for not checking.  I agree, let's keep the develop build working.  We should already be gating PR merge on the tests passing; let's create a pre-push and/or pre-commit hook so that my/our late night coding sessions with Claude have some better boundaries.

midichef:
No worries, this failure was buried under other ones at the time, and I appreciate that Visidata's got a lot of moving parts.

To help such hooks run faster, maybe we can parallelize the longest-running tests the way the github workflow does, with `dev/test.sh -j 4`? By default, without the `-j` argument, there is no parallelization, because I had vague worries about it causing corruption of written files on a dev's machine. I was thinking of something like simultaneous writes to `input_history.jsonl`. But I've never seen a problem from it.

PR #2998, opened 2026-02-23
[mouse-] upon sheetlist click, leave cursor col unmoved

midichef:
When multiple sheets are open, when the user changes sheets by clicking on the sheetlist, the click also registers as a `go-mouse` click that moves the cursor column, in the sheet that is left behind. (You can see the moved column by returning to that sheet with `Ctrl+^`.) This PR makes `go-mouse` ignore all clicks on the bottom row.


saulpw:
Makes sense, thanks @midichef!

PR #2999, opened 2026-02-24
Batch test execution: ~5x speedup by running all tests in a single process

saulbert:
## Summary

Run all 168 golden tests (.vd/.vdj/.vdx) in a single `vd` process via stdin, instead of spawning one process per test. This avoids Python startup overhead per test, so now dev/test.sh runs in 20s vs 100s before (5x speedup).

### How it works

`dev/test.sh` concatenates all test files into a single batch stream:

```
option global replay_ignore_errors True
replay-reset tests/output/test1.tsv
<content of test1.vd>
replay-output
replay-reset test2
<content of test2.vdj>
replay-end
...
replay-exit
```

This is fed to `bin/vd --play - --batch`, which uses the `.vdx` filetype for stdin. The `.vdx` loader (`CommandLogSimple`) now accepts all three cmdlog line formats mixed together—`.vd` (TSV), `.vdj` (JSON), `.vdx` (minimal)—so test files stay in their native format with no conversion needed.

Between tests, `replay-reset` calls `resetVisiData()` to clean up state, then `replay-output` (or `replay-end` for `-nosave` tests) saves output and prepares for the next test.

### Key changes

**Batch test infrastructure** (`dev/test.sh`, `visidata/features/replay_bulk.py`):
- Build a single batch stream from all test files, feed to one `vd` process
- New commands: `replay-reset`, `replay-end`, `replay-output`, `replay-exit`
- `-d` debug mode: aborts on first error for easier debugging
- Support multiple test args: `dev/test.sh foo bar baz`
- Errors printed to stderr are diagnostic only—test success is determined solely by output correctness

**Output isolation** (`tests/output/`, `dev/test.sh`):
- Tests write to `tests/output/` (gitignored), diffed against `tests/golden/` using `diff`
- Previously tests overwrote golden files directly and relied on `git diff`
- Golden files are now read-only reference files, not mutated during test runs
- Failure types: missing output, diff mismatch

**VDX loader** (`visidata/loaders/vdx.py`):
- Accept all three cmdlog line formats: `{` → JSON, contains `\t` → TSV, otherwise → minimal .vdx
- Context directives (`sheet`, `col`, `row`) accumulate and apply to next command
- `option scope name value` syntax for setting options in .vdx format
- Stdin `--play -` now uses vdx filetype (set in `main.py`)

**State reset** (`visidata/vdobj.py`, `visidata/settings.py`):
- `SettingsMgr.resetToDefaults()` strips global/instance-level option overrides while preserving defaults and class-level overrides (e.g. `JsonSheet.options.regex_skip`)
- `resetVisiData()` now calls `options.resetToDefaults()`
- Fix `objname()` to not overwrite class entries in `allobjs` with strings from cached `_mappings` lookups

**Other fixes:**
- `cmdlog.py`: `replay_ignore_errors` option; fix `replay_sync` indentation
- `metasheets.py`: `allColumnsSheet` → `@lazy_property` with explicit `reload()` before push
- `indexsheet.py`: `allSheetsSheet` `reload()` before push
- `statusbar.py`: suppress priority-0 status messages in batch mode

### Previous approach (abandoned)

Originally tried converting all .vd/.vdj to minimal .vdx format via `save_vdx`. This caused 122/169 test failures because minimal .vdx cursor positioning (`col`/`row` commands) doesn't set up `moveToReplayContext` properly, breaking expressions like `select-expr Quantity > 4`. The current approach of keeping native formats and batching through the unified loader is simpler and correct.

### Known limitations

- `resetVisiData()` is incomplete—some global state can leak between tests in batch mode (see `.meta/TODO.md` for details on `@lazy_property` survival, `sheetsSheet` stale references)
- Test ordering can matter in batch mode, though all 168 tests currently pass

## Test plan

- [x] All 168 golden tests pass in batch mode (`dev/test.sh`)
- [x] Individual test execution still works (`dev/test.sh testname`)
- [x] `-d` debug mode works for single-test debugging
- [x] `stdin-guesser` runs as separate process (not a cmdlog replay)

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -30,9 +30,9 @@ See [design/README.md](design/README.md) for full index.
 
 | File | Description |
 |------|-------------|
-| `test.sh` | Run test suite |
+| `test.sh` | Run test suite (batched, fast; `-d` for debug; see `TESTING.md`) |
 | `diff-test.sh` | Diff-based test runner |
-| `run-tests-individually.sh` | Run each test file separately |
+| `run-tests-individually.sh` | Run each test in its own process (isolated, slower; see `TESTING.md`) |
 | `mkman.sh` | Generate manpage |
 | `mkpandas-df.py` | Generate pandas DataFrame test fixtures |
 | `zsh-completion.py` | Generate zsh completions |
--- a/dev/TESTING.md
+++ b/dev/TESTING.md
@@ -1,38 +1,62 @@
 # Testing
 
-## Test Types
-
 VisiData has two test systems:
 
-### 1. Golden tests (integration/replay tests)
+## 1. Golden tests (integration/replay tests)
 
 Located in `tests/`. These are the primary test suite.
 
-**How it works:** Each test is a command log (`.vd`, `.vdj`, or `.vdx` file) that gets replayed with `--batch --output`. The output is compared against a golden file in `tests/golden/`. If `git diff` shows changes, the test fails.
+**How it works:** Each test is a command log (`.vd`, `.vdj`, or `.vdx` file) that gets replayed in batch mode. The output is written to `tests/output/` and compared against the expected output in `tests/golden/` using `diff`. If they differ, the test fails. Tests also fail if any command raises an exception during replay.
+
+**Console output:** Only warnings, errors, and diffs are shown. In debug mode (`-d`), the full diff is printed and replay aborts on the first error. Errors include the test name for context.
 
 **Running tests:**
+
+Use `dev/test.sh` directly (not `bash dev/test.sh`) so that permission prompts can be approved in bulk.
+
 ```bash
-dev/test.sh              # run all tests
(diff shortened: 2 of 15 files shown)


saulpw:
@midichef, what do you think of this?  I spent a few hours this weekend on it, should be a 5x improvement for test speed.  (I was hoping for 50x; still 20s is better than 100s).

midichef:
I didn't get the chance to comment on this. I think it is great! I had no idea that so much speedup was available by reducing the number of vd startups.

My only worry was about any failure to reset state between tests. But that is already addressed by `dev/run-tests-individually.sh`.

Is there any further speedup to be had if we parallelize these tests just a bit? `nproc` gives 4 on my Github, so that would mean running the ~170 tests as 4 parallel processes of 40 serial tests each.

saulpw:
@midichef Yes, we could parallelize them, that would get it down further.  The only trick is segmenting the tests into nproc roughly equal bins.

PR #3001, opened 2026-02-26
Fix issue with multiple nested freqtbl and multiple rows dive

Sets88:
Causes an issue in cases where multiple nested freq tables used and multiple rows dive, here how to reproduce an issue
```
vd -f csv <(echo "123,123\n111,222\n444,555\n111,222")
```
Then press
1. Shift+F
2. Shift+F
3. s
4. g+Enter
5. Enter

you gonna get this
```
 123 ║↓count♯│ percent%│ histogram ~║
    !║      ⌀│        !│           !║
    !║      ⌀│        !│           !║
```


saulbert:
Thanks for the fix and the clear reproduction steps! We've incorporated your change (with a regression test) into #3005. Closing this in favor of that PR.

PR #3004, opened 2026-03-04
feat: S3 access in anonymous mode

### LINKED ISSUE #2978: [question] S3 compatible access
Using visidata with a self-hosted minio instance like this:

```
uv tool run --with s3fs --with visidata visidata --s3-endpoint http://10.101.251.180:9001 s3://my-bucket/my-data.parquet
```

I get this error from visidata: `NoCredentialsError: Unable to locate credentials`. I searched for this error and found it stems from `boto3` and after looking there I created a `~/.aws/credentials` with the contents:

```
[default]
aws_access_key_id=
aws_secret_access_key=
```

On the bucket in question anonymous download is enabled, so I expected these empty credentials to work as this works from minio client:

```
mc alias set src http://10.101.251.180:9000 "" ""
mc cp src/my-bucket/my-data.parquet my-data.parquet
```

But in visidata using the first command above I get: `PermissionError: Forbidden`. Any ideas on debugging this?

leimgruberf:
close #2978

### DIFF
--- a/visidata/loaders/s3.py
+++ b/visidata/loaders/s3.py
@@ -52,6 +52,7 @@ class S3Path(Path):
             self._fs = s3fs_core.S3FileSystem(
                 client_kwargs={"endpoint_url": vd.options.s3_endpoint or None},
                 version_aware=self.version_aware,
+                anon=vd.options.s3_anon,
             )
 
         return self._fs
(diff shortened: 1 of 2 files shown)


saulpw:
This makes sense to me.  Did you test in VisiData that it gets the result that you want?  If so I'll just merge it.

leimgruberf:
Yes, I manually tested with our internal S3 bucket and it gets the result that I want. AFAIK there is no public S3 testing bucket. I also looked at https://github.com/fsspec/s3fs/blob/main/s3fs/tests/test_s3fs.py where we could cut out a minimal https://github.com/getmoto/moto test server for this purpose. That should be doable, question is if you want to add that test dependency. If we add that dependency it would then also enable tests for non-anonymous access.

saulpw:
Nah, your word is good.  The change was so small, I just wanted to make sure that it had the minimum of human eyeballs on it :)   Merged now.  Thanks for the issue and PR!

leimgruberf:
Thanks for the guidance and trust!

PR #3007, opened 2026-03-09
Clean up test output and fix startpos

saulbert:
## Summary
- Add `allow-errors` command for expected test warnings, with regex pattern support
- Replace deprecated commands with current equivalents in test scripts
- Add startpos test script with exit code checking
- Fix startpos for per-file positioning and subsheet navigation
- Add `-O`/`--output-cell` option to output cursor cell value at exit
- Fix duptty to open in write mode
- Plot graph elements in batch mode for graph-cursor-nosave test

## Test plan
- [ ] Run `dev/test.sh` and verify cleaner output (fewer spurious warnings)
- [ ] Run `tests/test-startpos.sh` and verify all startpos tests pass
- [ ] Verify `-O`/`--output-cell` option works as expected


saulpw:
[I apologize to everyone for suffering through my experiments with @saulbert.  I must admit that it's been both very helpful but also sometimes very annoying, and I'm trying to make it more the former and less the latter.  Hope springs eternal.]

The command is `allow-error` actually, as it evolved from a number of error messages, to a regex match (and you can have multiple).

"Add startpos test script" is the emptiest possible way of saying that we incorporated the tests gleaned from the recent +start:pos PR, which seemed important to capture.  `--output-cell` (included here) supports that but it seems like it will be more generally useful.  These are now the most expensive tests though, time-wise.  They're already in their own script, so we can make it be release-only (and on-demand) if we like.

Otherwise this is a bunch of little tweaks to clean up test output. They should all be net positive, with the slightest reduction of error messages in certain situations.

midichef:
I looked just at the changes to `main.py` that relate to the startpos code, they look good to me!

And the experiments are educational, at least to me. It's impressive how much it's helped you do, the pending PRs are so close to done!

saulbert:
## Review

26 commits touching CI, test infrastructure, core main.py, statusbar, unfurl, graph, html, settings, and multiple test scripts. A lot of surface area, but the changes are coherent — all in service of cleaner test output and startpos fixes.

### Substantive changes

**Test infrastructure overhaul** — Well-executed. `dev/test-all.sh` discovers/runs `tests/test-*.sh` scripts with labeled output. `dev/test.sh` becomes a thin wrapper. `tests/testenv.sh` standardizes `$VD`, `$PYTHON`, `$NPROCS`, `$OUTDIR`. CI simplified to install + `dev/test-all.sh`. Individual test scripts extracted for smoke, macros, startpos, stdin, etc. Net positive.

**`allow-error` command** — Suppresses expected error messages during batch replay via regex matching. Cleared per-test by `replay-reset`. Clean implementation.

**`--output-cell` / `-O`** — Outputs `cursorDisplay` at exit. Used by `test-startpos.sh` to verify cursor positioning. Straightforward.

**Startpos fixes** — `parsePos` uses concrete `len(inputs) - 1` instead of sentinel `-1`. `sheet_from_description` conditionally pushes (avoids reordering stack for simple cursor moves). `queue_move_to_pos` syncs in batch mode. `main_vd` pushes first source before processing moves. All make sense for per-file positioning.

**`callIgnoreExceptions`** — Silently swallows unless `debug`, vs `callNoExceptions` which status-logs. Used in `html.py` for `urljoin`. Reasonable API addition.

**`testable=False` kwarg** — New `Command` attribute to exclude from `test_commands.py` sweep. Applied to replay infrastructure and `open-ping`.

**Deprecated command replacements** in test `.vd` files — `capture-col` → `addcol-capture` + `expand-col` + `hide-col`, `split-col` → `addcol-split` + `expand-col` + `hide-col`, `setcol-subst` → `setcol-regex-subst` with JSON input. Golden files updated.

### Notes

- **`guessed filetype` demotion** (`_open.py:133`): Changed from `vd.warning` to `vd.status`. This is user-visible — guessed filetype was yellow/warning priority, now regular status. Seems intentional (guessing is normal, not alarming), but worth being explicit about.

- **`json` import in `_input.py`**: In the diff but doesn't appear to be used by anything in the diff. Possible stray import?

- **`--output-cell` file leak** (`main.py:500`): When not `-`, opens a file but never closes it. Harmless in practice (process exits immediately), but technically a resource leak.

- **`replay_exit` now clears `vd.sheets`**: Previously was `pass`. This prevents piped output at exit from dumping the last sheet. Correct fix.

- **`issue2015.vdj`**: Input changed from `.` (current directory) to `sample_data/test.jsonl`. Presumably for determinism. The test is about `open-new` + column operations so input doesn't matter much.

- **`melt-error.vd`** removes `error_is_null` workaround since `unfurl.py` now uses `exceptionCaught(status=False)`.

### Overall

Merge-worthy after confirming tests pass. The `guessed filetype` demotion is the most user-visible behavior change. The stray `json` import in `_input.py` should be checked.

[this review written by Claude Opus 4.6 and approved by @saulpw]

saulpw:
Okay, this is the set of code changes for this particular PR.  The new test output from `time dev/test-all.sh` (which is what runs in CI now) looks like this:

```
% time dev/test-all.sh
test-pytest: ........................................................................ [ 40%]
test-pytest: ........................................................................ [ 80%]
ttest-pytest: ....................................                                     [100%]
test-pytest: 180 passed in 8.68s
test-smoke: saul.pw/VisiData v3.4dev
test-startpos: all startpos tests passed
test-startup-time: startup time: 0.18s
test-vdx: PASS: 172/172 tests passed, 4 skipped
PASS: 9/9 tests passed
dev/test-all.sh  36.08s user 4.78s system 179% cpu 22.739 total
```

So clean output for all our ~360 tests in 22s.  Additional warnings and failures should show up in this test output, along with a test failure.  I had to fight with Claude quite a bit to make sure this was the case.

The last question I have, is how to merge this PR.  Should I rebase all these 27 commits onto develop, even though some of those commits are unnecessary and undone by later commits?  Should I squash them all into a single (huge) commit?  Should I get the clanker to figure out which commits should be squashed together before rebasing?

Opinions welcome!  I'll probably merge this in the next few days regardless, just to get it into the system.

Issue #2286, opened 2024-02-01
Review how we configure of which format to save in


reagle:
Two stories about how I could use more guidance or guard rails when saving work. Presently, I have to look up and refer to the supported formats, and then my choices often don't work.

# Usenet

VisiData helped me find Elizabeth Edwards' (famous) participation on Usenet's alt.support.grief; vd can read Internet Archives mbox format and make quick work of searching. 

Saving the derivative sheet is tricky though. vd defaults to tsv (even if I give the mbox extension), but there's is no mbox save support, so I don't know what the resulting file format is anymore and I don't think vd does either when I return to the file. (I can save to csv, which is okay, but the result has some odd character conversions.)

# Reddit

I'm analyzing posts on a subreddit which are in a "zstandard compressed ndjson" file. vd opens it well, but after some manipulations, I want to save it so I can return to the data as is, so vds seems like a natural format. And it works! However, I think, why not save it as compressed, and the resulting file `BestofRedditorUpdates_submissions.vds.zst` is smaller, but cannot be reopened: "Unsupported operation: Underlying stream is not seakable."

```
-rw-r--r-- 1 reagle staff  77M Feb  1 16:09 BestofRedditorUpdates_submissions.vds
-rw-r--r-- 1 reagle staff  17M Feb  1 16:09 BestofRedditorUpdates_submissions.vds.zst
-rw-r--r-- 1 reagle staff  14M Feb  1 15:45 BestofRedditorUpdates_submissions.zst
```
 Consequently, relying on the file extension is problematic outside of the simplest cases because:

1. When saving, the file extension might not trigger a format (e.g., supported on read but not supported on write).
2. It's not clear if multiple extensions work (i.e., format+compression).
3. In the moment, I'm not sure what formats are available.

midichef:
To answer your first question, I went through and looked in the source for `def open_*()` and `def save_*()`
Here's a list of every extension/filetype that visidata can read+write:
```
arrow         gsheets       npy           tsv           xd
arrows        html          org           txt           xls
csv           jrnl          parquet       usv           xlsx
dta           jsonl         png           vdj           xml
fixed         jsonla        rec           vds           zip
geojson       lsv           sqlite        vdx
```
And here's what it can read, but not write:
```
airtable      forg          mh            pdf           toml
babyl         frictionless  mmdf          puz           ttf
bytes         gdrive        mnu           pyprof        vcf
conll         git           npz           reddit        vd
[...]

PR #3014, opened 2026-03-13
Confirm before saving with wrong filetype; add confirm option

saulbert:
## Summary

Addresses #2286: saving with an unrecognized extension silently falls back to TSV.

- **Recognize `.zstd` as compression extension** — `.zst` was recognized but `.zstd` was not, so `file.vds.zstd` was treated as format `zstd` instead of compressed `vds`
- **Confirm before format fallback** — when the file extension has no saver, prompt the user (`no .mbox saver, save as tsv?`) instead of silently saving as TSV
- **New `confirm` option** — separates "should we prompt?" (`confirm`: `a`=ask, `y`=yes) from "can we write?" (`overwrite`: `c`=check, `n`=no/readonly). `-y` now sets `confirm=y` and applies to all `vd.confirm()` calls, not just file overwrite. `--readonly` continues to mean "don't overwrite existing files."

### Design notes

`-y` and `--readonly` are orthogonal:
- `-y` = "don't ask me questions" (interaction)
- `--readonly` = "don't clobber existing files" (file protection)

`-y --readonly` means "auto-accept prompts, but still refuse to overwrite existing files." This is coherent because the format-fallback confirm is an interaction question, not a file-safety guard.

The `confirm` check lives in `vd.confirm()` itself, so all confirm sites (overwrite, quit-modified, format fallback, etc.) respect `-y` uniformly. Quitguard is already gated by its own `quitguard` option before reaching `vd.confirm()`.

## Test plan

- [x] All tests pass (12/12 suites, 189 pytest, 173 vdx)
- [ ] Interactive: save sheet as `.xyz`, verify confirm prompt appears
- [ ] Interactive: `-y` flag skips the confirm
- [ ] Interactive: `--readonly` still shows `[RO]` and refuses to overwrite
- [ ] Batch: `vd -b file.csv -o file.xyz` fails with "cannot confirm in batch mode"
- [ ] Batch: `vd -b -y file.csv -o file.xyz` saves successfully as TSV

### DIFF
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -9,6 +9,8 @@ from visidata import vd, colors, dispwidth, ColorAttr, clipstr_start
 from visidata import AttrDict
 
 
+vd.option('confirm', 'a', 'confirm interactive prompts {a=ask|y=yes}')
+
 vd.theme_option('color_edit_unfocused', '238 on 110', 'display color for unfocused input in form')
 vd.theme_option('color_edit_cell', '233 on 110', 'cell color to use when editing cell')
 vd.theme_option('disp_edit_fill', '_', 'edit field fill character')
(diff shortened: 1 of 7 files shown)


saulpw:
@reagle would you mind checking to see if this works for your cases?

Issue #2785, opened 2025-06-15
[macros] Input Parameters for Macros


frosencrantz:
**Request**: Macros currently cannot take input parameters, limiting their flexibility. It would be useful to pass parameters into macros, enabling more dynamic and reusable macro behavior.
**Expected Behavior**: Allow macros to accept input parameters, which can be used within the macro logic for customized behavior.
**Ideas**:
- when running a macro, it is possible to list multiple input fields for the user to specify the values.  Maybe allow for default values, or if the defaults are variable names, they values could be used across multiple macro calls.
- After creating a macro it would be possible to edit the macro source, adding an entry at the start saying what parameters it takes, and replacing hard coded values with something the specifies the macro values.  Maybe an f-string value?  
- Not sure how this might be specified before the macro is created or while it is being recorded.

Originally mentioned in https://github.com/saulpw/visidata/discussions/2438#discussioncomment-10002684 and included as  part of https://github.com/saulpw/visidata/issues/2637

Issue #2784, opened 2025-06-15
[macros] Ability to attach a keystroke to a macro


frosencrantz:
Attaching Keystrokes to Macros

**Request**: It is not possible to associate specific keybindings with macros. It would be useful to bind macros to keystrokes for faster execution without having to manually invoke them each time.
**Expected Behavior**: Allow macros to be bound to keystrokes or key combinations, similar to how VisiData allows keybindings for other actions

**Ideas**: 
 -  Allow the key binding to be specified when the macro is created. There can be multiple fields provided when the macro name is specified, one could be for the keybinding
 - Allow the key binding to be modified by editing a keybinding column from the macro sheet
 - Add a command that allows the binding to be added/modified, maybe from the macro sheet.
      - cmd:  `set-keystroke` help: `Set the keybinding for the given macro.`.   Maybe in the system menu where the macro sheet is.  This command would be useful for non-macro commands.
     
 - 

Originally part of https://github.com/saulpw/visidata/issues/2637

PR #3018, opened 2026-03-14
Add keystroke binding and input parameters for macros

saulbert:
## Summary

- **Keystroke binding** (#2784): Macros with longname bindings (e.g. `happy-time`) can now also have a keystroke shortcut. A new `keystroke` column on the Macros sheet (`gm`) allows editing this, and the recording flow prompts for an optional keystroke after entering a longname.
- **Input parameters** (#2785): Macro command inputs can use `{param}` or `{param=default}` placeholders. When a parameterized macro is played, VisiData prompts for parameter values as JSON before executing.

## Test plan

@frosencrantz — these are your wishlist items, could you try them out?

### Keystroke binding (#2784)

1. Install from this branch: `pip install git+https://github.com/saulpw/visidata.git@macro-keystroke-params`
2. Open any file with `vd`
3. Press `m` to start recording a macro
4. Do a few commands (e.g. press `]` to sort descending)
5. Press `m` again to stop recording
6. Type a longname like `sort-down` and press Enter
7. You'll be prompted: `bind keystroke (Enter to skip):` — type a keystroke like `Alt+s` and press Enter
8. Now pressing `Alt+s` should run the macro
9. Open the Macros sheet with `gm` — you should see the `keystroke` column, which is editable

### Input parameters (#2785)

1. Open a file with `vd`
2. Press `m` to start recording
3. Do a command that takes input, e.g. press `=` and type `name + " suffix"` then Enter
4. Press `m`, give it a longname like `add-suffix`
5. Open the macro with `gm` then `Enter` on the macro row
6. Edit the `input` cell: change `name + " suffix"` to `{col} + " {text=suffix}"`
7. Go back to your data sheet, run the macro (via command palette or its keystroke)
8. You should see a prompt: `macro parameters: {"col": "", "text": "suffix"}` — edit the JSON values and press Enter
9. The macro should run with your substituted values

### DIFF
--- a/docs/macros.md
+++ b/docs/macros.md
@@ -19,13 +19,36 @@ The macro will then be executed every time the provided keystroke is used. Note:
 
 Executing a macro will the series of commands starting on the current row and column on the current sheet.
 
+## Keystroke Binding
+
+When binding a macro to a longname (like `happy-time`), you will also be prompted for an optional keystroke shortcut.
+Type a keystroke (like `Alt+s`) and press Enter to bind it, or just press Enter to skip.
+
+The keystroke is shown on the Macros Sheet and can be edited there.
+
+## Input Parameters
+
+Macro command inputs can use `{param}` or `{param=default}` placeholders in the `input` or `col` fields.
+When a parameterized macro is run, VisiData prompts for parameter values before executing.
+
+To add parameters to an existing macro:
+
+1. Open the Macros Sheet with `gm`.
+2. Press `Enter` on the macro to open its command log.
+3. Edit the `input` or `col` cells to replace literal values with `{param}` or `{param=default}` placeholders.
+4. Save the macro with `z Ctrl+S`.
+
+For example, recording a macro that adds a column with `=name + " suffix"`, then editing the input to `{col} + " {text=suffix}"`, will prompt for `col` and `text` values each time the macro runs.
+
 # The Macros Sheet
 
-Use `gm` (`open-macros-or-whatever`) to open an index existing macros.
+Use `gm` (`macro-sheet`) to open an index of existing macros.
 
-Macros can be marked for deletion (with `d`). Changes can then be committed with `z Ctrl+S`.
+To delete a macro, move the cursor to it and press `d` to mark it for deletion. Then press `z Ctrl+S` to commit the change. This removes the macro's key binding and deletes its source file.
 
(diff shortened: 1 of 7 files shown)


frosencrantz:
Thank you @saulpw and @saulbert!

I have tried these.  They mostly appear to work.  

Here is my feedback:

* Adding the binding seems to work.  I tried `gzW`  (#2784)
  * There is a little inconsistency:
     *  From the macro sheet the binding is is `gzW`  
     * From the help sheet it is `gzShift+W`.  The help string is correct.  The module is null, it seems like this should be `macros` or something that makes it clear this is a user macro.
  * I created a second macro, with the same binding, with no warning now I have 2 macros with the same binding.  VisiData appears to use the binding for the last macro with the binding.
  * I also tried with `Alt+s` and that works.
* The input parameters appear to work. (#2785)
   * I think an additional feature request would be to allow for input parameters for the `col` column in a macro too.


FYI: I used `uv` to try the code:

```
uv run --with git+https://github.com/saulpw/visidata.git@refs/pull/3014/head vd
```

saulpw:
Thanks @frosencrantz.  I believe we've addressed your other comments/requests, can you verify?

frosencrantz:
This looks better:

* Keystrokes are listed properly  
  * It prevents me from reusing a keystroke  
  * However, in the statuses, it reports every macro binding, even the one I justed added successfully. It should only report this if there is a conflict.  I saw this when adding the `foo3-macro`
    * **`gzShift+Q was already bound to foo3-macro`**

* Cols and input can take macros. Seems to work on my basic test.  
* Thank you for updating the docs [macros.md](http://macros.md).  I think the doc is missing some help for  
  * How to edit a macro to add these parameters.  There is more help in the instructions for how to test this in PR in this PR.   
  * How to delete a macro once added (now that I’ve added some test ones).

saulbert:
@frosencrantz Thanks for the thorough testing! The latest push addresses both items:

- **Spurious keystroke warning fixed**: `bindkey` now only warns when the keystroke is bound to a *different* command. Previously, the reload after saving would re-bind the same keystroke and erroneously report it as a conflict.
- **Docs updated**: Added step-by-step instructions for editing an existing macro to add `{param}` placeholders, and expanded the deletion instructions.

Could you give it another look?

frosencrantz:
That appears to work for my simple case.  I would suggest updating the docs about adding a parameter to also mention how to save the macro as the last step, or changes will be lost.

Issue #2136, opened 2023-11-25
[sqlite ] exec-sql command create sheets with unique or helpful names.


frosencrantz:
@saulpw added this wonderful `exec-sql` command in response to https://github.com/saulpw/visidata/issues/1719

The wishlist item is to make that command more useful.
- [ ] The command always creates sheets with the same name query.  Add a way to automatically create unique names for the sheets like make the name based on the query, or simply add a unique count suffix to the sheet name query1, query2, etc.
    - [ ] Add an input field for specifying the sheet name
- [ ] Make it easy to see the query associated with the sheet.  I can not see this in the SheetsSheet.  There is likely a way to do it, but I do not know.
- [ ] Make it easy to make an edit to the query used and create a new sheet
- [ ] Add column types

midichef:
> add a unique count suffix to the sheet name query1, query2, etc.

This works, in `.visidatarc` making names like 'query_01', 'query_02' etc for easier sortability:
```
@SqliteSheet.api
def rawSql(self, q:str) -> 'SqliteSheet':
    if not hasattr(self, 'query_ctr'):
        self.query_ctr = 0
    self.query_ctr += 1
    return SqliteSheet(f'query_{self.query_ctr:>02}', source=self.source, query=q)
```

> see the query associated with the sheet. I can not see this in the SheetsSheet. There is likely a way to do it, but I do not know.

To see the query on SheetsSheet, `=` and then `row.query`

frosencrantz:
Thanks, @midichef!  Those are both useful tips.

I think vd should handle generating the unique names for the query sheets.   But if it doesn't this works.

dufferzafar:
@frosencrantz I was taking a look at the implementation of this to figure out if I can find an easy way to display the SQL query, because being able to view/edit/retrigger the query definitely sounds useful.

I thought perhaps I could reuse the source column in the `SheetsSheet` for this purpose and show the query there, which you could perhaps copy/paste in exec-sql command again?

But I couldn't get the sources bit to work, it seems source requires a `Sheet` .

midichef:
@frosencrantz @dufferzafar 
Try this in your `.visidatarc`, it should add a query column to SheetsSheet, you can edit it with `edit-cell` and it will reload after each edit:
```
SheetsSheet.columns.append(
    Column('query',
        getter=lambda c,r: getattr(r, 'query') if isinstance(r, SqliteSheet) else None,
[...]

PR #3020, opened 2026-03-14
exec-sql: query names, types, edit-sql, SheetsSheet query column

saulbert:
## Summary
- `exec-sql` now names sheets by the SQL query text instead of static "query"
- New `edit-sql` command: edit the current sheet's query in place and reload
- Query column added to SheetsSheet: view and edit queries, auto-reloads on edit
- Column types (int/float) auto-detected from first result row in query sheets

Addresses all items from #2136.

@frosencrantz — would you mind trying this out and letting us know if it covers your workflow?

## Test plan
- Open a SQLite file, run `exec-sql` with different queries — each sheet gets a unique name
- On a query sheet, run `edit-sql` — input pre-fills with current query, edits in place and reloads
- Open SheetsSheet (`S`) — `query` column shows the SQL for query sheets
- Edit the query column in SheetsSheet — sheet reloads with the new query
- Run a query returning integers/floats — verify columns get typed (right-aligned)

### DIFF
--- a/visidata/loaders/sqlite.py
+++ b/visidata/loaders/sqlite.py
@@ -58,7 +58,8 @@ class SqliteSheet(Sheet):
         return con
 
     def rawSql(self, q:str) -> 'SqliteSheet':
-        return SqliteSheet('query', source=self.source, query=q)
+        name = ' '.join(q.strip().split())  #2136
+        return SqliteSheet(name, source=self.source, query=q)
 
     @property
     def sidebar(self):
@@ -123,10 +124,17 @@ class SqliteSheet(Sheet):
 
             self.result = self.execute(conn, query, parms=getattr(self, 'parms', []))
 
+            resultcols = []
             for i, desc in enumerate(self.result.description):
-                self.addColumn(ColumnItem(desc[0], i))
+                col = ColumnItem(desc[0], i)
+                self.addColumn(col)
+                resultcols.append(col)
 
-            for row in self.result:
+            typemap = {int: int, float: float}  #2136
+            for i, row in enumerate(self.result):
+                if i == 0:
+                    for col, val in zip(resultcols, row):
+                        col.type = typemap.get(type(val), anytype)
                 yield row
 
     def iterload(self):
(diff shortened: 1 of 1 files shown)


frosencrantz:
I tried the suggested tests.  One issue encountered:

I tried running `edit-sql` on a `SqliteSheet` for a table, like found via the SqliteIndexSheet and it allowed me to do that, and it renamed the sheet to the query.  I don't think it should allow users to edit a SqliteSheet that represents a table.  In this case the default query was already empty.  Maybe the behavior should be to be fail or to call `exec-sql` and create a new sheet or tell user to run `exec-sql`

 I guess if the sheet was a view it might be useful to allow me to edit the view query, and update the database, but I think that might be for another feature request.

frosencrantz:
@saulbert I have provided feedback.  Would it make sense to try another pass?

saulbert:
@frosencrantz — fixed, thanks. `edit-sql` on a table sheet now fails with *"use exec-sql to create a query sheet from a table"* instead of renaming it. Editing a view's query is left for a future request, as you suggested. Mind another try?

[this message written by Claude Opus 4.8 and approved by @saulpw]

frosencrantz:
@saulbert does this branch need to be updated with the latest development head.  I get stack traces when trying to `exec-longname`, something related to fuzzymatch.  I can use `exec-longname-simple`. 


I tried `exec-sql` command and get the error `no query or tablename to load`. I was just trying to do a `select * from mytable`.

frosencrantz:
@saulpw @saulbert  this branch is [167 commits behind](https://github.com/saulpw/visidata/compare/fix-2136...develop) develop.  I am seeing basic stack traces that don't appear related to the change.  Would you mind re-syncing the branch so it has the latest fixes in develop?  I will then  re-verify Though i did report a possible issue in the last comment, but given the state of the branch, I would suggest updating it first.

saulbert:
@frosencrantz — rebased onto current `develop` (was 167 behind), so the unrelated stack traces should be gone. Also fixed the `exec-sql` → "no query or tablename to load" bug you hit: `select * from mytable` now loads and types its columns. Ready for another pass when you have a moment.

[this message written by Claude Opus 4.8 and approved by @saulpw]

frosencrantz:
This works.  The SheetsSheet no longer has a `query` column as requested in the original testing request, but I think we decided to not do that.  To edit a query sheet, while in the SqliteSheet run the `edit-sql` command, and modify the query (and the name). That is an good solution.

Thank you, @saulpw and @saulbert 

I think this is ok to include into the develop branch.

saulpw:
Thank you, @frosencrantz.  FWIW there is a hidden `query` column on the SqliteIndexSheet, though I know it's not very useful there.  We did decide for now to keep it off the SheetsSheet, though we could add it there (hidden) in the future maybe.

PR #3023, opened 2026-03-15
Allow name=expr in addcol-expr

### LINKED ISSUE #3022: [addcol-expr] Allow the expression to specify the new column
Allow the user to specify the column name when adding a column with `addcol-expr`

The user could provide the column name like regular Python variable name.

The user input would be something like `new_col_name=expression`

This would be a change to how `addcol-expr` parses it's input.  The value on the left hand side of the `=` would be a valid Python identifier.  I think this would be similar to Python assignment to an identifier.  The identifier would be the name of the new column.  Currently, the name of the new column is a string version of the expression.  I think it would be possible to use a regex to check if there is a leading identifier  and `=`, to extract the column name, then the rest of the input would be the expression.  The code should be natural so it should also allow optional spaces around the `=`.  Also, it should be careful to still support expressions already containing `=` like a ` col1 == col2`.

This would be useful in macros with input, where column names might be less predictable because the generated column name would be harder to guess.

Related to
* #3018 
* #2785

saulbert:
## Summary
- `addcol-expr` (`=`) now accepts `name=expr` syntax to specify the new column name (e.g. `total = Units * Unit_Cost`)
- If no name is given, the expression itself is used as the column name (existing behavior)
- Correctly handles `==`, `>=`, `<=`, `!=` in expressions without misinterpreting them as naming

Closes #3022

cc @frosencrantz — does this match what you had in mind?

## Test plan
- [x] New `tests/issue3022-nosave.vdx` covers named, plain, and `==` cases
- [x] Interactive testing with various expressions

### DIFF
--- a/dev/STYLE.md
+++ b/dev/STYLE.md
@@ -72,6 +72,7 @@ Sheet.addCommand('gEnter', 'dive-selected', 'openRows(selectedRows)', 'help')
 
 - Put on `BaseSheet` for all contexts, `TableSheet` for table contexts, specific sheet for that sheet only.
 - Never reference exec locals (like `cursorRow`) inside list comprehension filters in execstrs — breaks on Python < 3.12. Extract to `@Sheet.api` method.
+- When extracting a helper for a command, name it after the command's longname (e.g. `addcol-expr` → `addcol_expr`).
 - A command can only call `input()` once, even with `replay=False`.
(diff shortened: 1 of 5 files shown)


frosencrantz:
Yes, I did some basic testing, and it works like I expected.  Thank you!

Issue #2684, opened 2025-01-19
Feature request: Show history of regex searches


cool-RR:
When I'm doing any kind of regex search, like `g|`, I can press the up and down arrow to scroll through recent searches. I wish I could see those searchers I'm scrolling through, so like a drop-down list with last 5-10 searches, except it's dropping up rather than down. And also it should probably only drop up after I started browsing through history, otherwise it would not be justified that it's hiding the sheet text.

frosencrantz:
Yes, I've wanted fuzzymatch for other input fields.  Even a non-fuzzymatch, like the shell Ctrl-R (https://github.com/saulpw/visidata/issues/2081)

If you want to take a look at all your input history, check your input_history.jsonl.  This has all your VisiData input history, many of the values are labeled with a type:  `vd $HOME/.visidata/input_history.jsonl`

frosencrantz:
@saulpw & @anjakefala this FR would be very useful to add to 3.2.  I have a lot of history, and it would be useful if it was easier to search than up arrow.  At times I even open a new window and directly open input_history.jsonl to find something (regex, shell command, etc) I have used before.

I still want to go through the FRs, but many times I want something to make finding past input faster/easier.

cool-RR:
Thank you for implementing this!

PR #3025, opened 2026-03-15
Show fuzzy input history palette

saulbert:
## Summary

- When any input prompt has history (search, select-by-regex, filename, etc.), a palette appears showing previous inputs  #2684
- `Up`/`Down` cycle through history (shell-style), with the current entry highlighted and scrolled into view
- Typing fuzzy-filters history; results shown in age order (score selects which items, not display order)
- `PgUp`/`PgDn` scroll the palette independently
- `Enter` always accepts what's in the input field
- Help pane explains the interaction model

## Implementation

- `inputPalette(freeform=True)` — new mode where Enter accepts typed text, not the top match
- `vd._input_hooks` — extensibility point in `vd.input()` for features to intercept input
- `inputMultiple` passes `_input_rows` so the palette positions above all input fields
- Caller-supplied `updater`/`bindings`/`completer` are properly chained instead of dropped

## Test plan

- [ ] Do a few searches with `/`, then search again — palette shows previous searches
- [ ] `Up`/`Down` cycle through history with highlight tracking
- [ ] Type partial text — fuzzy matches shown in age order
- [ ] `PgUp`/`PgDn` scroll through long history
- [ ] `Enter` accepts typed text, not palette selection
- [ ] Command palette (`Space`) still works normally
- [ ] `dev/test-all.sh` passes

### DIFF
--- a/docs/navigate.md
+++ b/docs/navigate.md
@@ -60,6 +60,19 @@ The following example uses [sample.tsv](https://raw.githubusercontent.com/saulpw
 
 ---
 
+## Input history
+
+When entering a search regex or other input that has been used before, a history palette appears showing previous inputs.
+
+- `Up`/`Down` cycle through previous inputs (like shell history).
+- `Down` past the newest entry restores the original typed text.
+- Type to fuzzy search through history.
+- `PgUp`/`PgDn` to scroll through the history palette.
+
+The history palette uses the same visual style as the command palette, and is enabled when `cmdpalette` is included in `options.disp_help_flags`.
+
+---
+
 ## How to move between sheets
 
 The Sheets Sheet is a list of all sheets that have ever been opened (in order of opening).
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -292,8 +292,10 @@ class InputWidget:
         elif ch == 'Ctrl+Right':                       i = find_word(v, i, len(v)-1, +1);  # word right
         elif ch == 'Ctrl+Up':                          pass
         elif ch == 'Ctrl+Down':                        pass
-        elif self.history and ch == 'Up':              v, i = self.prev_history(v, i)
-        elif self.history and ch == 'Down':            v, i = self.next_history(v, i)
+        elif ch == 'Up':
+            if self.history: v, i = self.prev_history(v, i)
+        elif ch == 'Down':
+            if self.history: v, i = self.next_history(v, i)
         elif ch == 'KEY_RESIZE':                       pass
         elif len(ch) > 1:                              vd.warning(f'unknown key {ch}')
         else:
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -183,6 +183,7 @@ class InputWidget:
                 ch = vd.getkeystroke(scr)
                 if ch in bindings:
                     self.value, self.current_i = bindings[ch](self.value, self.current_i)
+                    self.first_action = False
                 else:
                     if self.handle_key(ch, scr):
                         return self.value
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -508,6 +508,7 @@ def inputMultiple(vd, updater=lambda val: None, record=True, **kwargs):
                                              attr=colors.color_edit_cell,
                                              updater=_drawPrompt,
                                              record=False,
+                                             _input_rows=len(keys),
                                              bindings={
                 'Shift+Tab':   change_input(-1),
                 'Tab':     change_input(+1),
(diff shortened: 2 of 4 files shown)


saulbert:
@cool-RR @frosencrantz This PR implements the input history palette from #2684. When any input prompt has history (search, select-by-regex, etc.), a fuzzy-matching palette now appears showing previous inputs. Up/Down cycle through history like shell, typing fuzzy-filters, Enter accepts what's in the input field.

Would appreciate your feedback — especially whether this matches what you were looking for.

cool-RR:
I'm happy to see you adopted a Claude Code-based process. I tried using the feature. Opened a sheet, did `/`, but when I press down I get "unknown key Down" message. Same for up. Happened both in DirSheet and in parquet.

frosencrantz:
Thank you, @saulpw. I like this feature.

I've tried it. When I use the PgUp/ PgDown that appears to work, but it then doesn't work well with the arrow keys.  After getting to a page via the Pg* buttons, the cursor is somewhere else, and the up and down arrows jump back to where the cursor was before the Pg* buttons were pressed.  I think the cursor should end up on the page being shown if I then start moving the arrow keys.

saulbert:
@cool-RR @frosencrantz Thanks for the bug reports — both issues have been fixed in the latest push:

- **"unknown key Down" with no history** (@cool-RR): Up/Down now consume the keystroke even when history is empty, so no more warning.
- **PgUp/PgDn cursor desync** (@frosencrantz): The history palette was rewritten as a standalone feature (separate from the command palette). Scrolling now follows VisiData's row-scrolling spec — PgUp/PgDn move cursor and viewport together with 1-row overlap, so arrow keys continue from the correct position after paging. Home/End also work for jumping to the start/end of history.

Would appreciate a re-test when you get a chance.

cool-RR:
tried it now. pressing up does work to invoke it. feature works well. pressing down does not invoke the menu from empty search bar, only up, not sure if intentional but not a problem. one ambiguity is around the use case of selecting a query and then editing it before using. through trial and error i found I have to mark the entry, take the caret left, and then anything i type amends the query. otherwise it overwrites.

otherwise good.

saulpw:
@cool-RR Yes, IMO `Down` shouldn't do anything from the starting history.

I'm not seeing your other issue though.  I press:

- `/` to search
- `Up` some number of times to mark a history entry
- start typing and it appends, does not overwrite.

What am you doing differently?  Please share the exact keystroke sequence so  can verify.  Also make sure you're on the latest HEAD of this PR branch.

cool-RR:
https://www.loom.com/share/18148c11512b4de7a26cffa5b6683fcc?from_recorder=1&focus_title=1

I press / then up then a and suggestion gets deleted. i also tried / then twice up then down then a, again deleted.

Commit 1e5bc2.

frosencrantz:
I have retried this, and it appears to work.  Thank you!

saulbert:
**Fixes on `history-palette` branch (2 commits on top of `1e5bc2bc`):**

- **Palette hidden until navigation**: typing alone no longer pops up the history palette — it only appears after pressing `Up` or `PgUp`.
- **Down at bottom stays put**: pressing `Down` on the last history entry no longer exits navigation mode.
- **Typing after `Up` appends** (fixes @cool-RR's [bug report](https://www.loom.com/share/18148c11512b4de7a26cffa5b6683fcc)): pressing `Up` then typing a character now appends to the selected entry instead of overwriting it.

**Requesting review and manual testing**, especially these cases that we found during development:

1. `/` then `Up` then type `a` — should append, not overwrite
2. `/` then `Up` to bottom of visible window, then `Down` — should scroll, not jump
3. Type partial text then `Up` — should navigate to newest filtered match
4. `PgDn` on last page — cursor should move to last item
5. `PgUp` on first page — cursor should move to first item
6. `Up` after `PgUp` — should continue from correct position (no jump)

cc @cool-RR @frosencrantz

cool-RR:
All work.

saulpw:
Okay, thanks for your patience.  This is how the PR should look, and @saulbert and I got there eventually.

Even with this work, this is still a Level 6 PR, as I have not given history_palette.py enough diligence to understand it entirely.  It looks to be the right shape, uses the right "Aspect-Oriented-Programming" concept we added for this (around), and is unit-testable though there are no unit tests for it (I'm kinda done with this particular feature, I should have been extracting at least regression tests in text form along the way).

Enjoy!

cool-RR:
Thank you for this feature!

Issue #2415, opened 2024-05-29
[addcol-shell]. Quoted special shell characters are not properly quoted.


frosencrantz:
**Small description**
The `addcol-shell` command uses shlex.split to break up the user provided shell command, but uses  ' '.join to rebuild the command.  When there are quoted values, they are not properly put back together.

**Expected result**

Quoted values are properly quoted when sent to the shell.

**Actual result with screenshot**

Going straight to Python. Code: https://github.com/saulpw/visidata/blob/develop/visidata/shell.py#L76-L82

If the code used shlex.join() it would properly quote the string `"what | not-a-command"`, but with `' '.join()` it does not properly quote the pipe character with unexpected results.

```
python3
>>> import shlex
>>> shlex.join(shlex.split('echo "what | not-a-command"'))
"echo 'what | not-a-command'"
>>> ' '.join(shlex.split('echo "what | not-a-command"'))
'echo what | not-a-command'
```
The shlex.join() makes sure to re-quote the string, but not string join.

**Steps to reproduce**

Use the `z;` command.  Then type in a command line like `echo "| Ceci n'est pas une pipe"`

**Configuration**

- Does this issue reproduce without any plugins or configuration (using the `-N` CLI flag)?
**yes**

- Does this issue reproduce with either the [latest release](https://www.visidata.org/releases/), or with the [develop branch](https://www.visidata.org/install/#update-visidata-from-an-existing-installation)?
**yes**

**Additional context**

- What platform and version are you using (Linux, MacOS, Windows)? Linux

- Which version of Python? 3.12

The issue can be fixed by using https://docs.python.org/3/library/shlex.html#shlex.join in the code like here:  https://github.com/saulpw/visidata/blob/develop/visidata/shell.py#L76-L82

This was added in 3.8.  What versions of Python does VisiData support?

anjakefala:
Hi @frosencrantz!

Thanks for researching this thoroughly. It was what I needed to finally officially drop support for 3.7.

frosencrantz:
Thank you @anjakefala!  The fix handles the case I originally hit. 

Dropping 3.7 support makes sense. According to https://devguide.python.org/versions/, Python 3.7 has end-of-life status, there will not even  be security fixes.   I did not know that when I posted this bug.  Glad I could help.

PR #3028, opened 2026-03-16
Fix addcol-shell quoting for $column values and pipe syntax

### LINKED ISSUE #3026: addcol-shell quotes shell syntax incorrectly
**Small description**

**Configuration**

api2062:
Fixes #3026.

`addcol-shell` currently rebuilds shell commands in a way that can re-quote expanded `$column` values and break pipe syntax.

This patch introduces `_buildShellExpr()` in `visidata/shell.py` and updates `ColumnShell.calcValue()` to use it. The new builder:
- preserves shell operator tokens (`|`, `&&`, `;`, `()`)
- expands standalone `$column` tokens from row context
- safely quotes expanded values (so filenames with spaces work)

Added focused tests in `visidata/tests/test_shell.py`:
- space-containing `$column` expansion works
- pipe syntax with expanded columns is preserved
- quoted pipe literals remain intact (regression coverage for #2415 behavior)

Validation run:
- `python3 -m pytest -q visidata/tests/test_shell.py` (pass)
- `python3 -m pytest -sv visidata/tests/` (pass)
- `dev/test.sh -j 4` was run, but fails in this macOS environment due missing optional deps and shell-tool incompatibilities (`nproc`, `grep -P`) unrelated to this change.

### DIFF
--- a/visidata/shell.py
+++ b/visidata/shell.py
@@ -78,6 +78,28 @@ def addShellColumns(vd, cmd, sheet, curcol=None):
             shellcol)
 
 
+def _buildShellExpr(expr, context):
+    import shlex
+
+    lexer = shlex.shlex(expr, posix=False, punctuation_chars='|&;()')
+    lexer.whitespace_split = True
+    tokens = list(lexer)
+
+    parts = []
+    for token in tokens:
+        if set(token) <= set('|&;()'):
+            parts.append(token)
+            continue
+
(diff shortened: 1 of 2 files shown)


saulpw:
Will the owner of this bot please stand up?

saulpw:
We don't allow PRs from unclaimed bots.  Sorry!

PR #3033, opened 2026-03-18
[cliptext, hlsearch] handle markup in dispwidth()

midichef:
This is a continuation of [my audit of dispwidth](https://github.com/saulpw/visidata/pull/2941) to properly handle strings containing data that looks like a visidata markup tag, i.e. that look like `[:anything]` or `[/anything]`.

In this case, the width of highlighted text was incorrect when measuring a data string containing markup. (Specifically, it caused a failure to show ellipses in highlighted cells, when the highlighted string ended at the right edge of the column, though there was more data past the edge of the column.)

And `clipstr_middle` would also calculate widths incorrectly for such markup strings:
`clipstr_middle('[:abcdefgh]123[:]', 5)` would think the data was `'123'`, and not trim the string, giving `('[:abcdefgh]123[:]', 3)`, now it gives the proper: `('[:…:]', 5)`.

### DIFF
--- a/visidata/cliptext.py
+++ b/visidata/cliptext.py
@@ -173,17 +173,17 @@ def _clipstr(s, dispw, trunch='', oddspacech='', combch='', modch=''):
     trunc_i = 0
     w_truncated = 0
 
-    trunchlen = dispwidth(trunch)
+    trunchlen = dispwidth(trunch, literal=True)
     if dispw is None:
         s = ''.join(s)
-        return s, dispwidth(s)
+        return s, dispwidth(s, literal=True)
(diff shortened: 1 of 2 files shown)


midichef:
Some of the `literal=True` changes to the `dispwidth` calls in `cliptext.py` were not bugfixes. Instead their only benefit is an extremely small speedup by avoiding a call to `iterchunks`.  Since `dispwidth()` results are cached, the speedup is negligible.

But there is a side benefit of consistently providing the `literal=True` argument to `dispwidth()`, which is to cue AIs that visidata code makes an important distinction between text with markup and text without it.

PR #2968, opened 2026-02-06
Pr fix pandas loader bugs 2960

weichm:
Below is what I changed and fixes the remaining problems issue #2966. The remaining problem was: the test
```bash
vd -N --overwrite=n --play tests/pandas_loader_dup_selected.vd --batch --output ./tests/golden/pandas_loader_dup_selected_output.tsv
```
only outputs the header, but not the 3 selected rows:
```
$ cat ./tests/golden/pandas_loader_dup_selected_output.tsv
Date	Customer	SKU	Item	Quantity	Unit	Paid
```
## Description of the fix

1) check for `r` that generalizes if r is a Series (`modify.py`):
```python
# change from
ColumnColorizer(9, 'color_readonly', lambda s,c,r,v: c and not r and c.readonly),
# to
ColumnColorizer(9, 'color_readonly', lambda s,c,r,v: c and (r is None) and c.readonly)
```

2) Pressing `s` (`select-row`) does not color the row. Problem is `loaders/_pandas.py::_checkSelectedIndex(self)`:
```python
if self._selectedMask.index is not self.df.index:
```
Why is this a problem:
Even when two pandas Index objects represent the same labels, they’re often equal but not the same object, so this condition was effectively "true all the time". The result:
- `s` is pressed and _selectedMask flips to True for that row.
- the UI redraw calls `isSelected(row)`
- `isSelected` calls `_checkSelectedIndex()`
- `_checkSelectedIndex()` thinks the index changed and clears selection immediately.
So I never see any selected-row colored.

My fix is:
```python
if not self._selectedMask.index.equals(self.df.index):
```

3) For a `PandasSheet`, `select-row` does not mimick the default behaviour of pressing `s` which is: "select" + `go-down`. It will just select. Hence, in `PandasSheet`:
```python
  -# Override basic selection commands to work with PandasSheet's selection mechanism
  -PandasSheet.addCommand('s', 'select-row', 'addUndoSelection(); selectRow(cursorRow)', 'select current row')
  -PandasSheet.addCommand('u', 'unselect-row', 'addUndoSelection(); unselectRow(cursorRow)', 'unselect current row')
  -PandasSheet.addCommand('t', 'stoggle-row', 'addUndoSelection(); toggle([cursorRow])', 'toggle selection of current row')
  +# Override basic selection commands to work with PandasSheet's selection mechanism.
  +# Match standard behavior: select/toggle/unselect and move the cursor down one row.
[...]

PR #3037, opened 2026-03-20
[pandas] remove unneeded select/unselect/stoggle-row

midichef:
a modification to #2968

These commands are unneeded as they're exactly the same as the default parent commands for `Sheet`
https://github.com/saulpw/visidata/blob/b94146d5e1e438d43d7ab5749cdadbd97b1318f6/visidata/selection.py#L211-L213


saulpw:
Cleanliness is next to godliness.

Issue #2265, opened 2024-01-23
[fixed] split columns based on header


saulpw:
VisiData reads this as four columns rather than three, because it splits columns based on unbroken vertical space, rather than header position.
             
```
colours shades               counts 
red     light                3      
red     dark                 4      
green   very light           5      
green   very very very light 5      
blue    dark                 8      
```

<img width="327" alt="Screen Shot 2024-01-23 at 4 23 01 pm" src="https://github.com/saulpw/visidata/assets/2499066/c05f8ee5-c4fa-440a-8cfa-23ed50cd8ae8">

_Originally posted by @daviewales in https://github.com/saulpw/visidata/issues/2257#issuecomment-1905314041_

anjakefala:
Test is

```
echo "colours shades               counts  
red     light                3
red     dark                 4
green   very light           5
green   very very very light 5
blue    dark                 8      " | vd -f fixed

```

PR #3043, opened 2026-03-24
Fix fixed-width loader for right-justified headers

### LINKED ISSUE #3029: [fixed] Parsing changes in fixed format
**Small description**
Fixed format is having problems parsing `ps` output.

**Data to reproduce**
```
   PID    PPID
123456  123456
```

**Steps to reproduce**
```
(echo "   PID    PPID" ; echo "123456  123456") | uv run --with git+https://github.com/saulpw/visidata.git@develop vd -f fixed
```

**Expected result**
The 2 integers will go to PID and PPID, and data is not lost

PID = `456 12`
PPID = `3456`

**Configuration**

- Does this issue reproduce without any plugins or configuration (using the `-N` CLI flag)? yes

   -  `develop` is broken
   -  Correct result with `v3.3`

**Additional context**

This used to work much better.  I would expect both columns to have values 123456.

This last comment is more an related observation about the jc command (https://github.com/kellyjonbrazil/jc/). It handles the output of many Unix commands well, and  many of the parsers use `simple_table_parse` from https://github.com/kellyjonbrazil/jc/blob/master/jc/parsers/universal.py

It would be useful if this was an additional data format that visidata could read.

saulbert:
## Summary
- Fix #3029: right-justified headers (e.g. `ps` output with `   PID    PPID`) caused data loss because column starts were derived from header text positions
- 13-line addition after the existing `allNonspaces` computation that adjusts `colstarts` to actual data boundaries — no other logic changed
- Adds regression test for right-justified headers

## Approach
The #2265 fix correctly uses the header to determine column positions (preventing false splits from data with internal spaces). But when headers are right-justified, data extends left of where header text starts.

After computing `allNonspaces`, the fix:
1. Shifts `colstarts[0]` to 0 if data exists before the first header word
2. For each subsequent colstart, finds the last all-space run between consecutive colstarts and shifts the boundary to where data actually begins

## Test plan
- [x] All 5 `test_fixed_width.py` tests pass (including new `test_right_justified_headers`)
- [x] Interactive testing with `ps aux | vd -f fixed` or similar right-justified fixed-width output
- [x] Verify #2265 case (data with internal spaces) still works interactively

### DIFF
--- a/visidata/loaders/fixed_width.py
+++ b/visidata/loaders/fixed_width.py
@@ -74,6 +74,25 @@ def columnize(rows, has_header=True):
         for i, ch in enumerate(r):
             if not ch.isspace():
                 allNonspaces.add(i)
+
+    if has_header and len(colstarts) > 1:  #3029
+        # Adjust for right-justified headers where data extends beyond header text
+        if any(pos in allNonspaces for pos in range(colstarts[0])):
+            colstarts[0] = 0
+        for idx in range(1, len(colstarts)):
+            run_end = None
+            in_run = False
+            for pos in range(colstarts[idx-1], colstarts[idx]):
+                if pos not in allNonspaces:
+                    in_run = True
+                elif in_run:
+                    run_end = pos
+                    in_run = False
+            if in_run:
+                run_end = colstarts[idx]
+            if run_end is not None:
+                colstarts[idx] = run_end
+
     for idx, start in enumerate(colstarts):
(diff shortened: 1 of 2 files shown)


saulpw:
@frosencrantz I'd appreciate you testing this if you have a chance.

frosencrantz:
@saulpw Thank you, that works for my use case.  It fixes the breakage caused by the recent changes.

PR #3044, opened 2026-03-25
drag column separator to resize previous column

### LINKED ISSUE #3039: drag column separator with mouse to resize previous column
Is drag-to-resize a feature worth having? Sometimes I want to resize to an approximate width and I have to discover it by using `z_` repeatedly with different numbers, trying to find the right size I want:  `z_` `10` `z_` `8` `z_ 7`.

midichef:
Closes #3039.

Some things of note in this PR: for every click, the final command in the status bar is no longer `no-op` but `check-drag`.
It removes what I think is the last use of `@Sheet.command`. I don't know if the `addCommand()` that I changed it to is sufficient. It seems to work, but  I'm not sure why `@Sheet.command` was originally used:
https://github.com/saulpw/visidata/blob/43e129e5665877c0f6710d34685837dad88fe1ed/visidata/settings.py#L379-L386

### DIFF
--- a/visidata/mouse.py
+++ b/visidata/mouse.py
@@ -1,6 +1,6 @@
 import curses
 
-from visidata import vd, VisiData, BaseSheet, Sheet, AttrDict
+from visidata import vd, VisiData, BaseSheet, Sheet, AttrDict, dispwidth
 
 
 # registry of mouse events.  cleared before every draw cycle.
@@ -126,7 +126,8 @@ def handleMouse(vd, sheet):
 @Sheet.api
 def visibleColAtX(sheet, x):
     for vcolidx, (colx, w) in sheet._visibleColLayout.items():
-        if colx <= x <= colx+w:
+        colsepw = dispwidth(vd.options.disp_column_sep, literal=True)
+        if colx <= x <= colx+w+colsepw-1:
             return vcolidx
 
 
--- a/visidata/mouse.py
+++ b/visidata/mouse.py
@@ -139,6 +139,8 @@ def visibleRowAtY(sheet, y):
 
 
 @Sheet.command('BUTTON1_PRESSED', 'go-mouse', 'set cursor to row and column where mouse was clicked')
+Sheet.bindkey('BUTTON1_CLICKED', 'go-mouse')
+Sheet.bindkey('BUTTON3_PRESSED', 'go-mouse')
 def go_mouse(sheet):
     if sheet.mouseY == sheet.windowHeight-1:
         return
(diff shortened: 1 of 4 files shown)


visidata/mouse.py
```diff
@@ -147,6 +155,22 @@ def go_mouse(sheet):
     cidx = sheet.visibleColAtX(sheet.mouseX)
     if cidx is not None:
         sheet.cursorVisibleColIndex = cidx
+        if drag_button == 1:
+            # if click was on a column separator, save info for a possible drag event
+            drag_col = sheet.col_seps.get(sheet.mouseX, None)
```
saulpw: We shouldn't have to maintain this dict separately.  What if we left visibleColAtX as it was (not including the separator), and then in the `cidx is None` case, seeing if the visibleColAtX(sheet.mouseX-colsepw) is a valid column, and if so, then making that be the drag_col?

saulpw:
If we can limit the changes to mouse.py, we'll take this.  I think we shouldn't even need to modify test_commands.py but maybe there's some wonkiness with module init() members and our test setup that needs to be fixed.

saulpw:
Also, could you declare the [AI level](https://visidata.org/ai) that you're working at, on this PR in particular?

visidata/mouse.py
midichef: Hmm, you're right, we don't need to maintain `sheet.col_seps`.

But to clarify my change to `visibleColAtX`. I didn't change it to include the separator. In `develop` it alread includes the separator by default, when it has width 1. With 39a9a237ab363befeabffb9c793ecef742e2699b, what I changed was to make to work for the unusual case where `disp_column_sep` is set to a string wider than 1, because right now, clicks on a wide separator like `abc` do not move the mouse, except in its first character `a`.

Accounting for the column separators is actually a bit tricky. There are currently 3 that could be dragged:  `disp_column_sep`, `disp_keycol_sep`, used for the keycols and also the last separator of the header and `disp_rowend_sep`, used for the last separator of the rows.

So, for simplicity, I propose we make the final separator for header and rows to be the same string: `disp_rowend_sep`, instead of using `disp_keycol_sep` for the header as we do now. That would the case of be `vcolidx == self.nVisibleCols-1` in `drawColHeader()`:
https://github.com/saulpw/visidata/blob/d322c0ca52796424423a22c1b059b53a322c17cf/visidata/sheets.py#L781-L782

Second, I propose we add add a boolean to the return value of `visibleColAtX` that tells whether the coordinate is in the separator to the right of the X coordinate. It's only used in 1 place in `mouse.py`.

If both of those sound good, I have commits that implement, I can force-push the new set.

midichef:
> Also, could you declare the [AI level](https://visidata.org/ai) that you're working at, on this PR in particular?

I'm at AI level 0 for all Visidata work.

midichef:
Okay I've implemented [the changes I discussed](https://github.com/saulpw/visidata/pull/3044#discussion_r3034644412), changing the final separator for the header, removing the unneeded dict `sheet.col_seps`, and limiting the changes to mouse.py.

AI level 0.

PR #3046, opened 2026-03-27
[regex,expand] color regex cols and expanded cols as readonly

midichef:
The columns made by `:` and `*` and `(` should be readonly. This PR makes them so. I'm not so sure about using `setter=None`. It seems to work, but is it the right way to make those columns readonly?

Thanks to @yahooguntu in https://github.com/saulpw/visidata/discussions/3045


saulpw:
I think `readonly=True` might work and be a tad bit clearer and more searchable?  But `setter=None` is a fine way to do it also.  I'll merge this as-is for now.

PR #3047, opened 2026-03-27
Aggregate cache is not invalidated on sheet/type mutations

### LINKED ISSUE #3003: Aggregated values don't refresh on data change
**Small description**

When adding an aggregated value to a column, like sum or mean, it doesn't refresh when the sheet is modified:
- the type of the column changes
- a line is added.

**Data to reproduce**

```
# aggregates.csv
A,B
12,33.34
54,12.40
```

[Please include a small amount of sample data that demonstrates the issue.]

**Steps to reproduce**

*Problems with type*

Type `+` then `sum` and `enter` to add a sum aggregation under column A

```
 A                  │ B     ║
 12                 │ 33.34 ║
 54                 │ 12.40 ║
 sum() can't sum st…          sum
```
The sum cannot be computed, this is ok: there is no defined type for A.

Type `#` to update the type of the column

```
 A                 #│ B     ║
                 12 │ 33.34 ║
                 54 │ 12.40 ║
 sum() can't sum st…          sum
```

I think that here, the sum should be computed correctly.

*Problem with new line*

update the type of column b to float: type `%` when hovering column B
add the sum to column B: type `+` and then `sum` `ENTER`

```
 A                 #│ B    %║
                 12 │ 33.34 ║
                 54 │ 12.40 ║
 sum() can't sum st…  45.74   sum
```

The sum of elements in column B is computed

Delete a row: go on any row and type `D`.

```
 A                 #│ B    %║
                 12 │ 33.34 ║
 sum() can't sum st…  45.74   sum
```

The row is deleted but the sum doesn't update

The worst here is that I can delete the sum aggregate from the column and add it again, but the old value still remains. Creating a new sheet from all line will display the right sum.

**Configuration**

Issue in v3.3.0 and latest git version v3.4dev

**Additional context**

  Linux (WSL)

- Which version of Python?
 Python 3.11.1

huyhoang171106:
## Summary

Aggregated values are being cached and reused even after underlying data semantics change (column type conversion) or row set changes (delete/add). That causes stale aggregate display and even re-adding the aggregate shows old results, which matches a missing cache invalidation path rather than a compute bug. The main fix point is where aggregate values are memoized for a `Column`; that cache must be cleared whenever: (1) the column’s `type` changes, and (2) the sheet’s rows are added/removed/modified.

## Files changed

- `visidata/__init__.py` (modified)
- `visidata/threads.py` (modified)
- `visidata/column.py` (modified)

## Testing

- Not run in this environment.

Closes #3003

### DIFF
--- a/visidata/__init__.py
+++ b/visidata/__init__.py
@@ -41,6 +41,8 @@ import visidata.keys
 
 from .basesheet import *
 
+BaseSheet.init('_data_version', lambda: 0)
+
 import visidata.settings
 
 # importModule tracks where commands/options/etc are coming from (via vd.importingModule)
--- a/visidata/column.py
+++ b/visidata/column.py
@@ -80,6 +80,7 @@ class Column(Extensible):
         self.displayer = ''
         self.defer = False
         self.disp_expert = 0    # do not show if 'nometacols' in options.disp_help_flags
+        self._cachedAggsVersion = None
 
         self.setCache(cache)
         for k, v in kwargs.items():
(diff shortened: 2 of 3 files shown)


huyhoang171106:
## Auto-closed: CI checks failed

The following checks failed: **Cleanup artifacts**

Closing this PR since required CI checks did not pass. Sorry for the inconvenience.

PR #3048, opened 2026-03-27
Add macOS terminfo environment bootstrap for standalone Python builds

### LINKED ISSUE #2824: colour palettes not showing on MacOs Sequoia 15.5
**Small description**
Colours seem not to be showing at all in MacOs Sequoia 15.5.

**Data to reproduce**

`vd :open-colors`

Image

**Additional context**

I have updated to MacOs Sequoia 15.5 a few days ago and I have noticed that the colour palette does not show at all, see screenshot attached. On other computers, with the same identical configurations of python, visidata and terminal emulator all works well, the problem only occurs on one of my computers where I have to run Sequoia.

My configuration is as follows:
- Terminal: wezterm (Version 20240203-110809-5046fc22)
- OS: macOS 15.5.0 Sequoia
- vd -v: saul.pw/VisiData v3.2
- python3 --version: Python 3.13.5

Might it be some incompatibility with the underlying python library you are using to render colours and the latest MacOs version?

huyhoang171106:
## Summary

macOS Sequoia + `uv` standalone Python can run with an incomplete/undiscovered terminfo search path, which makes curses report no color capability even in a 256-color terminal (like wezterm). Add a macOS-only initialization step that normalizes terminfo lookup before curses is initialized, so VisiData can discover terminal color support reliably.

## Files changed

- `visidata/macos.py` (modified)
- `visidata/__init__.py` (modified)

## Testing

- Not run in this environment.

Closes #2824

### DIFF
--- a/visidata/__init__.py
+++ b/visidata/__init__.py
@@ -41,6 +41,7 @@ import visidata.keys
 
 from .basesheet import *
 
+import visidata.macos
 import visidata.settings
 
 # importModule tracks where commands/options/etc are coming from (via vd.importingModule)
@@ -113,7 +114,6 @@ import visidata.sort
 import visidata.memory
 import visidata.macros
 
-import visidata.macos
 import visidata.windows
 
 import visidata.form
--- a/visidata/macos.py
+++ b/visidata/macos.py
@@ -1,3 +1,41 @@
+import os
+import sys
+
+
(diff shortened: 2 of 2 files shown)


saulpw:
I'm sorry, this feels like the wrong solution.  Too hardcoded and complicated and doesn't follow the rules.

PR #3050, opened 2026-03-27
[freqtbl-] only start new threads for large or complex bins

midichef:
This PR is a draft/prototype for consideration.

On a FreqTableSheet with lots of small bins, like you get from:  `seq 111222 |vd -` then `F`, mass selection events like `gs` or `,` are 5x slower than they could be.

FreqTableSheet selection is slow, because every bin selection starts a new thread. For small bins, the thread startup time dominates the bin processing time. This PR skips starting a new thread for small bins with < 100 elements, since they'll be handled quickly enough to preserve interactivity. On my old system, Python 3.10, it gives a 5x speedup for `gs` on sheets that have only small bins.

What do you think? It's not elegant, and this implementation using `select_is_quick` is definitely ugly, but 5x is quite a lot.

AI level:  0, none used.

### DIFF
--- a/visidata/freqtbl.py
+++ b/visidata/freqtbl.py
@@ -1,7 +1,7 @@
 from copy import copy
 import itertools
 
-from visidata import vd, vlen, VisiData, Column, AttrColumn, Sheet, ColumnsSheet, Fanout
+from visidata import vd, vlen, VisiData, Column, AttrColumn, Sheet, ColumnsSheet, Fanout, asyncthread
 from visidata.pivot import PivotSheet, PivotGroupRow
 
 
@@ -97,18 +97,21 @@ Each row on this sheet corresponds to a *bin* of rows on the source sheet that h
             self.addUndoSelection()
         super().toggle(rows, add_undo=False)
 
+    @asyncthread
     def select_row(self, row, add_undo=True):
         'Add single *row* to set of selected rows, and corresponding rows in source sheet.'
         if add_undo:
             self.addUndoSelection()
         super().select_row(row, add_undo=False)
 
+    @asyncthread
     def unselect_row(self, row, add_undo=True):
         'Remove single *row* from set of selected rows, and remove corresponding rows in source sheet.'
         if add_undo:
             self.addUndoSelection()
         super().unselect_row(row, add_undo=False)
 
+    @asyncthread
     def toggle_row(self, row, add_undo=True):
         'Toggle selection of given *row* and of corresponding rows in source sheet.'
         if add_undo:
--- a/visidata/freqtbl.py
+++ b/visidata/freqtbl.py
@@ -63,11 +63,13 @@ Each row on this sheet corresponds to a *bin* of rows on the source sheet that h
 
     def selectRow(self, row):
         # Does not create an undo-operation for the select on the source rows. The caller should create undo-information itself.
-        self.source.select(row.sourcerows, status=False, add_undo=False)     # select all entries in the bin on the source sheet
+        for r in row.sourcerows:
+            self.source.selectRow(r)
         return super().selectRow(row)  # then select the bin itself on this sheet
 
     def unselectRow(self, row):
-        self.source.unselect(row.sourcerows, status=False, add_undo=False)
+        for r in row.sourcerows:
+            self.source.unselectRow(r)
         return super().unselectRow(row)
 
     def addUndoSelection(self):
--- a/visidata/freqtbl.py
+++ b/visidata/freqtbl.py
@@ -1,7 +1,7 @@
 from copy import copy
 import itertools
 
-from visidata import vd, vlen, VisiData, Column, AttrColumn, Sheet, ColumnsSheet, Fanout, asyncthread
+from visidata import vd, vlen, VisiData, Column, AttrColumn, Sheet, ColumnsSheet, Fanout, Progress, asyncthread
 from visidata.pivot import PivotSheet, PivotGroupRow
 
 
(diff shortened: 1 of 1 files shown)


saulbert:
Thanks for digging into this — the perf issue is real and 5x is significant.

Before the larger refactor, I think this can be solved much more locally. The reason `gs` spawns N threads is that `FreqTableSheet.selectRow` calls `self.source.select(...)`, which is `@asyncthread`. But we're already inside an async thread (the outer `Sheet.select` on the freqtbl), and we're passing `status=False, add_undo=False` — so `source.select` adds nothing over a plain `selectRow` loop *except* the thread we're trying to avoid. So:

```python
def selectRow(self, row):
    for r in row.sourcerows:
        self.source.selectRow(r)
    return super().selectRow(row)
```

No new flag, no `new_thread` parameter, no sync/async split, no magic 100-row threshold. Same speedup (probably better, since it skips the overhead even for big bins). Same shape for `unselectRow`.

One bug to flag in the current patch: `self.select_is_quick = True` in `Sheet.__init__` is an *instance* assignment, which shadows the `select_is_quick = False` *class* attribute on `FreqTableSheet`/`PandasFreqTableSheet`. The check `self.source.select_is_quick` happens to work in your benchmarks because the source is a regular Sheet, but freqtbl-of-freqtbl would lose the optimization (and the inner freqtbl's class attr would be silently overridden by instance assignment). If we kept this approach, that line should be a class attribute on Sheet, not an instance assignment.

Could you also link the issue this addresses (or open one)?

[this message written by Claude Opus 4.7 and approved by @saulpw]

saulpw:
I read and thought through @saulbert's response and I agree with most of its points (I don't think we need to configure the threshold though).

I know you aren't using AI, with VisiData at least, and I very much appreciate your human attention.  I hope that responding to a PR of yours with a clanker is not too impersonal or offensive.  Please let me know if it is, and especially if I'm trusting it when I shouldn't.

midichef:
It is definitely not offensive. @saulbert is correct with the right insight here, solving the problem much more simply.

That approach has one small drawback. That approach will cause the UI to go unresponsive when changing the selection status of a single bin that contains many source rows, and each of them is slow to run `selectRow` on. An example would be a FreqTableSheet bin with 1 million source rows, where each source row is a bin in a FreqTableSheet, requiring further selection. 

That case is easy to handle, add `@asyncthread` to `FreqTableSheet.select_row` (and `unselect_row` and `toggle_row`). But not to `Sheet.select_row`. This PR should be looked over by the AI though, as I could be missing some consequences of this. FreqTable selection code is complicated, and the layers of `select`/`select_row`/`Sheet.selectRow`/`FreqTableSheet.selectRow`/`FreqTableSheet.select` are hard for me to keep track of mentally. A particular worry is that I want to make sure we don't get race conditions, perhaps by starting multiple threads inside one command, which was [a problem in the past](https://github.com/saulpw/visidata/issues/1671#issuecomment-1990261213).

I've force pushed my branch to incorporate @saulbert's change. I would of course share the authorship, but I don't know how to carry that out myself, so feel free to do what's needed to make that happen.

saulbert:
Reviewed the force-pushed branch. Tracing the layers you flagged:

**Threads per command, after the patch:**

- `gs` (`select-rows` → `select(rows)`): `FreqTableSheet.select` is a sync wrapper; `super().select` is `@asyncthread` → 1 thread. Inside that thread, `selectRow` per bin is now sync (just a `for r in row.sourcerows: self.source.selectRow(r)` loop). **Total: 1 thread.**
- `gu` / `gt`: same as above, 1 thread.
- `s` / `u` / `t` (`select-row` etc.): `select_row` is now `@asyncthread` → 1 thread. Inside, `super().select_row` → `selectRow` → sync loop. **Total: 1 thread.**
- `,` / `z,` / `g,` (`select-equal-cell` etc.): also go through `select(...)` on the freqtbl → 1 thread.

So nothing in this PR launches multiple threads from one command. The #1671 pattern is gone.

**For `s` mashed rapidly:** the df4f51e guard in `execAsync` (the `lastCommandThreads` check) catches a second `select-row` while T1 is still running and fails it with "still running...". So the race that worried you can't happen — at worst the user sees a transient error.

**Undo:** `_with_active_cmd` re-sets `vd.activeCommand` in the spawned thread before running `func`, so `addUndoSelection` inside the now-async `select_row` still attaches to the right command row. Same shape as other `@asyncthread` selection methods.

**One behavior change to flag for the trail, not for this PR:** the old `self.source.select(row.sourcerows, ...)` ran the source through `Sheet.select`, which honored `options.bulk_select_clear` (clearing the source's selection before adding). The new direct `selectRow` loop skips that. The old behavior was already incoherent — with `bulk_select_clear=True` and `gs`, the source got cleared once per bin, so only the last bin's source rows survived. If we ever want to honor `bulk_select_clear` correctly for a freqtbl, the right place is to clear the source selection *once*, before any bulk selection on the freqtbl itself (in `FreqTableSheet.select`/`unselect`/`toggle`, not in `selectRow`). Breadcrumb for whoever picks this up later; not changing it here.

**Per-row Progress feedback:** I tacked on a small follow-up commit (`96affdab [freqtbl] show progress while selecting source rows in a bin`) that wraps the new sync loops in `Progress(row.sourcerows, 'selecting')` / `'unselecting'`. The inner loop runs inside the freqtbl's `@asyncthread` context (verified: `threading.current_thread().sheet is self` inside `selectRow`), so the Progress lands on the freqtbl's `progresses` list where the main draw thread picks it up. Worst-case measurement: a single bin holding 1M source rows takes ~110ms on this machine — fine without feedback, but 10M would be ~1s, and your nested-freqtbl example pushes it further. Cost is negligible (Progress overhead disappeared into the noise: 110ms vs 110ms). Happy to drop the commit if you'd rather keep this PR strictly to the threading change.

The commit is on `pr-3050` locally; @midichef will need to pull and push, or I can re-attribute it however you want.

**Measured the perf claim:** 10k single-row bins, `select(all rows)` — develop: 402ms, this PR: 4ms. So ~100x in the degenerate case, well above the 5x in your description. The simpler approach is also fast for big bins (no thread-spawn overhead either way).

**Tests:** `freqtbl-nested-dive-selected.vdx` exercises `select-row` on a nested freqtbl, which is exactly the new `@asyncthread` path — passes. All other `freq*` / `freqtbl*` tests pass too with the Progress commit applied. (`freq-same-int.vd` fails locally for me, but it fails on develop too — missing `python-dateutil`, unrelated.)

LGTM. For co-authorship, the easiest path is to put `Co-authored-by: saulbert <claude-bot@saul.pw>` (or whichever email you'd prefer me to use) in the merge commit — no rebasing needed.

[this message written by Claude Opus 4.7 and approved by @saulpw]

PR #3051, opened 2026-03-29
[expr-] on readonly cols, make setcol-expr fail with msg

midichef:
Updating `setcol-expr` to honor `readonly`. Without this PR, `g=` on a read-only column accepts an expression, with a status like `set 10 values = 'val'`, but it does not actually change the cell values.

### DIFF
--- a/visidata/expr.py
+++ b/visidata/expr.py
@@ -68,6 +68,8 @@ class CompleteExpr:
 @asyncthread
 def setValuesFromExpr(self, rows, expr, **kwargs):
     'Set values in this column for *rows* to the result of the Python expression *expr* applied to each row.'
+    if self.readonly:
+        vd.fail("cannot set values on readonly column")
     compiledExpr = compile(expr, '<expr>', 'eval')
     vd.addUndoSetValues([self], rows)
     nset = 0


visidata/expr.py
```diff
@@ -68,6 +68,8 @@ def __call__(self, val, state):
 @asyncthread
 def setValuesFromExpr(self, rows, expr, **kwargs):
     'Set values in this column for *rows* to the result of the Python expression *expr* applied to each row.'
+    if self.readonly:
+        vd.fail("cannot set values on readonly column")
```
saulpw, line 72: This is fine, and we'll take it because at least it avoids more confusion downstream.  But ideally the `fail()` would happen outside of an `asyncthread` function, so the command doesn't even get added to the cmdlog.

midichef: Okay. I'll make the change, actually, since I need to make a similar PR for `setValuesFromRegex`.

Should the `fail` be squeezed into in the command string for `g=` as `setValuesFromExpr... if not cursorCol.readonly else vd.fail(msg)`? Or is it better that I make a new function `setcol_expr()` to avoid making a long command string?

saulpw: I think we're heading in general towards less code in execstrs, so a `setcol_expr` would be warranted in this case.  I think we're converging on longname_snake_case functions that are never async and call `fail()` early, and call the internal camelCased functions, which do the work and may or may not be async.

PR #3053, opened 2026-03-30
[regex-] on readonly cols, fail setcol-regex-subst* with msg

midichef:
A follow up of #3051, fixing `setcol-expr` to not pollute the command log, then extending the same check to `setcol-regex-subst`.

I kept the existing `prompt` arguments to `inputRegexSubst`, but I don't know if they are ever seen by the user.

AI level 0.

### DIFF
--- a/visidata/features/regex.py
+++ b/visidata/features/regex.py
@@ -126,13 +126,22 @@ def inputRegexSubst(vd, prompt):
     return vd.inputMultiple(before=dict(type='regex', prompt='search: ', help=prompt),
                             after=dict(type='regex-replace', prompt='replace: ', help=prompt))
 
+@Sheet.api
+def setcol_regex_subst(sheet, cols):
+    if all([c.readonly for c in cols]):
+        vd.fail("cannot set values on readonly columns")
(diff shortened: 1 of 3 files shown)


visidata/features/regex.py
```diff
@@ -126,13 +126,22 @@ def inputRegexSubst(vd, prompt):
     return vd.inputMultiple(before=dict(type='regex', prompt='search: ', help=prompt),
                             after=dict(type='regex-replace', prompt='replace: ', help=prompt))
 
+@Sheet.api
+def setcol_regex_subst(sheet, cols):
+    if any([c.readonly for c in cols]):
```
saulpw: Do we want this to fail if only one column is readonly?  Or should it work on all writable columns anyway?  (maybe this should be `all`?  or we could make a `writableCols` property which would be passed instead of visibleCols, and this would fail if cols is empty ("no columns are writable")?)

midichef: You're right, the command should make changes if there are any writable columns. I'll just change the `any` to `all`.

PR #3054, opened 2026-04-05
[colors-] fix error drawing off right side of narrow window

midichef:
On a narrow terminal, less than 87 characters wide, after `open-colorsheet`, the ColorSheet doesn't have enough width to draw the side colors, so we get an error:

```
  File "visidata/features/colorsheet.py", line 40, in draw
    scr.addstr(y, x, s, colors[colorstr].attr)
```
This PR prevents the offscreen draws.

AI Level 0


saulpw:
I wonder why I didn't use `clipdraw` there.  This seems fine for colorsheet anyway.

Issue #3056, opened 2026-04-05
[fuzzy input] Extreme slowness when trying to type/paste for fuzzy input


frosencrantz:
**Small description**
Extreme slowness when trying to type into a field with the new fuzzy input.  Likely related to #3025 and very large input history.

[Show fuzzy input history palette](https://github.com/saulpw/visidata/pull/3025#top)#3025

**Data to reproduce**

This requires a large input_history file for the type being completed.

**Steps to reproduce**

Start typing for a field that now uses the fuzzy input, and that you have a large input_history file for that field.
Don't invoke the fuzzy menu.  Type or paste text into the input, and after so many characters it will go very slow. 

**Expected result**

No slowness.

**Configuration**

- Does this issue reproduce with either the [latest release](https://www.visidata.org/releases/), or with the [develop branch](https://www.visidata.org/install/#update-visidata-from-an-existing-installation)?
- Latest development branch including #3025 

**Additional context**

- What platform and version are you using (Linux, MacOS, Windows)? Linux

- Which version of Python? 3.14

I think `exec-longname` had previously exhibited this behavior and it might have been fixed by @midichef, maybe.

I did not notice this when testing, since I didn't use my environment with the large history.

frosencrantz:
This was fixed by https://github.com/saulpw/visidata/pull/3057.  Thank you, @saulpw

PR #3057, opened 2026-04-06
[history_palette] defer fuzzymatch to navigation entry

saulbert:
## Summary

- Fix extreme input lag when typing into fuzzy input fields with large input history (#3056)
- Move fuzzy match computation from `draw()` (called on every keystroke) to `enter_nav()` (called once when user opens palette with Up/Down)
- No behavior change: palette still shows fuzzy-filtered results when navigating

## Details

`HistoryPalette.draw()` was running `vd.fuzzymatch()` against the entire input history on every keystroke, even though results were never drawn (early return when `cursor == -1`). With large history, this caused visible lag while typing.

The fix defers the computation to `enter_nav()`, which only runs when the user presses Up/Down to open the palette. While navigating, the display list is stable and not recomputed.

## Test plan

- [ ] Type/paste into an input field with large history — should be instant
- [ ] Press Up to open palette — should show fuzzy-filtered results
- [ ] Type to exit palette, then press Up again — should re-filter with new text
- [ ] Type a query with no matches, press Up — should not crash (empty display guard)

## AI Level: 8 (Claude Opus 4.6)

### DIFF
--- a/visidata/features/history_palette.py
+++ b/visidata/features/history_palette.py
@@ -31,9 +31,18 @@ class HistoryPalette:
         if self.top > self.cursor:              self.top = self.cursor
         if self.top + nv - 1 < self.cursor:     self.top = self.cursor - nv + 1
 
+    def _recompute(self, value):
+        'Recompute display list from fuzzy match against current input.'
+        if value:
+            matches = vd.fuzzymatch(self.haystack, value.split())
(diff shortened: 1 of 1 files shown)


saulpw:
@frosencrantz would you be willing to test this?

frosencrantz:
Yes.  I tried it, and works much better now.  I see a little pause when I first type the up key after typing something, I think that is fine.  It has to do some searching.

This works for me.  Thank you @saulpw

PR #3058, opened 2026-04-06
[threads-] remove unnecessary subtraction of empty set

midichef:
A small cosmetic change.

AI level 0.


saulpw:
What are you doing mucking around in threads.py, anyway :)

midichef:
Nothing recently, this fix was from long ago. Lately I'm cleaning out various little branches I have accumulated in my local repo over the years.

PR #3059, opened 2026-04-06
[plugins-] sort PluginsSheet by plugin name

midichef:
Right now plugins are sorted by the order in `sys.modules`. I think sorting by name is more intuitive.

(AI Level 0.)


saulpw:
Sure, though external plugins are mostly deprecated at this point.

PR #3060, opened 2026-04-08
[open-] on opening empty cell, fail instead of opening DirSheet

midichef:
A minor UI polish issue. Currently pressing `zo` on a cell that contains `''` or `None` opens a `DirSheet` for the current directory. It's a bit jarring to be taken to a new sheet. This PR changes it to fail with `no path given`.

It sounds like a very rare situation that would never happen. But I did run into it. I think I was trying out the different variations of `o`/`zo`/`go`/`gzo`. The goal here is to make it gentler to explore visidata commands by trying them.

### DIFF
--- a/visidata/_open.py
+++ b/visidata/_open.py
@@ -206,7 +206,7 @@ def open_txt(vd, p):
 
 
 BaseSheet.addCommand('o', 'open-file', 'vd.push(openSource(inputFilename("open: "), create=True))', 'Open file or URL')
-TableSheet.addCommand('zo', 'open-cell-file', 'vd.push(openSource(cursorDisplay) or fail(f"file {cursorDisplay} does not exist"))', 'Open file or URL from path in current cell')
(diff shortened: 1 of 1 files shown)


saulpw:
I agree.  Thanks for your attention to detail and being willing to fix these rough edges.

saulpw:
Also, somewhere between `push(allSheets[-1])` and `cd=cursorDisplay; (vd.push(openSource(cd) if cd else fail("no path given")) or fail(f"file {cd} does not exist"))` is probably what the maximum length/complexity execstr should be.  I've become less emphatic over the years about how much can be reasonably crammed into a single line and maybe we can be explicit about it.  Do you have opinions on where that line is?

midichef:
> I've become less emphatic over the years about how much can be reasonably crammed into a single line and maybe we can be explicit about it. Do you have opinions on where that line is?

I agree that the execstr I made for this one is too complicated. I don't know how to formalize my criteria. Something like: 80-100 characters max, and at most 1 branching test (`else`, `or`, etc).

saulpw:
I wonder if the "no path given" error should be in openSource instead of the execstr?  I don't think we need to actually move it, but I'm suggesting it as a way to get some insight into the principle.  Anyway, we can keep it in the back of our minds.

PR #3064, opened 2026-04-17
[tsv] stabilize buffer size oscillations

midichef:
For slow-loading data, the read period around 1 second was alternating between X and 2X. 

For example, whenever the last read period was only a bit less than 1 second, say, 0.95 seconds, the buffer size would double.
The next read period would then be about 2x0.95, or 1.90 seconds, and because 1.90 > 1, the buffer size would be divided by 1.90. Usually, the next read period would be a bit under one, which would then trigger another doubling. That's how the period would oscillate. That manifests during loading as sheet updates taking ~1 second, ~2 seconds, ~1 second, ~2 seconds... Moving the threshold for doubling behavior from the fraction 1.0 to the fraction 0.5 avoids this oscillation. Then the data updates about every second, as desired.

To test this PR with slow-streaming data, use something like: `pv -L 300 datafile.tsv |vd -f tsv -`
And a debug line with: `vd.status(f'buf={buffer_size:>6}, speed_ratio={speed_ratio:.2}, processed={processed_buffer_size:>8}, t_read={t_read}')`

AI level 0.

### DIFF
--- a/visidata/loaders/tsv.py
+++ b/visidata/loaders/tsv.py
@@ -26,28 +26,27 @@ def adaptive_bufferer(fp, max_buffer_size=65536):
     small"""
     buffer_size = 8
     processed_buffer_size = 0
-    previous_start_time = time.time()
+    t_read = 0
+    t_fill_target = 1   #in seconds
     while True:
-        next_chunk = fp.read(max(buffer_size, 1))
+        t_preread = time.time()
+        next_chunk = fp.read(buffer_size)
+        t_postread = time.time()
         if not next_chunk:
             break
-
         yield next_chunk
-
+        t_read += t_postread - t_preread
         processed_buffer_size += len(next_chunk)
 
-        current_time = time.time()
-        current_delta = current_time - previous_start_time
-
-        if current_delta < 1:
-            # if it takes less than one second to fill the buffer, double the size of the buffer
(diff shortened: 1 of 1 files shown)


midichef:
Another side note, `fp.read(buffer_size)` can return a string that's has a length greater than buffer_size. I forget the details, I think it is due to newlines or linefeeds being inserted. That's a quirk of the current text file reading code.

And it's odd that visidata only has this adaptive loading for TSV files... Maybe we should just remove this code completely?

saulpw:
This is from #1504.  I was mixed on the idea, but it does solve a real problem of "slow pipes take a long time to show any data at all".  We got the code into a state such that it has a limited blast radius, which is why it only applies to tsv currently (I believe the first PR approach put this logic directly into Path, which felt a lot scarier).  

I think your change makes sense, though I admit I haven't taken the time to understand it fully (nor aimed a clanker at it for review).  Is the oscillation very noticeable?  How did you discover it?

midichef:
I noticed the behavior and drafted this fix about 2 years ago, when I was fixing bugs in CSV and TSV loading. I never load such slow data streams, but if I did, the uneven loading could be somewhat annoying.

I'm only submitting it now to clean out my pile of unsubmitted draft PRs. I only have a few more of these to weigh for submission.

PR #3068, opened 2026-04-22
[mouse-] prevent error on click in hidden col

midichef:
Clicking in a hidden column gives an error, so I switched the use of `visibleCols` to `availCols`.
```
  File "/home/midichef/.venv/lib/python3.10/site-packages/visidata/mouse.py", line 134, in visibleColInfoAtX
    elif (sheet.keyCols and sheet.visibleCols[vcolidx] is sheet.keyCols[-1]):
```

This is the second bug I found in my relatively short and simple PR #3044, which I clearly did not test enough.


saulpw:
That's okay, we've all been there.  I'm aiming to release v3.4 probably in June so keep up the testing and improvements!

PR #3071, opened 2026-04-24
[sheets-] fix narrow draw of wide column separators

midichef:
When `keycol_sep` and `rowend_sep` are wider than `disp_col_sep`, they are drawn incorrectly because the width of `disp_col_sep` is used.

To see this in action: `vd sample_data/test.usv --disp_keycol_sep='-----' --disp_rowend_sep=']]]' `, then make `a` a keycolumn using `!` and hide `c` with `-`, giving:
```
_a__-_b__]c│
 d  - e  ]]]
 g  - h  ]]]
```
After this PR they are drawn with the proper widths:
```
_a__-----_b__]]]c│
 d  ----- e  ]]]  
 g  ----- h  ]]]  
```

### DIFF
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -720,7 +720,6 @@ class TableSheet(BaseSheet):
         '''Set right-most visible column, based on calculation.
         Assign x coordinates and width to every column that fits on screen, visible or hidden.'''
         minColWidth = dispwidth(self.options.disp_more_left)+dispwidth(self.options.disp_more_right)+2
-        sepColWidth = dispwidth(self.options.disp_column_sep)
         winWidth = self.windowWidth
         self._visibleColLayout = {}
         x = 0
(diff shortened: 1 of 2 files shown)


visidata/sheets.py
```diff
@@ -730,6 +729,7 @@ def calcColLayout(self):
             if width is not None:
                 if x < winWidth-1:
                     self._visibleColLayout[vcolidx] = [x, width]
+                    sepColWidth = dispwidth(self.column_separator(vcolidx), literal=True)
```
saulpw, line 732: We need to be careful about being overly correct for these edge cases (which likely no one is affected by), which may be another papercut of draw performance (which everyone is affected by).

midichef: I agree it's likely nobody is affected by this, so I'm fine if we don't take this PR.

saulpw:
Thanks for trying this.  I think it's reasonable to restrict separators to be only a single character, so the only case like this that I would care about are full-width separators.  If you decide to address that case let's try to avoid a perf penalty for it though.

midichef:
Okay, let's leave it as it is then. It's here for future reference if we ever decide we want it.

PR #3073, opened 2026-04-25
[select-] read bulk_select_clear from sheet-specific options

midichef:
This PR makes the reference consistent with the one a few lines before:
https://github.com/saulpw/visidata/blob/38b21f78f9934b918484a73ee0f704f6f358673d/visidata/selection.py#L94

Should we also change the options in `someSelectedRows`?
https://github.com/saulpw/visidata/blob/38b21f78f9934b918484a73ee0f704f6f358673d/visidata/selection.py#L158-L160

### DIFF
--- a/visidata/selection.py
+++ b/visidata/selection.py
@@ -96,7 +96,7 @@ def select(self, rows, status=True, progress=True, add_undo=True):
     for r in (Progress(rows, 'selecting') if progress else rows):
         self.selectRow(r)
     if status:
-        if options.bulk_select_clear:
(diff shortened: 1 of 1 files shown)


saulpw:
Yes, we should use `self.options.some_selected_rows` on line 160 too.  We should use `vd.options.batch` for the other one, as the `batch` option can only be set globally--it shouldn't matter in practice but semantically it's more correct.  Either way the global `options` object is deprecated and all uses should be converted to vd.options or sheet.options eventually.

PR #3076, opened 2026-05-05
[modify-] clarify prompt for addcol-bulk

midichef:
The prompt is a bit unclear for `addcol-bulk`. This PR makes it clear that the input should be an integer. Just like `add-rows` does.
https://github.com/saulpw/visidata/blob/38b21f78f9934b918484a73ee0f704f6f358673d/visidata/modify.py#L348


saulpw:
Good point.  Although it makes me wonder if `gza` should alternatively be able to take a space-separated list of column names...

midichef:
I like it, I think taking a space-separated list of names would be useful.

PR #3077, opened 2026-05-06
[modify- rownum-] move @asyncthread to be after @api

midichef:
When `@asyncthread` comes before `@api`, the UI freezes while the decorated function runs. `@api` should come first, then `@asyncthread`, as it does in the rest of the codebase.

However, in practice, this PR changes nothing about how visidata behaves, because the buggy `Sheet.putChanges()` is currently overridden by every subclass that would otherwise call it.

### DIFF
--- a/visidata/experimental/rownum.py
+++ b/visidata/experimental/rownum.py
@@ -2,8 +2,8 @@ from visidata import *
 from functools import wraps, partial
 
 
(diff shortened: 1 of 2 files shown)


saulpw:
Thank you!  I wonder how difficult it would be to make a linter that can catch things like this.

PR #3095, opened 2026-05-22
[canvas-] in render, cancel render_async

midichef:
Hitting `Ctrl+L` during a slow-plotting graph causes the message:  `still running **render_async** from previous command`.

This PR fixes it to interrupt the previous render, and begin anew. The problem is that the old `plotAll_async` was changed to `render_async`, but not in the thread cancellation.

### DIFF
--- a/visidata/canvas.py
+++ b/visidata/canvas.py
@@ -722,7 +722,7 @@ class Canvas(Plotter):
     def render(self, h, w):
         'resets plotter, cancels previous render threads, spawns a new render'
         self.needsRefresh = False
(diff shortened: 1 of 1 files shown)


saulpw:
How are you finding these? Just in regular usage?

midichef:
I noticed it a couple of years ago when I was making changes to the graph-drawing code, and I kept a note of it. At the time I didn't submit a fix because I didn't understand threading well, and I didn't think such a small fix was worth risking breaking something.

PR #3100, opened 2026-05-22
[settings-] allow setting dict options from command-line

midichef:
The motivation here is to be able to easily switch `options.http_req_headers` from the command line.

For example, the SEC EDGAR database requires a User-Agent that has an email address, so this PR lets me set that via: `vd --http-req-headers="{ 'User-Agent': 'Example Company user@example.com' }"`

### DIFF
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -405,12 +405,19 @@ def editText(vd, y, x, w, attr=ColorAttr(), value='',
             vd.setLastArgs(v)
 
     if value:
+        t = type(value)
         if isinstance(value, (int, float)) and v[-1] == '%':  #2082
             pct = float(v[:-1])
             v = pct*value/100
-
-        # convert back to type of original value
-        v = type(value)(v)
+            # convert back to type of original value
+            v = t(v)
+        elif isinstance(v, str) and isinstance(value, (list, tuple, dict)):
+            import ast
+            v = ast.literal_eval(v)
+            if not isinstance(v, t):
+                vd.warning(f'error parsing string `{v}` into {t}')
+        else:
+            v = t(v)
 
     return v
 
--- a/visidata/_input.py
+++ b/visidata/_input.py
@@ -415,7 +415,7 @@ def editText(vd, y, x, w, attr=ColorAttr(), value='',
             import ast
             v = ast.literal_eval(v)
             if not isinstance(v, t):
-                vd.warning(f'error parsing string `{v}` into {t}')
+                vd.fail(f'error parsing string `{v}` into {t}')
         else:
             v = t(v)
 
--- a/visidata/settings.py
+++ b/visidata/settings.py
@@ -221,7 +221,7 @@ class OptionsObject:
                 import ast
                 value = ast.literal_eval(value)
                 if not isinstance(value, t):
-                    vd.warning(f'error parsing string for `{optname}` into {t}')
+                    vd.fail(f'error parsing string for `{optname}` into {t}')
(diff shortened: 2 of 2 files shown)


saulpw:
Thanks for the use case, I agree we need something like this.  But I would rather configure it at the options level, so that you can set the option to that same string in the interface or in .visidatarc, and it will auto-parse into a dict.  Does that make sense?

midichef:
That makes sense. So for that, we can add a field to `vd.options()` like `parse_type` (or maybe just `type`):
```
vd.option('http_req_headers', ..., parse_type=dict)
```
I'm still a bit shaky on how the various `_set` and `_get` work in the various options objects, but it seems to work as desired when I set it up like so:
```
def option(vd, name, default, description, replay=False, sheettype=BaseSheet, help:str='', cli_only=False, parse_type=None):
...
+    opt.parse_type = parse_type
...
class Option
    def __init__(self, name, value, description='', module='', help=''):
...
+        self.parse_type = None

class OptionsObject:
    def set(self, optname, value, obj='global', cmdlog=True):
...
+            pt = self._get(optname, 'default').parse_type
             if value is None and curval is not None:
                 return self.unset(optname, obj=obj)
+            elif pt in (list, tuple, dict) and isinstance(value, str):
+                import ast
+                value = ast.literal_eval(value)
+                if not isinstance(value, pt):
+                    vd.warning(f'error parsing option `{optname}` into {pt}')
```

Is this `pt = self._get(optname, 'default').parse_type` the right way to check the parse type from `OptionsObject.set(self, optname, value, obj='global', cmdlog=True)` when obj is `'global'`?

saulpw:
We shouldn't even need a separate `parse_type`; all other conversions use `type(opt.default)`.  We'd add a level of possible indirection there, since `dict` can't parse a string, but the idea is just that the default value would be a dict itself: `vd.option('http_req_headers', {}, ...)`

midichef:
Well, the default value is already a dict, which gives enough type information, so the current https://github.com/saulpw/visidata/pull/3100/commits/26f44af750be0abef6b93b221037b12d535e8f81 does handle setting the value to a string, in two situations, at the command-line and within .visidatarc.

The one situation left to handle is editing from the options sheet. To handle that, we'd need to change `editText` to similarly construct the dict from the edited string. It's simple to limit this behavior to options sheets, since `editOption` passes the option value object using `editCell(value=currentValue)`. Normal sheets only pass a string to `editCell` for the `value` parameter. So I'll push a commit to support options sheet cell-editing too.

saulpw:
> I'm still a bit shaky on how the various _set and _get work in the various options objects, but it seems to work as desired when I set it up like

BTW I apologize for this.  I'm not proud of the internal API of the OptionsObject and SettingsMgr.

visidata/_input.py
```diff
@@ -405,12 +405,19 @@ def editText(vd, y, x, w, attr=ColorAttr(), value='',
...
-
-        # convert back to type of original value
-        v = type(value)(v)
+            # convert back to type of original value
+            v = t(v)
+        elif isinstance(v, str) and isinstance(value, (list, tuple, dict)):
+            import ast
+            v = ast.literal_eval(v)
+            if not isinstance(v, t):
+                vd.warning(f'error parsing string `{v}` into {t}')
```
saulpw: I think this should be `fail`, since the value won't be set (and cmdlog replay should abort instead of sailing through).

midichef: Makes sense. I'll do the same for the analogous code in `OptionsObject.set()`.

PR #3104, opened 2026-05-22
[mainloop-] make sheet variable accessible in refactored code

midichef:
When mainloop code was moved into its own function in 897184a1dca9dba527c2b90fdab122a1e0edf42e, `sheet` wasn't brought along. This PR fixes that variable access..

I caught this using the `ruff` linter. I've used ruff on visidata on the past. But since the v3.4 release is coming up, I have given the codebase another inspection. Visidata is great for inspecting linter output since it's JSON:  `ruff check --output-format=json |vd`

### DIFF
--- a/visidata/mainloop.py
+++ b/visidata/mainloop.py
@@ -252,7 +252,7 @@ def mainloop(vd, scr):
             sheet.longname = ''
             prefixWaiting = False
 
-        vd._playNextQueuedCommand()
+        vd._playNextQueuedCommand(sheet)
 
         vd.callNoExceptions(sheet.checkCursor)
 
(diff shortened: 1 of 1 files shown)


visidata/mainloop.py
saulpw, line 272: hm, was this just broken then?  how did we not see this before?

midichef: Yes. I'm not sure if it ever gets called. In a quick glance at the code, `vd.nextCommands` seems to always contain strings or `CommandLogRow`:
https://github.com/saulpw/visidata/blob/8cb57c64d6f5b67cf13acb817ebf3052d5173645/visidata/cmdlog.py#L22
But the `cmd` that gets passed to `replayOne` above, is an element of `vd.nextCommands`, and it has `getattr()` called on it, so seemingly it's always a `CommandLogRow`, never a `string`:
https://github.com/saulpw/visidata/blob/8cb57c64d6f5b67cf13acb817ebf3052d5173645/visidata/cmdlog.py#L268-L271

So I wonder why `cmd` is treated like it may be in only 3 categories: `dict`, `list`, or other:
https://github.com/saulpw/visidata/blob/8cb57c64d6f5b67cf13acb817ebf3052d5173645/visidata/mainloop.py#L268-L272

PR #3106, opened 2026-05-23
[seaborn-] skip slow imports on every plot after the first plot

midichef:
This PR speeds up plots using Seaborn.

The imports of `matplotlib.pyplot` and `seaborn` were being done fresh on every call to `plot*-ext`, instead of just the first call. On my system that added a delay of about 2 seconds each time. After this PR the delay only happens on the first call.


saulpw:
vd.importExternal should be caching the package, just like how `import` does, so subsequent re-imports should take no time at all.  Rather than manually caching it, let's fix vd.importExternal properly.

midichef:
`importExternal` currently is caching the packages, but it's doing so inside the spawned process, where `ext_plot_seaborn`  is run with `multiprocessing`. The next spawned Seaborn process can't see those cached packages.

saulpw:
I see, sorry for not reading more carefully.  Thanks for submitting this!

PR #3116, opened 2026-05-29
[terminal-] hide Alt+[Shift+I as no-op binding on chooser

midichef:
In the command palette, `no-op` is bound to an ugly key sequence.
no-op-key

I use `bindkey` to an empty string that comes as the first binding to `no-op`, to make it blank:
no-op-key-after

I didn't test this much yet, but it seems fine in my initial use today.

### DIFF
--- a/visidata/tuiwin.py
+++ b/visidata/tuiwin.py
@@ -22,5 +22,6 @@ def getrootxy(vd, scr):  # like scr.getparyx() but for all ancestor scrs
     return px, py
 
 
+vd.bindkey('', 'no-op')              # needed to hide the next 2 long key combinations from showing up on palette chooser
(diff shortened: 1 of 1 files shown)


visidata/tuiwin.py
```diff
@@ -22,5 +22,6 @@ def getrootxy(vd, scr):  # like scr.getparyx() but for all ancestor scrs
     return px, py
 
 
+vd.bindkey('', 'no-op')              # needed to hide the next 2 long key combinations from showing up on palette chooser
```
saulpw, line 25: Oh, I guess because `no-op` is bound at definition time with an empty string, but `addCommand` ignores an empty string.

PR #3117, opened 2026-05-29
[tests] report elapsed time in each test script summary

saulbert:
Add an `elapsed()` helper to `tests/testenv.sh` and append `in <N>s` to each
test script's pass/fail summary line, so per-script wall time is visible in CI
and local runs. `test-perf.sh` also now collects all per-test times into one
summary line.

[PR opened via Claude Code at @saulpw's request]


saulpw:
I want to do this automatically for tests so that we're encouraged to keep overall test time down, and to see which tests contribute an outsized amount of time to the process.

PR #3118, opened 2026-05-29
[sheets] show sheet name in cursor status line

saulbert:
`cursorStatus` now prefixes the sheet name, so the left status bar shows which
sheet the cursor row/col refers to (useful when sheets share the screen).

[PR opened via Claude Code at @saulpw's request]


saulpw:
In reference to #3110

midichef:
This shows the sheet name for `show-cursor`. But it doesn't change it to appear in any passive display like the left status bar. Was this PR intended to do that?

PR #3121, opened 2026-05-31
[join] don't crash on unhashable keycol values  #3099

### LINKED ISSUE #3099: joins fail for unhashable cells
**Small description**
During joins, `JoinKeyColumn` stores keycol values in a `set()` to detect type conflicts:

https://github.com/saulpw/visidata/blob/67ec128928098489fcff9315b02fe446f2a2e42e/visidata/features/join.py#L153-L156

But that gives an error for unhashable types like lists and dicts:
```
  File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/features/join.py", line 156, in calcValue
    vals.add(c.getTypedValue(row[c.sheet]))
TypeError: unhashable type: 'list'
```

**Data to reproduce**
Join two sheets where the key column cells contain an identical list.

I won't suggest a solution as I still don't understand how to compare Visidata's typed objects. The system of wrappers around the various typed values (errors, etc) is still opaque to me.

saulbert:
## Summary

Joining sheets whose key columns hold unhashable typed values (lists, dicts) raised `TypeError: unhashable type: 'list'` from `JoinKeyColumn.calcValue` at `visidata/features/join.py:156`, where a `set()` was used to dedupe the typed key values across source columns.

This swaps the set for a list with pairwise equality so the mismatch warning still works without requiring the values to be hashable.

Closes #3099.

## Test plan
- [x] Added `test_join_unhashable` in `visidata/features/join.py` (auto-discovered by `visidata/tests/test_features.py`).
- [x] `dev/test-all.sh` — 11/11 suites pass (pytest 230, +1).
- [x] Manual repro from the issue (two JSONL sheets keyed on a list cell) now joins cleanly and shows the list as the JoinKeyColumn value.

### DIFF
--- a/visidata/features/join.py
+++ b/visidata/features/join.py
@@ -150,14 +150,16 @@ class JoinKeyColumn(WritableColumn):
         self.keycols = keycols
 
     def calcValue(self, row):
-        vals = set()
-        for i, c in enumerate(self.keycols):
+        vals = []
+        for c in self.keycols:
             if row[c.sheet] is not None:
-                vals.add(c.getTypedValue(row[c.sheet]))
-        if len(vals) != 1:
+                v = c.getTypedValue(row[c.sheet])
+                if not any(v == existing for existing in vals):  #3099
+                    vals.append(v)
+        if len(vals) > 1:
             keycolnames = ', '.join([f'{col.sheet.name}:{col.name}' for col in self.keycols])
             vd.warning(f"source key columns ({keycolnames}) have different types")
-        return vals.pop()
+        return vals[0]
 
     def putValue(self, row, value):
         for i, c in enumerate(self.keycols):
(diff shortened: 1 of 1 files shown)


visidata/features/join.py
```diff
@@ -150,14 +150,16 @@ def __init__(self, name='', keycols=None, **kwargs):
...
     def calcValue(self, row):
-        vals = set()
-        for i, c in enumerate(self.keycols):
+        vals = []
+        for c in self.keycols:
             if row[c.sheet] is not None:
-                vals.add(c.getTypedValue(row[c.sheet]))
-        if len(vals) != 1:
+                v = c.getTypedValue(row[c.sheet])
+                if not any(v == existing for existing in vals):  #3099
```
saulpw, line 157: so this is an explicit loop over `vals` with `==` whereas we were relying on the semantics of `set` to do the deduplication.  I believe that `set()` does this much faster for larger N but also that our N is necessarily a very small number--at most the number of different values with the same display value.  Maybe a comment but otherwise I'm inclined to just take this to solve the immediate problem.

PR #3127, opened 2026-06-05
use "visidata" prefix for temp files/dirs

hydrargyrum:
it makes it clearer to who those temp files belong

### DIFF
--- a/visidata/clipboard.py
+++ b/visidata/clipboard.py
@@ -105,7 +105,7 @@ def syscopyCells_async(sheet, cols, rows, filetype):
 
     import tempfile
     with io.StringIO() as buf:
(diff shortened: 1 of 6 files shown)


frosencrantz:
This seems like it would be helpful in finding temporary files created by VisiData.  I would suggest to make the prefix, and maybe even the suffix configurable by user options.

saulbert:
Thanks @hydrargyrum! This landed in develop in a generalized form in d3feb614: a `tmp_prefix` option (default `visidata-`) used by new `vd.TempFile()`/`vd.TempDir()` helpers that now wrap all temp file/dir creation, so the prefix is configurable in one place.

[this message written by Claude Opus 4.8 and approved by @saulpw]

PR #3137, opened 2026-06-09
[undo] columns-sheet resize + undo-sync (#3102, #3133)

### LINKED ISSUE #3102: make column resizing undoable
I often want to undo column resizes, so [I implemented undo for resizes](https://github.com/saulpw/visidata/discussions/2801). Should it be part of default visidata?

### LINKED ISSUE #3133: Undo not sync'ed between sheet and its columns-sheet
This behaviour may be non-trivial to modify, but similar to https://github.com/saulpw/visidata/issues/3122 not sync'ing the hasBeenModified property between sheet and its columns-sheet , there is a disconnect for the undo function also...

_First I delete a column in the columns-sheet for Sheet X._
- This will undo while still inside columns-sheet

_If I then close the columns-sheet_ with `quit-sheet`, returning to Sheet X:
- I can't undo column delete
- If I open columns-sheet anew, it won't undo as it's not the original instance, giving "nothing to undo on current sheet"

_If I don't `quit-sheet` the columns-sheet, but switch back to Sheet X:_
- If I undo from Sheet X, it does seem to attempt it -- however it's always right at the end of the current stack of undo commands. However it doesn't carry it out, instead showing this error when it reaches the attempt: `ValueError: list.remove(x): x not in list`. (Out of curiosity I tried with another columns-sheet edit, a type change, and this did not even make the erroneous attempt in Sheet X).
- If I return to the **very** same columns-sheet, I can undo the column delete

Is cooperative undo functionality between sheet and its columns-sheet possible for changes made inside the columns-sheet? I realize it might be involved and would be happy even if it's on the future roadmap...

saulbert:
Fixes #3102. Fixes #3133.

Undo-subsystem followups to #3136. Two commits:

**#3102 — make column resize undoable.** Width changes always record an undo now, not just hide/unhide. (Draw-time auto-width bypasses the setter, so no layout noise.)

**#3133 — sync columns-sheet edits with the source sheet's undo.** A single-source `ColumnsSheet`'s cmdlog now delegates to its source, so adding/deleting/reordering/retyping a column on the columns-sheet can be undone from the source sheet, survives closing the columns-sheet, and works from a freshly reopened columns-sheet.

The enabler: building a sheet's initial columns no longer records undofuncs. The columns-sheet leaked ~44 of them because it has 18 class-defined columns built synchronously in `__init__`->`resetCols` inside the command (a normal sheet has 0 and builds them in the async loader). Those construction undos are meaningless -- `addColumn` already gated `setModified` on `index is not None` but not its `addUndo`. So `resetCols` and the metasheet loader now run under `vd.suppressUndo()` (thread-local, since `reload` is async). With the creating command carrying no undofuncs, `undo()` skips it and the cmdlog delegation is clean.

Verified: all four scenarios from #3133 (undo in-sheet, from source, after close, from reopened sheet) work; user undos (addcol, resize, row delete) intact; 235 pytest + 11/11 golden (incl. macro/replay/undo) green.

Depends on #3136 (stacked); will retarget to `develop` once it merges.

Generated with Claude Code

### DIFF
--- a/visidata/cmdlog.py
+++ b/visidata/cmdlog.py
@@ -1,7 +1,7 @@
 import threading
 
 from visidata import vd, UNLOADED, namedlist, vlen, asyncthread, globalCommand, date
-from visidata import VisiData, BaseSheet, Sheet, ColumnAttr, VisiDataMetaSheet, JsonLinesSheet, TypedWrapper, AttrDict, Progress, ErrorSheet, CompleteKey, Path
+from visidata import VisiData, BaseSheet, Sheet, ColumnAttr, VisiDataMetaSheet, JsonLinesSheet, TypedWrapper, AttrDict, Progress, ErrorSheet, CompleteKey, Path, ColumnsSheet
 import visidata
 
 vd.option('replay_wait', 0.0, 'time to wait between replayed commands, in seconds', sheettype=None)
@@ -422,6 +422,14 @@ def cmdlog_sheet(sheet):
     return c
 
 
+@ColumnsSheet.property
+def cmdlog_sheet(sheet):
+    'Edits to a single source sheet log (and undo) against that source. #3133'
+    if len(sheet.source) == 1 and isinstance(sheet.source[0], BaseSheet):
+        return sheet.source[0].cmdlog_sheet
+    return super(ColumnsSheet, sheet).cmdlog_sheet
+
+
 @BaseSheet.property
 def shortcut(self):
     if self._shortcut:
@@ -443,7 +451,8 @@ def shortcut(self):
 def cmdlog(vd):
     if not vd._cmdlog:
         vd._cmdlog = CommandLogJsonl('cmdlog', rows=[])  # no reload
-        vd._cmdlog.resetCols()
+        with vd.suppressUndo():  # building a sheet's layout is not an undoable user action
+            vd._cmdlog.resetCols()
         vd.beforeExecHooks.append(vd._cmdlog.beforeExecHook)
     return vd._cmdlog
 
--- a/visidata/undo.py
+++ b/visidata/undo.py
@@ -1,7 +1,10 @@
 import itertools
+import contextlib
 from copy import copy
 
 from visidata import vd, options, VisiData, BaseSheet, UNLOADED
+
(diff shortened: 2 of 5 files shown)


visidata/cmdlog.py
```diff
@@ -422,6 +422,14 @@ def cmdlog_sheet(sheet):
     return c
 
 
+@ColumnsSheet.property
+def cmdlog_sheet(sheet):
+    'Edits to a single source sheet log (and undo) against that source. #3133'
+    if len(sheet.source) == 1 and isinstance(sheet.source[0], BaseSheet):
+        return sheet.source[0].cmdlog_sheet
+    return BaseSheet.cmdlog_sheet.fget(sheet)
```
saulpw: I suppose this the preferred way to defer to a base class property?  Should this be `Sheet` the instead (immediate parent)?

visidata/metasheets.py
```diff
@@ -71,12 +71,13 @@ def nSourceCols(self):
         return sum(vs.nCols for vs in self.source)
 
     def loader(self):
-        if len(self.source) == 1:
-            self.rows = self.source[0].columns
-            self.cursorRowIndex = self.source[0].cursorColIndex
-            self.columns[0].hide()  # hide 'sheet' column if only one sheet
-        else:
-            self.rows = [col for vs in self.source for col in vs.visibleCols if isinstance(vs, Sheet) and vs is not self]
+        with vd.suppressUndo():  # building the metasheet's own layout is not an undoable user action
```
saulpw: Can we wrap with suppressUndo instead at the loader() and resetCols() callsites?  This seems perhaps universally applicable.

visidata/undo.py
```diff
@@ -1,8 +1,13 @@
 import itertools
+import contextlib
+import threading
 from copy import copy
 
 from visidata import vd, options, VisiData, BaseSheet, UNLOADED
 
+# thread-local: resetCols/reload run in async threads; a global flag would race
+_undo_suppressed = threading.local()
```
saulpw: This does not seem like a good way to track this.  We can put the bool on `vd` and just let the GIL flow.

PR #3140, opened 2026-06-10
[sheets-] add command to change precision displayed for floats

midichef:
I often want to change the number of decimal points shown for floats, so this PR makes a convenience command for that bound to `g%`.

This is just a prototype command for discussion. I'm unsure about the name `set-float-precision`. There is only one other command that starts with `set-`, which is `set-aspect` for `Canvas` (`z_`).
Also, should we try to handle the case where the user has `disp_float_fmt` starting with `%`, like `%.02f`, which causes floats to be formatted using `locale.format_string`? That is, should we parse their `disp_float_fmt` to see whether it is of the form `%.02f` or the form `{:.02f}` ?

### DIFF
--- a/visidata/features/change_precision.py
+++ b/visidata/features/change_precision.py
@@ -41,11 +41,10 @@ def setcol_precision(col, amount:int):
         vd.fail('column type must be numeric or date')
 
 @VisiData.api
-def setcol_precision_input(vd):
+def setopt_precision(vd, precision):
      m = re.fullmatch(FLOAT_FORMAT_RE, vd.options.disp_float_fmt)
      if not m:
          vd.fail('could not parse disp_float_fmt')
-     precision = max(0, int(vd.input("float precision=", value="2")))
      if m[1]:
          vd.options.disp_float_fmt = '{:.' + str(precision) + 'f}'
      elif m[2]:
@@ -58,4 +57,4 @@ vd.addMenuItems('''
 
 Sheet.addCommand('Alt+-', 'setcol-precision-less', 'cursorCol.setcol_precision(-1)', 'show less precision in current column')
 Sheet.addCommand('Alt++', 'setcol-precision-more', 'cursorCol.setcol_precision(1)', 'show more precision in current column')
-Sheet.addCommand('g%', 'setcol-precision-input', 'vd.setcol_precision_input()', 'change the number of decimal places shown for floats')
+Sheet.addCommand('g%', 'setcol-precision-input', 'precision = max(0, int(vd.input("float precision=", value="2"))); vd.setopt_precision(precision)', 'change the number of decimal places shown for floats')
--- a/visidata/features/change_precision.py
+++ b/visidata/features/change_precision.py
@@ -5,6 +5,8 @@ import re
 
 from visidata import VisiData, vd, Sheet, Column, floatsi, currency, date
 
+FLOAT_FORMAT_RE = r'\{:\.([0-9]+)f\}|%\.([0-9]+)f'
+
 date_fmtstrs = [
     '%Y',
     '%Y-%m',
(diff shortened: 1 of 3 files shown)


saulpw:
Are you aware of `setcol-precision-more` in `visidata/features/change_precision.py`?  Seems like this should go in that file.  `setcol-precision-input`?

(Also not sure about `setcol-` here, as it doesn't strictly change the values; but it does change how they're represented in a text format so maybe it makes sense?)

midichef:
No, I didn't know about those! Okay, I'll force-push a version of this same `g%` command in `change_precision.py` called `setcol-precision-input`.

`setcol-` does make sense, as in my tests, using these `setcol-precision-*` commands (including the new proposed one) does changed the saved float string, for .tsv, though not for .json. That is, `0.456` in a float column when the `%` precision is set to 1 saves to tsv as `0.5`, but saves to json as `0.456`. Is this expected?

saulpw:
Yes, that's expected.  I've thought about forcing the imprecision down to .json too but that seems like it could be problematic, and would at least have to be gated by an option.  So unless/until someone comes along who actually wants it that behavior, we should leave it as it is.

midichef:
Okay.

Should this command parse the existing `disp_float_fmt` to match whichever of the two format string styles it uses? With something like `m = re.fullmatch(r'\{:\.0\df\}', options.disp_float_fmt)` vs `r'%\.0\df'`?

saulpw:
Hmm..the existing `setcol-precision` commands only support `%` fmtstr.  The easiest route is to copy that behavior; it would be nice to support both formats, but then it probably should also work with these setcol-precision commands too.

midichef:
Okay I added the ability to support both format to `setcol-precision-*`. Also `disp_float_fmt` has an unnecessary `0` in the default value of `{%:.02f}` which [isn't needed](https://docs.python.org/3/library/string.html#grammar-token-format-spec-precision), so I've pushed a commit to get rid of the `0`.

visidata/features/change_precision.py
```diff
@@ -28,12 +28,26 @@ def setcol_precision(col, amount:int):
...
+                col.fmtstr =       f'%.{max(0, int(m[2]) + amount)}f'
     else:
         vd.fail('column type must be numeric or date')
 
+@VisiData.api
+def setcol_precision_input(vd):
+     m = re.fullmatch(r'\{:\.([0-9]+)f\}|%\.([0-9]+)f', vd.options.disp_float_fmt)
+     if not m:
+         vd.fail('could not parse disp_float_fmt')
+     precision = max(0, int(vd.input("float precision=", value="2")))
```
saulpw: Technically this is fine because `input` is in the function name, but by convention we put the `input()` call in the execstr.

visidata/features/change_precision.py
```diff
@@ -28,12 +28,26 @@ def setcol_precision(col, amount:int):
         if col.fmtstr == '':
             col.fmtstr = f'%.{2 + amount}f'
         else:
-            precision_str = re.match(r'%.([0-9]+)f', col.fmtstr)
-            if not precision_str is None:
-                col.fmtstr = f'%.{max(0, int(precision_str[1]) + amount)}f'
+            m = re.fullmatch(r'\{:\.([0-9]+)f\}|%\.([0-9]+)f', col.fmtstr)
```
saulpw: Let's pull this regex into a global constant since we use it in two places and it's pretty complicated.

visidata/features/change_precision.py
midichef: Oh, right, I forgot about that convention. I've moved the `input()` call.

visidata/features/change_precision.py
midichef: done!

saulpw:
Thanks for the contribution, @midichef !

PR #3143, opened 2026-06-12
[sort-] prevent sortcol from having multiple priorities #3142

### LINKED ISSUE #3142: sort-asc-add & sort-desc-add skip sort levels when run more than once on same column
If you run `sort-asc-add` or `sort-desc-add` on the same column multiple times, it will use up phantom sort levels, including advancing to the next display character.

Run a combination of those commands 6 times on the same column (or your qty of sort levels), and you will see there will be no more levels left.

If you do it twice, then a third time on a different column, you will see that column's sort symbol will be that of the third sort level.

midichef:
Fixes #3142.

After this PR, when use runs `sort-asc-add` on a column that's already a sortcol, it will be moved to the end of the `_ordering`, changing the column to have the lowest priority. An alternate design choice would be to leave the ordering unchanged. But my reasoning was that, since the user is running `sort-asc-add`, they do want to change something about the sort order.

### DIFF
--- a/visidata/sort.py
+++ b/visidata/sort.py
@@ -41,7 +41,11 @@ def orderBy(sheet, *cols, reverse=False, change_column=False, save_cmd_input=Tru
         sheet._ordering = new_ordering
         do_sort = True
     else:
+        sortcols = [sortcol for (sortcol,rev) in sheet._ordering]
         for c in cols:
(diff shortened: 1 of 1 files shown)


saulpw:
Thanks, @midichef, this looks good to me!

PR #3144, opened 2026-06-14
[sort-] update sortcol list when a column is deleted

midichef:
When a column is a sort column, and is deleted, it stays in the `sheet._ordering` list. That means that future additions of columns with `z[` will use the wrong symbols.

This PR modifies `contract-cols` and `ColumnsSheet.delete_row()` to filter out such columns after the deletion is done.

Should the implementation use the `@after` decorator instead?

### DIFF
--- a/visidata/features/expand_cols.py
+++ b/visidata/features/expand_cols.py
@@ -164,6 +164,7 @@ def contract_source_cols(sheet, cols):
     for vs in sheet.source:
         vd.addUndo(setattr, vs, 'columns', vs.columns)
         vs.columns[:] = [c for c in vs.columns if c not in cols]
+        vs.validate_sortcols()
     return ret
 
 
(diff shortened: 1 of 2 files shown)


saulpw:
> Should the implementation use the @after decorator instead?

Yes, I would like this to be self-contained in `sort.py` if possible.  At a glance it seems both functions can be moved with `@after` or `@api`.

midichef:
Okay, I moved `delete_row` into `sort.py` by changing it to use `@after`.

I couldn't do that for `contract_cols`. There are two reasons. First, I got this error:
```
  File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/extensible.py", line 81, in wrappedfunc
    r = oldfunc(*args, **kwargs)
TypeError: 'NoneType' object is not callable
```
I believe that is a reference problem, caused by the file import order in `__init__.py`:  `sort.py` is imported before `features/expand_cols.py`. The `@after` in `sort.py` can't find any `contract_cols` yet since the one in `features/expand_cols.py` hasn't been imported.

A second independent problem is that  the original `contract_cols` has `@asyncthread`, and I can't guarantee it's done before running the second `contract_cols` (the one with `@after`).

saulpw:
Okay, thank you, that's the important one.  I don't mind keeping a reference to a function in a "core" module (sort.py) in a "feature" module.

midichef:
I've added sortcol validation to the more obscure command `contract-source-cols`. I also fixed a quirk in that command where it removed rows, then left them visible in the selection count afterward.

Issue #3148, opened 2026-06-20
[hide-col] When hiding the last column the current column ends up on the first hidden col rather than the last unhidden col


frosencrantz:
**Small description**
When hiding the last column the current column ends up on the first hidden col rather than the last unhidden column.  After hiding the rightmost column the current column should end up on the last unhidden column.

**Data to reproduce**

This works on any sheet with more than 1 column.

**Steps to reproduce**

Go the last column, hide it with `hide-col` with `-`.

**Expected result**

After hiding the rightmost column the current column should end up on the last unhidden column.  I am listing this as a bug, though I can see how this might be considered a feature request.  The `hide-col` command usually does the right thing, it hides the current column, and the current column ends up on the next unhidden column to the right.  This makes it very easy to hide/remove a multiple columns quickly.  Many times I am doing this to get down to one column so I can process and save the one column, so it would be helpful if the current column ended up on the last one unhidden column, rather than the first unhidden column.

Note there is the `save-col` and `save-col-keys` for the ability to save a single column or a subset of columns, though I don't remember this, and there is not a default binding for these.

saulpw:
Thanks for the bug report, @frosencrantz.  I'm curious, how do you (and other users) feel about the 'hidden' columns being shown at the right of the sheet?  It's been a few versions now and I have mixed feelings myself.

frosencrantz:
Overall I am happy with it. It makes it obvious there are hidden columns, and it is easy to quickly view their contents, and unhide them.   Hiding columns is easy and unchanged with`-`, but unhiding use to be a lot harder.  I would be ok if it was possible to toggle hidden columns.   Or if you were interested in exploring doing this in some different way. 

I think this is  just one of those edge cases, where it could behave a little better.  I think I found some other cases, the one that comes to mind is if you hit `F` on a hidden column the column is still hidden/missing in the FrequencySheet.

Though wondering if other people have thoughts.

frosencrantz:
It sometimes ends up being really easy to end up in the hidden columns like in this case, and being thrown a little. Or I feel using the right key it sometimes becomes a little to easy to get into the hidden columns.  Those times I think it would be great if there was some sort of tactile feedback to let me know I jumped over the line. I know that is not possible with a terminal, but it is sort of my gut feeling on it, if that helps.

midichef:
[...]

PR #3150, opened 2026-06-22
[sheets-] for hidden cols, draw first letter of name

midichef:
restores hidden column first letter, as mentioned in https://github.com/saulpw/visidata/issues/3148#issuecomment-4763174839

I'm not sure how I feel about this UI. It feels more suggestive and accurate to have `…` added? We'd have to give hidden columns a width of 2 though.

### DIFF
--- a/visidata/sheets.py
+++ b/visidata/sheets.py
@@ -763,6 +763,8 @@ class TableSheet(BaseSheet):
             # when cursor showing a hidden column
             if vcolidx >= self.nVisibleCols and vcolidx == self.cursorVisibleColIndex:
                 width = self.options.default_width
+            elif col.hidden:  #non-cursor hidden cols have room for 1 letter + truncator  #3150
+                return 2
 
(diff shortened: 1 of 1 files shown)


saulpw:
Yeah, I think I tried the single letter at first and didn't like it.  I'm open to making them width=2, so we can have `A…` showing.

midichef:
Okay, here's a prototype with width 2. I'm not sure if I like it, to be honest.

I definitely don't like the single letter header, which can be explored by running v3.1 as `uvx --from visidata==3.1.1 vd`.

I also moved the `vrows` list slice, so it's skipped for columns that already have widths. It only saves about 30 microseconds total on my system in all the draws till an idle timeout.

midichef:
I think I prefer the current `…` header for hidden columns. Usually when I want to hide a column, I want to less clutter on my sheet, so I don't want distracting header entries for `A…|B…|C…`  etc.

PR #3151, opened 2026-06-22
[csv] skip blank lines  #3085

### LINKED ISSUE #3085: tsv drops empty lines, csv keeps them
**Small description**
Blank lines in CSV and TSV files are treated differently. CSV keeps them as cells holding null and TSV ignores them.

**Steps to reproduce**
`echo "col\n1\n\n3" |vd -f csv`
`echo "col\n1\n\n3" |vd -f tsv`

**Expected result**
I'm not sure which behavior is more expected, but I'd expect that both formats should treat blank lines the same.

**Context**
The TSV behavior comes from this line:
https://github.com/saulpw/visidata/blob/8f0b8d4393dec30b8a4d516df93e8dd2d752d390/visidata/loaders/tsv.py#L86-L87

Jonathan-Haddock:
Skip blank lines in the CSV loader so it matches the TSV loader (#3085).

`csv.reader` returns `[]` for a blank line and has no option to skip them, so `iterload()` now drops empty rows — the approach @midichef suggested on the issue. A `,,` line parses to `['', '', '']` (a real row), not `[]`, so it's kept.

Added a test: blank lines are dropped (matching tsv), and a `,,` row is preserved.

Fixes #3085.

---
I've read and agree to the CAA: https://visidata.org/caa

- [x] AI Level (0-10): 5 — bots coded, human understands completely
    - [x] Model: Claude Opus 4.8 (via Claude Code)

### DIFF
--- a/visidata/loaders/csv.py
+++ b/visidata/loaders/csv.py
@@ -74,13 +74,11 @@ class CsvSheet(SequenceSheet):
                     row = next(rdr)
                 except csv.Error as e:
                     e.stacktrace=stacktrace()
-                    yield [TypedExceptionWrapper(None, exception=e)]
-                    continue
+                    row = [TypedExceptionWrapper(None, exception=e)]
                 except StopIteration:
                     return
-                if not row:  # blank line; the tsv loader skips these too
-                    continue  #3085
-                yield row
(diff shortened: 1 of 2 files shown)


saulpw:
Thanks for the contribution, @Jonathan-Haddock !

Issue #1929, opened 2023-06-20
How do I use vdsql with ibis?


anjakefala:
### Discussed in https://github.com/saulpw/visidata/discussions/1928

<div type='discussions-op-text'>

<sup>Originally posted by @dufferzafar June 19, 2023</sup>
I have `ibis-framework[mysql]` installed, but when I use `vdsql 'mysql://DSN'` vd tries to go through the standard mysql loader. How do I prevent that from happening & use the `ibis` loader?

Traceback:

```python

saul.pw/VisiData v2.12dev
Traceback (most recent call last):
  File "/home/dufferzafar/.local/bin/vdsql", line 8, in <module>
    rc = main()
  File "/home/dufferzafar/.local/lib/python3.8/site-packages/visidata/apps/vdsql/__main__.py", line 19, in main
    main.vd_cli()
  File "/home/dufferzafar/.local/lib/python3.8/site-packages/visidata/main.py", line 372, in vd_cli
    rc = main_vd()
  File "/home/dufferzafar/.local/lib/python3.8/site-packages/visidata/main.py", line 333, in main_vd
    run(vd.sheets[0])
  File "/home/dufferzafar/.local/lib/python3.8/site-packages/visidata/vdobj.py", line 33, in _vdfunc
    return getattr(visidata.vd, func.__name__)(*args, **kwargs)
  File "/home/dufferzafar/.local/lib/python3.8/site-packages/visidata/extensible.py", line 57, in wrappedfunc
    return oldfunc(*args, **kwargs)
  File "/home/dufferzafar/.local/lib/python3.8/site-packages/visidata/extensible.py", line 57, in wrappedfunc
    return oldfunc(*args, **kwargs)
  File "/home/dufferzafar/.local/lib/python3.8/site-packages/visidata/mainloop.py", line 288, in run
    ret = vd.mainloop(scr)
  File "/home/dufferzafar/.local/lib/python3.8/site-packages/visidata/mainloop.py", line 181, in mainloop
    self.draw_all()
  File "/home/dufferzafar/.local/lib/python3.8/site-packages/visidata/mainloop.py", line 115, in draw_all
    vd.drawSheet(vd.win1, ss1[0])
  File "/home/dufferzafar/.local/lib/python3.8/site-packages/visidata/mainloop.py", line 29, in drawSheet
    sheet.ensureLoaded()
[...]

PR #3152, opened 2026-06-23
[open] -f filetype overrides url scheme dispatch  #3126

### LINKED ISSUE #3126: Sql loader not triggered when using vd -f vdsql mysql url
I expected this cmd to use mysql sql loader but it still uses the vd default one:

```bash
uvx --from git+https://github.com/saulpw/visidata.git@develop \
 --with mysqlclient --with ibis-framework --with packaging --with pandas --with pyarrow --with pyarrow_hotfix --with numpy \
vd -f vdsql mysql://root:@127.0.0.1/central
```

saulbert:
Fixes #3126.

`vd -f vdsql mysql://…` (and more generally any `-f <filetype>` against a URL)
silently ignored the requested filetype: `openPath()` dispatches URLs purely by
scheme (`openurl_<scheme>`), so the explicit `-f` never got a chance to select a
loader.

This makes an explicit `-f <filetype>` with a dedicated `open_<filetype>` take
precedence over url-scheme dispatch. With no `-f`, scheme dispatch is unchanged.

Note: this only fixes the dispatch. For `-f vdsql` to resolve, `open_vdsql` must
be registered, i.e. the `vdsql` app/plugin has to be imported in the `vd`
session (running the `vdsql` binary, or loading it as a plugin) — by design the
builtin url loaders are only overridden when running `vdsql` directly (#1929).

Added a unit test covering both directions (filetype wins; no-filetype falls
back to scheme).

[written by Claude Opus 4.8 and reviewed by @saulpw]

### DIFF
--- a/visidata/_open.py
+++ b/visidata/_open.py
@@ -101,6 +101,13 @@ def openPath(vd, p, filetype=None, create=False):
     filetype = filetype or p.options.filetype  # resolve from path instance, Path class, global  #1710
 
     if p.scheme and not p.has_fp():
+        # an explicit filetype with a dedicated open_<filetype> overrides url-scheme dispatch  #3126
+        if filetype:
+            ft = filetype.lower()
+            openfunc = getattr(vd, 'open_'+ft, None) or vd.getGlobals().get('open_'+ft, None)
+            if openfunc:
+                return openfunc(p)
+
         schemes = p.scheme.split('+')
         openfuncname = 'openurl_' + schemes[-1]
 
--- a/visidata/tests/test_open.py
+++ b/visidata/tests/test_open.py
@@ -1,0 +1,16 @@
+from visidata import Path, vd
+
+
+class TestOpenPath:
+    def test_filetype_overrides_url_scheme(self):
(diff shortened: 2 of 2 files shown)


saulpw:
@aszenz Would you mind testing this for your use case?

aszenz:
> @aszenz Would you mind testing this for your use case?

Didn't work for me

```
 uvx --from git+https://github.com/saulbert/visidata.git@fix-3126-vdsql-filetype \
             --with mysqlclient --with ibis-framework --with packaging --with pandas --with pyarrow --with pyarrow_hotfix --with numpy \
             vd -f vdsql mysql://root:@127.0.0.1/central
```

It still used the default loader

saulbert:
The dispatch half of this landed on develop as ac1073eb.

@aszenz your retest failed because the fix only worked when the vdsql app was already loaded, and plain `vd` doesn't load it.  1d8a6fcd on develop fixes that: `-f vdsql` now imports the app on demand.  Please retest on develop and let us know if you have any more issues.

[this message written by Claude Fable 5 and approved by @saulpw]

PR #3156, opened 2026-06-28
[tests] automate .visidatarc + option-precedence manual tests

saulbert:
## What

Converts two items from `dev/checklists/manual-tests.md` into automated golden shell tests, and removes those items from the manual checklist:

- **Item 8 — `.visidatarc`** → `tests/test-options-config.sh`
- **Item 12 — Options precedence** → `tests/test-options-precedence.sh`

## Tests

**`test-options-config.sh`** (4 checks)
- numeric option, global
- numeric option, sheet-specific override (`CsvSheet.options.default_width` wins over global on a csv sheet)
- boolean option, global
- string option, global

**`test-options-precedence.sh`** (6 checks)
- native default (no rc) → builtin `20`
- visidatarc sets the value → `33`
- `--config` selects which rc loads → `77`
- CLI overrides visidatarc (`--default-width 99`)
- CLI option is global without `-g` (a file opened at runtime inherits it)
- sanity: without the flag, the runtime-opened file is the native default

## How it works

Both read the **resolved** option in sheet scope via an `addcol-expr` column dumped with `-O -` (`cursorFullDisplay`), and every invocation loads an isolated `--config` so results never depend on the developer's real `~/.visidatarc`. Auto-discovered by `dev/test-all.sh` (`tests/test-*.sh`); each runs in ~0.5s. Verified to fail (exit 1) on wrong expectations, not just pass vacuously.

## Notes for review

- Each script header quotes **verbatim** the manual-test text it replaces, so the checklist→test mapping lives next to the code.
- Two residual pieces intentionally stay out of these tests:
  - `motd_url` (former item 8) — network fetch, interactive-only; still manual. Currently preserved only as a note in the config-test header — flag if it needs a home in the manual pass.
  - The `-f`/xlsx filetype sub-cases (former item 12) describe **pre-#3139** behavior. Since #3139, `-f` attaches to the source path, never as a sheet option, so the OptionsSheet no longer shows `xlsx`. That per-path behavior is covered by `tests/test-filetype.sh` (manual item 13).

### DIFF
--- a/tests/test-options-config.sh
+++ b/tests/test-options-config.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+# Test .visidatarc option loading: numeric/boolean/string, global + sheet-specific.
+#
+# Automates this former manual test (removed from dev/checklists/manual-tests.md):
+#   8. .visidatarc
+#       - numerical, boolean and string option
+#       - sheet-specific and global
+#       - motd_url                          <- STILL MANUAL: network fetch, interactive-only
+#
+# Mechanism: open a data file in batch, add an expr column that evaluates the
+# resolved option in sheet scope (options.NAME), and dump the cursor cell with
+# `-O -`. addcol-expr lands the cursor on the new column, so no movement needed.
+# stderr passes through (visible); $() captures only the stdout value.
+source tests/testenv.sh
+FAIL=0
+
+RC=$(mktemp)
+trap 'rm -f "$RC"' EXIT
+cat > "$RC" <<'EOF'
+options.motd_url = ''
+options.default_width = 33
+options.clean_names = True
+options.disp_currency_fmt = '#%.02f'
+CsvSheet.options.default_width = 44
+EOF
+
+VDC="$PYTHON -m visidata --config $RC --visidata-dir tests/.visidata"
+
+CSV=sample_data/benchmark.csv
+TSV=sample_data/sample.tsv
+
+# resolve EXPR FILE -> resolved value of python expr (sheet scope) at cursor
+resolve() {
+    $VDC --batch "$2" --play <(printf 'addcol-expr _t=%s\n' "$1") -O -
+}
+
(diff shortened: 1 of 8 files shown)


tests/test-options-config.sh
```diff
@@ -0,0 +1,58 @@
...
+check() {
+    local desc="$1" expected="$2" got="$3"
+    if [[ "$got" != "$expected" ]]; then
+        echo "FAIL: $desc (expected '$expected', got '$got')"
+        FAIL=1
+    fi
+}
+
+# numeric option, global value resolves on a non-csv sheet
+check "numeric global" 33 "$(resolve options.default_width $TSV)"
```
saulbert, line 46: Automates manual-tests.md item 8 (.visidatarc): numeric (global + CsvSheet sheet-specific override on L48), boolean (L50), and string (L52) options. The `motd_url` sub-bullet of item 8 stays manual (network/interactive) and is noted in the header.

tests/test-options-precedence.sh
```diff
@@ -0,0 +1,77 @@
...
+check() {
+    local desc="$1" expected="$2" got="$3"
+    if [[ "$got" != "$expected" ]]; then
+        echo "FAIL: $desc (expected '$expected', got '$got')"
+        FAIL=1
+    fi
+}
+
+# native default (rc sets nothing): default_width builtin default is 20
+check "native default" 20 "$($VDE --batch $CSV --play <(dw) -O -)"
```
saulbert, line 59: Automates manual-tests.md item 12 (Options precedence): native default < visidatarc (L61) < CLI (L65); `--config` selects which rc loads (L63); CLI options apply globally without `-g` (L67, with the no-flag sanity check on L70). Item 12's `-f`/xlsx-filetype sub-cases describe pre-#3139 behavior and are covered by tests/test-filetype.sh (manual item 13), not here.

PR #3186, opened 2026-07-20
[arrow] load IPC streaming format from stdin  #3184

### LINKED ISSUE #3184: The arrow IPC loader does not accept stdin
(Apologies if this should be a feature/loader request but it is slightly unexpected behaviour - empty sheet with no error)

**Small description**

- `ArrowSheet.iterload` opens the data source by filename, something like`with pa.OSFile(str(self.source), 'rb') as fp:`
- this bypasses visidata's Path wrapper, so when the source is stdin the string is literally `-`
- using `self.source.open_bytes()` (like the other binary loaders) would make Arrow IPC streams readable from a pipe

**Data to reproduce**

Visidata exits without ever reading the pipe:

```bash
 > uv run --with pyarrow python -c 'import pyarrow as pa, sys
   t = pa.table({"a": [1, 2], "b": ["x", "y"]})
   with pa.ipc.new_stream(sys.stdout.buffer, t.schema) as w: w.write_table(t)' | vd -b --filetype arrows -
Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>
BrokenPipeError: [Errno 32] Broken pipe
```

If I write that same output to file I can open it with the arrow loader!

I can add more motivation if you’d like but basically this unlocks arrow ipc into vd via unix pipes.

P.s. thanks for such a great tool.

jhjn:
Fixes #3184

The arrow loader opened its source by filename `pa.OSFile(str(self.source)` which bypasses visidata's Path wrapper. `vd -f arrows -` tried to open a file named `-` and silently loads an empty sheet.

Changes:
- `open_arrow` / `open_arrows` now tag the sheet with `ipc_format='file'/'stream'`, so the loader knows which format was requested.
- For arrows, read the IPC stream directly via `self.source.open_bytes()` (matching other binary loaders) - streaming works on unseekable sources: `... | vd -f arrows -`.
- Keep existing behaviour for arrow.

Adds `visidata/tests/test_arrow.py` which fails on develop and passes with this fix.

### DIFF
--- a/visidata/loaders/arrow.py
+++ b/visidata/loaders/arrow.py
@@ -7,13 +7,13 @@ from visidata import Sheet, VisiData, TypedWrapper, anytype, date, vlen, Column,
 @VisiData.api
 def open_arrow(vd, p):
     'Apache Arrow IPC file format'
-    return ArrowSheet(p.base_stem, source=p)
+    return ArrowSheet(p.base_stem, source=p, ipc_format='file')
 
 
 @VisiData.api
 def open_arrows(vd, p):
     'Apache Arrow IPC streaming format'
-    return ArrowSheet(p.base_stem, source=p)
+    return ArrowSheet(p.base_stem, source=p, ipc_format='stream')
 
 
 def arrow_to_vdtype(t):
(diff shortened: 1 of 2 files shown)


visidata/loaders/arrow.py
```diff
@@ -57,15 +57,22 @@ def arrow_to_vdtype(t):
...
-            with pa.OSFile(str(self.source), 'rb') as fp:
-                self.coldata = pa.ipc.open_file(fp).read_all()
-        except pa.lib.ArrowInvalid:
-            with pa.OSFile(str(self.source), 'rb') as fp:
+        if self.ipc_format == 'stream':
+            # non-seeking, so this works on unseekable sources like stdin
+            with self.source.open_bytes() as fp:
                 self.coldata = pa.ipc.open_stream(fp).read_all()
+        else:
+            try:
```
saulpw, line 70: I think we need either the ipc_format check or the try/except block, but not both?  Which model did you use to create this PR, and what AI level is this at?

PR #3193, opened 2026-08-01
[seaborn-] prevent PicklingError seen on Python v3.14

### LINKED ISSUE #3192: PicklingError using Plot > Graph > replot using matplotlib
**Small description**
I get an error trying to plot a simple x by y graph.

**Data to reproduce**
I've tried with floats and ints, nothing special.

**Steps to reproduce**

Just follow https://www.visidata.org/docs/graph/ instructions and select Plot > Graph > replot using matplotlib

**Expected result**

I expect a file saved somewhere

the statuses output:

plotting 54 rows using matplotlib
PicklingError: Can't pickle <function <lambda at 0x7f0b80707690>: it's not found as visidata.modify.<lambda>

Image

**Configuration**

yes
I was not able to test in other versions.

I'm running: 
saul.pw/VisiData v3.4
Python 3.14.6
extra/python-seaborn 0.13.2-3 (Arch Linux)

midichef:
Closes #3192.

This PR simply restores the behavior used by Python versions before 3.14. So it's an acceptable workaround.

For a long-term solution, we should think whether visidata is affected by any of the [problems caused by using 'fork'](https://stackoverflow.com/questions/64095876/multiprocessing-fork-vs-spawn/66113051#66113051). To get the default `'forkserver'` running may be quite involved, because the multiprocessing docs say:

>The arguments to [Process](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process) usually need to be picklable so they can be passed to the child process.

and the arguments to `Process` in this case include the `vd` object:
https://github.com/saulpw/visidata/blob/cc10ddf3bdadae2e9dd2fd2219ce728219e4170b/visidata/features/graph_seaborn.py#L18

### DIFF
--- a/visidata/features/graph_seaborn.py
+++ b/visidata/features/graph_seaborn.py
@@ -15,7 +15,11 @@ def plot_seaborn(vd, rows, xcols, ycols):
     pyplot = vd.importExternal('matplotlib.pyplot', 'matplotlib')
     seaborn = vd.importExternal('seaborn')
     import multiprocessing
-    mp = multiprocessing.Process(target=ext_plot_seaborn, args=(vd, rows, xcols, ycols))
+    if multiprocessing.get_start_method() == 'forkserver':  #3192  Python >= 3.14
+        pfunc = multiprocessing.get_context('fork').Process
+    else:
+        pfunc = multiprocessing.Process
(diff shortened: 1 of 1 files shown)


saulpw:
Okay, I see.  Unfortunately `vd` object probably will never be picklable so we might need a different multiprocessing solution for things like this in the future.

PR #3197, opened 2026-08-14
[columns-] refresh allColumnsSheet source on columns-all (#3191)

### LINKED ISSUE #3191: [vd.allColumnsSheet] allColumnsSheet is a cached lazy_property, and doesn't update after first use.
**Small description**

`columns-all` (`gC`) shows the sheet stack as it was the *first* time it was invoked in the session, not the current stack. Sheets opened afterwards never appear.

`vd.allColumnsSheet` is a `lazy_property`, so the `ColumnsSheet` object — and with it the `source=vd.stackedSheets` list, which is evaluated once at construction — is cached for the life of the session:

https://github.com/saulpw/visidata/blob/develop/visidata/metasheets.py#L100-L102

The command does call `vs.reload()`, but `reload()` re-runs the loader over that same stale `source` list; nothing recomputes `source` from the current stack.

`columns-selected` is unaffected — it constructs a fresh `ColumnsSheet` on each invocation.

**Data to reproduce**

`tests/data1.tsv` and `tests/data2.tsv` from the repo.

**Steps to reproduce**

```
open-file tests/data1.tsv
columns-all
quit-sheet
open-file tests/data2.tsv
columns-all
```

Interactively: `vd -N tests/data1.tsv`, press `gC`, press `q`, then `Space open-file tests/data2.tsv`, then `gC`.

**Expected result**

The second `gC` lists the columns of both data1 and data2, with a `sheet` column distinguishing them.

The second `gC` lists only data1's columns. The `sheet` column is also hidden, which confirms the loader took the `len(self.source) == 1` branch — i.e. `source` is still `[data1]`:

```
name	width	type	fmtstr	value	aggregators	sortorder
Key	5			1
A	4			a1
B	4			b1
```

No error or stack trace.

The caching is directly observable:

```python
>>> vd.sheets = [a]
>>> [s.name for s in vd.allColumnsSheet.source]
['A']
>>> vd.sheets = [b, a]
>>> [s.name for s in vd.allColumnsSheet.source]   # expected ['B', 'A']
['A']
```

**Configuration**

- Does this issue reproduce without any plugins or configuration (using the `-N` CLI flag)? Yes — reproduced with `-N`.

- Does this issue reproduce with either the latest release, or with the develop branch? Reproduced on `develop` (v3.4dev, commit 6235d5a8). `allColumnsSheet` has been a cached `lazy_property` since aa2fae09 (2020-06-29); `source=vd.stackedSheets` arrived in f8202c6b (2021-03-18, first released in v2.10), so releases from v2.10 on are likely affected.

**Additional context**

- Which version of Python? 3.9.6.

- Which terminal are you using? N/A — reproduced in `--batch` replay, not a display issue.

A fix could reassign `source` in the command before reloading, e.g. `vs.source = vd.stackedSheets`, keeping the singleton but refreshing its source. The comment above the command ("copy vd.sheets so that ColumnsSheet itself isn't included") suggests the exclusion of the ColumnsSheet from its own source was the motivation for the snapshot; the loader already guards that with `vs is not self`.

This is AI discovered and generated.  This was found during the investigation for https://github.com/saulpw/visidata/issues/3190.  I have reviewed the description, did some basic verification, and think this is a valid issue.

DSeaStar:
## Summary
- `columns-all` (`gC`) reused a cached `allColumnsSheet` whose `source` was a snapshot of `vd.stackedSheets` from the first invocation, so sheets opened later never appeared.
- Snapshot `source=vd.stackedSheets` on each `gC` before `reload()`, and unhide the `sheet` column when more than one source sheet is present (it stays hidden after a prior single-sheet load).

Fixes #3191.

## Test plan
- [x] `pytest visidata/tests/test_columns_sheet.py` — first `gC` on one sheet, then open a second sheet and `gC` again; both sheets' columns are listed and the `sheet` column is visible.
- [ ] Interactively: `vd -N tests/data1.tsv`, `gC`, `q`, `o tests/data2.tsv`, `gC` — second Columns Sheet lists data1 and data2.

### DIFF
--- a/visidata/metasheets.py
+++ b/visidata/metasheets.py
@@ -77,6 +77,9 @@ Other commands (not specific to Columns Sheet):
             self.columns[0].hide()  # hide 'sheet' column if only one sheet
         else:
             self.rows = [col for vs in self.source for col in vs.visibleCols if isinstance(vs, Sheet) and vs is not self]
+            # reused all_columns sheet may have hidden this on a prior single-sheet load
+            if self.columns[0].hidden:
+                self.columns[0]._width = None
 
     def newRow(self):
         c = type(self.source[0])._coltype()
@@ -129,8 +132,8 @@ def join_cols(sheet):
     destSheet.addColumn(c, index=sheet.cursorRowIndex)
 
 
-# copy vd.sheets so that ColumnsSheet itself isn't included (for recalc in addRow)
-globalCommand('gC', 'columns-all', 'vs=vd.allColumnsSheet; vs.reload(); vd.push(vs)', 'open Columns Sheet: edit column properties for all visible columns from all sheets on the sheets stack')
+# copy current stackedSheets so the cached allColumnsSheet sees newly opened sheets (#3191)
+globalCommand('gC', 'columns-all', 'vs=vd.allColumnsSheet; vs.source=vd.stackedSheets; vs.reload(); vd.push(vs)', 'open Columns Sheet: edit column properties for all visible columns from all sheets on the sheets stack')
 
 Sheet.addCommand('C', 'columns-sheet', 'vd.push(ColumnsSheet(name+"_columns", source=[sheet]))', 'open Columns Sheet: edit column properties for current sheet')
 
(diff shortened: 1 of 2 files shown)


saulpw:
AI without obeying our PR template, auto-close.

PR #3198, opened 2026-08-15
[dedupe-] add command to select duplicate cells in a single column

midichef:
The motivation for this PR is for when I am skimming a dataset. It'd be helpful sometimes to highlight the first time a new value appears in the current column.

In existing visidata, to do this can be cumbersome:  get rid of any key columns, make the current column the key column, run `select-duplicate-rows`, restore the previous key columns.

After this PR, I could just run `select-col-duplicates` and `zt`. We could also make it a builtin:
```
TableSheet.addCommand(None, "select-col-distincts", "sheet.select_col_duplicates(cursorCol, duplicates=False)", "select each row where the cell in the current column is not a duplicate of a prior cell")
```
Should I add this command to this PR? It needs a better name, though.

And for completeness, should I add a `dedupe-col` command, that's analogous to `dedupe-rows`, that ignores keycols and uses only the values in the current column?

### DIFF
--- a/visidata/features/dedupe.py
+++ b/visidata/features/dedupe.py
@@ -27,7 +27,7 @@ from copy import copy
 from visidata import Sheet, TableSheet, asyncthread, Progress, vd, ItemColumn
 
 
-def gen_identify_duplicates(sheet):
+def gen_identify_duplicates(sheet, cols=None):
     """
     Takes a sheet, and returns a generator yielding a tuple for each row
     encountered. The tuple's structure is `(row_object, is_dupe)`, where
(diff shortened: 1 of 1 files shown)


visidata/features/dedupe.py
```diff
@@ -79,6 +79,15 @@ def select_duplicate_rows(sheet, duplicates=True):
 
     vd.status(f"selected {sel_count}{more_str} {sheet.rowtype}")
 
+@Sheet.api
+@asyncthread
+def select_col_duplicates(sheet, col, duplicates=True):
+    gen = gen_identify_duplicates(sheet, [col])
+    prog = Progress(gen, gerund="selecting", total=sheet.nRows)
```
saulpw: Here it's immaterial, but we stylistically we prefer the Progress on the same line as the `for`.  If not, it should use `with Progress(...) as prog` so that the Progress exits scope after the loop.

saulpw:
This change means that the behavior of the new command depends on the state of the sheet (in particular the key cols), which might be surprising for some batch cases.  I think this is fine, because the surprise would require programmatically defining keycols (risking no keycols) which is not straightforward.

Sorry for the conflict, I merged these in the wrong order.

visidata/features/dedupe.py
midichef: okay, I'll make that change.

midichef:
Okay, I've resolved the merge conflict and put the `Progress` on the same line as the `for`.

How do you feel about `select-col-distincts`, as a useful command? The reason I'd want a command is to cue people in that it's an available operation. It's not immediately obvious from the existence `select-col-duplicates` that you can invert the selection and you have distincts. On the other hand, we would already have 21 `select-*` commands, which is a lot, and `select-col-distincts` would make that 22.

PR #3199, opened 2026-08-15
[dedupe-] don't crash on unhashable typed values #3196

### LINKED ISSUE #3196: dedupe can't handle unhashable types
**Small description**
Dedupe can't handle cells with unhashable types like lists:
```
  File "/home/midichef/.venv/lib/python3.12/site-packages/visidata/features/dedupe.py", line 51, in gen_identify_duplicates
    is_dupe = vals in seen
TypeError: unhashable type: 'list'
```

But `FreqTable` can handle hashable types, because it groups items by their string representation:
https://github.com/saulpw/visidata/blob/ea9eecbc5d4be22dca627d911ea1a6347101cb75/visidata/pivot.py#L233

(I edited this issue to extract much of it into a separate issue #3219)

VXNCXNX:

- [x] [AI Level](https://visidata.org/ai) (0-10): **8**
    - [x] Model and version of AI used (required for AI levels 4+): Claude Opus 5, via Claude Code
    - [x] Human operator (if bot account): @VXNCXNX

Being straight about that number: the bug was found, diagnosed and coded by the bot, and the human operator set the goal and reviewed the result rather than typing the code. Please weigh it accordingly. The change itself is four lines and copies a pattern you already merged.

Fixes #3196.

## What's broken

`dedupe-rows` and `select-duplicate-rows` raise `TypeError: unhashable type: 'list'` on any sheet where a checked column holds a list or dict, which is ordinary for JSON and JSONL data.

## The fix

`gen_identify_duplicates` built a `set` of value tuples, and a tuple containing a list cannot be hashed.

This is the same bug class as #3099, which you fixed in `join.py` by swapping the set for a list and comparing with `==`. This copies that, including the linear scan, rather than inventing a hash-with-fallback hybrid:

```python
if not any(v == existing for existing in vals):  #3099
```

The cost is that dedupe becomes O(n^2) in the number of distinct rows. Say the word if you would rather have the hybrid, where hashable values keep the set and only unhashable ones fall back.

Left alone: the other half of #3196, where dedupe and FreqTable disagree on `1` versus `1.0`. That is a semantics call rather than a crash.

## Verification

`test_dedupe_unhashable` in `dedupe.py`, mirroring `test_join_unhashable`, asserting the duplicate flags come out `[False, True, False]` for rows `[1,2]`, `[1,2]`, `[3,4]`. It fails with the original `set` version, on the exact `TypeError` from the issue. Full suite is 253 passed against 252 before, the difference being that test.

### DIFF
--- a/visidata/features/dedupe.py
+++ b/visidata/features/dedupe.py
@@ -24,7 +24,7 @@ __author__ = "Jeremy Singer-Vine <jsvine@gmail.com>"
 
 from copy import copy
 
-from visidata import Sheet, TableSheet, asyncthread, Progress, vd
+from visidata import Sheet, TableSheet, asyncthread, Progress, vd, ItemColumn
 
 
 def gen_identify_duplicates(sheet):
@@ -45,12 +45,14 @@ def gen_identify_duplicates(sheet):
     else:
         cols_to_check = sheet.keyCols
 
-    seen = set()
+    seen = []
     for r in sheet.rows:
         vals = tuple(col.getValue(r) for col in cols_to_check)
-        is_dupe = vals in seen
+        # use a list with == comparison instead of a set, since vals may
+        # contain unhashable types like list or dict (e.g. from JSON) #3196
+        is_dupe = any(vals == existing for existing in seen)
         if not is_dupe:
-            seen.add(vals)
+            seen.append(vals)
         yield (r, is_dupe)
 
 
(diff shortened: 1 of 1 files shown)


saulpw:
Possible performance issue: this turns dedupe from O(n\*log m) where n is number of rows and m is number of distinct values, to O(n\*m).  We'll accept it, as it's better I think to work more slowly than not at all, but we may need to revisit at some point.

midichef:
This is unfortunately too slow. On my (old) system it takes about 15 seconds to dedupe 25,000 rows, and about 60 seconds for 50,000 rows. Anyone currently using dedupe has been using it on hashable types, probably strings, so this is a terrible downgrade for current users.

How about trying the AI's proposed "hash-with-fallback hybrid"?

saulpw:
Thanks for checking it on a live system, @midichef.

PR #3205, opened 2026-08-17
Warn when go-row-number / go-col-number are out of bounds

### LINKED ISSUE #3166: [go-row-number] / [go-col-number] do not warn on out-of-bounds
The input accepted by `go-row-number` and `go-col-number` does not validate or warn when out-of-bounds...

If input exceeds the existing number of columns or rows, VD will round off and go to the last col/row, though without warning. If this behaviour is to stay (should it?), then it should definitely warn. Negative values do the same (round to zero index), and even though neg values are an unlikely input, it might be more consistent to warn with that one too?

jennifferweslowski-design:
`zr` / `zc` still clamp to the first or last row or column (vim-like). They now also warn when the number is negative or past the last index, so it is obvious the jump did not land on the requested cell.

Fixes #3166

### DIFF
--- a/visidata/features/go_col.py
+++ b/visidata/features/go_col.py
@@ -60,8 +60,22 @@ def nextColName(sheet, show_cells=True):
     vd.warning(f'found no column with name: {name}')
 
 
+@Sheet.api
+def goColNumber(sheet, n):
+    'Move cursor to 0-based visible column *n*. Warn and clamp if out of bounds.'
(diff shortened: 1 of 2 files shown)


jennifferweslowski-design:
CLA Assistant reports that all committers have signed, but the `license/cla` check is still pending. Is there anything else I need to do, or could the check be re-run?

PR #3207, opened 2026-08-20
[color-] prevent early crash on colorless terminals like vt100

midichef:
Closes #3206.

I did a quick skim of the `curses` library to check that catching `ValueError` is specific enough to just the sitaution we're interested in, and it looks like it is; `init_pair()` only raises `ValueError` when `COLOR_PAIRS` is too small.
https://github.com/python/cpython/blob/bf95881b84662d3dcd35143495b9880962164fee/Modules/_cursesmodule.c#L6332

Issue #3206, opened 2026-08-18
COLOR_PAIRS-1 crash while opening


yarikoptic:
**Small description**

Similar manifestations happened in the distant past:

- https://github.com/saulpw/visidata/issues/1227

**Data to reproduce**

unfortunately can't be shared  and not sure if matters

**Steps to reproduce**

```
> visidata /tmp/out.tsv
... screen clears here
Traceback (most recent call last):
  File "/dartfs/rc/lab/D/DBIC/DBIC/CON/asmacdo/mechababs/.venv/bin/visidata", line 10, in <module>
    sys.exit(vd_cli())
  File "/dartfs/rc/lab/D/DBIC/DBIC/CON/asmacdo/mechababs/.venv/lib64/python3.11/site-packages/visidata/main.py", line 389, in vd_cli
    rc = main_vd()
  File "/dartfs/rc/lab/D/DBIC/DBIC/CON/asmacdo/mechababs/.venv/lib64/python3.11/site-packages/visidata/main.py", line 348, in main_vd
    run(vd.sheets[0])
  File "/dartfs/rc/lab/D/DBIC/DBIC/CON/asmacdo/mechababs/.venv/lib64/python3.11/site-packages/visidata/vdobj.py", line 34, in _vdfunc
    return getattr(visidata.vd, func.__name__)(*args, **kwargs)
  File "/dartfs/rc/lab/D/DBIC/DBIC/CON/asmacdo/mechababs/.venv/lib64/python3.11/site-packages/visidata/extensible.py", line 66, in wrappedfunc
    return oldfunc(*args, **kwargs)
  File "/dartfs/rc/lab/D/DBIC/DBIC/CON/asmacdo/mechababs/.venv/lib64/python3.11/site-packages/visidata/extensible.py", line 66, in wrappedfunc
    return oldfunc(*args, **kwargs)
  File "/dartfs/rc/lab/D/DBIC/DBIC/CON/asmacdo/mechababs/.venv/lib64/python3.11/site-packages/visidata/extensible.py", line 66, in wrappedfunc
    return oldfunc(*args, **kwargs)
  File "/dartfs/rc/lab/D/DBIC/DBIC/CON/asmacdo/mechababs/.venv/lib64/python3.11/site-packages/visidata/mainloop.py", line 325, in run
    ret = vd.mainloop(scr)
[...]

Issue #1232, opened 2021-12-08
[develop] Trouble running VisiData with `TERM` set to 'xterm'


anjakefala:
(This might be a completely separate error from @ajkerrigan, neverless I can't run visidata without exporting the TERM variable.)

I ran into a similar issue today. I can reproduce/fix my particular error on Docker.

You might want to try running `TERM=xterm-256color visidata` on your system.

## to reproduce
```
$ docker run -it --rm --entrypoint /bin/bash python:3
$ pip install visidata
$ vd
Traceback (most recent call last):
  File "/usr/local/bin/vd", line 6, in <module>
    visidata.main.vd_cli()
  File "/usr/local/lib/python3.10/site-packages/visidata/main.py", line 327, in vd_cli
    rc = main_vd()
  File "/usr/local/lib/python3.10/site-packages/visidata/main.py", line 292, in main_vd
    run(vd.sheets[0])
  File "/usr/local/lib/python3.10/site-packages/visidata/vdobj.py", line 34, in _vdfunc
    return func(visidata.vd, *args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/visidata/mainloop.py", line 333, in run
    ret = vd.mainloop(scr)
  File "/usr/local/lib/python3.10/site-packages/visidata/mainloop.py", line 201, in mainloop
    self.draw_all()
  File "/usr/local/lib/python3.10/site-packages/visidata/mainloop.py", line 105, in draw_all
    vd.draw_sheet(vd.win1, ss1[0])
  File "/usr/local/lib/python3.10/site-packages/visidata/mainloop.py", line 30, in draw_sheet
    scr.bkgd(' ', colors.color_default)
  File "/usr/local/lib/python3.10/site-packages/visidata/color.py", line 62, in __getattr__
    return self.get_color(optname).attr
  File "/usr/local/lib/python3.10/site-packages/visidata/color.py", line 153, in get_color
    r = self.colorcache[optname] = self._colornames_to_cattr(colornamestr, precedence)
  File "/usr/local/lib/python3.10/site-packages/visidata/color.py", line 121, in _colornames_to_cattr
    fg, bg, attrlist = self.split_colorstr(colornamestr)
  File "/usr/local/lib/python3.10/site-packages/visidata/color.py", line 93, in split_colorstr
    if self._get_colornum(x) is not None:   # only set known colors
  File "/usr/local/lib/python3.10/site-packages/visidata/color.py", line 114, in _get_colornum
[...]

PR #3209, opened 2026-08-23
[color-] derive color and pair limits from terminal  #3206

saulbert:
Follow-up to #3207 (see [discussion](https://github.com/saulpw/visidata/pull/3207#issuecomment-5383452515)), finishing the other half of #3206: on terminals with 8 colors and 64 pairs (`TERM=xterm`, `TERM=ansi`), `_get_colornum()` probed color availability with `init_pair(255, ...)`, which raises for every color when `COLOR_PAIRS` is 64 — so the whole UI drew colorless (also the old #1232).

- `_get_colornum()` checks the color number against `curses.COLORS` instead of allocating a probe pair
- `_get_colorpair()` derives its recycling limit from `curses.COLOR_PAIRS` instead of assuming 256, and folds the #3207 ValueError guard into the existing `return 0` fallback (graceful default colors rather than re-raising)
- unit tests for 8-color and colorless terminals against a mocked curses, verified failing before the fix

Verified under a pty with real terminfo: `TERM=vt100` runs monochrome and exits cleanly, `TERM=xterm`/`ansi` now emit 8-color escapes where they previously had none, `TERM=xterm-256color` output unchanged.

AI Level: 7 (Claude Fable 5, operated by @saulpw)

### DIFF
--- a/visidata/color.py
+++ b/visidata/color.py
@@ -170,14 +170,11 @@ class ColorMaker:
         if r is None:
             return None
 
-        try: # test to see if color is available
-            curses.init_pair(255, r, 0)
-            self.colorpair_cache[colorname] = r
-            return r
-        except curses.error:
-            return None  # not available
-        except ValueError:  # Python 3.10+  issue #1227; also for terminals with no colors  #3206
-            return None
+        if not 0 <= r < getattr(curses, 'COLORS', 0):
+            return None  # not available  #1227 #3206
+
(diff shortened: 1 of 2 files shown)


saulpw:
Given the comment on #3206 by @frosencrantz, I'm going to take this which should hopefully be more of a complete solution.  This should be part of 3.4.1.

